Issue #20, opened 2021-05-10
Create opensearch.org Maven repository


dblock:

mch2:
Would like to take a stab at this.  We have an issue to do this for lucene as well [here](https://github.com/opensearch-project/OpenSearch/issues/607). As a first step, am thinking of putting together a small cdk project to create the resources.  Will then look at updating infra to publish a nightly.

camerski:
@mch2 has been prototyping some options, and it looks like we have a path forward for automatically publishing everything through Sonatype with no human intervention needed. We probably don't need a self-hosted repository.

mch2:
Updated with a simple configuration to push to our Sonatype repo with ./gradlew publish.  I tested this with a locally running instance of nexus.

I do not think we need [#41] anymore, closing that out.

After this is added to CI, I am thinking we can extend this with a custom script in opensearch-build that builds/publishes each repo we would like to push.

mch2:
We did a successful test push here of only the opensearch repo - https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch.  @camerski is in the process of adding this to the nightly.

dblock:
See https://discuss.opendistrocommunity.dev/t/maven-repository-artifacts-for-plugin-development/6406 for another ask for this.

frotsch:
Pls. also refer to https://github.com/opensearch-project/OpenSearch/issues/939

frotsch:
Why only snapshots and not also the released artifacts?

For plugin developers it's IMHO an absolute must have. Local maven repo does not work well with distributed CI environments etc ...

So I like to ask to publish all the artifacts (releases and snapshots) to maven central like elastic did it/does it.

dblock:
> Why only snapshots and not also the released artifacts?
> 
> For plugin developers it's IMHO an absolute must have. Local maven repo does not work well with distributed CI environments etc ...
> 
> So I like to ask to publish all the artifacts (releases and snapshots) to maven central like elastic did it/does it.

Absolutely. I edited the issue accordingly.

frotsch:
https://discuss.opendistrocommunity.dev/t/how-will-the-opensearch-plugins-be-delivered-in-ga-release/6250

dblock:
[...]

PR #41, opened 2021-05-26
Add publish snapshot script

mch2:

### Description
Add a simple publish script to push pre-built/signed artifacts to a sonatype nexus repository.
This is intended do be used with https://aws.oss.sonatype.org/content/repositories/snapshots.

The script takes in a directory that should contain org/opensearch projects to publish.  It intentionally searches only for poms in this namespace and pushes only to a snapshot repository.

Tested with a local instance of nexus running at localhost:8081 and prebuilt/signed dummy artifacts.

note - this uses curl instead of the maven-deploy plugin to deploy.  The maven-deploy plugin auto generates a timestamp for each artifact, but it does not deploy signature files.  So that signatures match the artifact/pom names, any timestamp/build id should be appended during packaging.
 
### Issues Resolved
Partial - https://github.com/opensearch-project/opensearch-build/issues/20
 
### Check List

### DIFF
--- a/scripts/publish-snapshot.sh
+++ b/scripts/publish-snapshot.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
+
 ###### Information ############################################################################
+#  SPDX-License-Identifier: Apache-2.0
+#
+#  The OpenSearch Contributors require contributions made to
+#  this file be licensed under the Apache-2.0 license or a
+#  compatible open source license.
+#
 # Name:          publish-snapshot.sh
 # Language:      Shell
 #
--- a/scripts/publish-snapshot.sh
+++ b/scripts/publish-snapshot.sh
@@ -11,8 +11,7 @@
 # Language:      Shell
 #
 # About:         Deploy opensearch artifacts to a sonatype snapshot repository.
-#                The given directory is intended to be the root directory of a maven repository containing ./org/opensearch artifacts.
-#                This script will search POM files under ./org/opensearch.
+#                This script will search POM files under the passed in directory.
 #                If found, pom, jar, and signature files will be deployed to the org/opensearch namespace.
 #
 # Prerequisites: The given directory must be the parent directory of org/opensearch artifacts.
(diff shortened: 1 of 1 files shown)


scripts/publish-snapshot.sh
```diff
@@ -0,0 +1,93 @@
...
+    echo "No signature file found for jar, skipping ${jar}..."
+    continue
+  fi
+
+  echo "Validating signatures for - ${pomsig} and ${jarsig}"
+
+  gpg --verify-files "${pomsig}" "${jarsig}"
+
+  if [ $? -ne 0 ]; then
+    echo "Invalid signature on artifacts, skipping ${pom}"
```
camerski, line 79: You need a `continue` here.

Actually, it might be better to fail the whole script if any signature checks fail...otherwise it's hard to automatically verify that everything got uploaded correctly. I think I want a zero exit code to mean "everything was uploaded", not "everything that wasn't invalidated was uploaded".

camerski (commented):
Mostly minor stuff. The "invalid signature" logic needs a fix, but feel free to do the easy fix there.

scripts/publish-snapshot.sh
```diff
@@ -0,0 +1,93 @@
...
+#
+# Prerequisites: The given directory must be the parent directory of org/opensearch artifacts.
+#                Environment variables must be set:
+#                SONATYPE_ID/SONATYPE_PASSWORD - repository credentials
+#                SNAPSHOT_HOST - repository host
+#
+#
+# Usage:         ./publish-snapshot.sh <directory>
+#
+###############################################################################################
```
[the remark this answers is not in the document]
mch2: oops, will add it.

scripts/publish-snapshot.sh
mch2: Ok.  Will update this to validate the signature on all artifacts before any are pushed so that we don't end up with partial uploads.

mch2: If that is the case let's make sure we are executing this script one project at a time (that may have n artifacts to push) vs one execution to push multiple projects under org/opensearch.

Meaning CI execute this separately for each repo built.

camerski: Agreed.

gaiksaya: Can we add that in `About` section of the script? This can be easily missed

mch2: updated the About section

Issue #43, opened 2021-06-03
[Enhancement] Add a new deploy script for quick cluster creation


peterzhuamazon:
**Is your feature request related to a problem? Please describe.**
The original deployment script in ODFE is too complicated and takes a long time to run.
It also does not work outside of the existing GitHub Runners CI/CD workflows.
https://github.com/opendistro-for-elasticsearch/opendistro-build/blob/master/.github/scripts/setup_runners_service.sh

**Describe the solution you'd like**
We plan to recreate the script with a simpler and unified approach, so that the script can be run in any *nix environment.
The deployment should also be quick.
And it should include a cleanup process.

**Describe alternatives you've considered**

**Additional context**

PR #44, opened 2021-06-03
Add deploy single cluster tar scripts

peterzhuamazon:

### Description
Add deploy single cluster tar scripts
This script can deploy a single OpenSearch+Dashboards cluster for testing within 60 seconds.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/43
 
### Check List


scripts/deploy_cluster_tar_single.sh
```diff
@@ -0,0 +1,174 @@
...
+        c)
+            cleanup
+            exit
+            ;;
+        t)
+            TYPE=$OPTARG
+            ;;
+        s)
+            ENABLE_SECURITY=$OPTARG
+            ;;
```
camerski: Where is ENABLE_SECURITY used? I don't see any other references to it in this PR.

scripts/deploy_cluster_tar_single.sh
```diff
@@ -0,0 +1,174 @@
...
+echo "cluster.initial_master_nodes: [\"init-master\"]" >> config/opensearch.yml
+echo "cluster.name: opensearch-${VERSION}-linux-x64" >> config/opensearch.yml
+echo "network.host: 0.0.0.0" >> config/opensearch.yml
+echo "plugins.destination.host.deny_list: [\"10.0.0.0/8\", \"127.0.0.1\"]" >> config/opensearch.yml
+echo "script.context.field.max_compilations_rate: 1000/1m" >> config/opensearch.yml
+echo "webservice-bind-host = 0.0.0.0" >> plugins/opensearch-performance-analyzer/pa_config/performance-analyzer.properties
+
+# Start OpenSearch
+echo -e "\nStart OpenSearch"
+nohup ./opensearch-tar-install.sh > opensearch.log 2>&1 &
```
camerski: nohup will leave child processes running even if we ctrl-c out of this script. It would be better to put in a `wait` at the end of the script, similar to what the Docker entrypoint does.

Come to think of it, can you refactor this and the Docker entrypoint so they use the same startup and shutdown logic? Re-using that piece will make this script and the Docker image *way* more maintainable.

scripts/deploy_cluster_tar_single.sh
```diff
@@ -0,0 +1,174 @@
...
+    echo Clean Up
+    echo Kill Existing OpenSearch/Dashboards Process
+    (kill -9 `ps -ef | grep -i [o]pensearch | awk '{print $2}'` > /dev/null 2>&1) || echo -e "\tClear OpenSearch Process"
+    (kill -9 `ps -ef | grep -i [n]ode | awk '{print $2}'` > /dev/null 2>&1) || echo -e "\tClear Dashboards Process"
+    (kill -9 `ps -ef | grep -i [p]erformance | awk '{print $2}'` > /dev/null 2>&1) || echo -e "\tClear PerformanceAnalyzer Process"
+
+    echo Check PID List
+    (ps -ef | grep -i [o]pensearch) || echo -e "\tNo OpenSearch PIDs"
+    (ps -ef | grep -i [n]ode) || echo -e "\tNo Dashboards PIDs"
+    (ps -ef | grep -i [p]erformance) || echo -e "\tNo PerformanceAnalyzer PIDs"
```
camerski: This will kill all `node` processes, which is not what we want. If I have an unrelated `node` process running on my machine then I don't want it killed.

Also, we should be sending a SIGTERM, not a SIGKILL, so the processes have an opportunity to exit gracefully.

I recommend using the approach in the Docker entrypoint, where the script keeps track of the child PIDs and terminates them explicitly.

scripts/deploy_cluster_tar_single.sh
```diff
@@ -0,0 +1,174 @@
...
+
+# Start Dashboards
+echo -e "\nStart Dashboards"
+cd $DIR/opensearch-dashboards/bin
+nohup ./opensearch-dashboards > opensearch-dashboards.log 2>&1 &
+
+# Wait for start
+echo -e "\nSleep 30"
+sleep 30
+echo Startup Complete
```
camerski: It would be better to query the server every 5 seconds or so and look for a success response. Waiting 30 seconds doesn't guarantee that the server is up, and if the server comes up faster then we shouldn't be waiting around for it.

scripts/opensearch-onetime-setup.sh
```diff
@@ -0,0 +1,50 @@
...
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+
+# This script performs one-time setup for the OpenSearch tarball distribution.
+# It installs a demo security config and sets up the performance analyzer
+
+OPENSEARCH_HOME=`dirname $(realpath $0)`; cd $OPENSEARCH_HOME
```
camerski, line 30: I think it would be better to ask the caller to specify OPENSEARCH_HOME, otherwise this script is too tightly coupled with the `deploy_cluster_tar_single.sh` script, and the user *must* place this script in the root of their installation. I'm not sure we want that.

Specify OPENSEARCH_HOME in the `deploy_cluster_tar_single.sh` script and export it to this one, and make this script fail with an error if OPENSEARCH_HOME is unset.

scripts/opensearch-onetime-setup.sh
```diff
@@ -0,0 +1,50 @@
...
+
+##Security Plugin
+SECURITY_PLUGIN="opensearch-security"
+bash $OPENSEARCH_HOME/plugins/$SECURITY_PLUGIN/tools/install_demo_configuration.sh -y -i -s
```
camerski, line 34: You don't need the SECURITY_PLUGIN variable. It's not adding any value.

scripts/opensearch-onetime-setup.sh
```diff
@@ -0,0 +1,50 @@
...
+
+##Perf Plugin
+PA_PLUGIN="opensearch-performance-analyzer"
```
camerski, line 37: You don't need this either.

scripts/deploy_cluster_tar_single.sh
peterzhuamazon: That is the point it setup the cluster, start the process, and exit with the process keep running.

peterzhuamazon: If user want to stop it just run the script again with `-c` param.

scripts/deploy_cluster_tar_single.sh
peterzhuamazon: We do not have security disable param yet, so please do not approve it since it is still in progress.

scripts/deploy_cluster_tar_single.sh
camerski: See the next comment for why that is not OK

scripts/opensearch-onetime-setup.sh
peterzhuamazon: That is miss.

scripts/deploy_cluster_tar_single.sh
camerski: Can you pull it out of this PR then? Make a new PR with the security stuff once this one is approved.

peterzhuamazon:
I straight up copy the same onetime setup from docker part.
If we want to tweak onetime setup then the docker part needs changes as well.

camerski:
> I straight up copy the same onetime setup from docker part.
> If we want to tweak onetime setup then the docker part needs changes as well.

The refactoring comment was referring to the Docker entrypoint, not the onetime-setup step. In Docker-land, onetime-setup is only executed as part of the Docker build, not at runtime.

scripts/deploy_cluster_tar_single.sh
peterzhuamazon: Removed.

scripts/deploy_cluster_tar_single.sh
peterzhuamazon: We can find a better way to retrieve possible PID. Will improve.

scripts/deploy_cluster_tar_single.sh
peterzhuamazon: Not really liking that since Dashboards specifically will return 200 even if it is not up.

scripts/opensearch-onetime-setup.sh
peterzhuamazon: Need more debug and possible changes on docker then.

scripts/opensearch-onetime-setup.sh
peterzhuamazon: Need more debug and possible changes on docker then.

scripts/deploy_cluster_tar_single.sh
camerski: You'll need to look at the response content, not just the status code.

dblock:
Since I love renaming all the things :) Is a "single cluster" a misnomer, is this a "single node cluster" or a "single node"?

Will this run on any linux host or inside a docker container? How would one reuse this without doing something very damaging on their host?

scripts/opensearch-onetime-setup.sh
camerski: Not sure why GitHub included line 34 in the context here...the comment was specifically referring to the PA_PLUGIN variable. It's only used once, on line 38, and not exported, so you could just change line 38 to say "opensearch-performance-analyzer" rather than $PA_PLUGIN.

You definitely do need line 34.

peterzhuamazon:
Close this due to a lot of changes for the demo purposes.

PR #70, opened 2021-06-24
Add tarball build scripts

gaiksaya:
### Description
This PR provides build script to build opensearch tarball. 
Also adds manifest file required to download the required artifacts
 
### Issues Resolved
 
### Check List
- [x] Tested locally and on dev-desktop for x64


scripts/manifest.yml
```diff
@@ -0,0 +1,30 @@
+---
+version: 1.0.0.0-rc1
+
+download_urls:
+  OpenSearch: https://artifacts.opensearch.org/snapshots/core/opensearch/1.0.0/opensearch-min-1.0.0-SNAPSHOT-linux-x64-186.tar.gz
```
dblock, line 5: Consider structuring these in separate configuration files, or as a hierarchy like so:

```
products:
  opensearch:
    libs:
    plugins:
  opensearch-dashboards:
    plugins:
```

camerski: +1

gaiksaya: Sure will make the change. This script only concentrates on opensearch. We have different script for dashboards. For manifest I think I'll restructure it as hierarchy so there is low maintenance. Let me know if you think otherwise

gaiksaya: Made the changes

dblock:
Look good! Leaving to @camerski or others for appoval/CR.

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,159 @@
...
+            ;;
+    esac
+done
+
+if [ -z "$PLATFORM" ]; then
+    PLATFORM="linux"
+fi
+
+if [ "$PLATFORM" != "linux" ]; then
+    echo "error: This platform is not supported yet!"
```
camerski, line 76: Include the value of $PLATFORM in the error message so the user knows what mistake they made

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,159 @@
...
+    usage
+    exit 1
+fi
+
+if [ -z "$ARCHITECTURE" ]; then
+    ARCHITECTURE="x64"
+fi
+
+if [ "$ARCHITECTURE" != "x64" ] || [ "$ARCHITECTURE" != "arm64" ]; then
+    echo "error: This architecture is not supported yet!"
```
camerski, line 86: Include the value of $ARCHITECTURE in the error message

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,159 @@
...
+    usage
+    exit 1
+fi
+
+if [ -z "$MANIFEST_FILE" ]; then
+    echo "Please specify the right path to manifest file"
```
camerski, line 92: How about "You must specify a manifest file". The current message implies that a path was given but was somehow incorrect.

You might want to add a check that the manifest file exists and fail the script if it doesn't.

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,159 @@
...
+mkdir -p ${WORKING_DIR}/plugins_downloads
+IFS=$'\n'
+cd ${WORKING_DIR}/plugins_downloads
+for plugin in ${OPENSEARCH_PLUGINS_URLS}; do
+  aws s3 cp $plugin .
+done
+
+# Download k-NN lib
+aws s3 cp $LIBS .
+unzip opensearch-knnlib-*-$PLATFORM-$ARCHITECTURE.zip
```
camerski, line 113: This should be able to handle any number of libs. If someone adds a second lib file to the manifest, this script will not behave the way we want it to because it is hard-coded to knn.

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,159 @@
...
+
+cd ..
+
+# Download OpenSearch core TAR
+wget -q ${OPENSEARCH_CORE_URL}
+tar -xzf opensearch-min-*
+rm *tar.gz
```
camerski, line 120: This tightly couples the script to the filename in the manifest. I'd recommend this instead:

```
wget -q ${OPENSEARCH_CORE_URL} -O opensearch.tgz
tar -xzf opensearch.tgz
rm opensearch.tgz
```

That way there is no way for a glob to accidentally pick up an unrelated tarball, and it doesn't matter if the manifest specifies a differently-named file (for example, I might want to build a bundle from `camerski-testing-opensearch.tar.gz`)

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,159 @@
...
+cd opensearch-*
+for plugins in $OPENSEARCH_PLUGINS_URLS; do
+    plugin=`basename $plugins`
+  ./bin/opensearch-plugin install --batch file:../plugins_downloads/$plugin
+done
+
+echo "Plugins installed are:\n"
+ls -tlr plugins
+
+# Setup PA
```
camerski, line 132: Avoid acronyms in comments

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,159 @@
...
+
+# Copy the tarball installation script 
+cp $REPO_ROOT/release/tar/linux/opensearch-tar-install.sh .
+
+# Setup k-NN-library
+mkdir -p plugins/opensearch-knn/knnlib
+cp ../plugins_downloads/opensearch-knnlib-*/libKNNIndexV2_0_11.so plugins/opensearch-knn/knnlib 
+
+
+# Create TAR and Upload TAR to S3
```
camerski, line 154: You're not uploading to S3

release/tar/linux/opensearch-tar-build.sh
gaiksaya: `usage` has all the information. Called that function on all error messages

release/tar/linux/opensearch-tar-build.sh
gaiksaya: usage has all the information. Called that function on all error messages

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Will add it!

release/tar/linux/opensearch-tar-build.sh
gaiksaya: I did make it like that intentionally because it each library is handled differently. Check below how the knn behaves. I'll consider that as an future enhancement when we add new libs

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Makes sense. I did that in CDK not sure why I skipped here. Will make the changes.

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Removing the comments

scripts/manifest.yml
```diff
@@ -0,0 +1,32 @@
+---
+version: 1.0.0.0-rc1
```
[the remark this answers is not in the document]
gaiksaya: Done

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Done

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Done

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Done

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Added

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,175 @@
...
+    echo "error: The given file doesnot exist. Please check the specified path"
+    usage
+    exit 1
+fi
+
+SCHEMA_VERSION=`yq eval '.manifest.schema-version' $MANIFEST_FILE`
+echo $SCHEMA_VERSION
+
+if [ $SCHEMA_VERSION != 1.0 ]; then
+    echo "error: This script only supports manifest schema version 1.0. Please use the manifest file with right schema version"
```
camerski, line 109: Awesome :)

My bash-fu is weak; do you need to put `1.0` in quotes?

Might be worth putting `$SCHEMA_VERSION` in quotes, too, in case it contains a space.

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,175 @@
...
+    usage
+    exit 1
+fi
+
+REPO_ROOT=`git rev-parse --show-toplevel`
+VERSION=`yq eval '.version' $MANIFEST_FILE`
+WORKING_DIR="opensearch-working-directory"
+OPENSEARCH_CORE_URL=`yq eval '.products.opensearch.opensearch-min' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
+OPENSEARCH_PLUGINS_URLS=`yq eval '.products.opensearch.plugins' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
+LIBS=`yq eval '.products.opensearch.libs' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
```
peternied, line 119: Can we refactor the `sed | sed` steps in a way that they can be reused?

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,175 @@
...
+# Setup Performance Analyzer Agent
+cp -r plugins/opensearch-performance-analyzer/performance-analyzer-rca .
+chmod -R 755 performance-analyzer-rca
+mv bin/opensearch-performance-analyzer/performance-analyzer-agent-cli ./bin
+rm -rf ./bin/opensearch-performance-analyzer
+
+
+# Make sure the data folder exists and is writable
+if [ ! -d "data" ]; then 
+    mkdir data
```
peternied, line 159: Consider rewriting as `mkdir data 2>/dev/null` to so we can avoid the comment/if statement

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,175 @@
...
+fi
+chmod 755 data/
+
+# Copy the tarball installation script 
+cp $REPO_ROOT/release/tar/linux/opensearch-tar-install.sh .
```
peternied, line 164: I am not a fan of using current directory in scripts, how would you feel about using a variable instead?

That way any directory structure is explicitly declared and mapped

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,175 @@
...
+    PLATFORM="linux"
+fi
+
+if [ "$PLATFORM" != "linux" ]; then
+    echo "error: $PLATFORM platform is not supported yet!"
+    usage
+    exit 1
+fi
+
+if [ -z "$ARCHITECTURE" ]; then
```
peternied, line 82: Architecture should be required or it will error out - avoids being unclear about what is produced

peternied: Offline recap:
What if we moved these parameters into the manifest itself?

> As the artifacts listed are platform specific it doesn't make much sense to change the way you produce the product to be `./opensearch-tar-build.sh -f manifest.yml -a x64` and then do `./opensearch-tar-build.sh -f manifest.yml -a arm64` and you get the exact same but bundled up in an incorrectly named bundle output.

This would effective make the only parameter to this script be the manifest file.  Then it might make sense as convention to suggestion `manifest.linux.x64.yml` which is duplicated in the lines of the manifest file

release/tar/linux/opensearch-tar-build.sh
peternied: Offline recap: based on the usage of `set -e` at the top of the file this suggestion could introduce a hard to detect bug! Might be useful to have a script style overview where we figure out what styles make the most sense for us as a team - I'll drive setting this up

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -0,0 +1,175 @@
...
+ls -tlr plugins
+
+# Setup Performance Analyzer Agent
+cp -r plugins/opensearch-performance-analyzer/performance-analyzer-rca .
+chmod -R 755 performance-analyzer-rca
+mv bin/opensearch-performance-analyzer/performance-analyzer-agent-cli ./bin
+rm -rf ./bin/opensearch-performance-analyzer
+
+
+# Make sure the data folder exists and is writable
```
peternied, line 157: Offline: Could we require the performance analyzer to fix this by adding a .gitkeep with the directly path they need?   It is creating a special step added to decoupled code, a great way to add complexity that we will have trouble resolving in the future.

release/tar/linux/opensearch-tar-build.sh
camerski: `mkdir -p` does not generate an error if the directory already exists. Use that and get rid of the `if`.

release/tar/linux/opensearch-tar-build.sh
camerski: +1

gaiksaya: How does this look?

```
manifest:
  schema-version: 1.0
  platform: linux
  architecture: x64
```

peternied: I would be inclined to make platform part of the product (where a product is OpenSearch) as there are components that are interpreted and can run on any platform, so it isn't the manifest that has these properties but the produced artifact that does.

```
manifest:
   schema-version: 1.0
   description: Creates the OpenSearch bundle of core engine and default plugins
   product:
      name: OpenSearch
      platform: linux
      architecture: x64
```

Added some extra fields that make the manifest a little more verbose.  That prose isn't used by the script, but it could be useful if you've got a big stack of these manifest and you want some historical context

camerski: Platform and architecture don't belong in the `manifest` section, because they're not describing the manifest itself. I think they should be under each `product`. That will allow us to have one manifest that can build all four products (core, dashboards, core-arm, dashboards-arm) with the same product version number:

```
products:
  OpenSearch-linux-x64:
    platform: linux
    architecture: x64
    ...
  OpenSearch-linux-arm64:
    platform: linux
    architecture: arm64
    ...
```

Alternatively, a separate `build` section might make sense:

```
manifest:
  schema-version: 1.0
build:
  version: 1.0.0
  platform: linux
  architecture: x64
products:
  ...
```

The first option gives us build-the-whole-world-with-one-manifest, the second option gives us simpler manifests with no need for any logic about how to generate output filenames. Now that I write that sentence down, I think option 2 makes the most sense. Keep it simple.

gaiksaya: Right! It would be better to NOT hard code the key (eg: OpenSearch-linux-x64). Rather keep it in generalized. Will go with option 2 for now. Any future enhancements can be a part of manifest schema-version `2.0`

release/tar/linux/opensearch-tar-build.sh
gaiksaya: done

release/tar/linux/opensearch-tar-build.sh
gaiksaya: done

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Done

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Done

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -69,57 +52,60 @@ while getopts ":ha:p:f:" arg; do
...
 OPENSEARCH_CORE_URL=`yq eval '.products.opensearch.opensearch-min' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
 OPENSEARCH_PLUGINS_URLS=`yq eval '.products.opensearch.plugins' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
 LIBS=`yq eval '.products.opensearch.libs' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
 
 # # Download plugins to local 
-mkdir -p ${WORKING_DIR}/plugins_downloads
+mkdir -p $BASE_DIR/plugins_downloads
+cd $BASE_DIR
+WORKING_DIR=`pwd`
+echo "working dir is: ${WORKING_DIR}"
```
peternied, line 111: Helpful!

peternied (approved):

release/tar/linux/opensearch-tar-build.sh
peternied: @gaiksaya Could we create an issue on this to follow up?

gaiksaya: Created one! Feel free to add/edit the right words. https://github.com/opensearch-project/performance-analyzer/issues/29

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -37,8 +28,6 @@ function usage() {
     echo -e "-f MANIFEST_FILE\tPath/to/manifest.yml file"
     echo ""
     echo "Optional arguments:"
```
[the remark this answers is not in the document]
gaiksaya: Done

camerski (approved):

Issue #34, opened 2021-05-18
OpenSearch / OpenSearch Dashboards - TAR distribution (ARM64)


bbarani:
Is your feature request related to a problem? Please describe.
No

Describe the solution you'd like
Provide signed (ARM64)Tar artifact for OpenSearch / OpenSearch Dashboards

Describe alternatives you've considered
Generating the artifact using the source code

Additional context
Currently there are no Tar artifacts for OpenSearch / OpenSearch Dashboards

PR #72, opened 2021-06-25
Add manifest for ARM64 support tarball

gaiksaya:

### Description
This manifest file will be used while creating ARM64 tarball
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/34
 
### Check List


scripts/manifest.arm64.yml
```diff
@@ -0,0 +1,32 @@
+---
+version: 1.0.0.0-rc1
```
camerski, line 2: 1.0.0-rc1? If we're following semver then it should be three digits.

gaiksaya: The entire manifest will change per release, this is just a sample at this moment. Should I change it to 1.0.0?

scripts/manifest.arm64.yml
```diff
@@ -0,0 +1,32 @@
...
+    plugins:
+        - s3://staging.opensearch.amazon.com/plugins-temp/dashboard_plugins/queryWorkbenchDashboards-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/dashboard_plugins/anomalyDetectionDashboards-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/dashboard_plugins/securityDashboards-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/dashboard_plugins/alertingDashboards-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/dashboard_plugins/indexManagementDashboards-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/dashboard_plugins/notebooksDashboards-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/dashboard_plugins/ganttChartDashboards-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/dashboard_plugins/reportsDashboards-1.0.0.0-rc1-linux-arm64.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/dashboard_plugins/traceAnalyticsDashboards-1.0.0.0-rc1.zip
```
camerski, line 32: Consider adding a `schema-version` field to the manifest so we can evolve the format.

gaiksaya: Done

camerski (approved):

scripts/manifest.arm64.yml
```diff
@@ -0,0 +1,33 @@
+---
+manifest:
+  schema-version: 1.0
+version: 1.0.0
+products:
+  opensearch: 
+    opensearch-min: https://artifacts.opensearch.org/snapshots/core/opensearch/1.0.0/opensearch-min-1.0.0-SNAPSHOT-linux-arm64-189.tar.gz
```
peternied, line 7: Considering this string `opensearch-min-1.0.0-SNAPSHOT-linux-arm64-189.tar.gz`

* `opensearch-min` name
* `1.0.0` version
* `SNAPSHOT` build flavor/type?
* `linux` operating system
* `arm64` platform - Where is this naming from? sometimes I've seen it as `aarch64` other times `arm`
* `189` build number - Should this be hardcoded in a manifest?
* `.tar.gz` file extension.

scripts/manifest.arm64.yml
```diff
@@ -0,0 +1,33 @@
...
+        - s3://staging.opensearch.amazon.com/plugins-temp/opensearch_plugins/opensearch-security-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/opensearch_plugins/opensearch-performance-analyzer-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/opensearch_plugins/opensearch-index-management-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/opensearch_plugins/opensearch-knn-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/opensearch_plugins/opensearch-anomaly-detection-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/opensearch_plugins/opensearch-reports-scheduler-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/opensearch_plugins/opensearch-asynchronous-search-1.0.0.0-rc1.zip
+        - s3://staging.opensearch.amazon.com/plugins-temp/opensearch_plugins/opensearch-notebooks-1.0.0.0-rc1.zip
+  opensearch-dashboards: 
+    opensearch-dashboards-min: https://artifacts.opensearch.org/snapshots/core/opensearch-dashboards/1.0.0/opensearch-dashboards-min-1.0.0-SNAPSHOT-linux-arm64-110.tar.gz
```
peternied, line 23: Should the build numbers be different?  If these are different artifacts shouldn't they be in separate manifests?

scripts/manifest.arm64.yml
```diff
@@ -0,0 +1,33 @@
+---
```
peternied, line 1: Should there be a header of some kind?

peternied (commented):
Is there another manifest for x64 architectures?

scripts/manifest.arm64.yml
gaiksaya: Manifest files will change each release/build. The plugins/core will provide the link to get their artifacts from. We would just download them and bundle

scripts/manifest.arm64.yml
gaiksaya: A manifest is collection of all the download links to get the raw artifacts that can be bundled together into the final product. It is a central location to know what to get from where.

scripts/manifest.arm64.yml
gaiksaya: Nope Its just `yml` file :)

scripts/manifest.arm64.yml
peternied: Offline recap: the artifacts are updated by an external source, these are expected to change over time and not necessarily align.  That said - OS and OSB in the same file might be worth considering changing as the processes and people updating the sections of this file will be different.

peternied (approved):

scripts/manifest.arm64.yml
camerski: I think we should be creating a new manifest file for each release, not updating the old one. Per-release manifests makes it super easy to reproduce an old build, and makes code review simpler because we're not attempting to diff two different versions. Each release is a stand-alone thing.

gaiksaya: In that case I'll rename this as `manifest.linux.arm64.example.yml`. The `example` can be replaced with right version in future

PR #64, opened 2021-06-22
Add standard test related scripts to build repo

peterzhuamazon:
Signed-off-by: Peter Zhu <zhujiaxi@amazon.com>

### Description
Add standard test related scripts to build repo.
This PR include all the related integTest scripts for deploy cluster, setup mavenlocal, as well as integTest dockerfile.
These are all the things we need to run integTest in opensearch/dashboards.

 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #75, opened 2021-06-29
Add tarball build script for opensearch dashboards

gaiksaya:

### Description
Add tarball build script for opensearch dashboards. Supports both arm64 and x64
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/34
 
### Check List
- [x] Tested on dev desktop(x64) and ec2(arm64)


release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,147 @@
...
+function usage() {
+    echo "Usage: $0 [args]"
+    echo -e "Supported manifest schema-version: 1.0"
+    echo ""
+    echo "Arguments:"
+    echo -e "-f MANIFEST_FILE\t[Required] Path/to/manifest.yml file"
+    echo -e "-h help"
+}
+
+while getopts ":ha:p:f:" arg; do
```
peterzhuamazon, line 34: What is `a:` for?

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,147 @@
...
+fi
+
+if [ "$ARCHITECTURE" != "x64" ] && [ "$ARCHITECTURE" != "arm64" ]
+then
+    echo "error: $ARCHITECTURE architecture is not supported yet! Only supported architectures are x64 and arm64"
+    usage
+    exit 1
+fi
+
+if [[ ( "$ARCHITECTURE" == "arm64" && ( "$UNDERLYING_ARCH" != "aarch64" && "$UNDERLYING_ARCH" != "arm64" )) ||  ( "$ARCHITECTURE" == "x64" && ( "$UNDERLYING_ARCH" != "x86_64" && "$UNDERLYING_ARCH" != "amd64" )) ]]
```
peterzhuamazon, line 98: Same call outs as in the #76: https://github.com/opensearch-project/opensearch-build/pull/76#pullrequestreview-696254839

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,147 @@
...
+    echo "Please use the appropriate build environment"
+    exit 1
+fi
+
+REPO_ROOT=`git rev-parse --show-toplevel`
+ROOT=`dirname $(realpath $0)`; echo $ROOT; cd $ROOT
+DIR_NAME="opensearch-dashboards-$VERSION"
+WORKING_DIR="$ROOT/$DIR_NAME"
+TARGET_DIR="$ROOT/target"
+PLUGINS_TEMP=`mktemp -d`
```
peterzhuamazon, line 110: Need to add a trap delete entry to properly cleanup the temp folder, or you will end up with infinite temp folder with plugins.

peterzhuamazon: Recommend waiting for my #64 to get in so you can use the lib to properly cleanup these.

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,147 @@
...
+for plugin in ${OPENSEARCH_DASHBOARDS_PLUGINS_URLS}; do
+  aws s3 cp $plugin $PLUGINS_TEMP/
+done
+
+# Download OpenSearch Dashboards min tarball
+wget -q ${OPENSEARCH_DASHBOARDS_CORE_URL} -O opensearch-dashboards.tar.gz
+tar -xzf opensearch-dashboards.tar.gz --strip-components=1 --directory "$WORKING_DIR" && rm -rf opensearch-dashboards.tar.gz
+
+# Install plugins to OpenSearch
+for plugins in $OPENSEARCH_DASHBOARDS_PLUGINS_URLS; do
```
peterzhuamazon, line 128: Very confusing use of variable in `plugins`, which is really just a single url of that plugin, use `plugin_url` is better.

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,147 @@
...
+
+# Tar the bundle and clean up
+rm -rf $PLUGINS_TEMP
+cd $ROOT
+echo "Generating the tarball"
+tar -czf $TARGET_DIR/opensearch-dashboards-$VERSION-$PLATFORM-$ARCHITECTURE.tar.gz $DIR_NAME
+cd $TARGET_DIR
+shasum -a 512 opensearch-dashboards-$VERSION-$PLATFORM-$ARCHITECTURE.tar.gz > opensearch-dashboards-$VERSION-$PLATFORM-$ARCHITECTURE.tar.gz.sha512
+shasum -a 512 -c opensearch-dashboards-$VERSION-$PLATFORM-$ARCHITECTURE.tar.gz.sha512
+rm -rf ${WORKING_DIR}
```
peterzhuamazon, line 147: You dont rm TARGET_DIR as well?

peterzhuamazon: Suggest the cleanup should happen earlier, as well as later in the script.

gaiksaya: Target contains the final artifact and shasum we cannot remove that

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Missed it. Will add the removal of temp dir

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Answered

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: It was for arch. Will remove a and p as well. Thanks for catching

gaiksaya: Done

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Removed the temp dir but makes sense to remove the folders when the code fails. Will look into it.

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Changed

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,147 @@
...
+
+if [[ ( "$ARCHITECTURE" == "arm64" && ( "$UNDERLYING_ARCH" != "aarch64" && "$UNDERLYING_ARCH" != "arm64" )) ||  ( "$ARCHITECTURE" == "x64" && ( "$UNDERLYING_ARCH" != "x86_64" && "$UNDERLYING_ARCH" != "amd64" )) ]]
+then
+    echo "error: The underlying architecture($UNDERLYING_ARCH) doesnot match the product architecture($ARCHITECTURE) that you are trying to build"
+    echo "Please use the appropriate build environment"
+    exit 1
+fi
+
+REPO_ROOT=`git rev-parse --show-toplevel`
+ROOT=`dirname $(realpath $0)`; echo $ROOT; cd $ROOT
```
peternied, line 106: Did you want to keep this echo $ROOT?  If so might want to add context in the echoed message

gaiksaya: Nope not needed! will remove it

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Added trap

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,153 @@
...
+done
+
+if [ -z "$MANIFEST_FILE" ]; then
+    echo "error: You must specify the manifest file"
+    usage
+    exit 1
+fi
+
+if [ ! -f "$MANIFEST_FILE" ]; then
+    echo "error: The given file doesnot exist. Please check the specified path"
```
camerski, line 62: "Error: The file '${MANIFEST_FILE}' does not exist"

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,153 @@
...
+UNDERLYING_ARCH=`uname -p`
+
+if [ -z "$VERSION" ]; then
+    echo "error: Please specify the version in the manifest file"
+    usage
+    exit 1
+fi
+
+if [ "$SCHEMA_VERSION" != "1.0" ]; then
+    echo "error: This script only supports manifest schema version 1.0. Please use the manifest file with right schema version"
```
camerski, line 80: "Error: This script only supports manifest schema version 1.0, but ${MANIFEST_FILE} is version '${SCHEMA_VERSION}'"

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,153 @@
...
+    usage
+    exit 1
+fi
+
+if [ "$PLATFORM" != "linux" ]; then
+    echo "error: $PLATFORM platform is not supported yet! Only supported platform is linux"
```
camerski, line 86: Good error message

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,153 @@
...
+fi
+
+if [ "$ARCHITECTURE" != "x64" ] && [ "$ARCHITECTURE" != "arm64" ]
+then
+    echo "error: $ARCHITECTURE architecture is not supported yet! Only supported architectures are x64 and arm64"
+    usage
+    exit 1
+fi
+
+if [[ ( "$ARCHITECTURE" == "arm64" && ( "$UNDERLYING_ARCH" != "aarch64" && "$UNDERLYING_ARCH" != "arm64" )) ||  ( "$ARCHITECTURE" == "x64" && ( "$UNDERLYING_ARCH" != "x86_64" && "$UNDERLYING_ARCH" != "amd64" )) ]]
```
camerski, line 98: Why don't we just figure out ARCHITECTURE from UNDERLYING_ARCH? Then the user doesn't need to specify the architecture. If you want x86, run this script on an x86 box. If you want arm, run it on arm.

What you have is fine for now; this is a suggestion for future improvements.

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,153 @@
...
+then
+    echo "error: The underlying architecture($UNDERLYING_ARCH) doesnot match the product architecture($ARCHITECTURE) that you are trying to build"
```
camerski, line 100: Take care with spacing: `architecture($UNDERLYING_ARCH)` and `doesnot`.

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,153 @@
...
+DIR_NAME="opensearch-dashboards-$VERSION"
+WORKING_DIR="$ROOT/$DIR_NAME"
+TARGET_DIR="$ROOT/target"
+PLUGINS_TEMP=`mktemp -d`
+OPENSEARCH_DASHBOARDS_CORE_URL=`yq eval '.products.opensearch-dashboards.opensearch-dashboards-min' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
+OPENSEARCH_DASHBOARDS_PLUGINS_URLS=`yq eval '.products.opensearch-dashboards.plugins' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
+
+TRAP_SIG_LIST="TERM INT EXIT CHLD"
+function delete_temp_folders() {
+rm -rf $WORKING_DIR $PLUGINS_TEMP $TARGET_DIR
```
camerski, line 116: Indent this so it's obvious that we're in function scope.

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,153 @@
...
+REPO_ROOT=`git rev-parse --show-toplevel`
+ROOT=`dirname $(realpath $0)`; cd $ROOT
+DIR_NAME="opensearch-dashboards-$VERSION"
+WORKING_DIR="$ROOT/$DIR_NAME"
+TARGET_DIR="$ROOT/target"
+PLUGINS_TEMP=`mktemp -d`
+OPENSEARCH_DASHBOARDS_CORE_URL=`yq eval '.products.opensearch-dashboards.opensearch-dashboards-min' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
+OPENSEARCH_DASHBOARDS_PLUGINS_URLS=`yq eval '.products.opensearch-dashboards.plugins' $MANIFEST_FILE | sed s/^-// | sed -e 's/^[[:space:]]*//'`
+
+TRAP_SIG_LIST="TERM INT EXIT CHLD"
```
camerski, line 114: We don't want to trap SIGCHLD, do we? SIGCHLD will fire whenever this script executes a command, e.g. `wget` or `tar`.

release/tar/linux/opensearch-dashboards-tar-build.sh
```diff
@@ -0,0 +1,153 @@
...
+function delete_temp_folders() {
+rm -rf $WORKING_DIR $PLUGINS_TEMP $TARGET_DIR
+}
+trap delete_temp_folders $TRAP_SIG_LIST
+
+# # Download plugins to local 
+mkdir $WORKING_DIR
+mkdir -p $TARGET_DIR
```
camerski, line 122: Why `mkdir` for one and `mkdir -p` for the other?

camerski (commented):
Mostly looks pretty good. A couple of suggestions and a question about SIGCHLD.

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Because WORKING_DIR is removed at the end of the script or by trap if the script fails. We want TARGET_DIR to exist even after script finishes. Usecase: User runs both dashboards and opensearch script. The script should not fail saying it already exists. (Same with what we did with /data folder) The reason for target folder is we have 2 artifacts 1 bundle and 1 sha512. It would easily get lost of not put in the folder. Dont want user to search through his folders for artifacts

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Done

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Done

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Noted as an enhancement

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: There is just one space

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Done

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Removed

release/tar/linux/opensearch-dashboards-tar-build.sh
camerski: Exactly...there should be two more spaces :) Between `architecture` and `($UNDERLYING_ARCH)`, and between `does` and `not`.

Totally not a blocker for this change.

release/tar/linux/opensearch-dashboards-tar-build.sh
camerski: Do we want the script to fail if $WORKING_DIR already exists? It's OK if we do, but it's not clear to me that we want that behavior.

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: done :)

release/tar/linux/opensearch-dashboards-tar-build.sh
gaiksaya: Well in that case, wont the script remove the folder during trap? If the folder is already there and we say exit this part of the code will remove it right? Referring to this part of the code: https://github.com/opensearch-project/opensearch-build/pull/75/files#diff-341213f87e3420ebd2beca37f8e09c45f47c74c39cfc32b7b47678d290362ae0R116

gaiksaya: Would it be safe to add `-p` to $WORKING_DIR as well?

camerski: Adding `-p` is safe. The way `mkdir` works is that it will fail with an error if the path already exists, unless you specify `-p`.

So the question is: Why are we failing if WORKING_DIR exists, but not failing if TARGET_DIR exists? If the answer is "we should fail on both", that's fine. If the answer is "we should not fail on either", that's also fine. If the answer is "WORKING_DIR is different to TARGET_DIR and here is why we fail for WORKING_DIR but not TARGET_DIR", that's also fine - just add a comment explaining why.

gaiksaya: Got it. I'll add `-p` to be safe.

gaiksaya: Done

peternied (approved):

camerski (approved):

PR #76, opened 2021-06-29
Additional check for build env and handle target directory creation

gaiksaya:

### Description
Additional check for build env (the build architecture should support the product architecture)
Handle target directory creation (Use case: User wants to build both opensearch and dashboards. Earlier script throws an error saying `target` directory already exists
 
### Check List


release/tar/linux/opensearch-tar-build.sh
```diff
@@ -93,6 +94,13 @@ then
     exit 1
 fi
 
+if [[ ( "$ARCHITECTURE" == "arm64" && ( "$UNDERLYING_ARCH" != "aarch64" && "$UNDERLYING_ARCH" != "arm64" )) ||  ( "$ARCHITECTURE" == "x64" && ( "$UNDERLYING_ARCH" != "x86_64" && "$UNDERLYING_ARCH" != "amd64" )) ]]
```
peterzhuamazon, line 97: You mean `( "$UNDERLYING_ARCH" != "aarch64" || "$UNDERLYING_ARCH" != "arm64" )`?
Since a server cannot be both aarch64 and arm64 at the same time.

peterzhuamazon: Actually confused a bit why it is a problem with underlying arch.
If user building arm64 on x64 and failed I can get it.
However, build arm64 on aarch64 is not a concern since they are internally the same arch, just name differences.

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -105,7 +113,7 @@ LIBS=`yq eval '.products.opensearch.libs' $MANIFEST_FILE | sed s/^-// | sed -e '
 
 # # Download plugins to local 
 mkdir $WORKING_DIR
-mkdir $TARGET_DIR
+mkdir -p $TARGET_DIR
```
peterzhuamazon, line 116: Why does TARGET_DIR have -p while WORKING_DIR does not?

peterzhuamazon (changes requested):
Line `33` what is `a:` for?

peterzhuamazon (changes requested):
Need to add a trap delete entry to properly cleanup the temp folder, or you will end up with infinite temp folder with plugins.

Do not use rm -rf later on since you cannot guarantee the script wont quit in the middle of the run.

Recommend waiting for my #64 to get in so you can use the lib to properly cleanup these.

peterzhuamazon (changes requested):
Line `132`: Very confusing use of variable in `plugins`, which is really just a single url of that plugin, use `plugin_url` is better.

peterzhuamazon (changes requested):
Last line:
You dont rm TARGET_DIR as well?
Suggest the cleanup should happen earlier, as well as later in the script.

release/tar/linux/opensearch-tar-build.sh
gaiksaya: yes aarch64 and arm64 for building arm is okay. That is what this logic does. Tested it on arm64 and x64. You can try too. If manifest contains arch as `arm64` and underlying system is neither aarch64 `&&` arm64 then throw this error

release/tar/linux/opensearch-tar-build.sh
gaiksaya: I mentioned the use case in PR description

release/tar/linux/opensearch-tar-build.sh
```diff
@@ -93,6 +94,13 @@ then
     exit 1
 fi
 
+if [[ ( "$ARCHITECTURE" == "arm64" && ( "$UNDERLYING_ARCH" != "aarch64" && "$UNDERLYING_ARCH" != "arm64" )) ||  ( "$ARCHITECTURE" == "x64" && ( "$UNDERLYING_ARCH" != "x86_64" && "$UNDERLYING_ARCH" != "amd64" )) ]]
+then
+    echo "error: The underlying architecture($UNDERLYING_ARCH) doesnot match the product architecture($ARCHITECTURE) that you are trying to build"
+    echo "Please use the appropriate build environment"
```
peternied, line 100: `appropriate` doesn't tell someone that sees this failure how they can resolve it, could we be specific about the kind of environment that should be setup, e.g. `To build arm64 you need to be on an arm64 architecture environment?`

peternied (approved):

release/tar/linux/opensearch-tar-build.sh
gaiksaya: Done

camerski (approved):

peterzhuamazon (approved):

PR #79, opened 2021-07-07
Fix trap handling on successful exits

gaiksaya:

### Description
Trap function should delete temporary folders only when there is an unsuccessful exit. This PR implements the same
 
### Issues Resolved
 
### Check List
- [x] Tested locally with both successful and unsuccessful (exit 1) exits


peternied (approved):
I am approving, but this is a problem, this 'simple script' is growing and growing.  I don't believe that shell script is the only option - and this is an illustration of why this might be a unmaintainable option over time.

gaiksaya:
> I am approving, but this is a problem, this 'simple script' is growing and growing. I don't believe that shell script is the only option - and this is an illustration of why this might be a unmaintainable option over time.

I agree! Complicated code is very difficult to maintain in shell. We can think about other options in future for this.

peterzhuamazon (approved):

PR #98, opened 2021-07-24
Add OpenSearch/Dashboards Multi-Arch Docker Images

### LINKED ISSUE #30: OpenSearch / OpenSearch Dashboards Multi-Arch Docker image (X64 + ARM64)
Is your feature request related to a problem? Please describe.
No

Describe the solution you'd like
Provide (ARM64)Docker images via DockerHub for OpenSearch / OpenSearch Dashboards (With plugins)

Describe alternatives you've considered
Generating the image using the source code

Additional context
Currently there are no (ARM) Docker images for OpenSearch / OpenSearch Dashboards

peterzhuamazon:
### Description
Add OpenSearch/Dashboards Multi-Arch Docker Images
 
### Issues Resolved
#30 
 
### Check List


release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,141 @@
+#!/bin/bash
+
+set -e
+
+# Import libs
+. ../../lib/shell/file_management.sh
```
gaiksaya, line 6: Can you use absolute path here? If the script is run from different folder this can cause issue

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,141 @@
...
+
+# Check buildx status
+echo -e "\n* Check buildx status"
+docker buildx ls | grep $BUILDER_NAME
+docker ps | grep $BUILDER_NAME
+
+
+# Copy configs
+cp -v config/${PRODUCT}/* $DIR/
+cp -v ../../scripts/opensearch-onetime-setup.sh $DIR/
```
gaiksaya, line 115: Same. Use absolute path from root dir

release/docker/build-image-multi-arch.sh
peterzhuamazon: Then user should run from the default folder.
Because if they run from other folder not only this line will fail the cp line and other download line will also fail.
I dont want to tie this script to the git command to find out the repo root.

release/docker/build-image-multi-arch.sh
peterzhuamazon: Same as above.

release/docker/build-image-multi-arch.sh
gaiksaya: How is that different from all other scripts we use? Also inorder to use this script they have to clone, fork the repo anyway. The thing is it can be easily missed and any change directory later can cause the file not found exception.

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,141 @@
...
+}
+while getopts ":ho:v:f:p:a:r:" arg; do
+    case $arg in
+        h)
+            usage
+            exit 1
+            ;;
+        o)
+            TARBALL=`realpath $OPTARG`
+            ;;
```
gaiksaya, line 52: Minor can ignore: `o` is quite confusing. Usually specifies the output file. Can we use something else?

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,141 @@
...
+    for index in ${!ARCHITECTURE_ARRAY[@]}
+    do
+        arch_uname=`echo ${ARCHITECTURE_ARRAY[$index]} | sed 's/x64/x86_64/g;s/arm64/aarch64/g'`
+        echo -e "\nCopying ${PRODUCT} arch ${ARCHITECTURE_ARRAY[$index]} version ${VERSION} from ${TARBALL_ARRAY[$index]}"
+        cp -v ${TARBALL_ARRAY[$index]} ${DIR}/${PRODUCT}-${arch_uname}.tgz
+    done
+    ls -l $DIR
+fi
+
+# Build multi-arch images
```
gaiksaya, line 138: Can you convert all comments into log info (echo statements)? Will be easy to troubleshoot if the script if failing somewhere. Also it lets the user know the progress of the script

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,141 @@
...
+    echo ""
+    echo "This script is used to build the OpenSearch Docker image with single architecture (x64 or arm64). It prepares the files required by the Dockerfile in a temporary directory, then builds and tags the Docker image."
+    echo "--------------------------------------------------------------------------"
+    echo "Usage: $0 [args]"
+    echo ""
+    echo "Required arguments:"
+    echo -e "-v VERSION\tSpecify the OpenSearch version number that you are building, e.g. '1.0.0' or '1.0.0-beta1'. This will be used to label the Docker image. If you do not use the '-o' option then this tool will download a public OPENSEARCH release matching this version."
+    echo -e "-f DOCKERFILE\tSpecify the dockerfile full path, e.g. dockerfile/opensearch.al2.dockerfile."
+    echo -e "-p PRODUCT\tSpecify the product, e.g. opensearch or opensearch-dashboards, make sure this is the name of your config folder and the name of your .tgz defined in dockerfile."
+    echo -e "-a ARCHITECTURE\tSpecify the multiple architecture you want to add to the multi-arch image, separate by comma, e.g. 'x64,arm64'."
```
gaiksaya, line 24: Please add checks to see if the user has entered valid args. eg: ar64 also starts running the script which should not

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,141 @@
...
+    echo -e "-r REPOSITORY\tSpecify the docker repository name in the format of '<Docker Hub RepoName>/<Docker Image Name>:<Tag Name>', due to multi-arch image either save in cache or directly upload to Docker Hub Repo, no local copies."
```
gaiksaya, line 25: If this is really required add in the help message. If not add it in optional args

release/docker/build-image-multi-arch.sh
peterzhuamazon: @camerski would love your thoughts on this, I remember you mention it is ok to state that the script can only be run within a specific directory. If that is the case I will add more instructions in README about it.

release/docker/build-image-multi-arch.sh
peterzhuamazon: What would you suggest? I dont think -t would be better.

release/docker/build-image-multi-arch.sh
peterzhuamazon: No need, there is already outputs on each section for each command, which is redundant to add all again.

release/docker/build-image-multi-arch.sh
peterzhuamazon: It wont run to end because it will be disabled later on due to wget/docker command failure.

release/docker/build-image-multi-arch.sh
peterzhuamazon: They are all required parameters to run the scripts.

release/docker/build-image-multi-arch.sh
gaiksaya: "convert them" to echo statements for logging, Its easy to when you have logs. for debugging

release/docker/build-image-multi-arch.sh
gaiksaya: Early failure is better than late failures. Else it becomes frustrating when you don't know what error you made

release/docker/build-image-multi-arch.sh
gaiksaya: `-t` works. @camerski any thoughts?

release/docker/build-image-multi-arch.sh
peterzhuamazon: This is a good enhancement but not functionality breaking, and it has outputs from all existing commands that runs in the script. I would argue it already have all the things we need for debugging, even cp command it using -v param to print out.

release/docker/build-image-multi-arch.sh
peterzhuamazon: It requires consistent editing later if we need to add more arch.
I really dont see this as a issue as it only complicated the script.
It is good to add if the script never has a safe guard, but I doubt it is really needed because in your logic even version input needs to be checked, product name as well, then repository validation, etc.

camerski: I agree with Sayali. Fail fast.

release/docker/build-image-multi-arch.sh
camerski: Yeah, `-t` seems more natural.

release/docker/build-image-multi-arch.sh
camerski: Agree with Sayali in general: human-readable log messages are better than echoing commands. In this case, the message says "Downloading ${PRODUCT} arch ${ARCHITECTURE_ARRAY[$index]} version ${VERSION} from ${URL}", which is human-readable enough.

release/docker/build-image-multi-arch.sh
gaiksaya: The reason being I didn't specify the REPOSITORY and the script still ran. If it is hard core required please add here: https://github.com/opensearch-project/opensearch-build/pull/98/files#diff-b5bd2c1d433b43e933753221e7178a5cb2916e96ece90a632ca2c782eb221128R81

release/docker/build-image-multi-arch.sh
camerski: Ideally the script should work when executed from any working directory. `realpath $0` gives you the absolute location of the script, so you can use that to navigate to the `file_management.sh` dependency.

Basically, it's better to be able to run `/path/to/script -o tarball.tgz` than requiring `script -o /anotherpath/to/tarball.tgz`, because I'm more likely to be working in /anotherpath.

In the end it doesn't matter too much, though, just so long as the script fails gracefully if it can't source `file_management.sh`. No silent failures!

release/docker/build-image-multi-arch.sh
camerski: If it's required then there should be a check that a value has been provided, same as what you're doing on line 81 for the other required arguments.

release/docker/build-image-multi-arch.sh
peterzhuamazon: The script does not fail silently it will complain the file not found, which is good for an early failure. I will update the readme to ask people run the script within specific folder.

release/docker/build-image-multi-arch.sh
peterzhuamazon: Point taken, will make the changes and commit again.

peterzhuamazon: Changed.

release/docker/build-image-multi-arch.sh
peterzhuamazon: Since it is human-readable enough according to @camerski I will leave it as it is and we can improve later with another PR.

release/docker/build-image-multi-arch.sh
peterzhuamazon: Add checks.

release/docker/build-image-multi-arch.sh
peterzhuamazon: Missed, added -r.

dblock:
What do we want to do with this?

release/docker/build-image-multi-arch.sh
peterzhuamazon: Because this is a multi-node image, you cannot save a local image due to your local only have 1 CPU arch. The buildx is using a running container to build, thus cache it. I tried the cached methods but does not work as expected, so the only option is as soon as it builds, --push to the docker registry where it can store both arch layers.

release/docker/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+* After installation, verify whether or not you have Docker Engine by running:
+  ```
+  docker build --help
+  ```
+
+* Verify if you have Docker Desktop by running:
+  ```
+  docker buildx --help
+  ```
```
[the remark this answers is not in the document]
peterzhuamazon: This only available in docker desktop, not in docker engine, thus I use this to identify whether they have the requirements to generate multi-arch.

And where did you get the information on `docker builder toolkit`?

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,153 @@
...
+fi
+
+if [ "$ARCHITECTURE" != "x64" ] || [ "$ARCHITECTURE" != "arm64" ]
+then
+    echo "We only support 'x64' and 'arm64' as architecture name for -a parameter"
+    exit 1
+fi
+
+# Warning docker desktop
+if (! docker buildx version)
```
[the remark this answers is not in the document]
peterzhuamazon: I am not sure what you mean are you talking about mark down code block?
(! docker buildx version) will run the command in a subshell and return the buildx version.
This is to ensure that you have buildx before running the build.

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,153 @@
...
+
+# Copy configs
+cp -v config/${PRODUCT}/* $DIR/
+cp -v ../../scripts/opensearch-onetime-setup.sh $DIR/
+
+# Copy TGZ
+if [ -z "$TARBALL" ]; then
+    for index in ${!ARCHITECTURE_ARRAY[@]}
+    do
+        # No tarball file specified so download one
```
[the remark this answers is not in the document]
peterzhuamazon: There is already a downloading echo on line 136:
```
        echo -e "\nDownloading ${PRODUCT} arch ${ARCHITECTURE_ARRAY[$index]} version ${VERSION} from ${URL}"
```

I tweaked the comments with better grammar.

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,152 @@
...
+echo New workspace $DIR
+echo -e "\n* Prepare docker buildx"
+docker buildx use default
+docker buildx create --name $BUILDER_NAME --use
+docker buildx inspect --bootstrap
+
+# Check buildx status
+echo -e "\n* Check buildx status"
+docker buildx ls | grep $BUILDER_NAME
+docker ps | grep $BUILDER_NAME
```
camerski, line 122: What is the expected output from these two commands? If possible, can this script fail if the output is not what we expect?

camerski (commented):
Looks good, apart from the two boolean logic errors.

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,152 @@
...
+  echo "You must specify '-v VERSION', '-f DOCKERFILE', '-p PRODUCT', '-a ARCHITECTURE', '-r REPOSITORY'"
+  usage
+  exit 1
+else
+  echo $VERSION $DOCKERFILE $PRODUCT $ARCHITECTURE
+  IFS=', ' read -r -a ARCHITECTURE_ARRAY <<< "$ARCHITECTURE"
+  IFS=', ' read -r -a TARBALL_ARRAY <<< "$TARBALL"
+fi
+
+if [ "$PRODUCT" != "opensearch" ] || [ "$PRODUCT" != "opensearch-dashboards" ]
```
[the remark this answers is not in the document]
peterzhuamazon: fixed

release/docker/build-image-multi-arch.sh
```diff
@@ -0,0 +1,152 @@
...
+then
+    echo "Enter either 'opensearch' or 'opensearch-dashboards' as product name for -p parameter"
+    exit 1
+fi
+
+if [ "$ARCHITECTURE" != "x64" ] || [ "$ARCHITECTURE" != "arm64" ]
```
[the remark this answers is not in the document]
peterzhuamazon: fixed.

release/docker/build-image-multi-arch.sh
peterzhuamazon: ` docker buildx ls | grep $BUILDER_NAME ` will show whether a builder get created with the given name.
` docker ps | grep $BUILDER_NAME ` will show whether a builder container is provisioned for building multi-arch.

Both would fail if there is nothing in the grep:
```
$ echo 123 | grep 111

$ echo $?
1
```

camerski (approved):

release/docker/README.md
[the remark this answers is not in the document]
gaiksaya: https://docs.docker.com/buildx/working-with-buildx/

gaiksaya (approved):

release/docker/README.md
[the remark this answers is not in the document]
peterzhuamazon: This is a rather old documentation the new one is combined to docker desktop:
https://docs.docker.com/desktop/multi-arch/

release/docker/build-image-multi-arch.sh
camerski: OK, got it.

Issue #122, opened 2021-08-02
[META] Test orchestration framework


setiah:
This is a meta issue for defining the test orchestration framework as part of automated test infra

### Summary of Action plan
todo

### Roles and Responsibilities 
1. Orchestration framework will provide s3 location to publish logs as part of configuration parameters for a job. It is the responsibility of each job to publish its own logs to that s3 location.

### Tasks 
Please note these P0s are defined according to prod readiness for 1.1 release

#### P0

- [x] Define syntax of manifest file and S3 location. See https://github.com/opensearch-project/opensearch-build/issues/111
- [x] #173
- [x] Create _artifact reader_ stage, that reads artifacts from s3, reads manifest file for each artifact and extracts parameters to run testing jobs.
- [x] Create _test executor_ stage, that uses those parameters to kick off multiple parallel testing jobs on provided artifact.
- [x] #176
- [x] #175
- [x] #174
- [x] #317
- [x] #283 
- [x] #384
- [x] #262
- [x] #263 
- [x] #276 
- [x] #260 
- [x] Support notification/hooks to different systems #107 
- [x] #207 

P0 - non surge
- [x] #221

#### P1
- [x] #257
- [x] #1184
- [x] #211

PR #138, opened 2021-08-04
Prototype testing workflow skeleton

camerski:
### Description
Prototype of the test workflow. This is mostly stubbed-out code, but it shows the overall flow of reading a manifest file, setting up a test cluster, running `integtest.sh` for each component, and reporting results.

You can run the tool on a demo manifest that contains the OpenSearch 1.0.0 release artifact:

`python tester.py https://artifacts.opensearch.org/builds/opensearch/1.0.0/manifest.yml`

Make sure you have `pyyaml` installed (with `pip install pyyaml` or whatever package manager you prefer to use if you don't like pip).

The goal of this PR is to get feedback on the general approach, and to highlight the subtasks we should add to https://github.com/opensearch-project/opensearch-build/issues/122 for the next few days. Assuming people agree with this approach, I suggest the following subtasks in rough order of priority:

* Implement LocalTestCluster to stand up / tear down a test cluster on the local host.
* Implement IntegTestSuite so that it can capture test results and logs.
* Implement a mechanism for storing test results, logs, and metadata.
* Add failure handling and make sure `tester.py` cleans up (especially remote test clusters) and returns a suitable exit code.
* Add `tester.py` to a Jenkins or GitHub Actions workflow that can accept a manifest URL as an argument.
* Create and implement RemoteSingleNodeCluster to set up a remote single-node cluster using CDK.
* Create and implement RemoteMultiNodeCluster.
* Create and implement PerformanceTestSuite.
* Change `tester.py` to execute integ test suites in parallel.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/122
 
### Check List

### DIFF
--- a/workflows/testing/test_workflow/git.py
+++ b/workflows/testing/test_workflow/git.py
@@ -1,0 +1,19 @@
+import subprocess
+import tempfile
+
+class GitRepository:
+    def __init__(self, url, ref):
+        self._url = url
+        self._dir = tempfile.TemporaryDirectory()
+        self.execute(f'git init')
+        self.execute(f'git remote add origin {url}')
+        self.execute(f'git fetch --depth 1 origin {ref}')
+        self.execute(f'git checkout {ref}')
+        print(f'Checked out {url}@{ref} into {self._dir.name}')
+
+    def dir(self):
+        return self._dir.name
+
+    def execute(self, command):
+        print(f'Executing "{command}" in {self._dir.name}')
+        subprocess.check_call(command, shell=True, cwd=self._dir.name)
--- a/workflows/testing/test_workflow/integ_test.py
+++ b/workflows/testing/test_workflow/integ_test.py
@@ -1,0 +1,13 @@
+import os
+
+class IntegTestSuite:
+    def __init__(self, name, repo):
+        self._name = name
+        self._repo = repo
+
+    def execute(self, cluster):
+        script = self._repo.dir() + "/integtest.sh"
+        if (os.path.exists(script)):
+            repo.execute(f'sh integtest.sh -b {cluster.endpoint()} -p {cluster.port()}')
+        else:
+            print(f'{script} does not exist. Skipping integ tests for {self._name}')
--- a/workflows/testing/test_workflow/manifest.py
+++ b/workflows/testing/test_workflow/manifest.py
@@ -1,0 +1,45 @@
+import yaml
+
+class BundleManifest:
+    @staticmethod
+    def from_file(path):
+        with open(path, 'r') as file:
+            data = yaml.safe_load(file)
+            return BundleManifest(data)
+
+    @staticmethod
+    def from_text(text):
+        data = yaml.safe_load(text)
+        return BundleManifest(data)
+
+    def __init__(self, data):
+        schema_version = str(data['schema-version'])
+        if schema_version != '1.0':
+            raise ValueError(f'Schema version must be 1.0; found {schema_version}')
+        self._build_id = data['build']['build-id']
+        self._bundle_location = data['build']['location']
+        self._components = list(map(lambda entry: Component(entry), data['components']))
+
+    def build_id(self):
+        return self._build_id
+
+    def bundle_location(self):
+        return self._bundle_location
+
+    def components(self):
+        return self._components
+
+class Component:
+    def __init__(self, data):
+        self._name = data['name']
+        self._repository = data['repository']
+        self._commit = data['commit']
+
+    def name(self):
+        return self._name
+
+    def repository_url(self):
+        return self._repository
+
+    def commit_id(self):
+        return self._commit
--- a/workflows/testing/tester.py
+++ b/workflows/testing/tester.py
@@ -1,0 +1,34 @@
+import os
+import sys
+import tempfile
+import urllib.request
+from test_workflow.manifest import BuildManifest
+from test_workflow.test_cluster import LocalTestCluster
+from test_workflow.git import GitRepository
+from test_workflow.integ_test import IntegTestSuite
+
+if (len(sys.argv) < 2):
+    print("Usage: tester.py /path/to/manifest")
+    exit(1)
+
+with tempfile.TemporaryDirectory() as work_dir:
+    os.chdir(work_dir)
+
+    # Download the manifest
+    with urllib.request.urlopen(sys.argv[1]) as response:
+        text = response.read()
+        manifest = BuildManifest.from_text(text)
+
+    # Spin up a test cluster
+    cluster = LocalTestCluster(manifest.bundle_location)
+
+    # For each component, check out the git repo and run `integtest.sh`
+    for component in manifest.components():
+        print(component.name())
+        repo = GitRepository(component.repository_url(), component.commit_id())
+        test_suite = IntegTestSuite(component.name(), repo)
+        test_suite.execute(cluster)
+
+    cluster.destroy()
+
+    # TODO: Store test results, send notification.
--- a/workflows/testing/test_workflow/test_cluster.py
+++ b/workflows/testing/test_workflow/test_cluster.py
@@ -1,0 +1,12 @@
+class LocalTestCluster:
+    def __init__(self, bundle_location):
+        pass
+
+    def endpoint(self):
+        return 'localhost'
+
+    def port(self):
+        return 9200
+
+    def destroy(self):
+        pass


dblock:
Meta question: do we have other workflows/tools that are Python. I know the signer is. Just wondering whether we are trying to standardize this in any way?

workflows/testing/test_workflow/integ_test.py
```diff
@@ -0,0 +1,13 @@
...
+
+class IntegTestSuite:
+    def __init__(self, name, repo):
+        self._name = name
+        self._repo = repo
+
+    def execute(self, cluster):
+        script = self._repo.dir() + "/integtest.sh"
+        if (os.path.exists(script)):
+            repo.execute(f'sh integtest.sh -b {cluster.endpoint()} -e {cluster.port()}')
```
peterzhuamazon, line 11: -e?

camerski:
> Meta question: do we have other workflows/tools that are Python. I know the signer is. Just wondering whether we are trying to standardize this in any way?

At this point in time we have not had a discussion about standardization. This PR should not be construed as an endorsement of Python :)

setiah:
Thanks @camerski for proposing this. This is definitely one way to write orchestration layer but I am leaning towards a Jenkins pipeline based workflow due to a couple of reasons. It seems like in this approach, we might need to reinvent the wheel for a couple of great features that are available out of the box w/ Jenkins, for instance - 
1. Need to write our own state machine logic to run workflows.
2. Need to take dependency on DDB (or any other external service) to persist intermediate states. We will also need to add logic for DDB interactions. 
3. No visualization for workflows available
4. Need to write job parallelization logic inside this python workflow to run independent tests in parallel.
5. Not clear yet how we breakdown tests into separate jobs. If we choose to run everything as one job, it makes things too monolithic. This is a problem because when there is a job failure someone will need to actually dig into logs to check was it related to integ tests/ perf tests/ bwc or any other workflow that we run. A clear separation of jobs would help to better understand general characteristics such as which test suite is more flaky and needs fixing.

Jenkins pipeline seem to offer all of this and more out of the box. It seems to have easy way to pass params between stages.

workflows/testing/test_workflow/git.py
```diff
@@ -0,0 +1,20 @@
+import os
+import tempfile
+
+class GitRepository:
+    def __init__(self, url, ref):
+        self._url = url
+        self._dir = tempfile.TemporaryDirectory()
+        self.execute(f'git clone {url} {self._dir.name}')
```
peternied, line 8: Do we need to worry about non-public repositories needing alternative credentials?

workflows/testing/test_workflow/integ_test.py
```diff
@@ -0,0 +1,13 @@
...
+        else:
+            print(f'{script} does not exist. Skipping integ tests for {self._name}')
```
peternied, line 13: This seems like an exceptional state, shouldn't all repositories include integration tests?  I'd recommend throwing with a nasty error so they have to be dealt with

workflows/testing/test_workflow/manifest.py
```diff
@@ -0,0 +1,47 @@
+import yaml
+
+class BuildManifest:
+    @staticmethod
+    def from_file(path):
+        with open(path, 'r') as file:
+            hash = yaml.safe_load(file)
```
peternied, line 7: hash? Not sure what this naming convention means?

workflows/testing/test_workflow/manifest.py
```diff
@@ -0,0 +1,47 @@
...
+    def build_id(self):
+        return self._build_id
+
+    def bundle_location(self):
+        return self._bundle_location
+
+    def components(self):
+        return self._components
+
+class Component:
```
peternied, line 34: Save some lines with `@dataclass` or something similar?

workflows/testing/tester.py
```diff
@@ -0,0 +1,36 @@
...
+    with urllib.request.urlopen(sys.argv[1]) as response:
+        text = response.read()
+        manifest = BuildManifest.from_text(text)
+
+    # Spin up a test cluster
+    cluster = LocalTestCluster(manifest.bundle_location)
+
+    # For each component, check out the git repo and run `integtest.sh`
+    for component in manifest.components():
+        if (component.name() == 'opensearch'):
```
peternied, line 27: Why add this special case?

camerski:
All good points, @setiah. Let me explain why I approached it this way.

> 1. Need to write our own state machine logic to run workflows.

Correct. Definitely a downside of this approach.

> 2. Need to take dependency on DDB (or any other external service) to persist intermediate states. We will also need to add logic for DDB interactions.

Correct, but not really a big deal. DDB is pretty simple.

> 3. No visualization for workflows available

Not sure how much this matters.

> 4. Need to write job parallelization logic inside this python workflow to run independent tests in parallel.

This is not difficult.

> 5. Not clear yet how we breakdown tests into separate jobs. If we choose to run everything as one job, it makes things too monolithic. This is a problem because when there is a job failure someone will need to actually dig into logs to check was it related to integ tests/ perf tests/ bwc or any other workflow that we run. A clear separation of jobs would help to better understand general characteristics such as which test suite is more flaky and needs fixing.

This is where the fun starts :)

Jenkins pipelines are static - there is not (or, at least, I couldn't find) a way to dynamically change the number of stages at runtime. This is important for test runs: If there are `n` components in the bundle then we want to run `n` test suites, if the next bundle has `n + 1` components then the pipeline needs to be able to add a parallel step for the `n + 1th` component, without needing a code change for the pipeline.

That basically means that we need to build our own parallelism - the best the pipeline can do is have a step that executes `run-all-integ-tests.sh`. The other steps in the workflow (download and parse the manifest, check out all the repos) are pretty trivial and I'm not concerned about being able to restart a failed workflow for those pieces.

Test cluster lifecycle is also a bit tricky. I think we could manage that state using the Jenkins pipeline data storage model but I haven't prototyped that so it's unclear to me how well that will work, and what the sharp edges will be. I suspect we will need to store things like CloudFormation stack ARNs in external storage so that we have an easy way to terminate them if the Jenkins pipeline explodes.

All that said, I'm not opposed to doing it all in Jenkins. Let's prototype both and see where we land.

workflows/testing/tester.py
camerski: Because we don't have integtest.sh for this repo. It's a big repo, so checking out a ton of code that we won't use is a substantial waste of time.

workflows/testing/test_workflow/manifest.py
camerski: My python-fu is weak :) Thanks!

workflows/testing/test_workflow/manifest.py
camerski: It means the author of this code comes from a Ruby background. What's the convention in Python?

workflows/testing/test_workflow/integ_test.py
camerski: Eventually they should all include integration tests. We're not there yet, however.

workflows/testing/test_workflow/git.py
camerski: I'm happy to say we don't support non-public repositories.

workflows/testing/test_workflow/integ_test.py
peternied: I tend to lend on not included in components into bundling until they meet a basic compliance level.  Forces earlier investment via components and makes our 'management' of onboarding new components simpler.

Otherwise we have to scan the logs for this error message that means we have an untested component.

/2cents

workflows/testing/test_workflow/git.py
```diff
@@ -0,0 +1,20 @@
...
+        self.execute(f'git checkout {ref}')
+        print(f'Checked out {url}@{ref} into {self._dir.name}')
+
+    def dir(self):
+        return self._dir.name
+
+    def execute(self, command):
```
dblock, line 15: This doesn't fail when the command fails.

```python
import os
import tempfile
import subprocess

class GitRepository:
    def __init__(self, url, ref):
        self._url = url
        self._dir = tempfile.TemporaryDirectory()
        self.execute(f'git clone --branch {ref} --depth 1 {url} {self._dir.name}')
        print(f'Checked out {url}@{ref} into {self._dir.name}')

    def dir(self):
        return self._dir.name

    def execute(self, command):
        current = os.getcwd()
        print(f'Executing "{command}" in {self._dir.name}')
        os.chdir(self._dir.name)
        subprocess.check_call(command, shell = True)
        os.chdir(current)
```

workflows/testing/test_workflow/integ_test.py
dblock: This doesn't fail when integ tests fail.

workflows/testing/test_workflow/manifest.py
peternied: ohhh... I have no idea - hash to me means md5 or sha1...  Blob or data come to mind, maybe whatever yaml is called, if it was json i'd call it json.

workflows/testing/tester.py
peternied: > checking out a ton of code that we won't use

Hmm... is this worth doing a sparse checkout to make it faster? What kind of performance should we expect?

dblock:
I think for git checkout you can save yourself a lot of time and only check out depth 1.

```python
import os
import tempfile
import subprocess

class GitRepository:
    def __init__(self, url, ref):
        self._url = url
        self._dir = tempfile.TemporaryDirectory()
        self.execute(f'git init')
        self.execute(f'git remote add origin {url}')
        self.execute(f'git fetch --depth 1 origin {ref}')
        self.execute(f'git checkout FETCH_HEAD')
        print(f'Checked out {url}@{ref} into {self._dir.name}')

    def dir(self):
        return self._dir.name

    def execute(self, command):
        current = os.getcwd()
        print(f'Executing "{command}" in {self._dir.name}')
        os.chdir(self._dir.name)
        try:
            subprocess.check_call(command, shell = True)
        finally:
            os.chdir(current)
```

workflows/testing/test_workflow/integ_test.py
camerski: Yup, it's a prototype. Fault tolerance will come if/when we turn this into a production system.

workflows/testing/test_workflow/integ_test.py
camerski: Directionally, I agree. Since this is a prototype I'm happy to worry about that later.

camerski:
> I think for git checkout you can save yourself a lot of time and only check out depth 1. Here's a full class with error handling.

Oh, yeah, depth 1 is *way* faster...thanks!

workflows/testing/test_workflow/git.py
camerski: I didn't know about the `subprocess` library...that looks much better than `os.system`. Again, my python-fu is weak, so thanks for the pointer :)

workflows/testing/test_workflow/manifest.py
dblock: Should just be `data`, e.g. https://github.com/opensearch-project/opensearch-build/blob/c9885c0f7fe4fb6808cf3b34364ef91efcdeb825/tools/bundle-build/lib/manifest.py

workflows/testing/test_workflow/git.py
dblock: Beyond Python, make sure every method called fails loudly.

camerski: Yes, of course. Remember that "the goal of this PR is to get feedback on the general approach, and to highlight the subtasks we should add to #122 for the next few days". Fault tolerance is not so important for prototype code.

workflows/testing/tester.py
camerski: Changed to use `depth 1`.

workflows/testing/test_workflow/manifest.py
camerski: Done

workflows/testing/test_workflow/manifest.py
dblock: Here's a similar refactor: https://github.com/opensearch-project/opensearch-build/pull/150

dblock:
Let's also see if we can reuse the common parts, such as git and manifest implementations?

camerski:
> Let's also see if we can reuse the common parts, such as git and manifest implementations?

Yeah, we should do a big refactoring exercise once we have the overall workflow more-or-less ready.

Once this PR is approved I can get started on some of the refactoring and extracting of common functionality.

dblock:
> > Let's also see if we can reuse the common parts, such as git and manifest implementations?
> 
> Yeah, we should do a big refactoring exercise once we have the overall workflow more-or-less ready.
> 
> Once this PR is approved I can get started on some of the refactoring and extracting of common functionality.

You don't have to wait. Those .py classes that were written as part of #150 and #145 are directly reusable. Save yourself time rewriting/copying code.

workflows/testing/test_workflow/manifest.py
```diff
@@ -0,0 +1,45 @@
+import yaml
+
+class BundleManifest:
```
setiah, line 3: We can separate this out as a separate tool to read/write manifest file

camerski:
> > > Let's also see if we can reuse the common parts, such as git and manifest implementations?
> > 
> > 
> > Yeah, we should do a big refactoring exercise once we have the overall workflow more-or-less ready.
> > Once this PR is approved I can get started on some of the refactoring and extracting of common functionality.
> 
> You don't have to wait. Those .py classes that were written as part of #150 and #145 are directly reusable. Save yourself time rewriting/copying code.

Those changes in #150 and #145 also need to be refactored (calling a library `lib` is not a great idea). I would rather get this already-written code checked in so we can move fast on implementing the TestCluster and TestSuite classes, and refactor later.

workflows/testing/test_workflow/manifest.py
camerski: Yup, there will be a big refactoring exercise later to unify this code with the build workflow, and pull out common components. Code re-use is not top priority right now; working prototype is the goal.

peternied (changes requested):
I am happy to leave many issues open associated with this PR if we move it into a `prototype` folder so that we are sure what support level is associated with this code

camerski:
> I am happy to leave many issues open associated with this PR if we move it into a `prototype` folder so that we are sure what support level is associated with this code

Sure, I can do that.

setiah:
> Jenkins pipelines are static - there is not (or, at least, I couldn't find) a way to dynamically change the number of stages at runtime. This is important for test runs: If there are `n` components in the bundle then we want to run `n` test suites, if the next bundle has `n + 1` components then the pipeline needs to be able to add a parallel step for the `n + 1th` component, without needing a code change for the pipeline.
> 
> That basically means that we need to build our own parallelism - the best the pipeline can do is have a step that executes `run-all-integ-tests.sh`. The other steps in the workflow (download and parse the manifest, check out all the repos) are pretty trivial and I'm not concerned about being able to restart a failed workflow for those pieces.
 
Actually we can do parallelization at multiple levels. The first level is orchestration layer where Jenkins jobs can be made to run in parallel via pipeline. Orchestration framework can kick off additional parallel jobs for `n+1th` component. 
Another layer could be python tool itself, where the Jenkins orchestration can be made agnostic and the tool can have logic to parallelize multiple tests based on the bundle components. The possibility exists. We can come up with exact structure later when we are scaling test infra to support multiple bundles.

dblock:
> > I am happy to leave many issues open associated with this PR if we move it into a `prototype` folder so that we are sure what support level is associated with this code
> 
> Sure, I can do that.

I think this is unnecessarily moving stuff around. We are building a system that needs to work, AFAIK the criteria to merge this code is that it completes an end-to-end test successfully, code improvements and refactorings are should have's.

dblock:
> > Jenkins pipelines are static - there is not (or, at least, I couldn't find) a way to dynamically change the number of stages at runtime. This is important for test runs: If there are `n` components in the bundle then we want to run `n` test suites, if the next bundle has `n + 1` components then the pipeline needs to be able to add a parallel step for the `n + 1th` component, without needing a code change for the pipeline.
> > That basically means that we need to build our own parallelism - the best the pipeline can do is have a step that executes `run-all-integ-tests.sh`. The other steps in the workflow (download and parse the manifest, check out all the repos) are pretty trivial and I'm not concerned about being able to restart a failed workflow for those pieces.
> 
> Actually we can do parallelization at multiple levels. The first level is orchestration layer where Jenkins jobs can be made to run in parallel via pipeline. Orchestration framework can kick off additional parallel jobs for `n+1th` component.
> Another layer could be python tool itself, where the Jenkins orchestration can be made agnostic and the tool can have logic to parallelize multiple tests based on the bundle components. The possibility exists. We can come up with exact structure later when we are scaling test infra to support multiple bundles.

Is it a crazy idea to generate a Jenkins pipeline from a manifest?

camerski:
> code improvements and refactorings are should have's and can be deferred, but not forgotten. That said, we ought to keep ourselves to a high bar for code quality during development, and not defer what can be done right away.

@dblock Given that statement, what exactly in this PR do you consider to be a must-have before approving, and what do you think can be deferred?

mch2 (approved):

peternied (approved):

dblock:
> > code improvements and refactorings are should have's and can be deferred, but not forgotten. That said, we ought to keep ourselves to a high bar for code quality during development, and not defer what can be done right away.
> 
> @dblock Given that statement, what exactly in this PR do you consider to be a must-have before approving, and what do you think can be deferred?

You got two approvals so go for it.

PR #141, opened 2021-08-05
Add prototype for a component build system

### LINKED ISSUE #133: Script to build each component from manifest and assemble bundle.
This script should read a manifest and output all artifacts ready for upload.
Example.
/build/opensearch-dashboards-min-1.0.0-SNAPSHOT-linux-arm64-198.tar.gz <- min bundle
/build/opensearch-dashboards-1.0.0-SNAPSHOT-linux-arm64-198.tar.gz. <-- full bundle
/build/opensearch-sql-1.0.0.0-rc1.zip
/build/org/opensearch... <- maven artifacts

Input: to the script should be a manifest file location - format defined here #111 
Output: all required artifacts are written to ./build

- [x] Clone each component repository defined in the manifest
- [x] Build each component. This includes plugin zips and maven artifacts and place under a new folder with build id.  Note: We need to know if the version of a particular component is already published to maven central.  If this is the case we do not need to rebuild and include that artifact.
- [ ] Assemble the bundle itself and add to the /build directory. This is dependent on being able to write manifests - #134 

To make assembling maven artifacts easier, each repo can be published to maven local and copied from that location into /build.  All artifacts will be under org/opensearch.

mch2:
This is a first attempt at a central top-down build.

Given a config file for a particular bundle version, it clones each repository
and builds it with a given build script.

### Description
```
Usage: ruby build.rb ./configs/opensearch-1.0.0.yml
```

Component builds:
This PR includes a standard build script for projects using ./gradlew assemble and repo specific scripts for those that don't or have slightly different needs.
This PR is to get this started, the idea is that each repo would include a copy of this script under build/build.sh that tells us how to build their component.

Currently the order of components declared in Config matters because multiple repos depend on engine & common-utils and we do not yet have 1.0 in maven central.  These will be compiled first and the OpenSearch version passed along as a param to build.sh scripts.  The intention here is that repos will not depend on any unreleased version of a dependency. If they need to depend on the bleeding edge & compile against the version of those dependencies being built in this bundle build, they can do so.

Maven dependencies:
Each script can place their dependencies in <repo-name>-artifacts/maven to be picked up and published on their behalf.

### Issues Resolved
#133 
 
### Check List

### DIFF
--- a/bundle-build/config.rb
+++ b/bundle-build/config.rb
@@ -1,0 +1,40 @@
+##
+# This class is used to parse config used during a bundle build
+class Config
+    attr_reader :type, :version, :components
+    def initialize(hash)
+        @type = hash['type']
+        @version = hash['version']
+        @components = hash['components'].map{|component| Component.new(component)}
+    end
+
+    def self.getArchitecture()
+      arch=`uname -m`.strip
+      return case arch
+      when 'x86_64'
+          :'x64'
+      when 'aarch64', 'arm64'
+          :'arm64'
+      else
+          puts "Building on unsupported Architecture #{arch}, only x64 and arm64 are supported"
+          exit
+      end
+    end
+end
+
+class Component
+    attr_reader :name, :repository, :ref
+    def initialize(hash)
+        @name = hash['name']
+        @repository = hash['repository']
+        @ref = hash['ref']
+    end
+
+    def checkout()
+        `git clone #{repository} #{name}`
+        currentDir=Dir.pwd
+        Dir.chdir(name)
+        `git checkout #{ref}`
+        Dir.chdir(currentDir)
+    end
+end
--- a/bundle-build/configs/opensearch-1.0.0.yml
+++ b/bundle-build/configs/opensearch-1.0.0.yml
@@ -1,0 +1,48 @@
+type: opensearch
+version: 1.0.0
+components:
+  - name: OpenSearch
+    repository: https://github.com/opensearch-project/OpenSearch.git
+    ref: 1.0
+  - name: common-utils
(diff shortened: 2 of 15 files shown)


bundle-build/configs/opensearch-1.0.0.yml
```diff
@@ -0,0 +1,48 @@
+type: opensearch
```
dblock, line 1: name?

bundle-build/config.rb
```diff
@@ -0,0 +1,40 @@
...
+        @ref = hash['ref']
+    end
+
+    def checkout()
+        `git clone #{repository} #{name}`
+        currentDir=Dir.pwd
+        Dir.chdir(name)
+        `git checkout #{ref}`
+        Dir.chdir(currentDir)
+    end
```
dblock, line 39: There's no error handling here. All these commands fail silently, back-tick execution doesn't raise errors on negative exit code AFAIK.

dblock (changes requested):
- You need a Gemfile for any tool/script so we can lock dependencies and make things predictable and then execute anything with `bundle exec`.
- This is quickly becoming a mess. Replace the build.rb implementation to use gli, for example https://github.com/dblock/fue/blob/master/bin/fue. It will cut down the code a lot. You can take fue as a complete example.
- For error handling on command execution see https://github.com/dblock/fue/blob/master/lib/fue/shell.rb#L6 for an example.
- I love Ruby, but I wonder whether we're doing ourselves a disservice writing tools in many different languages? We have some scripts like signer written in Python.

dblock:
Let's close in favor of https://github.com/opensearch-project/opensearch-build/pull/145

Issue #111, opened 2021-07-28
Notifications/Hooks to different systems.


saratvemulapalli:

camerski:
The test system will need a way to get metadata about bundle builds.

First stab at a proposal:

1. Bundle builds have a unique ID and are stored in S3, under a path like `/path/to/bundles/<buildnumber>/<files>`.
2. `<files>` contains all the components of the bundle (jars, plugin zips, etc) and a manifest file.
3. The build system stores the bundle build in S3, then sends an SNS notification to the test system. The SNS message contains the bundle build ID and a link to the manifest URL. A simple JSON document would suffice: `{"buildid":<number>, "manifest":"path"}`
4. The manifest.yml looks like this:

```
schema-version: 1.0
product:
  type: opensearch | opensearch-min | dashboards | dashboards-min
  version: major.minor.patch
  platform: x64 | arm64
components:
  - name: string
    repository: github url
    commit: hash
    branch (optional): branch or tag name
  ...
build:
  snapshot: true | false
```

The `schema-version` field is to allow for evolving the manifest format over time.

The `product` element provides basic information about what this bundle represents.

The test system can inspect the `product` and `build` elements to determine which bundle test suites to run. For example, a snapshot opensearch-min bundle will have a fairly small set of tests (no plugins, no need for a full set of expensive performance tests for a snapshot build), whereas an opensearch non-snapshot bundle with 15 components will require a much larger test suite, including tests for each component.

The `snapshot` field could be included under the `product` element, but I feel that product information and build information are distinct attributes, so I prefer to split out the build information into its own structure.

Component information is included both for reproducibility and to inform the test system which packages need to be checked out in order to run per-component integration tests.

When the test system receives a notification it can load the manifest and start an appropriate test pipeline for the bundle.

saratvemulapalli:
I like this.
Couple of questions (Trying to understand):
1. How are dependencies listed? Would they be another product?
2. If yes to 1. How do we know which build-id was used to build the bundle for a product. i.e Probably the inner question is, is the `build-id` the id for bundle or for each product artifact the bundle was built out of?

camerski:
By dependencies, do you mean the individual plugins? Those are captured in the `components` section, one entry per plugin plus one entry for the core opensearch component. Plugins are not standalone products.

saratvemulapalli:
[...]

PR #145, opened 2021-08-05
Component-build system

dblock:
### Description

Rewrites https://github.com/opensearch-project/opensearch-build/pull/141 in Python and moves some things around.
 
### Check List


manifests/opensearch-bundle.1.0.yml
```diff
@@ -0,0 +1,53 @@
+---
+manifest:
```
mch2, line 2: Not critical for the first iteration of this, but we should standardize this with #111 which suggests schema-version: 1.0

I'm thinking the input manifest and the copy included in the bundle should follow the same schema with the input being a subset of a final manifest.

tools/bundle-build/lib/git.py
```diff
@@ -0,0 +1,25 @@
...
+        return self._dir.name
+
+    def execute(self, command):
+        current = os.getcwd()
+        print(f'Executing "{command}" in {self._dir.name}')
+        os.chdir(self._dir.name)
+        try:
+            subprocess.check_call(command, shell = True)
+        finally:
+            os.chdir(current)
```
camerski, line 25: Recommendation: Use the `cwd` argument to `check_call` rather than manually calling `chdir`.

manifests/opensearch-bundle.1.0.yml
dblock: Done

tools/bundle-build/lib/manifest.py
```diff
@@ -0,0 +1,50 @@
...
+        return self._name
+
+    def arch(self):
+        arch = subprocess.check_output(['uname', '-m']).decode().strip()
+        if arch == 'x86_64':
+            return 'x64'
+        elif arch == 'aarch64' or arch == 'arm64':
+            return 'arm64'
+        else:
+            raise ValueError(f'Unsupported architecture: ' + arch)
```
camerski: This function doesn't belong here. This class represents the data recorded in the manifest; it should not be computing anything.

Make `arch` a read-only attribute in the manifest and put the `uname` check into `build.py`, or, alternatively, when `build.py` determines the CPU architecture then it could update the manifest. Either way, this logic should be in `build.py` or in a class that handles CPU architecture.

tools/bundle-build/build.py
```diff
@@ -0,0 +1,22 @@
...
+    manifest = BuildManifest.from_file(sys.argv[1])
+    build = manifest.build()
+    print(f'Building {build.name()} ({build.arch()})')
+
+    os.chdir(work_dir)
+
+    for component in manifest.components():
+        print(f'=== Building {component.name()} ...')
+        component.checkout()
+        component.build(build.version(), build.arch())
```
mch2, line 22: If i'm reading this correctly, its building in the temp dir but not copying component artifacts into a common output directory.  Each build.sh is currently placing artifacts in <repo-dir>/<component-name>-artifacts.

This should put all components into a specified output directory

mch2: Also on that note I'd like the output to be something like this.

```
./build
   <min-artifact>
   ./plugins
       <plugin-zips>
   ./libs
```

This makes bundle assembly easier to write.

Not critical for this PR, I can follow up with that change

mch2: I will update this to pass an output dir reference to each build.sh so that it can place those components directly where they need to go.

tools/bundle-build/README.md
```diff
@@ -0,0 +1,15 @@
+## Bundle Build
+
+Builds the OpenSearch bundle.
+
+The OpenSearch repo is built first, followed by common-utils. These dependencies are published to maven local, and subsequent project builds pick up those.
```
peternied, line 5: Seems like this is based on the order in the manifest rather than a special case, is that correct?

scripts/bundle-build/standard-gradle-build/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+outputDir=$(basename $PWD)-artifacts
+mkdir $outputDir
+./gradlew assemble --no-daemon --refresh-dependencies -Dbuild.snapshot=false -DskipTests=true -Dopensearch.version=$1
+
+zipPath=$(find . -path \*build/distributions/*.zip)
```
peternied, line 7: Error handling if there are no zips?

scripts/bundle-build/standard-gradle-build/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+outputDir=$(basename $PWD)-artifacts
+mkdir $outputDir
+./gradlew assemble --no-daemon --refresh-dependencies -Dbuild.snapshot=false -DskipTests=true -Dopensearch.version=$1
```
peternied, line 5: What happens if gradlew build fails? (This scenario isn't isolated only to the default build script)

scripts/bundle-build/components/notifications/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+outputDir=$(basename $PWD)-artifacts
+mkdir -p $outputDir/maven
+cd notifications
+
+./gradlew publishToMavenLocal -PexcludeTests="**/SesChannelIT*" -Dopensearch.version=$1
```
peternied, line 7: Hmm, I'm slightly conerned about repo specific logic being in this common location that could be overlooked when considering testing / validating build configuration.  As this how bundles are built, thoughts on a way to make sure the typical normal build is identical?

peternied:
Great teamwork @dblock and @mch2 - making the dream work!  I've added some overarching design considerations, happy to approve with additional commits or responses on my comments.

scripts/bundle-build/components/notifications/build.sh
mch2: This is a build script that is specific to the notifications repo.  It should eventually get moved to that repository but it is here now to get this off the ground.

scripts/bundle-build/standard-gradle-build/build.sh
mch2: If a particular component fails to build, the entire bundle build should exit.

scripts/bundle-build/standard-gradle-build/build.sh
mch2: Maybe I should have named this file path differently,  This script will only be invoked for components that do not have a special case and product a zip at a build/distributions location.  If the event something is updated and there is no zip, then it should fail.

tools/bundle-build/README.md
mch2: Correct, it will build in the order defined in the manifest.  I called out these two specifically because almost all components rely on these being built and available in order to compile.

tools/bundle-build/build.py
dblock: With https://github.com/opensearch-project/opensearch-build/pull/145/commits/b97225e3921bf5d3dda0e97ebf89e3c4745fea3a you'll get a successful build up to performance-analyzer and this output in `artifacts`:

```
drwxrwxr-x  6 ubuntu ubuntu      4096 Aug  5 23:03 maven
drwxrwxr-x  2 ubuntu ubuntu      4096 Aug  5 23:04 plugins
-rw-rw-r--  1 ubuntu ubuntu 230910212 Aug  5 23:02 opensearch-min-1.0.0-linux-x64.tar.gz
```

... `plugins`:

```
-rw-rw-r-- 1 ubuntu ubuntu 10802657 Aug  5 23:04 opensearch-alerting-1.0.0.0.zip
-rw-rw-r-- 1 ubuntu ubuntu   972506 Aug  5 23:03 opensearch-job-scheduler-1.0.0.0.zip
-rw-rw-r-- 1 ubuntu ubuntu    19084 Aug  5 23:03 opensearch-job-scheduler-sample-extension-1.0.0.0.zip
-rw-rw-r-- 1 ubuntu ubuntu 39048243 Aug  5 23:04 opensearch-security-1.0.0.0.zip
-rw-rw-r-- 1 ubuntu ubuntu 17419492 Aug  5 23:03 opensearch-sql-1.0.0.0.zip
```

... `maven`:

```
drwxrwxr-x  4 ubuntu ubuntu 4096 Aug  5 23:02 common-utils
drwxrwxr-x 34 ubuntu ubuntu 4096 Aug  5 23:02 opensearch
drwxrwxr-x  4 ubuntu ubuntu 4096 Aug  5 23:03 opensearch-job-scheduler
drwxrwxr-x  4 ubuntu ubuntu 4096 Aug  5 23:03 opensearch-job-scheduler-spi
```

tools/bundle-build/lib/manifest.py
dblock: I moved arch into the newly added `BuildOutput`.

scripts/bundle-build/standard-gradle-build/build.sh
dblock: This will abort the shell script because `set -e` and the build will fail, https://intoli.com/blog/exit-on-errors-in-bash-scripts/

scripts/bundle-build/components/notifications/build.sh
dblock: We will be moving these into the individual repos at the very least.

mch2 (approved):

dblock:
Need a second approving review, maybe @peternied ?

dblock:
This is ready to merge.

I will tackle artifacts tomorrow unless someone beats me to it. The input manifest will contain exact things to pickup from the build of each component, like this:

```
components:
  - name: OpenSearch
    ref: 1.0
    artifacts:
      - opensearch-min-*.tar.gz 
      - maven/*
```

And the output manifest will have the full name:

```
components:
  - name: OpenSearch
    ref: 1.0
    sha: 5c0978d616273a8e3cc5b17538a734a044b881e5
    artifacts:
      - opensearch-min-1.0.0-linux-x64.tar.gz
```

And instead of `cp -r artifacts/...` from build output, will only copy what's declared.

scripts/bundle-build/standard-gradle-build/build.sh
dblock: I'll replace this to be picking up files explicitly.

camerski (approved):
Looks pretty good. There are some unused variables and missing `set -e`s in the various `build.sh` scripts, but those can be fixed when we move the scripts into the individual component repositories.

mch2:
> This is ready to merge.
> 
> I will tackle artifacts tomorrow unless someone beats me to it. The input manifest will contain exact things to pickup from the build of each component, like this:
> 
> ```
> components:
>   - name: OpenSearch
>     repository: https://github.com/opensearch-project/OpenSearch.git
>     ref: 1.0
>     artifacts:
>       - opensearch-min-*.tar.gz 
>       - maven/*
> ```
> 
> And the output manifest will have the full name:
> 
> ```
> components:
>   - name: OpenSearch
>     repository: https://github.com/opensearch-project/OpenSearch.git
>     ref: 1.0
>     sha: 5c0978d616273a8e3cc5b17538a734a044b881e5
>     artifacts:
>       - opensearch-min-1.0.0-linux-x64.tar.gz
> ```
> 
> And instead of `cp -r artifacts/...` from build output, will only copy what's declared.

The output of this file is going to be the input to our bundle build where we need to know if a component provides a plugin that needs to be installed.  I'd rather rely on plugins only copying into a set of folders what should be installed vs filtering with a wildcard/regex.  I think that the input files will get a bit crazy if thats the case. I suggest we follow a standard folder structure for build.sh scripts to place their artifacts to be copied.

/plugins
/bundles <- Only really used by OpenSearch repo
/maven
/libs <- used by knn off the top of my head, used for any lib that should be installed in the bundle

All the artifacts can be grouped and listed under these categories in the output, and we can easily parse that file to know what should be installed in a bundle.

input:
 ```
 components:
   - name: OpenSearch
     ref: 1.0
   - name: job-scheduler
    repository: https://github.com/opensearch-project/job-scheduler.git
    ref: 1.0.0.0
 ```

output:
 ```
 components:
   - name: OpenSearch
     ref: 1.0
     artifacts:
       - bundle: 
          -  opensearch-min-1.0.0-linux-x64.tar.gz
       - maven:
          - /path/to/file.jar
   - name: job-scheduler
    repository: https://github.com/opensearch-project/job-scheduler.git
    ref: 1.0.0.0
    artifacts:
      - plugin: opensearch-job-scheduler-1.0.0.0.zip
```

scripts/bundle-build/standard-gradle-build/build.sh
peternied: Forgive my ignorance, but where is the `set -e` invoked that will impact this script execution?  I see it was set in opensearch.sh, does that cascade into the other script executions as well?

dblock: It has been added as part of #150 :)

Issue #143, opened 2021-08-05
Create role for upload bundle artifacts to S3 bucket


peternied:

mch2:
Closing this, we have cloudformation code that sets this up here: https://github.com/opensearch-project/opensearch-build/tree/main/bundle-workflow/cloudformation

PR #146, opened 2021-08-06
Add CloudFormation template for creating AWS resources needed by the build-bundle-test workflows

camerski:
### Description
CloudFormation stack for managing the S3 bucket used by the build-bundle-test workflows and IAM users for each of the three stages.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/143
 
### Testing
Manually verified that the bucket and IAM users are created. Tested in the IAM policy simulator to verify that the users have the correct access to folders within the bucket.

### Check List

### DIFF
--- a/cloudformation/build-bundle-test-workflow.yml
+++ b/cloudformation/build-bundle-test-workflow.yml
@@ -1,0 +1,101 @@
+Description: Template for setting up resources for the build-bundle-test workflows
+Resources:
+  BuildBucket:
+    Type: "AWS::S3::Bucket"
+  WorkflowUser:
+    Type: "AWS::IAM::User"
+    Properties:
+      UserName: buildworkflow
+      Policies:
+        - PolicyName: AssumeRole
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              Effect: Allow
+              Action: "sts:AssumeRole"
+              Resource:
+                - !Sub "arn:aws:iam::${AWS::AccountId}:role/opensearch-build"
+                - !Sub "arn:aws:iam::${AWS::AccountId}:role/opensearch-bundle"
+                - !Sub "arn:aws:iam::${AWS::AccountId}:role/opensearch-test"
+  BuildRole:
+    Type: "AWS::IAM::Role"
+    Properties:
+      RoleName: opensearch-build
+      AssumeRolePolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Principal:
+              AWS:
+                - !GetAtt WorkflowUser.Arn
+            Action:
+              - 'sts:AssumeRole'
+      Policies:
+        - PolicyName: S3_Build_Write_Access
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              Effect: Allow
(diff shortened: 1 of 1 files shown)


cloudformation/build-bundle-test-workflow.yml
```diff
@@ -0,0 +1,59 @@
+Description: Template for setting up resources for the build-bundle-test workflows
+Resources:
+  BuildBucket:
+    Type: "AWS::S3::Bucket"
+  BuildUser:
+    Type: "AWS::IAM::User"
```
peternied, line 6: Why are we creating users?  Won't the scripts be run inside of the Jenkins system, or on a developer machine where they could assume a specific role.

cloudformation/build-bundle-test-workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+    Properties:
+      UserName: builder
+      Policies:
+        - PolicyName: S3_Write_Access
```
peternied, line 10: Rename to `S3_Build_Write_Access` for consistency.

cloudformation/build-bundle-test-workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+              Effect: Allow
+              Action: "s3:PutObject"
+              Resource:
+                - !Sub "arn:aws:s3:::${BuildBucket}/builds/*"
+  BundleUser:
+    Type: "AWS::IAM::User"
+    Properties:
+      UserName: bundler
+      Policies:
+        - PolicyName: S3_Build_Read_Access
```
peternied, line 23: Aren't these already public buckets - is granting read access needed?

peternied (changes requested):
Rather than have the CFN yml, what about making a fresh CDK project with multiple stacks to manage this first of many changes?

* Requesting changes on Users v Roles comment

cloudformation/build-bundle-test-workflow.yml
mch2: We would like to trigger the bundle builds & assembly fro Github Workflows which requires a User.

I don't think we need two sets of these creds though.  Perhaps we create a single user that is allowed to assume both these roles.

camerski: Yeah, I'll switch to roles and create a single "system" user that can have credentials in the CI system.

cloudformation/build-bundle-test-workflow.yml
camerski: Only the prod bucket will be public. For our beta stack and individual developer stacks everything should be private. Also, IAM roles still need permissions to invoke GetObject (S3's permissions model is more complicated than other AWS services...access is gated by both the bucket policy and the caller's permissions).

cloudformation/build-bundle-test-workflow.yml
camerski: Done

cloudformation/build-bundle-test-workflow.yml
camerski: Done

peternied (approved):

mch2 (approved):

Issue #149, opened 2021-08-06
Update build system to build knn


mch2:
Right now our build script for knn only includes the .zip but is not building and including knnlib.  The knnlib is required during bundle assembly.

This is because it is using the standard-gradle-build script in https://github.com/opensearch-project/opensearch-build/tree/main/scripts/bundle-build/.

- This component should have a custom script here - [here](https://github.com/opensearch-project/opensearch-build/tree/main/bundle-workflow/scripts/bundle-build/components). That builds both the lib and plugin zip.  The lib should be placed in /libs and plugin zip in /plugins.
- This component's snapshot build by passing -Dbuild.snapshot=true needs to work and output -SNAPSHOT artifacts.

Example of this being built is [here](https://github.com/opendistro-for-elasticsearch/k-NN/blob/main/.github/workflows/CD.yml#L82-L239)

Issue #148, opened 2021-08-06
Update build system to build performance-analyzer


mch2:
Currently Performance-analyzer is commented out in the [InputManifest](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/opensearch-bundle.1.0.yml) for 1.0.  

This component should be included with all the others but its build is currently failing if uncommented.

- This component requires performance-analyzer-rca to also be assembled and included within performance-analyzer-rca.zip. [Example](https://github.com/opensearch-project/opensearch-build/blob/main/release/tar/linux/opensearch-tar-build.sh#L147) of it being required during bundle assembly. 
- This component's snapshot build by passing -Dbuild.snapshot=true needs to work and output -SNAPSHOT artifacts.

xuezhou25:
About this build fail, it is because [spotless](https://github.com/opensearch-project/performance-analyzer-rca/blob/c1c32ab1f03e96fbbb03b520ba0de7d61d2d1618/build.gradle#L100) in RCA(when build PA, it will by default build RCA internal) required two branch to compare(for example main and 1.0). 
```
* What went wrong:
A problem occurred configuring root project 'tmp3gn5llmf'.
> No such reference 'origin/main'
```
After fix RCA by skip the spotless step, it can build successful.

mch2:
RCA is also failing its spotbugs check on main. -> https://github.com/opensearch-project/performance-analyzer-rca/pull/47

We can get by that when building RCA directly, but when we build the main plugin repo it is *always* cloning and building the main branch of RCA internally as part of its build scripts. https://github.com/opensearch-project/performance-analyzer/issues/45
This means we cannot bypass this step and the build fails.

This means that any build of perf analyzer fails, even for 1.1, until the above PR is merged.

mch2:
with the following snippet we can get by the spotbugs check in build.sh for each repo.

```
# remove this script when https://github.com/opensearch-project/performance-analyzer/issues/44 is fixed
git fetch origin main
git checkout main
# move back to initial ref
git checkout -
```

without the [spotbugs](opensearch-project/performance-analyzer-rca#47) fix we have no way of producing a successful build.

PR #161, opened 2021-08-06
Consolidate the build and test prototype workflows into one place

camerski:
### Description
This is a non-functional change. Files have been moved around and, in some cases, renamed, and there have been a couple of tweaks to relative paths, but no functional changes have been made.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/156

### Check List
- [X] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #162, opened 2021-08-09
Add Bundle assembly script to bundle-workflow

### LINKED ISSUE #133: Script to build each component from manifest and assemble bundle.

mch2:

### Description
This commit introduces an assembly step to build a full bundle.
Given a build manifest created with the build step, it will
install all plugins, create a bundle manifest, and create the tarball.

Other notes:
- Fixes OpenSearch core's build.sh script to correctly rename the distro to -min. 
- Generates a build-id during the build step and re-uses the same build-id in the bundle manifest.
- It is expected that the manifest path passed as input lives inside a directory containing build output from the build workflow.
- Bundles are output to a separate /bundles folder.
- Adds support to reference a public location where artifacts are hosted in the build manifest.  If env variable PUBLIC_ARTIFACT_URL is set, it will construct a url in the format > PUBLIC_ARTIFACT_URL/bundles/version/build-id/...
- It is expected that bundles will be copied to a /bundles folder, and builds to a /builds folder.

Example manifest with env variable set:

```
build:
  architecture: x64
  id: 41d5ae25183d4e699e92debfbe3f83bd
  location: https://artifacts.opensearch.org/bundles/1.0.0/41d5ae25183d4e699e92debfbe3f83bd/opensearch-1.0.0-linux-x64.tar.gz
  name: OpenSearch
  version: 1.0.0
components:
- commit_id: 8daa06f816a0a17694ff1c3b56977a9a18fdf1f9
  location: https://artifacts.opensearch.org/builds/1.0.0/41d5ae25183d4e699e92debfbe3f83bd/bundle/opensearch-min-1.0.0-linux-x64.tar.gz
  name: OpenSearch
  ref: 1.0
- commit_id: 7fad9529358259de529763c1c923fd947817a3bd
  location: https://artifacts.opensearch.org/builds/1.0.0/41d5ae25183d4e699e92debfbe3f83bd/plugins/opensearch-job-scheduler-1.0.0.0.zip
  name: job-scheduler
  ref: 1.0.0.0
  repository: https://github.com/opensearch-project/job-scheduler.git
```

if not set, absolute local paths are used:
```
build:
  architecture: x64
  id: 41d5ae25183d4e699e92debfbe3f83bd
  location: /Volumes/workplace/opensearch-build/bundles/opensearch-1.0.0-linux-x64.tar.gz
  name: OpenSearch
  version: 1.0.0
components:
- commit_id: 8daa06f816a0a17694ff1c3b56977a9a18fdf1f9
  location: /Volumes/workplace/opensearch-build/artifacts/bundle/opensearch-min-1.0.0-linux-x64.tar.gz
  name: OpenSearch
  ref: 1.0
- commit_id: 7fad9529358259de529763c1c923fd947817a3bd
  location: /Volumes/workplace/opensearch-build/artifacts/plugins/opensearch-job-scheduler-1.0.0.0.zip
  name: job-scheduler
  ref: 1.0.0.0
  repository: https://github.com/opensearch-project/job-scheduler.git
```

This script does not yet produce a fully functional bundle.  Knn and performance-analyzer require special configuration.  This should be done as part of  #148 and #149

 
### Issues Resolved
closes #133 
 
### Check List

### DIFF
--- a/bundle-workflow/assemble.sh
+++ b/bundle-workflow/assemble.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# Copyright OpenSearch Contributors.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+DIR="$(dirname "$0")"
+python3 "$DIR/python/assemble.py" $@
--- a/bundle-workflow/python/assemble.py
+++ b/bundle-workflow/python/assemble.py
@@ -1,0 +1,38 @@
+import os
+import sys
+import tempfile
+from assemble_workflow.bundle import Bundle
+from assemble_workflow.bundle_recorder import BundleRecorder
+from manifests.build_manifest import BuildManifest
+
+if (len(sys.argv) < 2):
+    print("Build an OpenSearch Bundle")
+    print("usage: assemble.sh /path/to/build_manifest")
+    exit(1)
+
+
+build_manifest_path = sys.argv[1]
+build_manifest = BuildManifest.from_file(build_manifest_path)
+build = build_manifest.build
+artifacts_dir = os.path.dirname(os.path.realpath(build_manifest_path))
+output_dir = os.path.join(os.getcwd(), 'bundle')
+os.makedirs(output_dir, exist_ok=True)
+
+with tempfile.TemporaryDirectory() as work_dir:
+    print(f'Bundling {build.name} ({build.architecture}) into {output_dir} ...')
+
+    os.chdir(work_dir)
+
+    bundle_recorder = BundleRecorder(build, output_dir, artifacts_dir)
+    bundle = Bundle(build_manifest, artifacts_dir, bundle_recorder)
+
+    bundle.install_plugins()
+    print(f'Installed plugins: {bundle.installed_plugins}')
+
+    #  Save a copy of the manifest inside of the tar
+    bundle_recorder.write_manifest(bundle.archive_path)
+    bundle.build_tar(output_dir)
+
+    bundle_recorder.write_manifest(output_dir)
+
+print(f'Done.')
--- a/bundle-workflow/python/assemble_workflow/bundle.py
+++ b/bundle-workflow/python/assemble_workflow/bundle.py
@@ -1,0 +1,85 @@
+# Copyright OpenSearch Contributors.
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import tarfile
+import tempfile
+import shutil
+import subprocess
+
+'''
+This class is responsible for executing the build of the full bundle and passing results to a bundle recorder.
+It requires a min tarball distribution where plugins will be installed and the path to an artifacts directory where 
+plugins can be found.
+'''
+
+
+class Bundle:
+
+    def __init__(self, build_manifest, artifacts_dir, bundle_recorder):
+        """
+        Construct a new Bundle instance.
+        :param build_manifest: A BuildManifest created from the build workflow.
+        :param artifacts_dir: Dir location where build artifacts can be found locally
+        :param bundle_recorder: The bundle recorder that will capture and build a BundleManifest
+        """
+        self.min_tarball = self.get_min_bundle(build_manifest.components)
+        self.plugins = self.get_plugins(build_manifest.components)
+        self.artifacts_dir = artifacts_dir
+        self.bundle_recorder = bundle_recorder
+        self.tmp_dir = tempfile.TemporaryDirectory()
+        self.installed_plugins = []
+        tmp_path = self.add_component(self.min_tarball, "bundle")
+        self.unpack(tmp_path, self.tmp_dir.name)
+        # OpenSearch & Dashboard tars will include only a single folder at the top level of the tar
+        self.archive_path = next(iter([file.path for file in os.scandir(self.tmp_dir.name) if file.is_dir()]))
+
+    def install_plugins(self):
+        for plugin in self.plugins:
+            print(f'Installing {plugin.name}')
+            self.install_plugin(plugin)
+        self.installed_plugins = os.listdir(os.path.join(self.archive_path, 'plugins'))
+
+    def install_plugin(self, plugin):
+        tmp_path = self.add_component(plugin, "plugins")
+        cli_path = os.path.join(self.archive_path, 'bin/opensearch-plugin')
+        self.execute(f'{cli_path} install --batch file:{tmp_path}')
+
+    def add_component(self, component, component_type):
+        rel_path = self.get_rel_path(component, component_type)
+        tmp_path = self.copy_component(rel_path, self.tmp_dir.name)
+        self.bundle_recorder.record_component(component, rel_path)
+        return tmp_path
+
+    def execute(self, command):
+        print(f'Executing "{command}" in {self.archive_path}')
+        subprocess.check_call(command, cwd=self.archive_path, shell=True)
+
+    def unpack(self, tar_path, dest):
+        with tarfile.open(tar_path) as tar:
+            tar.extractall(dest)
+
+    def build_tar(self, dest):
+        tar_name = self.bundle_recorder.tar_name
+        with tarfile.open(tar_name, "w:gz") as tar:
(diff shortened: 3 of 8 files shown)


bundle-workflow/README.md
```diff
@@ -6,8 +6,16 @@ The OpenSearch repo is built first, followed by common-utils. These dependencies
 
 ### Usage
 
+To build all components that make up a bundle from a build config:
 ```bash
-./tools/bundle-build/build.sh manifests/opensearch-bundle.1.0.yml
+./tools/bundle-workflow/build.sh manifests/opensearch-bundle.1.0.yml
```
dblock, line 11: I think the manifest got renamed as par of #161

bundle-workflow/assemble.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Copyright OpenSearch Contributors.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+DIR="$(dirname "$0")"
+python3 "$DIR/python/assemble.py" $@
```
dblock, line 9: I'd like us to rename this `python` directory because it's a property of the language, not a thing. Let's just call it `lib`?

bundle-workflow/python/lib/build_manifest.py
```diff
@@ -20,17 +20,25 @@ def __post_init__(self):
...
     dict: dict = None
+    location: str = None
 
     def __post_init__(self):
+        self.id = self.data['build-id'] if 'build-id' in self.data else uuid.uuid4().hex
         self.name = self.data['name']
         self.version = self.data['version']
         self.__set_dict()
 
+    def set_location(self, location):
```
dblock, line 35: Now you can set `.location = ` and call `set_location`, the first one won't have the desired side-effect. I think you want to copy the value of location into the `dict` elsewhere

bundle-workflow/python/lib/bundle.py
```diff
@@ -0,0 +1,49 @@
...
+        self.execute(f'{cli_path} install --batch file:{plugin.local_path}')
+
+    def execute(self, command):
+        print(f'Executing "{command}" in {self.path}')
+        subprocess.check_call(command, cwd = self.path, shell = True)
+
+    def unpack(self, dest):
+        tar = tarfile.open(self.min_bundle.local_path)
+        tar.extractall(dest)
+        tar.close()
```
dblock, line 41: try and finally this one

bundle-workflow/python/lib/bundle.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def list_installed_plugins(self):
```
dblock, line 43: Turn this into a property that returns a list of plugin paths, `print` outside of this class

dblock (approved):
Some should have's below, consider updating.

bundle-workflow/assemble.sh
mch2: updated this.  We already have a folder in there named lib, so I renamed that to src.

bundle-workflow/lib/assemble.py
```diff
@@ -0,0 +1,47 @@
...
+    for plugin in manifest.plugins:
+        print(f'=== Installing {plugin.name} ...')
+        plugin.download(work_dir)
+        bundle.install_plugin(plugin)
+
+    print(f'Installed plugins:')
+    print(bundle.plugins)
+
+    filename = "-".join([build.name.lower(), build.version, 'linux', output.arch]) + '.tar.gz'
+    build.dict['location'] = os.path.join(output.dest, filename)
```
dblock, line 39: Let’s add a todo to fix this? Seems equally weird to the old version.

dblock (approved):

bundle-workflow/lib/assemble.py
mch2: +1 will see how this looks with @camerski 's refactoring in #165.

bundle-workflow/python/assemble.py
```diff
@@ -0,0 +1,64 @@
...
+build_manifest = BuildManifest.from_file(build_manifest_path)
+build = build_manifest.build
+artifacts_dir = os.path.dirname(os.path.realpath(build_manifest_path))
+output_dir = os.path.join(os.getcwd(), 'bundle')
+os.makedirs(output_dir, exist_ok=True)
+
+# Fetch built bundle components
+plugins = get_plugins(build_manifest.components)
+min_tar = get_min_bundle(build_manifest.components)
+tar_name = get_tar_name(build)
```
dblock, line 36: Any reason why these shouldn't be properties of `BuildManifest` or (probably better) `Bundle`? This way you can do `bundle.install_plugins()` for example, too.

bundle-workflow/python/assemble_workflow/bundle.py
```diff
@@ -0,0 +1,74 @@
...
+    Construct a new Bundle instance.
+    :param min_tarball: The BuildManifest component representing the min_tarball distribution
+    :param artifacts_dir: Dir location where build artifacts can be found locally
+    :param bundle_recorder: The bundle recorder that will capture and build a BundleManifest
+    """
+    def __init__(self, min_tarball, artifacts_dir, bundle_recorder):
+        self.min_tarball = min_tarball
+        self.artifacts_dir = artifacts_dir
+        self.bundle_recorder = bundle_recorder
+        self.tmpDir = tempfile.TemporaryDirectory()
```
dblock, line 27: `tmp_dir`, this ain't Java ;)

dblock (approved):
LGTM. Only nits.

bundle-workflow/python/assemble_workflow/bundle.py
```diff
@@ -0,0 +1,74 @@
...
+'''
+class Bundle:
+
+    """
+    Construct a new Bundle instance.
+    :param min_tarball: The BuildManifest component representing the min_tarball distribution
+    :param artifacts_dir: Dir location where build artifacts can be found locally
+    :param bundle_recorder: The bundle recorder that will capture and build a BundleManifest
+    """
+    def __init__(self, min_tarball, artifacts_dir, bundle_recorder):
```
camerski, line 23: Pydoc goes after the declaration. It's the opposite of Javadoc :)

bundle-workflow/python/assemble_workflow/bundle_recorder.py
```diff
@@ -0,0 +1,82 @@
...
+            self.data['build'] = {}
+            self.data['build']['id'] = build_id
+            self.data['build']['name'] = name
+            self.data['build']['version'] = str(version)
+            self.data['build']['architecture'] = arch
+            self.data['build']['location'] = location
+            self.data['schema-version'] = '1.0'
+            # We need to store components as a hash so that we can append artifacts by component name
+            # When we convert to a BundleManifest this will get converted back into a list
+            self.data['components_hash'] = {}
```
camerski, line 66: Do you need to store the components as a hash? The bundler does not need to update components, so can you store them directly as a list and avoid the munging on lines 80-81?

dblock (approved):

Issue #160, opened 2021-08-06
Lock python dependencies for scripts


dblock:
**Is your feature request related to a problem? Please describe.**

See https://github.com/opensearch-project/opensearch-build/pull/159#pullrequestreview-724704235. 

**Describe the solution you'd like**

Add a `Pipfile` and lock dependencies, checking in `Pipfile.lock`, document that one needs to `python3 -m pipenv install` before running scripts to make the environment predictable.

PR #159, opened 2021-08-06
Adding troubleshooting section for bundle-build

peternied:
### Description
Including setup for install pyyaml if it is missing
 
### Check List
- [X] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #163, opened 2021-08-09
Lock Python dependencies using pipenv.

dblock:

### Description

Instead of documenting dependencies as suggested in https://github.com/opensearch-project/opensearch-build/pull/159 introduce a Pipfile that lists them along with auto-failure and suggestions of what to install when `python3`, `pip` or `pipenv` may be missing. 

Extract common parts of `build.sh` and `test.sh` into `run.sh` that takes an argument of what to run.
 
### Issues Resolved

Closes  https://github.com/opensearch-project/opensearch-build/issues/160.

### Check List

### DIFF
--- a/bundle-workflow/Pipfile
+++ b/bundle-workflow/Pipfile
@@ -1,0 +1,12 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+pyyaml = "*"
+
+[dev-packages]
+
+[requires]
+python_version = "3.8"
--- a/bundle-workflow/Pipfile
+++ b/bundle-workflow/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-pyyaml = "*"
+pyyaml = "~=5.4"
 
 [dev-packages]
 
--- a/bundle-workflow/python/test.py
+++ b/bundle-workflow/python/test.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 import os
 import sys
 import tempfile
--- a/bundle-workflow/run.sh
+++ b/bundle-workflow/run.sh
@@ -5,12 +5,16 @@
 
 set -e
 
-DIR="$(dirname "$0")"
-
 if [ -z "$1" ]; then (echo "syntax: run.sh [workflow.py]"; exit -1); fi
 command -v python3 >/dev/null 2>&1 || (echo "missing python3"; exit -1)
 command -v pip >/dev/null 2>&1 || (echo "missing python3-pip"; exit -1)
 command -v pipenv >/dev/null 2>&1 || (echo "missing pipenv"; exit -1)
 
+DIR="$(dirname "$0")"
+
+echo "Installing dependencies in $DIR ..."
+export PIPENV_PIPFILE="$DIR/Pipfile"
 python3 -m pipenv install
+
+echo "Running "$1" ${@:2} ..."
 python3 -m pipenv run "$1" ${@:2}
--- a/bundle-workflow/run.sh
+++ b/bundle-workflow/run.sh
@@ -8,9 +8,9 @@ set -e
 DIR="$(dirname "$0")"
 
 if [ -z "$1" ]; then (echo "syntax: run.sh [workflow.py]"; exit -1); fi
-command -v python3 >/dev/null 2>&1 || (echo "missing python3, try 'sudo apt install python3'"; exit -1)
-command -v pip >/dev/null 2>&1 || (echo "missing pip, try 'sudo apt install python3-pip'"; exit -1)
-command -v pipenv >/dev/null 2>&1 || (echo "missing pipenv, try 'sudo apt install pipenv'"; exit -1)
+command -v python3 >/dev/null 2>&1 || (echo "missing python3"; exit -1)
+command -v pip >/dev/null 2>&1 || (echo "missing python3-pip"; exit -1)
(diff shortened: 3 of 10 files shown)


bundle-workflow/python/test.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
```
dblock, line 1: Without these running under pyenv would cause https://stackoverflow.com/questions/27606653/oserror-errno-8-exec-format-error.

bundle-workflow/Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+pyyaml = "*"
```
camerski, line 7: We should use a specific version, or at least a version range. "*" is going to be fragile.

bundle-workflow/run.sh
```diff
@@ -0,0 +1,16 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+DIR="$(dirname "$0")"
+
+if [ -z "$1" ]; then (echo "syntax: run.sh [workflow.py]"; exit -1); fi
+command -v python3 >/dev/null 2>&1 || (echo "missing python3, try 'sudo apt install python3'"; exit -1)
+command -v pip >/dev/null 2>&1 || (echo "missing pip, try 'sudo apt install python3-pip'"; exit -1)
+command -v pipenv >/dev/null 2>&1 || (echo "missing pipenv, try 'sudo apt install pipenv'"; exit -1)
```
camerski, line 13: On line 15 we install pipenv, so we shouldn't fail on line 13 if it isn't already installed.

Also, `apt` is not the recommended way to install python dependencies, including pip. See https://pip.pypa.io/en/stable/installation/ for better options.

bundle-workflow/Pipfile
dblock: That's what Pipefile.lock is for. It locks at the current version that has been tested with. However, from the declaration point of view we think the future version of pyyaml will also work, until we know better. So there's no reason to specify a version here. It would be needed if we knew, for example, that version 3 works, but 4 breaks everything.

bundle-workflow/run.sh
dblock: > On line 15 we install pipenv, so we shouldn't fail on line 13 if it isn't already installed.

That's incorrect, on line 15 we install all the dependencies that are listed in Pipfile.lock, not pipenv.
 
> Also, `apt` is not the recommended way to install python dependencies, including pip. See https://pip.pypa.io/en/stable/installation/ for better options.

That's why we don't. I think you might want to read up on [pipenv](https://pipenv.pypa.io/en/latest/). We install python, pip and pipenv, which are all globally installed tools. Then we install all the other dependencies by running `pipenv install` from `Pipfile.lock`.

bundle-workflow/Pipfile
camerski: Please don't assume I haven't read the documentation before commenting.

Breaking changes from major version updates is exactly why we should specify a major version here. If pyyaml ever releases version 4 (which, presumably, is a breaking change because of the major version bump) then the next time someone runs `pipenv update` it will pull in v4 and break everything. If we specify `3` then `pipenv update` will pull in `3.4` or whatever is the latest, but not attempt to pull in `4` and everything will continue to work.

Changing a dependency to a new major version should be a conscious decision, not a side-effect.

dblock: > Please don't assume I haven't read the documentation before commenting.

Ask along if the docs aren't clear about how pipenv works.
 
> Breaking changes from major version updates is exactly why we should specify a major version here. If pyyaml ever releases version 4 (which, presumably, is a breaking change because of the major version bump) then the next time someone runs `pipenv update` it will pull in v4 and break everything. If we specify `3` then `pipenv update` will pull in `3.4` or whatever is the latest, but not attempt to pull in `4` and everything will continue to work.
 
> Changing a dependency to a new major version should be a conscious decision, not a side-effect.

Running `pipenv update` is that conscious decision and you have to test the outcome, then make a PR for the change to take effect. Nobody is running and making commits nilly willy of dependency changes. This is just by convention in every package manager, including pipenv: you put in these files the version as you know it, and let the tool figure out the best compatible matrix, write it into Pipfile.lock and then this becomes your current stable set.

dblock:
I see your point now @camerski. I ran `python3 -m pipenv install pyyaml~=5.4` to lock pyyaml at 5.x.

bundle-workflow/run.sh
camerski: We say "try 'sudo apt install pipenv'"...we shouldn't give that advice. We should direct people to the pipenv website for installation instructions. Same for line 12: apt is not the best way to install pip.

bundle-workflow/Pipfile
camerski: As discussed in person, we have a different view of what counts as a "conscious decision" about major version updates and how conservative our approach should be. Wildcards are fine; this is just a difference of opinion. No approach is inherently better than the other; there are just different tradeoffs.

> If I put a version in Pipfile, what do I need a Pipfile.lock for? That's not how the tool was designed.

The pipfile contains a version range (e.g. `>4`) and the Pipfile.lock contains the exact version that was resolved (e.g. `4.4.0`).

bundle-workflow/run.sh
dblock: Removed advice, updated to just say "missing pipenv" and such.

camerski (approved):

camerski (approved):

camerski (approved):

mch2 (approved):
lgtm

Issue #157, opened 2021-08-06
Refactor tester, maybe reuse BuildOutput manifest and other shared code


dblock:
**Is your feature request related to a problem? Please describe.**

- We have at least 2 tools using build output workflows, https://github.com/opensearch-project/opensearch-build/tree/main/workflows/testing and https://github.com/opensearch-project/opensearch-build/tree/main/tools/bundle-build. The python implementation for these manifests should probably be shared. See https://github.com/opensearch-project/opensearch-build/issues/155#issuecomment-894443601 for more ideas.
- The tester tool needs refactoring discussed in #138.

camerski:
Resolved by https://github.com/opensearch-project/opensearch-build/pull/170

PR #165, opened 2021-08-09
Add 'manifests' package that contains immutable input and build manifest classes

camerski:
### Description
Provides `InputManifest` and `BuildManifest` classes in the `manifests` package.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/bundle-workflow/python/manifests/__init__.py
+++ b/bundle-workflow/python/manifests/__init__.py
@@ -1,0 +1,1 @@
+# This page intentionally left blank
--- a/bundle-workflow/python/manifests/build_manifest.py
+++ b/bundle-workflow/python/manifests/build_manifest.py
@@ -1,0 +1,81 @@
+# Copyright OpenSearch Contributors.
+# SPDX-License-Identifier: Apache-2.0
+
+import yaml
+
+'''
+A BuildManifest is an immutable view of the outputs from a build step
+The manifest contains information about the product that was built (in the `build` section),
+and the components that made up the build in the `components` section.
+
+The format for schema version 1.0 is:
+schema-version: 1.0
+build:
+  name: string
+  version: string
+components:
+  - name: string
+    repository: URL of git repository
+    ref: git ref that was built (sha, branch, or tag)
+    commitid: The actual git commit ID that was built (i.e. the resolved "ref")
+    artifacts:
+      maven:
+        - maven/relative/path/to/artifact
+        - ...
+      plugins:
+        - plugins/relative/path/to/artifact
+        - ...
+      libs:
+        - libs/relative/path/to/artifact
+        - ...
+  - ...
+'''
+class BuildManifest:
+    @staticmethod
+    def from_file(path):
+        with open(path, 'r') as file:
+            return BuildManifest(yaml.safe_load(file))
+
+    def __init__(self, data):
+        self.version = str(data['schema-version'])
+        if self.version != '1.0':
+            raise ValueError(f'Unsupported schema version: {self.version}')
+        self.build = self.Build(data['build'])
+        self.components = list(map(lambda entry: self.Component(entry),
+                                   data['components']))
+
+    def to_dict(self):
+        return {
+            'schema-version': '1.0',
+            'build': self.build.to_dict(),
+            'components': list(map(lambda component: component.to_dict(),
+                                   self.components))
+        }
+
+    class Build:
+        def __init__(self, data):
+            self.name = data['name']
+            self.version = data['version']
+
+        def to_dict(self):
+            return {
+                'name': self.name,
+                'version': self.version
+            }
+
+    class Component:
+        def __init__(self, data):
+            self.name = data['name']
+            self.repository = data['repository']
+            self.ref = data['ref']
+            self.commitid = data['commitid']
+            self.artifacts = data['artifacts']
+
+        def to_dict(self):
+            return {
+                'name': self.name,
+                'repository': self.repository,
+                'ref': self.ref,
+                'commitid': self.commitid,
+                'artifacts': self.artifacts
+            }
--- a/bundle-workflow/python/manifests/build_manifest.py
+++ b/bundle-workflow/python/manifests/build_manifest.py
@@ -13,6 +13,7 @@ schema-version: 1.0
 build:
   name: string
   version: string
+  architecture: x64 or arm64
 components:
   - name: string
     repository: URL of git repository
@@ -56,11 +57,13 @@ class BuildManifest:
         def __init__(self, data):
             self.name = data['name']
             self.version = data['version']
+            self.architecture = data['architecture']
 
         def to_dict(self):
             return {
                 'name': self.name,
-                'version': self.version
+                'version': self.version,
+                'architecture': self.architecture
             }
 
     class Component:
--- a/bundle-workflow/python/manifests/build_manifest.py
+++ b/bundle-workflow/python/manifests/build_manifest.py
@@ -18,7 +18,7 @@ components:
   - name: string
     repository: URL of git repository
     ref: git ref that was built (sha, branch, or tag)
-    commitid: The actual git commit ID that was built (i.e. the resolved "ref")
+    commit_id: The actual git commit ID that was built (i.e. the resolved "ref")
     artifacts:
       maven:
         - maven/relative/path/to/artifact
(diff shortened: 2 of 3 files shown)


bundle-workflow/python/manifests/__init__.py
```diff
@@ -0,0 +1 @@
+# This page intentionally left blank
```
dblock, line 1: If you're going to write a sentence, starting with a capital T, finish it with a period :)

bundle-workflow/python/manifests/build_manifest.py
```diff
@@ -0,0 +1,81 @@
...
+The format for schema version 1.0 is:
+schema-version: 1.0
+build:
+  name: string
+  version: string
+components:
+  - name: string
+    repository: URL of git repository
+    ref: git ref that was built (sha, branch, or tag)
+    commitid: The actual git commit ID that was built (i.e. the resolved "ref")
```
dblock, line 20: Could we make this `commit_id` or `ref`? Otherwise we'll end up with things like `buildmanifestthatiscool`.

bundle-workflow/python/manifests/build_manifest.py
```diff
@@ -0,0 +1,81 @@
...
+'''
+A BuildManifest is an immutable view of the outputs from a build step
+The manifest contains information about the product that was built (in the `build` section),
+and the components that made up the build in the `components` section.
+
+The format for schema version 1.0 is:
+schema-version: 1.0
+build:
+  name: string
+  version: string
```
mch2, line 15: We should include platform/architecture here as well

dblock (commented):
- I think you also need a base `Manifest` class and sink common logic such as version into it. 
- I don't think we should merge this until the classes are used, otherwise it's not really a refactor.
- plus minor stuff.

bundle-workflow/python/manifests/build_manifest.py
camerski: Yes, we should.

mch2 (approved):
lgtm. with minor comment

camerski:
> * I think you also need a base `Manifest` class and sink common logic such as version into it.

Inheritance is not the right approach here, composition is. We could possibly extract some parts to compose, but it's not worth it at this point.

> * I don't think we should merge this until the classes are used, otherwise it's not really a refactor.

Merging dead code is totally OK so that the reviewers of the upcoming refactors can see how they fit together. If you feel strongly then I can hold off on the merge until the rest of the PRs are approved, but I don't think that is necessary.

bundle-workflow/python/manifests/build_manifest.py
camerski: `commitid` and `ref` are capturing different things. We only really *need* the `commitid`, but `ref` is helpful for humans to quickly know if this was a build from `main` or `1.0` or whatever branch it was.

I don't quite understand what you mean by `buildmanifestthatiscool`...can you explain?

bundle-workflow/python/manifests/__init__.py
camerski: I grammar the way I wants

bundle-workflow/python/manifests/build_manifest.py
mch2: I'd like to be able to feed BuildManifests back into the build system and rebuild components from the same sha.  Use case:  After the initial snapshot build we want to rebuild without a snapshot tag from the same sha's and promote to staging.  If we are using 'ref' from InputManifest, I'd rather use 'ref' and make guarantees that in a BuildManifest this will always be a sha

dblock:
> Merging dead code is totally OK so that the reviewers of the upcoming refactors can see how they fit together. If you feel strongly then I can hold off on the merge until the rest of the PRs are approved, but I don't think that is necessary.

It'd be fine as dead code. But I feel strongly because this is dead code that duplicates existing logic, making the codebase net worse. Finish the refactor of replacing the manifests in https://github.com/opensearch-project/opensearch-build/tree/main/bundle-workflow/python/lib with this code, and then we can merge?

camerski:
> > Merging dead code is totally OK so that the reviewers of the upcoming refactors can see how they fit together. If you feel strongly then I can hold off on the merge until the rest of the PRs are approved, but I don't think that is necessary.
> 
> It'd be fine as dead code. But I feel strongly because this is dead code that duplicates existing logic, making the codebase net worse. Finish the refactor of replacing the manifests in https://github.com/opensearch-project/opensearch-build/tree/main/bundle-workflow/python/lib with this code, and then we can merge?

Sure, that's fine. The rest of the refactoring is coming up this afternoon. I've done all the code changes; I'm just picking them apart into standalone PRs so that the reviews are easy.

bundle-workflow/python/manifests/build_manifest.py
dblock: I am just saying that `commitid` should be `commit_id`, underscore, because it's not 1 word.

A "commit ID" is a SHA-1 of all the changes, see https://stackoverflow.com/questions/29106996/what-is-a-git-commit-id. A "ref" is an indirect reference to a commit. The most direct way to reference a commit is via its SHA-1 hash, but it can be a branch, tag, etc.

So I think @mch2 you're saying all the right things and you want a "commit ID", and @camerski is right here to call that a commit ID and not a "ref".

camerski:
> > > Merging dead code is totally OK so that the reviewers of the upcoming refactors can see how they fit together. If you feel strongly then I can hold off on the merge until the rest of the PRs are approved, but I don't think that is necessary.
> > 
> > 
> > It'd be fine as dead code. But I feel strongly because this is dead code that duplicates existing logic, making the codebase net worse. Finish the refactor of replacing the manifests in https://github.com/opensearch-project/opensearch-build/tree/main/bundle-workflow/python/lib with this code, and then we can merge?
> 
> Sure, that's fine. The rest of the refactoring is coming up this afternoon. I've done all the code changes; I'm just picking them apart into standalone PRs so that the reviews are easy.

@dblock Assuming you're OK with that plan, can I have an approval? I obviously won't execute the merge until the other changes are approved, but I don't want to leave this change hanging.

dblock:
> @dblock Assuming you're OK with that plan, can I have an approval? I obviously won't execute the merge until the other changes are approved, but I don't want to leave this change hanging.

Code here looks OK to me. Let's see those other changes on top!

bundle-workflow/python/manifests/build_manifest.py
camerski: For that use case I would generate a new InputManifest from the BuildManifest, rather than reading a BuildManifest and trying to treat it as an InputManifest. Does that make sense?

dblock:
You should merge https://github.com/opensearch-project/opensearch-build/pull/162 first though because otherwise @mch2 will have to go back and redo that work on top of your upcoming changes.

bundle-workflow/python/manifests/build_manifest.py
mch2: Makes sense - I think a better approach for my use case is to create a separate promotion script that creates a new InputManifest using the commit_id in a built BundleManifest as a component's ref, and then feed it back to the build system.

camerski:
> You should merge #162 first though because otherwise @mch2 will have to go back and redo that work on top of your upcoming changes.

Discussed with @mch2 and agreed that this change can go first. I'll rebase #162 on top of this one.

camerski:
> > @dblock Assuming you're OK with that plan, can I have an approval? I obviously won't execute the merge until the other changes are approved, but I don't want to leave this change hanging.
> 
> Code here looks OK to me minus the naming of `commit_id`. Let's see those other changes on top!

@dblock Renamed.

bundle-workflow/python/manifests/build_manifest.py
gaiksaya: Should we add platform as well? linux/mac/windows?

camerski: Graah, yeah, platform is not the same as architecture. My bad :)

I have a few changes stacked on top of this...can I add platform as a new PR?

gaiksaya: Sure if you remember it! :)

gaiksaya (approved):
To do: Add platform in new PR

camerski:
Replaced with https://github.com/opensearch-project/opensearch-build/pull/170

PR #166, opened 2021-08-09
Add build_workflow library

camerski:
### Description
Adds `Builder`, `BuildRecorder`, and `GitRepository` classes that represent the build process for a component. This is a refactoring of the old classes in `lib` that simplifies and cleanly separates logic and responsibilities.

### Issues Resolved
 
### Check List


mch2:
Can we combine this with #168?  I think its easier to digest as a single PR

bundle-workflow/python/build_workflow/builder.py
```diff
@@ -0,0 +1,48 @@
+# Copyright OpenSearch Contributors.
```
dblock, line 1: This looks like a variation/reimplementation of https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/python/lib/component.py without dataclass.

bundle-workflow/python/build_workflow/builder.py
```diff
@@ -0,0 +1,48 @@
...
+        if build_script is None:
+            raise RuntimeError(f'Could not find build.sh script. Looked in {paths}')
+
+        build_command = f'{build_script} {version} {arch}'
+        self.git_repo.execute(build_command)
+        self.output.record_component(self.component_name, self.git_repo)
+
+    def export_artifacts(self):
+        artifacts_dir = os.path.realpath(os.path.join(self.git_repo.dir.name, 'artifacts'))
+        for artifact_type in ["maven", "plugins", "libs"]:
```
mch2, line 43: this is missing "bundle".  Opensearch core's build script will place the min bundle in that directory.

camerski: Ah, I missed that. Good catch!

camerski:
https://github.com/opensearch-project/opensearch-build/pull/170

PR #168, opened 2021-08-09
Add build workflow entry point

camerski:
### Description
`build.py` is the entry point to execute the build workflow. It takes in an input manifest and outputs build artifacts + an output manifest to the `artifacts` directory.

This change depends on https://github.com/opensearch-project/opensearch-build/pull/165 and https://github.com/opensearch-project/opensearch-build/pull/166. See those PRs for details of the manifests and build_workflow packages.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/bundle-workflow/python/build.py
+++ b/bundle-workflow/python/build.py
@@ -3,31 +3,57 @@
 
 import os
 import sys
+import subprocess
 import tempfile
-import urllib.request
-from lib.build_manifest import BuildManifest
-from lib.build_output import BuildOutput
+import yaml
+from manifests.input_manifest import InputManifest
+from build_workflow.build_recorder import BuildRecorder
+from build_workflow.builder import Builder
+from build_workflow.git_repository import GitRepository
 
 if (len(sys.argv) < 2):
     print("Build an OpenSearch Bundle")
     print("usage: build.sh /path/to/manifest")
     exit(1)
 
-with tempfile.TemporaryDirectory() as work_dir:
-    manifest = BuildManifest.from_file(sys.argv[1])
-    build = manifest.build
-    output = BuildOutput()
+component_scripts_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../scripts/bundle-build/components')
+default_build_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../scripts/bundle-build/standard-gradle-build/build.sh')
 
-    print(f'Building {build.name} ({output.arch}) into {output.dest} ...')
+def get_arch():
+    arch = subprocess.check_output(['uname', '-m']).decode().strip()
+    if arch == 'x86_64':
+        return 'x64'
+    elif arch == 'aarch64' or arch == 'arm64':
+        return  'arm64'
+    else:
+        raise ValueError(f'Unsupported architecture: {arch}')
+
+arch = get_arch()
+manifest = InputManifest.from_file(sys.argv[1])
+output_dir = os.path.join(os.getcwd(), 'artifacts')
+os.makedirs(output_dir, exist_ok = True)
+
+with tempfile.TemporaryDirectory() as work_dir:
+    print(f'Building in {work_dir}')
 
     os.chdir(work_dir)
 
-    for component in manifest.components:
-        print(f'=== Building {component.name} ...')
-        component.checkout()
-        component.build(build.version, output.arch)
-        component.export(output.dest)
+    build_recorder = BuildRecorder(output_dir, manifest.build.name, manifest.build.version, arch)
 
-    manifest_path = os.path.join(output.dest, 'manifest.yml')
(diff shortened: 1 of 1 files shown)


bundle-workflow/python/build.py
```diff
@@ -3,31 +3,57 @@
...
     exit(1)
 
-with tempfile.TemporaryDirectory() as work_dir:
-    manifest = BuildManifest.from_file(sys.argv[1])
-    build = manifest.build
-    output = BuildOutput()
+component_scripts_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../scripts/bundle-build/components')
+default_build_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../scripts/bundle-build/standard-gradle-build/build.sh')
+
+def get_arch():
```
mch2, line 22: we are going to need to reuse this during assembly, can we extract this somewhere with static access?

bundle-workflow/python/build.py
```diff
@@ -3,31 +3,57 @@
...
 if (len(sys.argv) < 2):
     print("Build an OpenSearch Bundle")
     print("usage: build.sh /path/to/manifest")
     exit(1)
 
-with tempfile.TemporaryDirectory() as work_dir:
-    manifest = BuildManifest.from_file(sys.argv[1])
-    build = manifest.build
-    output = BuildOutput()
+component_scripts_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../scripts/bundle-build/components')
```
mch2, line 19: If these are static paths it makes sense to me to build these paths directly inside of the Builder class.  Any reason these wouldn't be static?

bundle-workflow/python/build.py
dblock: This already exists in https://github.com/opensearch-project/opensearch-build/blob/2ac8a7a876f497d7870fc591dbf497f0606cd018/bundle-workflow/python/lib/build_output.py#L25

bundle-workflow/python/build.py
dblock: This code already exists in https://github.com/opensearch-project/opensearch-build/blob/2ac8a7a876f497d7870fc591dbf497f0606cd018/bundle-workflow/python/lib/component.py#L31

bundle-workflow/python/build.py
```diff
@@ -3,31 +3,57 @@
...
+
     for component in manifest.components:
-        print(f'=== Building {component.name} ...')
-        component.checkout()
-        component.build(build.version, output.arch)
-        component.export(output.dest)
-
-    manifest_path = os.path.join(output.dest, 'manifest.yml')
-    manifest.save_to(manifest_path)
-    print(f'Done.')
```
mch2, line 33: nit - I liked having this as a confirmation of success, can we put it at the end?

bundle-workflow/python/build.py
```diff
@@ -3,31 +3,57 @@
...
+                          repo,
+                          component_scripts_path,
+                          default_build_path,
+                          build_recorder)
+        builder.build(manifest.build.version, arch)
+        builder.export_artifacts()
+
+    output_manifest = build_recorder.get_manifest()
+    manifest_path = os.path.join(output_dir, 'manifest.yml')
+    with open(manifest_path, 'w') as file:
```
mch2, line 58: I think this logic belongs in BuildRecorder, something like build_record.write_manifest(path)

mch2 (changes requested):
I like the separation of concerns here.  Some comments to help clean it up a bit further and for reuse with bundling.  Could you please combine this with #166?

bundle-workflow/python/build.py
camerski: This code should not be in `lib/component.py`. Setting up the default paths is not the responsibility of the builder (or the now-defunct `component`) - it only consumes those paths.

Also, the relative paths in `component.py` are super fragile. Moving the logic out into build.py means `builder` can rely on absolute paths (or, more accurately, `builder` doesn't care whether they are relative or absolute).

bundle-workflow/python/build.py
camerski: Oh, man, that is soooo the wrong place to figure out the architecture!

camerski: > we are going to need to reuse this during assembly, can we extract this somewhere with static access?

Assembly will just read the architecture string from the build manifest, won't it?

bundle-workflow/python/build.py
camerski: Sure

bundle-workflow/python/build.py
mch2: I was thinking I would need to validate it during bundling but that process is architecture independent.

camerski:
Replaced with https://github.com/opensearch-project/opensearch-build/pull/170

PR #169, opened 2021-08-10
Delete the old build workflow code

camerski:
### Description
Cleans up all the old code that was refactored away in https://github.com/opensearch-project/opensearch-build/pull/165, https://github.com/opensearch-project/opensearch-build/pull/166, and https://github.com/opensearch-project/opensearch-build/pull/168.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/157
 
### Check List
- [X] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #170, opened 2021-08-10
Refactor the build workflow and extract common components for future sharing with the bundle and test workflows

camerski:
### Description
This is a rollup of all the changes in #165, #166, #168, and #169 for easier viewing.

At the time of writing, I have addressed all the comments from those other PRs. Once this PR is approved I will close the other 4 and merge this PR instead. Leaving them all open for now in case people have in-progress reviews.
 
### Issues Resolved
 
### Testing done
Ran the tool using the manifest in https://github.com/opensearch-project/opensearch-build/blob/main/manifests/opensearch-1.0.0.yml. Verified that the output manifest is essentially identical to the old output (some fields have been renamed or added, so they're not 100% identical) and that all artifacts are copied into the right places.

### Check List


bundle-workflow/python/build_workflow/builder.py
```diff
@@ -0,0 +1,48 @@
...
+        :param component_scripts_path: Where to look for individual component "build.sh" scripts for those repositories that don't contain their own "scripts/build.sh".
+        :param default_build_script: Path to a defauld build script to use when neither a per-repo "scripts/build.sh" nor a "component_scripts_path/<component>/build.sh" script can be found.
+        :param output: The output recorder that will capture build information and artifacts.
+        '''
+
+        self.component_name = component_name
+        self.git_repo = git_repo
+        self.component_scripts_path = component_scripts_path
+        self.default_build_script = default_build_script
+        self.output = output
```
mch2, line 26: can we rename this to build_recorder instead of output?

mch2 (changes requested):

bundle-workflow/python/build_workflow/builder.py
camerski: Done

mch2 (approved):

bundle-workflow/python/build_workflow/git_repository.py
```diff
@@ -0,0 +1,30 @@
...
+import subprocess
+import tempfile
+from dataclasses import dataclass
+
+'''
+This class checks out a Git repository at a particular ref into a temporary directory.
+Clients can obtain the actual commit ID by querying the "sha" attribute, and the temp directory name with "dir.name".
+'''
+@dataclass
+class GitRepository:
```
mch2, line 13: I missed this on my first pass.  We already have git_repository in lib/ that also should be removed.

camerski: Not sure how I missed that. BALEETED!

bundle-workflow/python/build_workflow/builder.py
```diff
@@ -0,0 +1,48 @@
...
+        '''
+
+        self.component_name = component_name
+        self.git_repo = git_repo
+        self.component_scripts_path = component_scripts_path
+        self.default_build_script = default_build_script
+        self.build_recorder = build_recorder
+
+    def build(self, version, arch):
+        paths = [os.path.realpath(os.path.join(self.git_repo.dir.name, 'scripts/build.sh')),
```
dblock, line 29: I would extract this into a `build_script` method. We don't write tests for this code, but if we did we'd want to make sure we explain the logic of selection of the build script.

There's also a smell here that we have 3 resolutions of a build path in what seems like the wrong place: one in the checked out code (belongs with the component), one in the override location (belongs with the component builder), and the default one (belongs in the bundle builder).

bundle-workflow/python/manifests/__init__.py
```diff
@@ -0,0 +1 @@
+# This page intentionally left blank
```
dblock, line 1: I'd prefer
`This hurts.` 
vs.
`This hurts`

bundle-workflow/python/manifests/build_manifest.py
```diff
@@ -0,0 +1,84 @@
...
+schema-version: 1.0
+build:
+  name: string
+  version: string
+  architecture: x64 or arm64
+components:
+  - name: string
+    repository: URL of git repository
+    ref: git ref that was built (sha, branch, or tag)
+    commit_id: The actual git commit ID that was built (i.e. the resolved "ref")
```
dblock, line 21: Nit: lowercase `The actual` to be consistent with the above.

dblock (approved):
This is good work. I made some suggestions, most are nits or nice to have's.

I would just merge this PR, it's much clearer and more readable than those individual ones.

dblock:
One more thing, add `#!/usr/bin/env python` on top of all executing .py scripts, see https://github.com/opensearch-project/opensearch-build/pull/163#discussion_r685261497

bundle-workflow/python/build_workflow/build_recorder.py
```diff
@@ -0,0 +1,70 @@
+# Copyright OpenSearch Contributors.
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import shutil
+import yaml
+from manifests.build_manifest import BuildManifest
+
+class BuildRecorder:
+    def __init__(self, output_dir, name, version, arch):
```
setiah, line 10: Can we pls add some documentation (pydoc) for this module

PR #187, opened 2021-08-11
Add ScriptFinder to abstract away the logic for finding build.sh and integtest.sh scripts

camerski:
Also refactor builder.py to use ScriptFinder

### Testing
Manually verified that build.py works correctly
 
### Check List

### DIFF
--- a/bundle-workflow/python/build_workflow/builder.py
+++ b/bundle-workflow/python/build_workflow/builder.py
@@ -9,31 +9,22 @@ It will notify the build recorder of build information such as repository and gi
 Artifacts found in "<build root>/artifacts/<maven|plugins|libs|bundle>" will be recognized and recorded.
 '''
 class Builder:
-    def __init__(self, component_name, git_repo, component_scripts_path, default_build_script, build_recorder):
+    def __init__(self, component_name, git_repo, script_finder, build_recorder):
         '''
         Construct a new Builder instance.
         :param component_name: The name of the component to build.
         :param git_repo: A GitRepository instance containing the checked-out code.
-        :param component_scripts_path: Where to look for individual component "build.sh" scripts for those repositories that don't contain their own "scripts/build.sh".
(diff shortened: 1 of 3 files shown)


dblock (approved):
I like it.

bundle-workflow/python/build_workflow/builder.py
```diff
@@ -9,31 +9,22 @@
...
     def build(self, version, arch):
-        paths = [os.path.realpath(os.path.join(self.git_repo.dir.name, 'scripts/build.sh')),
-                 os.path.realpath(os.path.join(self.component_scripts_path, self.component_name, 'build.sh')),
-                 self.default_build_script]
-
-        build_script = next(filter(lambda path: os.path.exists(path), paths), None)
-        if build_script is None:
-            raise RuntimeError(f'Could not find build.sh script. Looked in {paths}')
-
+        build_script = self.script_finder.find_build_script(self.component_name, self.git_repo.dir.name)
```
dblock, line 27: To be pedantic, for integ tests this isn't a "build script", it's an "integ test script". Maybe use "script" here.

bundle-workflow/python/paths/script_finder.py
```diff
@@ -0,0 +1,43 @@
...
+      * Root of the Git repository
+      * /scripts/<script-name> in the Git repository
+      * <component_scripts_path>/<component_name>/<script-name>
+      * <default_scripts_path>/<script-name>
+    '''
+
+    component_scripts_path: str
+    default_scripts_path: str
+
+    def find_build_script(self, component_name, git_dir):
```
dblock, line 21: Notice that both `find_build_script` and `find_integ_test_script` are identical minus the name of the script, `build.sh`, so these methods can combine in one common implementation.

camerski: I considered that, but decided that the client api should have two methods rather than asking the client to know the name of the script to fetch. That way, typos will result in a NoMethodError (or whatever it is that Python throws when it doesn't recognize a method name) rather than a RuntimeError saying it can't find `bui1d.sh`.

Internally it could totally be refactored into a private helper. Future PR!

mch2 (approved):

Issue #179, opened 2021-08-10
Build bundle 1.1-SNAPSHOT


mch2:
Support for passing `--snapshot` was added in #189. Assuming build works in #212. 

Fix snapshot builds in plugins, and change the plugin CI to build snapshot versions by default:

- [x] https://github.com/opensearch-project/common-utils/issues/56
fixed in https://github.com/opensearch-project/common-utils/pull/58
- [x] https://github.com/opensearch-project/sql/issues/177
fixed in https://github.com/opensearch-project/sql/pull/181
- [x] https://github.com/opensearch-project/alerting/issues/141
fixed in https://github.com/opensearch-project/alerting/pull/142
- [x] https://github.com/opensearch-project/index-management/issues/116
fixed in https://github.com/opensearch-project/index-management/pull/121
- [x] https://github.com/opensearch-project/k-NN/issues/78
fixed in https://github.com/opensearch-project/k-NN/pull/79
- [x] https://github.com/opensearch-project/security/issues/1403
fixed in https://github.com/opensearch-project/security/pull/1409
- [x] https://github.com/opensearch-project/anomaly-detection/issues/173
fixed in https://github.com/opensearch-project/anomaly-detection/pull/176
- [x] https://github.com/opensearch-project/asynchronous-search/issues/30
fixed in https://github.com/opensearch-project/asynchronous-search/pull/31
- [x] https://github.com/opensearch-project/dashboards-reports/issues/138
fixed in https://github.com/opensearch-project/dashboards-reports/pull/140
- [x] https://github.com/opensearch-project/dashboards-notebooks/issues/60
fixed in https://github.com/opensearch-project/dashboards-notebooks/pull/62
- [x] notifications
fixed in https://github.com/opensearch-project/notifications/pull/284
- [x] https://github.com/opensearch-project/performance-analyzer-rca/issues/48
fixed in https://github.com/opensearch-project/performance-analyzer-rca/pull/54
fixed in https://github.com/opensearch-project/performance-analyzer-rca/pull/47
fixed in https://github.com/opensearch-project/performance-analyzer-rca/pull/55
- [x] https://github.com/opensearch-project/performance-analyzer/issues/43
fixed in https://github.com/opensearch-project/performance-analyzer/pull/53
- [x] #349

dblock:
Support for flags added in #189.

Actual snapshot builds broken:

- [ ] https://github.com/opensearch-project/sql/issues/177
[...]

PR #189, opened 2021-08-11
Added support for passing snapshot build options.

dblock:
### Description

Added support for passing `--snapshot` to the bundle build, which cascades `-s` which cascades into `-dbuild.snapshot=true` into gradle/maven builds. Opened some issues to fix the actual snapshot builds for a handful of plugins, but that doesn't need to hold up this PR. 

Making the custom override scripts actually parse arguments so that we can extend them in the future with additional parameters and not rely on the fact that arg1 is OpenSearch version and arg2 is the architecture.

### Issues Resolved

Part of https://github.com/opensearch-project/opensearch-build/issues/179
 
### Check List

### DIFF
--- a/bundle-workflow/python/assemble.py
+++ b/bundle-workflow/python/assemble.py
@@ -1,20 +1,17 @@
 import os
-import sys
 import tempfile
+import argparse
 from assemble_workflow.bundle import Bundle
 from assemble_workflow.bundle_recorder import BundleRecorder
 from manifests.build_manifest import BuildManifest
 
-if (len(sys.argv) < 2):
-    print("Build an OpenSearch Bundle")
-    print("usage: assemble.sh /path/to/build_manifest")
-    exit(1)
+parser = argparse.ArgumentParser(description = "Assembly an OpenSearch Bundle")
+parser.add_argument('manifest', type = argparse.FileType('r'), help="Manifest file.")
+args = parser.parse_args()
 
-
-build_manifest_path = sys.argv[1]
-build_manifest = BuildManifest.from_file(build_manifest_path)
+build_manifest = BuildManifest.from_file(args.manifest)
 build = build_manifest.build
-artifacts_dir = os.path.dirname(os.path.realpath(build_manifest_path))
+artifacts_dir = os.path.dirname(os.path.realpath(args.manifest))
 output_dir = os.path.join(os.getcwd(), 'bundle')
 os.makedirs(output_dir, exist_ok=True)
 
--- a/bundle-workflow/python/assemble.py
+++ b/bundle-workflow/python/assemble.py
@@ -5,7 +5,7 @@ from assemble_workflow.bundle import Bundle
 from assemble_workflow.bundle_recorder import BundleRecorder
 from manifests.build_manifest import BuildManifest
 
-parser = argparse.ArgumentParser(description = "Assembly an OpenSearch Bundle")
+parser = argparse.ArgumentParser(description = "Assemble an OpenSearch Bundle")
 parser.add_argument('manifest', type = argparse.FileType('r'), help="Manifest file.")
 args = parser.parse_args()
 
--- a/bundle-workflow/python/build_workflow/build_recorder.py
+++ b/bundle-workflow/python/build_workflow/build_recorder.py
@@ -40,7 +40,7 @@ class BuildRecorder:
             self.data['build'] = {}
             self.data['build']['id'] = build_id
             self.data['build']['name'] = name
-            self.data['build']['version'] = str(version)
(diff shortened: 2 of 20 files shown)


bundle-workflow/python/build_workflow/build_recorder.py
```diff
@@ -35,13 +35,14 @@ def write_manifest(self, folder):
             yaml.dump(output_manifest.to_dict(), file)
 
     class BuildManifestBuilder:
-        def __init__(self, build_id, name, version, arch):
+        def __init__(self, build_id, name, version, arch, snapshot):
             self.data = {}
             self.data['build'] = {}
             self.data['build']['id'] = build_id
             self.data['build']['name'] = name
             self.data['build']['version'] = str(version)
```
mch2, line 43: Thoughts on also appending -SNAPSHOT to this version string if its a snapshot? 

This would have the benefit of automatically setting tar builds with -SNAPSHOT

bundle-workflow/scripts/bundle-build/standard-gradle-build/build.sh
```diff
@@ -5,13 +5,63 @@
...
+    echo "Error: You must specify the OpenSearch version"
+    usage
+    exit 1
+fi
+
+[[ "$SNAPSHOT" == "true" ]] && VERSION=$VERSION-SNAPSHOT
+[ -z "$OUTPUT" ] && OUTPUT=artifacts
+
+mkdir -p $OUTPUT
+echo ./gradlew assemble --no-daemon --refresh-dependencies -DskipTests=true -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT
```
mch2, line 59: is this echo intentional?

mch2 (approved):
lgtm, few minor questions.

I wouldn't worry too much about other repos not tagging snapshot builds unless they want their artifacts to end up in our maven snapshot repo.

bundle-workflow/python/assemble.py
```diff
@@ -1,20 +1,17 @@
...
+import argparse
 from assemble_workflow.bundle import Bundle
 from assemble_workflow.bundle_recorder import BundleRecorder
 from manifests.build_manifest import BuildManifest
 
-if (len(sys.argv) < 2):
-    print("Build an OpenSearch Bundle")
-    print("usage: assemble.sh /path/to/build_manifest")
-    exit(1)
+parser = argparse.ArgumentParser(description = "Assembly an OpenSearch Bundle")
```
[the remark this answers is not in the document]
dblock: Duh.

bundle-workflow/scripts/bundle-build/standard-gradle-build/build.sh
dblock: Yes, cause I can never tell what we're running where.

bundle-workflow/python/build_workflow/build_recorder.py
dblock: Added.

bundle-workflow/python/build_workflow/builder.py
```diff
@@ -24,8 +24,9 @@ def __init__(self, component_name, git_repo, component_scripts_path, default_bui
         self.component_scripts_path = component_scripts_path
         self.default_build_script = default_build_script
         self.build_recorder = build_recorder
+        self.output_path = 'artifacts'
```
camerski, line 27: Make this a constant declared at the class level:

```
class Builder:
  OUTPUT_PATH = 'artifacts'
```

camerski (approved):
This change looks fine, just a couple of suggested tweaks for the future.

There is, however, a lot going on in this PR: Changing from sys.argv to ArgumentParser, plumbing the snapshot flag through, changing the 'artifacts' directory name from a hard-coded string to a variable, changing the way manifests read from files, updating the component build scripts to consume the snapshot flag, some inconsequential style changes (such as putting `then` on the same line as the `if` in a shell script). Most of these changes are unrelated.

In future, consider splitting unrelated changes into their own PRs in order to reduce cognitive load on the reviewers, and to ensure that each commit is a small, self-contained change that can be easily reverted if it turns out there is a problem.

bundle-workflow/scripts/bundle-build/standard-gradle-build/build.sh
dblock: Great point. I'll remove the echo for now and address whether we want to `set -x -e` everywhere in a separate PR.

bundle-workflow/python/build_workflow/builder.py
dblock: I think this actually should become input as a fully qualified path, so you can redirect things and not always create a hard-coded artifacts dir from where you run things. I didn't want to pollute this PR with that, though.

PR #193, opened 2021-08-12
Copy opensearch-tar-install.sh installation script into bundle

### LINKED ISSUE #192: Bundle system - copy opensearch-tar-install script into bundle
The [opensearch-tar-install](https://github.com/opensearch-project/opensearch-build/blob/main/release/tar/linux/opensearch-tar-install.sh) script needs to be included with assembled bundles similar to [this](https://github.com/opensearch-project/opensearch-build/blob/main/release/tar/linux/opensearch-tar-build.sh#L156) shell script.

mch2:

### Description
This PR copies opensearch-tar-install.sh into an assembled bundle.
 
### Issues Resolved
closes #192 
 
### Check List

### DIFF
--- a/bundle-workflow/python/assemble.py
+++ b/bundle-workflow/python/assemble.py
@@ -6,6 +6,7 @@
 import os
 import tempfile
 import argparse
+import shutil
 from assemble_workflow.bundle import Bundle
 from assemble_workflow.bundle_recorder import BundleRecorder
 from manifests.build_manifest import BuildManifest
@@ -13,6 +14,12 @@ from manifests.build_manifest import BuildManifest
 parser = argparse.ArgumentParser(description = "Assemble an OpenSearch Bundle")
 parser.add_argument('manifest', type = argparse.FileType('r'), help="Manifest file.")
 args = parser.parse_args()
+
+tarball_installation_script = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../../release/tar/linux/opensearch-tar-install.sh')
+
+if not os.path.isfile(tarball_installation_script):
+    print(f'No installation script found at path: {tarball_installation_script}')
+    exit(1)
 
 build_manifest = BuildManifest.from_file(args.manifest)
 build = build_manifest.build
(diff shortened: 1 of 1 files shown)


camerski (approved):

dblock:
Works but feels hacky. Can we put this into a bundle manifest to pickup?

dblock (approved):

mch2:
> Works but feels hacky. Can we put this into a bundle manifest to pickup?

I'm not sure a manifest is the right spot for this, the only static manifest we can reference this in is the Input to the build system.  I could pass along the file to bundle.py and have it copied there?

dblock:
> > Works but feels hacky. Can we put this into a bundle manifest to pickup?
> 
> I'm not sure a manifest is the right spot for this, the only static manifest we can reference this in is the Input to the build system. I could pass along the file to bundle.py and have it copied there?

Why don't you like including it in the build manifest? It's the "installer script" that should appear in bundle once it's produced. So it's another artifact like a plugin IMO.

mch2:
> > > Works but feels hacky. Can we put this into a bundle manifest to pickup?
> > 
> > 
> > I'm not sure a manifest is the right spot for this, the only static manifest we can reference this in is the Input to the build system. I could pass along the file to bundle.py and have it copied there?
> 
> Why don't you like including it in the build manifest? It's the "installer script" that should appear in bundle once it's produced. So it's another artifact like a plugin IMO.

I'm not a fan because they are two separate steps.  The input manifest describes what to build and should only include the components that we want to build.  The build system will produce the min tar and the plugins that need to go into a bundle.  It's entirely possible that someone would want to build those components and not produce a bundle. The assembly process will gather those components and assemble a bundle if desired.  It seems odd to me to include a script required in the assembly process as input to the build system.

dblock:
First, this PR is good to go. Don't hold anything.

> > Why don't you like including it in the build manifest? It's the "installer script" that should appear in bundle once it's produced. So it's another artifact like a plugin IMO.
> 
> I'm not a fan because they are two separate steps. The input manifest describes what to build and should only include the components that we want to build. The build system will produce the min tar and the plugins that need to go into a bundle. It's entirely possible that someone would want to build those components and not produce a bundle, and install whichever plugins they choose. The assembly process will gather those components and assemble a bundle if desired. It seems odd to me to include a script required in the assembly process as input to the build system.

I see. It sounds like we want "additional artifacts" to be "assembled into the bundle". Maybe we need an assembly manifest to be supplied along with the output build manifest?

PR #197, opened 2021-08-12
Moved source code from python to src.

dblock:

### Description

It's strange to call a source code folder by the language in which the code is written. I propose `src`.
  
### Check List


dblock:
> We have source code written in at least two languages: Python and Shell. Where do you propose putting the shell code? Will it also go in `src`?

If there was such a thing, yes, but I think we actually don't.

The way I see it is that today we have "scripts you can invoke", such as `build.sh` and `assemble.sh` in the root of `bundle-workflow` because they can be invoked (not because they are bash scripts), and "scripts that override some behavior that otherwise belongs in projects", which is in `scripts` (not called `bash`).

camerski:
> > We have source code written in at least two languages: Python and Shell. Where do you propose putting the shell code? Will it also go in `src`?
> 
> If there was such a thing as scripts written in shell, yes, but I think we actually don't have any other than scripts that can be invoked.
> 
> The way I see it is that today we have "scripts you can invoke", such as `build.sh` and `assemble.sh` in the root of `bundle-workflow` because they can be invoked (not because they are bash scripts), and "scripts that override some behavior that otherwise belongs in projects", which is in `scripts` (not called `bash`).

The meta-question here is "why is `scripts` not a subdirectory of `src`?" It's all source code.

In any case, I would ask that you defer this change while the codebase is under heavy development. This is a disruptive change that impacts multiple developers with changes in flight, and the cost of rebasing and fixing the inevitable merge conflicts is pretty high. In a week or two we'll be finished with the bulk of those changes and it would make sense to do a big rename at that point.

dblock:
> The meta-question here is "why is `scripts` not a subdirectory of `src`?" It's all source code.

Because it's "data", as in overrides/configures some behavior like a manifest, and not part of the source code of the tool.

> In any case, I would ask that you defer this change while the codebase is under heavy development. This is a disruptive change that impacts multiple developers with changes in flight, and the cost of rebasing and fixing the inevitable merge conflicts is pretty high. In a week or two we'll be finished with the bulk of those changes and it would make sense to do a big rename at that point.

No problem. I'll leave this open.

dblock:
Closing for now, it's just adding noise.

PR #218, opened 2021-08-15
Add --component to build a single component.

dblock:

### Description

Got a bit tired of constantly commenting out components in the manifest while fixing individual components. You can now pass in `--component` and it will build just that component, skipping all the others. The build output also shows the last failure and gives a copy-paste to rebuild the last component that failed. 
 
### Check List

### DIFF
--- a/bundle-workflow/README.md
+++ b/bundle-workflow/README.md
@@ -38,10 +38,10 @@ Build Step:
 
 The following options are available.
 
-| name               | description                                                         |
-|--------------------|---------------------------------------------------------------------|
-| --snapshot         | Build a snapshot instead of a release artifact, default is `false`. |
-| --component [name] | Build a single component by name, e.g. `--component common-utils`.  |
+| name               | description                                                           |
+|--------------------|-----------------------------------------------------------------------|
+| --snapshot         | Build a snapshot instead of a release artifact, default is `false`.   |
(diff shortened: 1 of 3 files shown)


mch2:
I like the intent here but I think we need to clarify that this is to "Rebuild" a single component.  Right now build.sh scripts rely on their dependencies being built in-order before them and stored in maven local.  This was a trade-off that was made over each build script cloning and building their deps so that we can guarantee all components are built from the same unreleased version.

dblock:
> I like the intent here but I think we need to clarify that this is to "Rebuild" a single component. Right now build.sh scripts rely on their dependencies being built in-order before them and stored in maven local. This was a trade-off that was made over each build script cloning and building their deps so that we can guarantee all components are built from the same unreleased version.

Are you suggesting I adjust wording, or do you think more is needed, such as renaming the option?

mch2:
> > I like the intent here but I think we need to clarify that this is to "Rebuild" a single component. Right now build.sh scripts rely on their dependencies being built in-order before them and stored in maven local. This was a trade-off that was made over each build script cloning and building their deps so that we can guarantee all components are built from the same unreleased version.
> 
> Are you suggesting I adjust wording, or do you think more is needed, such as renaming the option?

I am suggesting adjusting the wording in the README, the option name is ok.

mch2 (approved):

camerski (approved):

Issue #136, opened 2021-08-03
Create Release workflow to move artifacts from staging to release


mch2:
This workflow should run in Jenkins infra but invoke a script in this repository.

This workflow should:
1. Sign all artifacts
2. Promote bundle artifacts from /release-candidates to /release folder.
3. Promote maven artifacts to a temporary staging repo in Sonatype.  This temp repo will be promoted on release day manually.

mch2:
A few notes here:

The input to this workflow should be a particular build id.
We will want to take a particular build id and fetch all S3 artifacts associated with it from /release-candidates and promote to /releases
I believe maven promotion could happen by using the nexus-staging-maven-plugin.
Example:

```
      mvn --settings=${mvn_settings}
        org.sonatype.plugins:nexus-staging-maven-plugin:1.6.5:deploy-staged-repository
        -DrepositoryDirectory=<some-local-path>                                               
        -DnexusUrl=http://localhost:8081/nexus/service/local/staging/deploy/maven2     
        -DserverId=test                                                       
        -DstagingProfileId=test
```

The artifacts from S3 will include a maven repo with the artifacts that should be promoted.  I believe we could just pass that as our repositoryDirectory, if now we will need to use the deploy plugin and create a new tmp local repo.

mch2:
This is created - see #348 for status on moving this to production.

PR #226, opened 2021-08-16
Signing artifacts from manifest

Poojita-Raj:
Signed-off-by: Poojita-Raj <poojiraj@amazon.com>

### Description
As part of release workflow, we can now use a manifest file to sign the needed bundle artifacts with the signed artifacts residing in /signed_artifacts. 
 
### Issues Resolved
Related to #136 
 
### Check List

### DIFF
--- a/bundle-workflow/README.md
+++ b/bundle-workflow/README.md
@@ -57,3 +57,24 @@ Artifacts will be updated as follows.
 #### Custom Install Scripts
 
 You can perform additional plugin install steps by adding an `install.sh` script. By default the tool will look for a script in [scripts/bundle-build/components](scripts/bundle-build/components), then default to a noop version implemented in [scripts/bundle-build/standard-gradle-build/install.sh](scripts/bundle-build/standard-gradle-build/install.sh).
+
+Step 3: Signing [optional]
+
+The signing step takes the manifest file created from the build step and signs all its component artifacts using the opensearch-signer-client. The input requires a path to the build manifest and is expected to be inside the artifacts directory with the same directories mentioned in the build step. 
+
+The following options are available. 
+
+| name        | description                                                         |
+|-------------|---------------------------------------------------------------------|
+| --component | The component name of the component whose artifacts will be signed  |
+| --type      | The artifact type to be signed. Currently one of 3 options:         |
+|             | 	1. plugins                                                  |
+|             |         2. maven                                                    |
+|             |         3. bundle                                                   |
+
+The signed artifacts (<artifact>.asc) will be found in the same location as the original artifact. 
+
+Signing step (to sign all artifacts):
+```bash
+./bundle_workflow/sign.sh artifacts/manifest.yml
+```
--- a/bundle-workflow/README.md
+++ b/bundle-workflow/README.md
@@ -3,6 +3,7 @@
         - [Custom Build Scripts](#custom-build-scripts)
     - [Assemble the Bundle](#assemble-the-bundle)
         - [Custom Install Scripts](#custom-install-scripts)
+    - [Signing Artifacts](#signing-artifacts)
 
 ## OpenSearch Bundle Workflow
 
@@ -58,9 +59,9 @@ Artifacts will be updated as follows.
 
 You can perform additional plugin install steps by adding an `install.sh` script. By default the tool will look for a script in [scripts/bundle-build/components](scripts/bundle-build/components), then default to a noop version implemented in [scripts/bundle-build/standard-gradle-build/install.sh](scripts/bundle-build/standard-gradle-build/install.sh).
 
-Step 3: Signing [optional]
+### Signing Artifacts
 
-The signing step takes the manifest file created from the build step and signs all its component artifacts using the opensearch-signer-client. The input requires a path to the build manifest and is expected to be inside the artifacts directory with the same directories mentioned in the build step. 
+The signing step (optional) takes the manifest file created from the build step and signs all its component artifacts using the opensearch-signer-client. The input requires a path to the build manifest and is expected to be inside the artifacts directory with the same directories mentioned in the build step. 
 
 The following options are available. 
 
@@ -68,9 +69,9 @@ The following options are available.
 |-------------|---------------------------------------------------------------------|
 | --component | The component name of the component whose artifacts will be signed  |
 | --type      | The artifact type to be signed. Currently one of 3 options:         |
-|             | 	1. plugins                                                  |
-|             |         2. maven                                                    |
-|             |         3. bundle                                                   |
+|             | 	1. plugins                                                      |
+|             |     2. maven                                                        |
+|             |     3. bundle                                                       |
 
 The signed artifacts (<artifact>.asc) will be found in the same location as the original artifact. 
 
--- a/bundle-workflow/README.md
+++ b/bundle-workflow/README.md
@@ -65,13 +65,10 @@ The signing step (optional) takes the manifest file created from the build step 
 
 The following options are available. 
 
-| name        | description                                                         |
-|-------------|---------------------------------------------------------------------|
-| --component | The component name of the component whose artifacts will be signed  |
-| --type      | The artifact type to be signed. Currently one of 3 options:         |
-|             | 	1. plugins                                                      |
-|             |     2. maven                                                        |
-|             |     3. bundle                                                       |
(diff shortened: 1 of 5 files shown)


bundle-workflow/python/sign.py
```diff
@@ -0,0 +1,37 @@
...
+from manifests.bundle_manifest import BundleManifest
+
+parser = argparse.ArgumentParser(description = "Sign artifacts")
+parser.add_argument('manifest', type = argparse.FileType('r'), help="Manifest file.")
+args = parser.parse_args()
+
+def sign(filename):
+    signatureFile = filename + ".asc"
+    signing_cmd = './opensearch-signer-client -i ' + filename + ' -o ' + signatureFile + ' -p pgp'
+    sp.run(signing_cmd.split(),cwd='opensearch-signer-client/src')
```
dblock, line 17: I believe this doesn't fail if the command fails, you'll need to `check_output`.

bundle-workflow/python/sign.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+
+import os
+import subprocess as sp
```
dblock, line 4: Let's standardize and just leave this as `subprocess`.

bundle-workflow/python/sign.py
```diff
@@ -0,0 +1,37 @@
...
+import tempfile
+import uuid
+import argparse
+from manifests.bundle_manifest import BundleManifest
+
+parser = argparse.ArgumentParser(description = "Sign artifacts")
+parser.add_argument('manifest', type = argparse.FileType('r'), help="Manifest file.")
+args = parser.parse_args()
+
+def sign(filename):
```
dblock, line 14: Refactor into a `Signer` class, similar to how we have `Builder`?

bundle-workflow/python/sign.py
```diff
@@ -0,0 +1,37 @@
...
+    signatureFile = filename + ".asc"
+    signing_cmd = './opensearch-signer-client -i ' + filename + ' -o ' + signatureFile + ' -p pgp'
```
dblock, line 16: You're building a cmd, then splitting it later. Build it as an array. Avoids quoting issues, too.

bundle-workflow/python/sign.py
```diff
@@ -0,0 +1,37 @@
...
+    verify_cmd = 'gpg --verify-files ' + signatureFile
+    sp.run(verify_cmd.split(),cwd='opensearch-signer-client/src')
+    sp.run(['mv',signatureFile,'signed_artifacts/'])
+
+manifest = BundleManifest.from_file(args.manifest)
+output_dir = os.path.join(os.getcwd(), 'signed_artifacts')
+os.makedirs(output_dir, exist_ok = True)
+build_id = os.getenv('OPENSEARCH_BUILD_ID', uuid.uuid4().hex)
+
+os.system('git clone https://github.com/opensearch-project/opensearch-signer-client.git')
```
dblock, line 27: Wrap this underneath signer, and use existing Git classes to avoid having to do explicit `git clone` or `rm -rf` signer at the end.

dblock (changes requested):
Some should have's.

bundle-workflow/python/signing_workflow/signer.py
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+'''
+This class is responsible for signing an artifact using the OpenSearch-signer-client and verifying its signature.
+The signed artifacts will be found in the same location as the original artifacts.
+'''
+class Signer:
+    def __init__(self, git_repo):
+        self.git_repo = git_repo
+        self.git_repo.execute("./bootstrap","src")
```
mch2, line 13: if these are executing within the repo, you should be able to execute ./src/bootstrap and not need to update the signature?

bundle-workflow/python/signing_workflow/signer.py
```diff
@@ -0,0 +1,24 @@
...
+        self.git_repo.execute("rm config.cfg","src")
+
+    def sign(self, filename):
+        signatureFile = filename + ".asc"
```
mch2, line 17: nit - signature_file

bundle-workflow/README.md
```diff
@@ -48,4 +48,26 @@ Bundle step:
 ```
 ### Scripts
 
-Each component build relies on a `build.sh` script that is used to prepare bundle artifacts for a particular bundle version that takes two arguments: version and target architecture. By default the tool will look for a script in [scripts/bundle-build/components](scripts/bundle-build/components), then in the checked-out repository in `build/build.sh`, then default to a Gradle build implemented in [scripts/bundle-build/standard-gradle-build](scripts/bundle-build/standard-gradle-build).
\ No newline at end of file
+Each component build relies on a `build.sh` script that is used to prepare bundle artifacts for a particular bundle version that takes two arguments: version and target architecture. By default the tool will look for a script in [scripts/bundle-build/components](scripts/bundle-build/components), then in the checked-out repository in `build/build.sh`, then default to a Gradle build implemented in [scripts/bundle-build/standard-gradle-build](scripts/bundle-build/standard-gradle-build).
+
+Step 3: Signing [optional]
```
dblock, line 53: Rebase, the README has undergone some structural surgery.

bundle-workflow/python/git/git_repository.py
```diff
@@ -25,6 +25,6 @@ def __post_init__(self):
         self.sha = subprocess.check_output(['git', 'rev-parse', 'HEAD'], cwd = self.dir.name).decode().strip()
         print(f'Checked out {self.url}@{self.ref} into {self.dir.name} at {self.sha}')
 
-    def execute(self, command):
+    def execute(self, command, dirname=""):
```
dblock, line 28: This looks like a subdirectory. Let's call it `subdirname` or something like that? 

Would it work with `None` instead of `""`? From the interface POV it's strange to pass an empty string when otherwise we mean `None`. Maybe just `if dirname: ...join` to make it explicit.

Nit: most code is styling with spaces around `=`, I know we want some kind of automatic linting, but in the meantime let's try to make things at least look a little consistent?

bundle-workflow/python/sign.py
```diff
@@ -1,37 +1,31 @@
...
-def sign(filename):
-    signatureFile = filename + ".asc"
-    signing_cmd = './opensearch-signer-client -i ' + filename + ' -o ' + signatureFile + ' -p pgp'
-    sp.run(signing_cmd.split(),cwd='opensearch-signer-client/src')
-    verify_cmd = 'gpg --verify-files ' + signatureFile
-    sp.run(verify_cmd.split(),cwd='opensearch-signer-client/src')
-    sp.run(['mv',signatureFile,'signed_artifacts/'])
+manifest = BuildManifest.from_file(args.manifest)
+basepath = os.path.dirname(os.path.abspath(args.manifest.name))
+git_repo = GitRepository("https://github.com/opensearch-project/opensearch-signer-client.git", "HEAD")
```
dblock, line 17: nit: maybe `opensearch_signer_git_repo`, but also this feels like an implementation detail and maybe belongs inside `Signer`?

bundle-workflow/python/sign.py
```diff
@@ -1,37 +1,31 @@
...
-os.makedirs(output_dir, exist_ok = True)
-build_id = os.getenv('OPENSEARCH_BUILD_ID', uuid.uuid4().hex)
-
-os.system('git clone https://github.com/opensearch-project/opensearch-signer-client.git')
-os.system('cp ./config.cfg ./opensearch-signer-client/src/')
-sp.run(['./bootstrap'],cwd='opensearch-signer-client/src')
-
-print(f'Signing {manifest.build.name} into {output_dir}')
 for component in manifest.components:
+    if (not args.component) or ( args.component and component.name == args.component):
```
dblock, line 21: Probably a bit clearer using the opposite condition and use `continue`. 

Like [here](https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/python/build.py#L36)

dblock (changes requested):
Consider nice-to/should-have's below, nothing major.
You do need to rebase these changes.

dblock (approved):
One more thing in readme before merging please.

mch2 (approved):

PR #228, opened 2021-08-16
Allow GitRepository to check out into a named directory

camerski:
### Description
Add a command-line parameter to "test.py" to allow for a fixed working directory. This makes debugging test script failures easier.

Fix typo in looking up the bundle tarball location.

### Testing
Manually verified that test.py and build.py still work correctly.
 
### Check List

### DIFF
--- a/bundle-workflow/python/git/git_repository.py
+++ b/bundle-workflow/python/git/git_repository.py
@@ -1,30 +1,39 @@
 # Copyright OpenSearch Contributors.
 # SPDX-License-Identifier: Apache-2.0
 
+import os
 import subprocess
 import tempfile
-from dataclasses import dataclass
+import shutil
 
-'''
-This class checks out a Git repository at a particular ref into a temporary directory.
-Clients can obtain the actual commit ID by querying the "sha" attribute, and the temp directory name with "dir.name".
-'''
-@dataclass
 class GitRepository:
-    url: str
-    ref: str
-    sha: str = None
-    dir: tempfile.TemporaryDirectory = None
+    '''
+    This class checks out a Git repository at a particular ref into a named directory (or temporary a directory if no named directory is given). Temporary directories will be automatically deleted when the GitRepository object goes out of scope; named directories will be deleted and recreated when the GitRepository is constructed.
+    Clients can obtain the actual commit ID by querying the "sha" attribute, and the temp directory name with "dir".
+    '''
+    def __init__(self, url, ref, directory = None):
+        self.url = url
+        self.ref = ref
+        if directory is None:
+            self.temp_dir = tempfile.TemporaryDirectory()
+            self.dir = self.temp_dir.name
+        else:
+            self.dir = directory
+            print(f'Deleting {self.dir}')
+            shutil.rmtree(self.dir, ignore_errors = True)
+            os.makedirs(self.dir, exist_ok = True)
 
-    def __post_init__(self):
-        self.dir = tempfile.TemporaryDirectory()
-        self.execute(f'git init')
-        self.execute(f'git remote add origin {self.url}')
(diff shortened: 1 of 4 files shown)


peternied (approved):

dblock:
I don't love this because leaving files around is a side-effect of defining the working directory. Why would the working directory behave differently from the default temporary directory? Instead, maybe we should have a `--preserve` or something like that that makes sure those folders aren't deleted. It has the positive side effect that multiple runs always start clean, and produce a new folder every time, and don't execute on top of each-other.

bundle-workflow/python/git/git_repository.py
```diff
@@ -1,30 +1,39 @@
...
 class GitRepository:
-    url: str
-    ref: str
-    sha: str = None
-    dir: tempfile.TemporaryDirectory = None
+    '''
+    This class checks out a Git repository at a particular ref into a named directory (or temporary a directory if no named directory is given). Temporary directories will be automatically deleted when the GitRepository object goes out of scope; named directories will be deleted and recreated when the GitRepository is constructed.
+    Clients can obtain the actual commit ID by querying the "sha" attribute, and the temp directory name with "dir".
+    '''
+    def __init__(self, url, ref, directory = None):
```
mch2, line 14: nit - doesn't the doc need to live after the method declaration?

bundle-workflow/python/test_workflow/integ_test_suite.py
```diff
@@ -6,9 +6,9 @@ def __init__(self, name, repo):
         self.repo = repo
 
     def execute(self, cluster):
-        script = self.repo.dir.name + "/integtest.sh"
+        script = self.repo.dir + "/integtest.sh"
```
mch2, line 9: nit, os.path.join?

mch2 (approved):
lgtm, some nits.  We should extend this option to build.py as well.

bundle-workflow/python/git/git_repository.py
camerski: That documentation is for the class

camerski:
> I don't love this because leaving files around is a side-effect of defining the working directory. Why would the working directory behave differently from the default temporary directory? Instead, maybe we should have a `--preserve` or something like that that makes sure those folders aren't deleted? That option would have the positive side effect that multiple runs always start clean, and produce a new folder every time, and don't execute on top of each-other. I am also worried that `--workdir .` or `--wordir ~` could overwrite files and create a ton of unexpectedly devastating problems.

Totally agree, but unfortunately Python's TempDirectory class does not have a way to do this (or, at least, not that I could find).

bundle-workflow/python/test_workflow/integ_test_suite.py
camerski: That's fixed in https://github.com/opensearch-project/opensearch-build/pull/230

dblock:
> > I don't love this because leaving files around is a side-effect of defining the working directory. Why would the working directory behave differently from the default temporary directory? Instead, maybe we should have a `--preserve` or something like that that makes sure those folders aren't deleted? That option would have the positive side effect that multiple runs always start clean, and produce a new folder every time, and don't execute on top of each-other. I am also worried that `--workdir .` or `--wordir ~` could overwrite files and create a ton of unexpectedly devastating problems.
> 
> Totally agree, but unfortunately Python's TempDirectory class does not have a way to do this (or, at least, not that I could find).

Like so:

```python
from contextlib import contextmanager
import tempfile
import shutil

@contextmanager
def TemporaryDirectory(delete = True):
    name = tempfile.mkdtemp()
    try:
        yield name
    finally:
        if delete:
            shutil.rmtree(name)
```

camerski:
Sure, someone can do that in a new PR since this one is already merged.

Issue #205, opened 2021-08-12
Implement LocalTestCluster


camerski:
Subtask of https://github.com/opensearch-project/opensearch-build/issues/177

Implement the LocalTestCluster class so we can run integ tests using an on-box test cluster.

camerski:
Forgot we also need to create a no-security cluster. Reopening.

PR #235, opened 2021-08-16
Implement LocalTestCluster

camerski:
### Description
Implements LocalTestCluster that downloads, unpacks, and starts a local OpenSearch node (via opensearch-tar-install.sh), and tears it down again when the tests are finished.

### Testing
Manual testing. I was able to run the SQL plugin integ tests.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/205
 
### Check List

### DIFF
--- a/bundle-workflow/python/test_workflow/test_cluster.py
+++ b/bundle-workflow/python/test_workflow/test_cluster.py
@@ -45,7 +45,8 @@ class LocalTestCluster:
             except TimeoutExpired:
                 print('Process failed to terminate even after SIGKILL')
                 raise
-        print(f'Process terminated with exit code {self.process.returncode}')
-        self.stdout.close()
-        self.stderr.close()
-        self.process = None
+        finally:
+            print(f'Process terminated with exit code {self.process.returncode}')
+            self.stdout.close()
+            self.stderr.close()
+            self.process = None
--- a/bundle-workflow/python/test_workflow/test_cluster.py
+++ b/bundle-workflow/python/test_workflow/test_cluster.py
@@ -1,6 +1,28 @@
+import os
+import tempfile
+import urllib.request
+import shutil
+import subprocess
+
 class LocalTestCluster:
-    def __init__(self, bundle_location):
-        pass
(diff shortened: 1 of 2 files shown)


bundle-workflow/python/test_workflow/test_cluster.py
```diff
@@ -9,4 +31,21 @@ def port(self):
...
+            print('Process did not terminate after 10 seconds. Sending SIGKILL')
+            self.process.kill()
+            try:
+                print('Waiting for process to terminate')
+                self.process.wait(10)
+            except TimeoutExpired:
+                print('Process failed to terminate even after SIGKILL')
+                raise
+        print(f'Process terminated with exit code {self.process.returncode}')
+        self.stdout.close()
```
dblock, line 49: If the process times out on termination, you'll be leaking a handle.

dblock (approved):
If the caller forgets to call `destroy`, you'll leave everything behind, consider adopting a pattern with `yield` ala https://github.com/opensearch-project/opensearch-build/blob/948ace083bf8e5fc017db062d1104575f4186327/bundle-workflow/python/system/temporary_directory.py and get rid of `try/finally`.

bundle-workflow/python/test_workflow/test_cluster.py
camerski: Good catch

camerski:
> If the caller forgets to call `destroy`, you'll leave everything behind, consider adopting a pattern with `yield` ala https://github.com/opensearch-project/opensearch-build/blob/948ace083bf8e5fc017db062d1104575f4186327/bundle-workflow/python/system/temporary_directory.py and get rid of `try/finally`.

Good idea. I'll learn some more Python and address that in a later PR.

bundle-workflow/python/test_workflow/test_cluster.py
camerski: Fixed

setiah (approved):

PR #239, opened 2021-08-17
Add k-nn build script and configure knnlib during bundle step.

mch2:

### Description
Add a build.sh script for k-nn.  This is so that knnlib is properly built and placed into /libs.
Updates bundle step to configure this lib if knn is included as a plugin.

Tested this by executing the build & assembly steps and unpacking the built tarball.
lib is placed at the correct location and ./opensearch-tar-install.sh can find it.
 
### Issues Resolved
closes 149
 
### Check List

### DIFF
--- a/bundle-workflow/python/assemble_workflow/bundle.py
+++ b/bundle-workflow/python/assemble_workflow/bundle.py
@@ -38,6 +38,8 @@ class Bundle:
         for plugin in self.plugins:
             print(f'Installing {plugin.name}')
             self.install_plugin(plugin)
+            if plugin.name == 'k-NN':
+                self.copy_knnlib(plugin)
         self.installed_plugins = os.listdir(os.path.join(self.archive_path, 'plugins'))
 
     def install_plugin(self, plugin):
@@ -83,3 +85,12 @@ class Bundle:
 
     def get_min_bundle(self, build_components):
         return next(iter([c for c in build_components if "bundle" in c.artifacts]), None)
+
+    def copy_knnlib(self, component):
+        local_path = self.get_rel_path(component, 'libs')
+        if local_path:
+            knnlib = os.path.join(self.artifacts_dir, local_path)
+            dest_path = os.path.join(self.archive_path, 'plugins/opensearch-knn/knnlib')
+            os.makedirs(dest_path, exist_ok=True)
+            dest = os.path.join(dest_path, os.path.basename(knnlib))
+            shutil.copyfile(knnlib, dest)
--- a/bundle-workflow/scripts/bundle-build/components/k-NN/build.sh
+++ b/bundle-workflow/scripts/bundle-build/components/k-NN/build.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+
+# Copyright OpenSearch Contributors.
+# SPDX-License-Identifier: Apache-2.0
+
+set -ex
+
+function usage() {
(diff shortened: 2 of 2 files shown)


bundle-workflow/python/assemble_workflow/bundle.py
```diff
@@ -38,6 +38,8 @@ def install_plugins(self):
         for plugin in self.plugins:
             print(f'Installing {plugin.name}')
             self.install_plugin(plugin)
+            if plugin.name == 'k-NN':
```
dblock, line 41: I'm confused why this is necessary. Why isn't the installation of the plugin itself doing this copy?

mch2: There are a couple reasons here that I can see of why this is currently installed separately.

1. This lib is not included as part of the plugin zip distribution.  It is a separate component in addition to the zip. 
2. If it were included in the zip, we need to update the[ install cli ](https://github.com/opensearch-project/OpenSearch/blob/19e54d662488a183c1388e66cfe7463a24c6aa7c/distribution/tools/plugin-cli/src/main/java/org/opensearch/plugins/InstallPluginCommand.java#L856) to support this.

mch2: There is also an additional step for setting up performance analyzer in our existing script.  This one-off logic should be included during the install step as well.
https://github.com/opensearch-project/opensearch-build/blob/main/release/tar/linux/opensearch-tar-build.sh#L147

dblock: Let's find a more generic way to execute these? Maybe along with build.sh we can add an install.sh?

dblock (approved):
Open an issue for extracting the k-NN plugin custom installation and I'm good to merge.

mch2:
> Open an issue for extracting the k-NN plugin custom installation and I'm good to merge.

Done in #251.  I like the idea of adding an install.sh at a minimum.

bundle-workflow/python/assemble_workflow/bundle.py
gaiksaya: Not sure if plugin team can handle these using post/pre-installation script. Example: https://github.com/opensearch-project/performance-analyzer/blob/main/packaging/deb/postinst

gaiksaya (approved):
Assuming that you confirmed with k-NN team about creating the `.so` library whether the process/command is same both architecture (x64 and arm64)

bundle-workflow/python/assemble_workflow/bundle.py
camerski: `install.sh` would be a good workaround until the OpenSearch plugin installer provides a way for plugin zips to include and install native libraries.

For more "why is this necessary", take a look at https://github.com/opensearch-project/opensearch-build/blob/main/release/tar/linux/opensearch-tar-install.sh to see how LD_LIBRARY_PATH is set to consume the native library. TL;DR: OpenSearch plugins don't really have a good way to include native libraries.

camerski (approved):

bundle-workflow/python/assemble_workflow/bundle.py
dblock: Interesting! I opened https://github.com/opensearch-project/opensearch-plugins/issues/75

PR #240, opened 2021-08-17
Remove demo certificates when DISABLE_INSTALL_DEMO_CONFIG is set to true

danpawlik:
### Description
The demo certificates exists in official Opensearch container images.
Sometimes it may raise an issue to start properly the opensearch
service. In that case, if DISABLE_INSTALL_DEMO_CONFIG variable is set to
true, demo certificates will be removed before opensearch will start.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/254
 
### Check List

### DIFF
--- a/release/docker/config/opensearch/opensearch-docker-entrypoint.sh
+++ b/release/docker/config/opensearch/opensearch-docker-entrypoint.sh
@@ -71,6 +71,9 @@ done < <(env)
 # will run in.
 export OPENSEARCH_JAVA_OPTS="-Dopensearch.cgroups.hierarchy.override=/ $OPENSEARCH_JAVA_OPTS"
 
+if [[ "${DISABLE_INSTALL_DEMO_CONFIG}" =~ [tT]rue ]]; then
+    rm /usr/share/opensearch/config/{esnode-key.pem,esnode.pem,kirk-key.pem,kirk.pem,root-ca.pem}
+fi
 
 # Start up the opensearch and performance analyzer agent processes.
 # When either of them halts, this script exits, or we receive a SIGTERM or SIGINT signal then we want to kill both these processes.


dblock:
I understand where you're coming from @danpawlik, but why is the demo config not working in the first place? Shouldn't we fix that instead of adding a feature that we'll have to document and support forever to get rid of it?

danpawlik:
@dblock  thanks for reply. 
In my case, I can not run Opensearch without removing the demo certificates, so IMHO the container images are broken.
From proposed solution, number 1 seems to be the best.
Should I port the functionality and replace my commit ?

nebulon42:
@dblock not sure that your suggested 1. will work because the demo files are in the image from the beginning and not created by the specified script. At least I tried to replace the script by an empty file like suggested in https://github.com/opendistro-for-elasticsearch/opendistro-build/issues/10 and it didn't change anything. So I gather that the script is not executed/the files are already there. Also the creation timestamps for the demo certificates suggest that they have been created together with the image.

danpawlik:
any news @dblock  ?

dblock:
To me an environment variable should not be permanently disabling (deleting) configuration, especially anything security-related, demo or not, at runtime. If this script runs in a production environment, `DISABLE_INSTALL_DEMO_CONFIG` would disable that, too, btw.

I don't know much about demo images or how they are created, but I'll do my best to help. @peterzhuamazon can you point us to the right direction?

peterzhuamazon:
> To me an environment variable should not be permanently disabling (deleting) configuration, especially anything security-related, demo or not, at runtime. If this script runs in a production environment, `DISABLE_INSTALL_DEMO_CONFIG` would disable that, too, btw.
> 
> I don't know much about demo images, or how they are created, but I'll do my best to help. @peterzhuamazon can you point us to the right direction?
> 
> It looks like the official images ship with a demo security certificate configuration? I like the idea that they would be generated when the container starts the first time. Either way, need something more robust than an environment variable to replace them.

We are doing the same practice by running install demo configuration script which comes from security repo. This script will be run either use tarball or use docker images.
I have yet to be able to reproduce the errors.
https://github.com/opensearch-project/security/blob/main/tools/install_demo_configuration.sh

Also, we change the way this script is run just like @dblock pointed out.
We will always run install demo configuration with a one time setup script here:
https://github.com/opensearch-project/opensearch-build/blob/main/scripts/opensearch-onetime-setup.sh#L16

If needed we can change this behavior back to this:
https://github.com/opendistro-for-elasticsearch/opendistro-build/blob/aabb00fb349e1eb1b916ed68c22fb5ef8be8abc7/elasticsearch/docker/build/elasticsearch/bin/docker-entrypoint.sh#L85

Let me know what you guys think about it.

Thanks.

nebulon42:
@peterzhuamazon I think reverting to the previous behaviour will be the only way without larger rework on the container images side. Only downside of this approach is that the changes of this script will go away on container termination unless the whole `/usr/share/opensearch` directory being persisted somewhere. But it ensures that the demo certificates are ephemeral which IMO is an important point. You can always delete the certs from disk in a tarball install but you cannot get them out of a container image without extending the image (or with this PR but I think this is only a workaround).

peterzhuamazon:
Got it. Let me work on it tomorrow. Thanks for feedbacks @nebulon42.

peterzhuamazon:
@nebulon42 you mind close this PR so I can open a new one with the original behavior?
Thanks.

nebulon42:
I think @peterzhuamazon meant to address @danpawlik in the last comment.

danpawlik:
Thanks @peterzhuamazon  and @nebulon42.
Closing PR.

peterzhuamazon:
> I think @peterzhuamazon meant to address @danpawlik in the last comment.

Yes, I apologize. And thanks both @danpawlik and @nebulon42 will send the new PR later.

peterzhuamazon:
Fixed now: https://github.com/opensearch-project/opensearch-build/pull/436

PR #246, opened 2021-08-17
Extract TestCluster interface

camerski:
Formalizing the TestCluster interface now that we have multiple people working on TestCluster implementations.

### Testing
Manually verified that `test.py` still works.

### Check List

### DIFF
--- a/bundle-workflow/python/test_workflow/test_cluster.py
+++ b/bundle-workflow/python/test_workflow/test_cluster.py
@@ -1,52 +1,41 @@
-import os
-import tempfile
-import urllib.request
-import shutil
-import subprocess
+import abc
 
-class LocalTestCluster:
-    def __init__(self, bundle_manifest):
-        self.manifest = bundle_manifest
-        self.work_dir = tempfile.TemporaryDirectory()
+class TestCluster(abc.ABC):
+    '''
+    Abstract base class for all types of test clusters.
+    '''
 
+    @abc.abstractmethod
     def create(self):
-        print(f'Creating local test cluster in {self.work_dir.name}')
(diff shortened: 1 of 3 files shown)


bundle-workflow/python/test_workflow/test_cluster.py
```diff
@@ -3,8 +3,26 @@
 import urllib.request
 import shutil
 import subprocess
+import abc
 
-class LocalTestCluster:
+class TestCluster(abc.ABC):
```
owaiskazi19, line 8: Can we pull the interface in a separate file? The other clusters like PerfTest, IntegTest can use the same

camerski: Sure

owaiskazi19 (approved):

bundle-workflow/python/test_workflow/test_cluster.py
```diff
@@ -1,52 +1,41 @@
...
-                self.process.wait(10)
-            except TimeoutExpired:
-                print('Process failed to terminate even after SIGKILL')
-                raise
-        finally:
-            print(f'Process terminated with exit code {self.process.returncode}')
-            self.stdout.close()
-            self.stderr.close()
-            self.process = None
+class ClusterCreationException(Exception):
```
owaiskazi19, line 37: nit: Can't we use this class to throw error while creating the cluster?

camerski: That's the idea, yeah. I'll update LocalTestCluster to throw this exception when it fails.

gaiksaya (approved):

PR #248, opened 2021-08-17
Do not nuke existing folders in git checkout.

dblock:

### Description

I did not enjoy accidentally using a path with files I cared about ;(
  
### Check List

### DIFF
--- a/bundle-workflow/python/git/git_repository.py
+++ b/bundle-workflow/python/git/git_repository.py
@@ -19,9 +19,7 @@ class GitRepository:
             self.dir = self.temp_dir.name
         else:
             self.dir = directory
-            print(f'Deleting {self.dir}')
-            shutil.rmtree(self.dir, ignore_errors = True)
-            os.makedirs(self.dir, exist_ok = True)
+            os.makedirs(self.dir, exist_ok = False)
 
         # Check out the repository
         self.execute(f'git init', True)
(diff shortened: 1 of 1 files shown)


gaiksaya:
I think the class description needs to be updated as well. https://github.com/opensearch-project/opensearch-build/pull/248/files#diff-667ead02eb7cd0e63a1df908d4cc8599c54def808ce002a4fc73b1f5f06ccab2R11

bundle-workflow/python/git/git_repository.py
```diff
@@ -19,9 +19,7 @@ def __init__(self, url, ref, directory = None):
             self.dir = self.temp_dir.name
         else:
             self.dir = directory
-            print(f'Deleting {self.dir}')
```
peternied, line 22: I think there was utility in being sure the folder was cleaned up? Maybe we add a default that causes this to barf unless you've set `allow-nuke=true`

peternied (approved):

bundle-workflow/python/git/git_repository.py
dblock: We checkout into a temporary folder, so these actually do get cleaned up. Making GitRepository act like a temporary folder itself seems unnecessary, but I would be fine if we needed that. The problem here is that this class is cleaning up something it didn't create!

dblock:
> I think the class description needs to be updated as well

Done.

mch2 (approved):

gaiksaya (approved):

PR #249, opened 2021-08-17
Build index-management that requires artifacts from previous components.

dblock:

### Description

Adds the capability for a script to consume previously built components. This is needed for index-management that needs a job-scheduler and notifications plugin to build. 

If you know a way to disable certain tests in that one to avoid needing these, I'd probably prefer that.
  
### Check List

### DIFF
--- a/bundle-workflow/scripts/bundle-build/components/index-management/build.sh
+++ b/bundle-workflow/scripts/bundle-build/components/index-management/build.sh
@@ -55,8 +55,26 @@ if [ -z "$VERSION" ]; then
     exit 1
 fi
 
+for plugin in job-scheduler notifications; do
+    PLUGIN_VERSION=$VERSION.0
+    [[ "$SNAPSHOT" == "true" ]] && PLUGIN_VERSION=$PLUGIN_VERSION-SNAPSHOT
+    ZIP=$DEST/plugins/opensearch-$plugin-$PLUGIN_VERSION.zip
+    if [[ ! -f $ZIP ]]; then
+        echo "error: missing $ZIP, $plugin must be built first"
+        exit 1
+    fi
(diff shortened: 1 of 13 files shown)


peternied (approved):
I am approving, but that is a lot of copypasta - I wish I knew a better way

dblock:
> I am approving, but that is a lot of copypasta - I wish I knew a better way

I know :(

bundle-workflow/scripts/bundle-build/components/index-management/build.sh
```diff
@@ -0,0 +1,81 @@
...
+
+for plugin in job-scheduler notifications; do
+    PLUGIN_VERSION=$VERSION.0
+    [[ "$SNAPSHOT" == "true" ]] && PLUGIN_VERSION=$PLUGIN_VERSION-SNAPSHOT
+    ZIP=$DEST/plugins/opensearch-$plugin-$PLUGIN_VERSION.zip
+    if [[ ! -f $ZIP ]]; then
+        echo "error: missing $ZIP, $plugin must be built first"
+        exit 1
+    fi
+    cp $ZIP src/test/resources/$plugin
```
mch2, line 67: Is this only required for a test gradle task?  In our build here we are skipping tests and relying on test steps to be run on merge/pr and later after a bundle is assembled during the test workflow.

dblock: Yes. I suppose I should spend a bit more time trying to figure out how to disable that :)

dblock:
Both these projects decided to continue checking in the .zips, and this was needed to run integration tests only, so we don't need this.

https://github.com/opensearch-project/index-management/pull/121
https://github.com/opensearch-project/dashboards-reports/pull/140

PR #253, opened 2021-08-18
Add security_enabled flag to LocalTestCluster

camerski:
### Description
Add 'work_dir' argument to LocalTestCluster so that a top-level temp directory can be wired in and the keep-or-delete logic stays in one place.

Change LocalTestCluster so that calls to 'create' do not return until the cluster is ready to take requests.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/205
 
### Testing
Manually verified that `test.py` works correctly with and without security enabled, and that the `--keep` flag does the right thing.

### Check List

### DIFF
--- a/bundle-workflow/python/test_workflow/local_test_cluster.py
+++ b/bundle-workflow/python/test_workflow/local_test_cluster.py
@@ -1,29 +1,33 @@
 import os
-import tempfile
 import urllib.request
-import shutil
 import subprocess
-from test_workflow.test_cluster import TestCluster
+import time
+import ssl
+import requests
+from test_workflow.test_cluster import TestCluster, ClusterCreationException
 
 class LocalTestCluster(TestCluster):
-    def __init__(self, bundle_manifest):
+    '''
+    Represents an on-box test cluster. This class downloads a bundle (from a BundleManifest) and runs it as a background process.
+    '''
+
+    def __init__(self, work_dir, bundle_manifest, security_enabled):
         self.manifest = bundle_manifest
-        self.work_dir = tempfile.TemporaryDirectory()
+        self.work_dir = os.path.join(work_dir, 'local-test-cluster')
+        os.makedirs(self.work_dir, exist_ok = True)
+        self.security_enabled = security_enabled
+        self.process = None
 
     def create(self):
-        print(f'Creating local test cluster in {self.work_dir.name}')
-        os.chdir(self.work_dir.name)
-        print(f'Downloading bundle from {self.manifest.build.location}')
-        urllib.request.urlretrieve(self.manifest.build.location, 'bundle.tgz')
-        print(f'Downloaded bundle to {os.path.realpath("bundle.tgz")}')
-        print('Unpacking')
-        subprocess.check_call('tar -xzf bundle.tgz', shell = True)
-        print('Unpacked')
+        self.download()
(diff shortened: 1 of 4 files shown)


bundle-workflow/python/test_workflow/local_test_cluster.py
```diff
@@ -51,3 +58,37 @@ def destroy(self):
...
+        subprocess.check_call(f'echo "plugins.security.disabled: true" >> {os.path.join(dir, "config", "opensearch.yml")}', shell = True)
+
+    def wait_for_service(self):
+        print('Waiting for service to become available')
+        url = self.url('/_cluster/health')
+
+        for attempt in range(10):
+            try:
+                print(f'Pinging {url} attempt {attempt}')
+                response = requests.get(url, verify = False, auth = ('admin', 'admin'))
```
camerski, line 86: Passing basic auth information for the non-security case is OK because the server just ignores it. Yes, I know it's sending a username and password over an unencrypted channel, but the default username and password are hardly secrets - they're right there in the code!

bundle-workflow/python/test_workflow/local_test_cluster.py
```diff
@@ -51,3 +58,37 @@ def destroy(self):
...
+        print('Unpacked')
+
+    def disable_security(self, dir):
+        subprocess.check_call(f'echo "plugins.security.disabled: true" >> {os.path.join(dir, "config", "opensearch.yml")}', shell = True)
+
+    def wait_for_service(self):
+        print('Waiting for service to become available')
+        url = self.url('/_cluster/health')
+
+        for attempt in range(10):
```
peternied, line 83: Suggestion: [retry](https://pypi.org/project/retry/) could trim this down, and make it easier if these files need more retry functionality in the future

peternied (approved):

bundle-workflow/python/test_workflow/local_test_cluster.py
camerski: Good idea. Future PR!

bundle-workflow/python/test_workflow/local_test_cluster.py
```diff
@@ -51,3 +58,37 @@ def destroy(self):
...
+            try:
+                print(f'Pinging {url} attempt {attempt}')
+                response = requests.get(url, verify = False, auth = ('admin', 'admin'))
+                print(f'{response.status_code}: {response.text}')
+                if response.status_code == 200 and '"status":"green"' in response.text:
```
gaiksaya, line 88: Is the status always going to be and need to be green? Esp for a single node cluster? I don't know answer to that but does using `status: green || yellow` makes sense? I have seen few times that the cluster is yellow at the start. More inputs required.

gaiksaya (approved):
The concern [here](https://github.com/opensearch-project/opensearch-build/pull/253#discussion_r690829586) can be addressed in another PR if needed. Approving for now

bundle-workflow/python/test_workflow/local_test_cluster.py
camerski: If a freshly-installed and just-started cluster is yellow then we have a problem :)

Also, I think we don't want to run integ tests against a yellow cluster. The tests are likely to be flaky if the service is not stable when the tests begin.

PR #261, opened 2021-08-18
Fail with a better error if bundle is missing from the output manifest.

dblock:

### Description

I had a partial manifest and got a weird error because the min bundle wasn't in it.
Fail with a better error if bundle is missing from the output manifest.
  
### Check List


mch2 (approved):

peternied (approved):
Seems like we should start authoring unit tests created this issue for follow up https://github.com/opensearch-project/opensearch-build/issues/264

Issue #251, opened 2021-08-17
Build System - Extract k-NN & PA agent custom setup 


mch2:
Right now k-NN and performance analyzer have some custom setup steps when a bundle is assembled.

Ex: https://github.com/opensearch-project/opensearch-build/pull/239

This should either be pulled out as part of plugin installation or a separate install script that is executed for each component with custom logic.

PR #267, opened 2021-08-19
Added support for custom install scripts.

dblock:

### Description

Introduces `install.sh` that will run during installation of plugins.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/251
 
### Check List

### DIFF
--- a/bundle-workflow/README.md
+++ b/bundle-workflow/README.md
@@ -48,6 +48,10 @@ Bundle step:
 ```bash
 ./bundle-workflow/assemble.sh artifacts/manifest.yml
 ```
-### Scripts
+### Custom Build Scripts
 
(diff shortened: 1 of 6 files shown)


dblock:
I think we should keep custom build.sh and install.sh together under the same folder for components, so maybe as a next step I'll rename `scripts/bundle-build` to just `scripts` and `standard-gradle-build` to `default`. WDYT?

mch2:
I'd been working on a slightly different version here - https://github.com/opensearch-project/opensearch-build/pull/268
It works nicely with k-nn but still working on assembling performance-analyzer properly.

The key difference here is there is no default script, it executes it as an additional bundle step if there is a script at path.

mch2 (approved):
I prefer this to my draft impl.  Your implementation in install.sh is cleaner, I forgot we already had libs on the artifact path.

peternied (approved):
> I think we should keep custom build.sh and install.sh together under the same folder for components, so maybe as a next step I'll rename scripts/bundle-build to just scripts and standard-gradle-build to default. WDYT?

I like this approach and the naming convention  - make it so!

PR #269, opened 2021-08-19
Simplify custom scripts tree.

dblock:

### Description

Today one has to look for build.sh, install.sh and integtest.sh in 3 different places for the same component. Each of those folders only has a single script, which seems too much structure.  This PR simplifies the tree of scripts and hides away repeated logic inside `ScriptFinder`.

```
bundle-workflow
 |----- scripts
              |----- components
                            |- index-management
                                      |- build.sh
                                      |- install.sh
                                      |- integtest.sh
              |----- default
                            |- build.sh
                            |- install.sh
                            |- integtest.sh
```
 
### Issues Resolved

Followup from https://github.com/opensearch-project/opensearch-build/pull/267#pullrequestreview-734212471
 
### Check List


bundle-workflow/python/paths/script_finder.py
```diff
@@ -7,16 +7,22 @@
...
+
+    For install.sh scripts, given a component name, it will look in the following locations, in order:
+      * <component_scripts_path>/<component_name>/<script-name>
+      * <default_scripts_path>/<script-name>
     '''
 
-    component_scripts_path: str
-    default_scripts_path: str
+    component_scripts_path: str = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../../scripts/components')
+    default_scripts_path: str = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../../scripts/default')
```
camerski, line 25: Relying on `__file__` and relative paths is going to be fragile...this is OK for now, but consider using a regular constructor that accepts an optional base path instead of using `@dataclass`. dataclass is not really appropriate for this class anyway - it's not a plain data object and it has no fields that should be readable by clients.

bundle-workflow/python/build.py
```diff
@@ -16,8 +16,8 @@
 
 args = BuildArgs()
 
-component_scripts_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../scripts/bundle-build/components')
-default_scripts_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../scripts/bundle-build/standard-gradle-build')
+component_scripts_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../scripts/components')
+default_scripts_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../scripts/default')
 script_finder = ScriptFinder(component_scripts_path, default_scripts_path)
```
camerski, line 21: Did you forget to change this to use the no-arg ScriptFinder?

bundle-workflow/python/paths/script_finder.py
```diff
@@ -7,16 +7,22 @@
...
+    For build.sh and integtest.sh scripts, given a component name and a checked-out Git repository, 
+    it will look in the following locations, in order:
       * Root of the Git repository
       * /scripts/<script-name> in the Git repository
       * <component_scripts_path>/<component_name>/<script-name>
       * <default_scripts_path>/<script-name>
+
+    For install.sh scripts, given a component name, it will look in the following locations, in order:
+      * <component_scripts_path>/<component_name>/<script-name>
+      * <default_scripts_path>/<script-name>
```
camerski, line 21: Why are we not looking in the Git repository for install scripts? Seems like install logic should be owned by the plugins, same as the build and test logic.

I guess if https://github.com/opensearch-project/opensearch-plugins/issues/75 gets implemented soon then we won't need custom install scripts for long. Consider this non-blocking.

bundle-workflow/python/build.py
dblock: Yup, thx.

bundle-workflow/python/paths/script_finder.py
dblock: I actually think this is going to equally fragile than specifying it 3 times in every calling code :)

bundle-workflow/python/paths/script_finder.py
dblock: This happens in the assembly phase where we no longer have source code checked out, we're building off output from a build. That output could contain the custom scripts, but that really is saying that that output should just install itself without custom scripts, which is either https://github.com/opensearch-project/opensearch-plugins/issues/75 or https://github.com/opensearch-project/k-NN/issues/80

dblock:
Addressed @camerski comments.

bundle-workflow/python/paths/script_finder.py
camerski: Ah, OK. Makes sense. Let's get #75 done so we can get rid of this workaround.

bundle-workflow/python/paths/script_finder.py
camerski: So lets get rid of at least one of those pieces of fragility :)

camerski (approved):

mch2 (approved):
LGTM - we will need to restore index-management or remove it to use the default script.

dblock:
> LGTM - we will need to restore index-management or remove it to use the default script.

I deleted it.

PR #272, opened 2021-08-19
Add core plugins zips to published artifacts

### LINKED ISSUE #200: Add Core plugins zips to published artifacts.
The plugins built with the core repo should be bundled as zips and hosted at artifacts.opensearch.org.

This task involves updating Opensearch's build.sh script to build and include these artifacts with the component's artifact list.

Right now we also have brittle logic that assumes any component with "plugins" in their artifacts list is a plugin repository.  This should be updated to identify the min bundle component in another way.  Perhaps with a separate artifact folder for "min-bundle"

adnapibar:
### Description
Create build artifacts (zips) for core plugins and copy those into `artifacts/core-plugins` directory.
 
### Issues Resolved
Closes #200 
 
### Check List

Signed-off-by: Rabi Panda <adnapibar@gmail.com>

### DIFF
--- a/bundle-workflow/README.md
+++ b/bundle-workflow/README.md
@@ -23,6 +23,7 @@ Artifacts will contain the following folders.
   bundle/ <- contains opensearch min tarball 
   maven/ <- all built maven artifacts
   plugins/ <- all built plugin zips
+  core-plugins/ <- all built core plugins zip
   manifest.yml <- build manifest describing all built components and their artifacts
 ```
 
@@ -44,7 +45,7 @@ Each component build relies on a `build.sh` script that is used to prepare bundl
 ./bundle-workflow/assemble.sh artifacts/manifest.yml
 ```
 
-The bundling step takes output from the build step, installs plugins, and assembles a full bundle into a `bundle` folder. The input requires a path to the build manifest and is expected to be inside the `artifacts` directory that contains `bundle`, `maven` and `plugins` subdirectories from the build step.
+The bundling step takes output from the build step, installs plugins, and assembles a full bundle into a `bundle` folder. The input requires a path to the build manifest and is expected to be inside the `artifacts` directory that contains `bundle`, `maven`, `plugins` and `core-plugins` subdirectories from the build step.
 
 Artifacts will be updated as follows.
 
--- a/bundle-workflow/python/build_workflow/builder.py
+++ b/bundle-workflow/python/build_workflow/builder.py
@@ -32,7 +32,7 @@ class Builder:
 
     def export_artifacts(self):
         artifacts_dir = os.path.realpath(os.path.join(self.git_repo.dir, self.output_path))
-        for artifact_type in ["maven", "bundle", "plugins", "libs"]:
+        for artifact_type in ["maven", "bundle", "plugins", "libs", "core-plugins"]:
             for dir, dirs, files in os.walk(os.path.join(artifacts_dir, artifact_type)):
                 for file_name in files:
                     absolute_path = os.path.join(dir, file_name)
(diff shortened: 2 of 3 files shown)


bundle-workflow/scripts/bundle-build/components/OpenSearch/build.sh
```diff
@@ -91,3 +91,16 @@ ARTIFACT_BUILD_NAME=opensearch-$VERSION$IDENTIFIER-$QUALIFIER.tar.gz
 ARTIFACT_TARGET_NAME=opensearch-min-$VERSION$IDENTIFIER-$QUALIFIER.tar.gz
 mkdir -p "${outputDir}/bundle"
 cp distribution/archives/$TARGET/build/distributions/$ARTIFACT_BUILD_NAME "${outputDir}"/bundle/$ARTIFACT_TARGET_NAME
+
+
+# Build core plugins
```
dblock, line 96: Comment seems redundant with the `echo` :)

bundle-workflow/scripts/bundle-build/components/OpenSearch/build.sh
```diff
@@ -91,3 +91,16 @@ ARTIFACT_BUILD_NAME=opensearch-$VERSION$IDENTIFIER-$QUALIFIER.tar.gz
 ARTIFACT_TARGET_NAME=opensearch-min-$VERSION$IDENTIFIER-$QUALIFIER.tar.gz
 mkdir -p "${outputDir}/bundle"
 cp distribution/archives/$TARGET/build/distributions/$ARTIFACT_BUILD_NAME "${outputDir}"/bundle/$ARTIFACT_TARGET_NAME
+
+
```
dblock, line 95: nit: leave 1 line

bundle-workflow/scripts/bundle-build/components/OpenSearch/build.sh
```diff
@@ -91,3 +91,16 @@ ARTIFACT_BUILD_NAME=opensearch-$VERSION$IDENTIFIER-$QUALIFIER.tar.gz
...
+# Build core plugins
+echo "Building core plugins..."
+mkdir -p "${outputDir}/core-plugins"
+cd plugins
+../gradlew assemble -Dbuild.snapshot=$SNAPSHOT
+for plugin in *; do
+  if [ -d "$plugin" ] && [ "examples" != "$plugin" ]; then
+    PLUGIN_ARTIFACT_BUILD_NAME="$plugin"-$VERSION$IDENTIFIER.zip
```
dblock, line 103: Does $plugin need to be quoted here?

bundle-workflow/scripts/bundle-build/components/OpenSearch/build.sh
```diff
@@ -91,3 +91,16 @@ ARTIFACT_BUILD_NAME=opensearch-$VERSION$IDENTIFIER-$QUALIFIER.tar.gz
...
+    cp "$plugin"/build/distributions/"$PLUGIN_ARTIFACT_BUILD_NAME" "../${outputDir}"/core-plugins/"$PLUGIN_ARTIFACT_BUILD_NAME"
```
dblock, line 104: This forces `outputDir` to always be a relative path and will break in unexpected ways. Instead, `cd ..` after assembling inside `plugins` and iterate over the output from the parent folder. This will remove the need for `..` relative path here.

dblock (changes requested):
The relative path for `outputDir` needs fixing, everything else is minor.

dblock:
Btw, if you say `Closes #200` in the PR body, it will auto-close that issue when this PR is merged.

bundle-workflow/scripts/bundle-build/components/OpenSearch/build.sh
adnapibar: Thanks :) Fixed.

dblock (approved):

gaiksaya:
Are we calling those as `core-plugins` or `native-plugins`?

adnapibar:
> Are we calling those as `core-plugins` or `native-plugins`?

Historically the community knows these as core plugins, so it makes sense to be consistent. Also, calling these "native" may create confusion that these plugins rely on JNI.

mch2 (approved):

PR #287, opened 2021-08-21
Adding initial bwc.py to test orchestrator

### LINKED ISSUE #275: Add support for running bwctest.sh

saratvemulapalli:
Signed-off-by: Sarat Vemulapalli <vemulapallisarat@gmail.com>

### Description
Adding support to run BWC tests in OpenSearch project.
All bwc tests will invoke `bwctest.sh` on each repository in OpenSearch-project.

See https://github.com/saratvemulapalli/OpenSearch/blob/main/bwctest.sh as an example.
 
### Issues Resolved
#275 
 
### Check List


bundle-workflow/python/test_bwc.py
```diff
@@ -0,0 +1,62 @@
...
+
+def execute_shell(command, work_dir):
+    return subprocess.check_output(command, cwd = work_dir, shell = True)
+
+def pull_component(component, work_dir):
+    GitRepository(component.repository, component.commit_id, os.path.join(work_dir, component.name))
+
+def run_tests(work_dir):
+    bwc_script = "bwctest.sh"
+    change_ownership = f"chmod +x {bwc_script}"
```
dblock, line 38: Git keeps permissions, so this shouldn't be needed.

bundle-workflow/python/test_bwc.py
```diff
@@ -0,0 +1,62 @@
...
+def is_component_test_supported(component):
+    if component.name == 'OpenSearch':
+        return True
+    else:
+        return False
+
+def execute_shell(command, work_dir):
+    return subprocess.check_output(command, cwd = work_dir, shell = True)
+
+def pull_component(component, work_dir):
```
dblock, line 33: If you need functions like these, refactor into a class such as Tester, then have the git repo as a property and call `.checkout()`

bundle-workflow/python/test_bwc.py
```diff
@@ -0,0 +1,62 @@
...
+    return args
+
+#TODO read from configuration file
+def is_component_test_supported(component):
+    if component.name == 'OpenSearch':
+        return True
+    else:
+        return False
+
+def execute_shell(command, work_dir):
```
dblock, line 30: Should be a method on an object.

bundle-workflow/python/test_bwc.py
```diff
@@ -0,0 +1,62 @@
...
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(description = "Test an OpenSearch Bundle")
+    parser.add_argument('--manifest', type = argparse.FileType('r'), help="Manifest file.")
+    parser.add_argument('--keep', dest = 'keep', action='store_true', help = "Do not delete the working temporary directory.")
+    args = parser.parse_args()
+    return args
+
+#TODO read from configuration file
+def is_component_test_supported(component):
```
dblock, line 24: Move to be a property of `component`.

bundle-workflow/python/test_bwc.py
```diff
@@ -0,0 +1,62 @@
...
+
+from manifests.bundle_manifest import BundleManifest
+from git.git_repository import GitRepository
+from test_workflow.local_test_cluster import LocalTestCluster
+from test_workflow.integ_test_suite import IntegTestSuite
+from paths.script_finder import ScriptFinder
+from system.temporary_directory import TemporaryDirectory
+
+
+def parse_arguments():
```
dblock, line 16: Refactor into a class, e.g. https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/python/build_workflow/build_args.py

dblock (changes requested):
I think before merging you need a runner ala `build.sh` that calls this and a bit of documentation in README. 

Some should have's.

bundle-workflow/python/test_bwc.py
saratvemulapalli: Eventually we would not have a check and all components will be tested. 
i.e if any component test fails or tests do not exist we will flag them.

bundle-workflow/python/test_bwc.py
saratvemulapalli: Yeah, definitely. This feels like bunch of other tests would need this as well.

bundle-workflow/src/test_workflow/test_args.py
```diff
@@ -0,0 +1,33 @@
...
+            "manifest", type=argparse.FileType("r"), help="Manifest file."
+        )
+        parser.add_argument("--component", help="Test a specific component")
+        parser.add_argument(
+            "--keep",
+            dest="keep",
+            action="store_true",
+            help="Do not delete the working temporary directory.",
+        )
+        args = parser.parse_args()
```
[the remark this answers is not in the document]
saratvemulapalli: Yeah the reason why I wanted to add a common test args is most of the test suites have manifest, component and keep. 
If you guys think its not helpful for integtest and perftests, I would be happy to rename it.

bundle-workflow/src/test.py
```diff
@@ -6,49 +6,48 @@
...
+                repo = test_component.checkout(os.path.join(work_dir, component.name))
+                test_suite = IntegTestSuite(component.name, repo)
+                test_suite.execute(cluster)
+        finally:
+            cluster.destroy()
+
+        # TODO: Store test results, send notification.
+
+
+def bwc_test_suite():
```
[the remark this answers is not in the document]
saratvemulapalli: Yeah agree that each job should have its own entry point. I'll go create `run_bwc_test.py` and `bwc_test.sh`
Yeah the idea this PR puts in is that, given a manifest and few options its takes care of all running all the suite of tests for all components. 

But thinking about it I like the idea of abstracting running the tests vs orchestrating the tests. 

The advantages of approach #338 proposes:
1. Tests for each component is abstracted into `bwc_test_suite.py` i.e each test run is abstracted and could have independent test recorder. 
2. Helps the runner to consolidate all the results into and if needed it could add retries later.

I'll merge this in and create a quick follow up to structure the code as we discussed here. 
What do you think?

bundle-workflow/src/test_workflow/test_args.py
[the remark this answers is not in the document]
dblock: If they are the same args used everywhere, why not!

bundle-workflow/src/test.py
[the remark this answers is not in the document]
setiah: sounds good to me @saratvemulapalli

setiah (approved):

bundle-workflow/src/test_workflow/test_args.py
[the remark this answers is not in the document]
setiah: we could check this in as is and raise another PR for the common inputs once #338 is checked in.

bundle-workflow/src/test_workflow/bwc_test_suite.py
```diff
@@ -0,0 +1,51 @@
...
+    keep: bool
+
+    def __init__(self, manifest, component, keep):
+        self.manifest = manifest
+        self.component = component
+        self.keep = keep
+
+    def run_tests(self, work_dir):
+        bwc_script = "bwctest.sh"
+        run_bwctests = f"./{bwc_script}"
```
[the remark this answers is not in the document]
saratvemulapalli: Thats a good point, sure definitely. 
I'll remove this in the follow up.

PR #290, opened 2021-08-22
Update signer repo url to require a github token.

mch2:

### Description
Update the signer url to take an access token so that it can be cloned from CI. The signer repo currently requires a github token to be cloned.
 
### Issues Resolved
related to setting up release workflow -  #136 
 
### Check List

### DIFF
--- a/bundle-workflow/python/signing_workflow/signer.py
+++ b/bundle-workflow/python/signing_workflow/signer.py
@@ -2,11 +2,11 @@
 
 # Copyright OpenSearch Contributors.
 # SPDX-License-Identifier: Apache-2.0
-import sys 
+import sys
+import os
 
 sys.path.insert(0,"../git")
 from git.git_repository import GitRepository
-
 
 '''
 This class is responsible for signing an artifact using the OpenSearch-signer-client and verifying its signature.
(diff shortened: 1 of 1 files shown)


dblock (changes requested):
Maybe best to check for existence if GITHUB_TOKEN since CI is very scenario specific and this would fail if the token is missing?

dblock (approved):

saratvemulapalli (approved):

bundle-workflow/python/signing_workflow/signer.py
```diff
@@ -2,22 +2,27 @@
...
 class Signer:
     def __init__(self):
-        self.git_repo = GitRepository("https://github.com/opensearch-project/opensearch-signer-client.git", "HEAD")
+        self.git_repo = GitRepository(self.get_repo_url(), "HEAD")
         self.git_repo.execute("./bootstrap", subdirname = "src")
         self.git_repo.execute("rm config.cfg", subdirname = "src")
 
+    def get_repo_url(self):
+        if "GITHUB_TOKEN" in os.environ:
+            return "https://${GITHUB_TOKEN}@github.com/opensearch-project/opensearch-signer-client.git"
```
saratvemulapalli, line 23: Curious why the signer client is closed source?

peternied (approved):

bundle-workflow/python/signing_workflow/signer.py
mch2: This is in the process of being opened up - [issue](https://github.com/opensearch-project/opensearch-signer-client/issues/5).

Issue #264, opened 2021-08-18
Bundle workflow should have unit tests


peternied:
We should start creating unit tests to validation functionality in the bundle workflow, and then can be picked up by the CI system when it is avaliable.

dblock:
https://github.com/opensearch-project/opensearch-build/issues/250 is related

PR #295, opened 2021-08-23
Adds CI with sorted imports, style guide and tests

### LINKED ISSUE #250: Add CI for this repo

We need CI to test things like build-workflow and things like #242.

GHA that runs unit and integration tests.

dblock:
### Description

- Adds CI
- Checks for sorted imports with `isort`
- Enforces style guide with `flake8`
- Runs pytest for tests, moved `python` into `src`
- Implemented tests for `system.arch`

(CR best reviewed commit-by-commit)
 
### Issues Resolved

Closes #250 and #264.  
 
### Check List


bundle-workflow/.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+ignore = E722
```
dblock, line 2: This is usage of `except` catch-all, which we use to trap any kind of build errors from components.

bundle-workflow/src/assemble_workflow/bundle_recorder.py
```diff
@@ -2,23 +2,32 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+from urllib.parse import urljoin
+
 import yaml
+
 from manifests.bundle_manifest import BundleManifest
```
Poojita-Raj, line 9: Is it convention to have sorted third party imports followed by sorted internal module imports?

bundle-workflow/src/assemble_workflow/bundle_recorder.py
```diff
@@ -2,23 +2,32 @@
...
-from urllib.parse import urljoin
 
 
 class BundleRecorder:
     def __init__(self, build, output_dir, artifacts_dir):
         self.output_dir = output_dir
         self.build_id = build.id
-        self.public_url = os.getenv('PUBLIC_ARTIFACT_URL', None)
+        self.public_url = os.getenv("PUBLIC_ARTIFACT_URL", None)
```
Poojita-Raj, line 16: Nit - Since, default value is optional and is None by default, we could omit it or make it clearer by adding the keyword.
````suggestion
self.public_url = os.getenv("PUBLIC_ARTIFACT_URL", default = None)
````

File: bundle-workflow/src/test_workflow/local_test_cluster.py [hunk 4]
```diff
@@ -0,0 +1,105 @@
...
+            try:
+                print(f"Pinging {url} attempt {attempt}")
+                response = requests.get(url, verify=False, auth=("admin", "admin"))
+                print(f"{response.status_code}: {response.text}")
+                if response.status_code == 200 and '"status":"green"' in response.text:
+                    print("Cluster is green")
+                    return
+            except requests.exceptions.ConnectionError:
+                print("Service not available yet")
+            time.sleep(10)
```
Poojita-Raj, line 104: Is there any reason why we attempt to check at equal intervals instead of using an exponential backoff strategy for retrying requests?

Poojita-Raj (approved):

bundle-workflow/src/assemble_workflow/bundle_recorder.py
dblock: Yes, this is output of defaults of `isort`. We can change the convention if we don't like it.

bundle-workflow/src/assemble_workflow/bundle_recorder.py
dblock: Let's leave this for a separate PR.

bundle-workflow/src/test_workflow/local_test_cluster.py
dblock: In this PR this file only moved from `python` to `src` and hasn't changed. 

Reading the code, there's no real good reason for timeout choices nor does it matter much. We're booting a service, then waiting for it to start. If anything, I would exponentially forward-on here because after waiting a long time it's likely the service will start "sooner".

gaiksaya:
Can we add `Pipfile.lock` in .gitignore?
Also why remove  `check-vulnerability-whitesource.yml` workflow?

dblock:
> Can we add `Pipfile.lock` in .gitignore and delete from this commit?

We do not want to. This file represents the locked python versions for our dependencies and creates a predictable consistent behavior every time - no dependency suddenly gets updated. Read about it at https://pipenv.pypa.io/en/latest/.

> Also why remove `check-vulnerability-whitesource.yml` workflow?

Does it ever run? Where and how? Looks disabled to me.

bundle-workflow/src/assemble_workflow/bundle_recorder.py
mch2: personally, I'm not a huge fan of this default but its a nit.

mch2 (approved):

bundle-workflow/src/assemble_workflow/bundle_recorder.py
dblock: Let's consider changing it later!

PR #303, opened 2021-08-23
Update sign.py to only sign specific file types

### LINKED ISSUE #302: Restrict sign workflow to only signing specific file types.
The signer right now signs every file in a passed in /maven folder.  This should be restricted to only signing specific file types.  As a result all metadata.xml and sha files are signed, which is unnecessary and causes a long running CI job.

mch2:

### Description
Restrict signer to sign only the following file types - ['.zip', '.jar', '.war', '.pom', '.module', 'tar.gz']
 
### Issues Resolved
closes #302 
 
### Check List


dblock (approved):

dblock:
Perfect opportunity for a unit test ;)

bundle-workflow/python/sign.py
```diff
@@ -15,20 +16,21 @@
...
     print(f'\nSigning {component.name}')
     for artifact_type in component.artifacts:
-        
+
         if args.type and args.type != artifact_type:
             continue
-        
-        artifact_list = component.artifacts[artifact_type]
+
+        artifact_list = [a for a in component.artifacts[artifact_type] if pathlib.Path(a).suffix in ACCEPTED_FILE_TYPES]
```
peternied, line 33: Might be useful for debugging to know what was excluded with a log message

peternied (approved):

bundle-workflow/tests/signing_workflow/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+#
+#  The OpenSearch Contributors require contributions made to
+#  this file be licensed under the Apache-2.0 license or a
+#  compatible open source license.
+
+import os
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../.."))
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../src"))
```
mch2, line 11: idk if there is a more elegant way to define this, but without /src it could not find the signer's import of git_repository.

dblock: I couldn't find a better way to do it.

You should have just one though, and in tests import `from src.git.git_repository import GitRepository`, note the `src.`. As written you may have a conflict in class names between `tests` and `src`.

bundle-workflow/tests/signing_workflow/test_sign.py
```diff
@@ -0,0 +1,33 @@
+import unittest
+from unittest.mock import MagicMock, call, patch
+
+from src.signing_workflow.signer import Signer
+
+
+class TestSigner(unittest.TestCase):
+
+    @patch('src.signing_workflow.signer.GitRepository')
```
dblock, line 9: Oh I like this!

bundle-workflow/src/signing_workflow/signer.py
```diff
@@ -17,11 +18,26 @@
...
+    def sign_artifacts(self, artifacts, basepath):
+        for artifact in artifacts:
+            if self.is_invalid_file_type(artifact):
+                print(f'Skipping signing of file ${artifact}')
+                continue
+            location = os.path.join(basepath, artifact)
+            self.sign(location)
+            self.verify(location + ".asc")
+
+    def is_invalid_file_type(self, file_name):
```
dblock, line 38: Can be a class method.

dblock (approved):
I think you should fix the double insert of paths in `__init__.py`, otherwise LGTM.

peternied (approved):

PR #307, opened 2021-08-24
Add Input Manifest for 1.0 maven release.

mch2:

### Description
This PR creates a small input manifest to push 1.0 artifacts to maven for the repo's included.
Going forward this will be part of the full bundle release workflow, but for 1.0.0 we do not have solid 1.0 refs for all repos nor do we need to rebuild the full bundle.

### Issues Resolved
One of the remaining steps for #20.
This will need to be set up in CI @bbarani and run as part of [this](https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/Jenkinsfile) jenkins job to build the artifacts.
The next step is to promote with [this](https://github.com/opensearch-project/opensearch-infra/pull/97/files) job.
 
### Check List


dblock:
Should we call this `-maven` to make it clearer rather than `-partial`?

Add this information to https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/README.md#opensearch-bundle-workflow. Maybe link the versions to the manifests from that README?

dblock (approved):
We/you/someone can also write a unit test that ensures that the manifest loads from Python code.

PR #311, opened 2021-08-24
Added unit tests for TemporaryDirectory, ScriptFinder and GitRepository.

dblock:
### Description

Added unit tests for `TemporaryDirectory`, `ScriptFinder` and `GitRepository`.

### Issues Resolved

Part of https://github.com/opensearch-project/opensearch-build/issues/264.
 
### Check List


bundle-workflow/tests/git/test_git_repository.py
```diff
@@ -0,0 +1,87 @@
...
+import tempfile
+import unittest
+from unittest.mock import MagicMock
+
+from src.git.git_repository import GitRepository
+
+
+class TestGitRepository(unittest.TestCase):
+    def setUp(self):
+        self.repo = GitRepository(
```
mch2, line 15: nit - Rather than cloning a dependency in this test, should we be mocking subprocess and ensuring it is invoked with the correct commands?

mch2 (approved):

bundle-workflow/tests/git/test_git_repository.py
dblock: You're right, it feels more like an integration test as is. But gets the job done, so leaving it as is for now in favor of covering more surface.

peternied (approved):

Issue #260, opened 2021-08-18
Copy manifest and bundle from S3


setiah:
Using a python script to download the manifest file and bundle from S3 to jenkins

- [x] Provide S3 read write methods as a utility for other modules
- [x] Provide s3 read/write role permission support for Jenkins and local dev

setiah:
This is done with #367

PR #316, opened 2021-08-25
Copy manifest and bundle from S3 to jenkins 

meghasaik:
### Description
[Describe what this change achieves]
 This change basically achieves two things: 
1. Copying the files from S3
2. Putting the files into S3
### Issues Resolved
Issue No. 260
 https://github.com/opensearch-project/opensearch-build/issues/260
### Check List


bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
...
+    # RoleSessionName: Name for the session
+    # bucket_bundle: Object path from where you want to get the bundle
+    # bucket_manifest: Object path from where you want to get the manifest
+    # path_tosave_Bundle: Path where you want to save the bundle at
+    # path_tosave_Manifest: Path where you want to save the manifest at
+    def __init__(self, RoleArn, RoleSessionName):
+        self.RoleArn = RoleArn
+        self.RoleSessionName = RoleSessionName
+
+    def get_S3Objects(self, bucketname, bucket_bundle, bucket_manifest, path_tosave_Bundle, path_tosave_Manifest):
```
owaiskazi19, line 24: Should be `bucket_name`,`path_to_save_bundle`,`path_to_save_mainfest`.
Ref: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # get_S3Objects:Used for getting the files from S3
+    # bucketname: Name of the bucket you want to get the file from
+    # RoleArn: Amazon Resource Name (ARN) of the role that you want to assume
+    # RoleSessionName: Name for the session
+    # bucket_bundle: Object path from where you want to get the bundle
+    # bucket_manifest: Object path from where you want to get the manifest
+    # path_tosave_Bundle: Path where you want to save the bundle at
+    # path_tosave_Manifest: Path where you want to save the manifest at
+    def __init__(self, RoleArn, RoleSessionName):
```
owaiskazi19, line 20: Should be `role_arn`, `role_session_name`

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
...
+        self.RoleArn = RoleArn
+        self.RoleSessionName = RoleSessionName
+
+    def get_S3Objects(self, bucketname, bucket_bundle, bucket_manifest, path_tosave_Bundle, path_tosave_Manifest):
+        sts_connection = boto3.client('sts')
+        acct_b = sts_connection.assume_role(
```
owaiskazi19, line 26: Let's name it as `assumed_role` for clarity

bundle-workflow/src/getfiles_S3bkt.py
```diff
@@ -0,0 +1,36 @@
...
+
+import argparse
+
+import read_write_s3
+
+
+class GetBundleManifest:
+    def parse_arguments(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument("-r", "--assume", help="Assume role")
```
owaiskazi19, line 17: Pass argument `a` to map it to `assume`

bundle-workflow/src/getfiles_S3bkt.py
```diff
@@ -0,0 +1,36 @@
...
+        parser.add_argument("-s", "--session", help="Session name")
+        parser.add_argument("-bkt", "--bktname", help="Bucketname")
```
owaiskazi19, line 19: Pass argument `b` and map it to `--bucket-name`

bundle-workflow/src/getfiles_S3bkt.py
```diff
@@ -0,0 +1,36 @@
...
+        parser.add_argument("-ob", "--obj_bundle", help="ObjectPath in Bucket for Bundle")
```
owaiskazi19, line 20: Pass argument `o` and map it to `--object-bundle`

bundle-workflow/src/getfiles_S3bkt.py
```diff
@@ -0,0 +1,36 @@
...
+        parser.add_argument("-om", "--obj_manifest", help="ObjectPath in Bucket for Manifest")
```
owaiskazi19, line 21: Should be `--object-manifest`

bundle-workflow/src/getfiles_S3bkt.py
```diff
@@ -0,0 +1,36 @@
...
+        parser.add_argument("-bp", "--bundlepath", help="Filepath for the bundle")
+        parser.add_argument("-mp", "--manifestpath", help="Filepath for the manifest")
+
+        args = parser.parse_args()
+
+        return args
+
+    def get_S3files(self, args):
```
owaiskazi19, line 29: Let's name it as `get_files_from_S3`

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import boto3
+
+
+class Read_Write_Files:
```
owaiskazi19, line 10: class name should be like ReadWriteFiles. 
Ref: https://www.python.org/dev/peps/pep-0008/#class-names

bundle-workflow/src/read_write_s3.py
meghasaik: As the function name? because boto3 can only take assume_role() as the function

bundle-workflow/Pipfile
```diff
@@ -9,6 +9,7 @@ requests = "~=2.26"
 isort = "~=5.9"
 flake8 = "~=3.9"
 pytest = "*"
+boto3 = "*"
```
owaiskazi19, line 12: Can we assign it with the latest stable version and not `*`?

bundle-workflow/src/getfiles_S3bkt.py
```diff
@@ -0,0 +1,36 @@
+# !/usr/bin/env python3
```
owaiskazi19, line 1: Name this file as `get_files_from_S3` or something similar following the same naming convention

owaiskazi19 (changes requested):
Follow naming convention as mentioned her: https://www.python.org/dev/peps/pep-0008/#prescriptive-naming-conventions

bundle-workflow/src/read_write_s3.py
owaiskazi19: Sorry if I was not clear before. Name `acct_b` as `assumed_role`

peternied (changes requested):
The surface area of this pull request has moved considerable from its initial revision.  By separating many changes into smaller pull requests this will allow for faster and better feedback to get approval on the pull requests.

I would recommend the follow kinds of pull requests for the work I am seeing done here:
* Overall workflow updates - when where is this code called and how does it tie back to the test systems, codifies the high level requirements, e.g. aws permissions, data in/out contracts 
* Remote File interactions
* File parsing mechanics
* Reporting UX 
* Transformation of local data into reporting format

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
...
+class ReadWriteFiles:
+
+    # get_S3Objects:Used for getting the files from S3
+    # bucket_name: Name of the bucket you want to get the file from
+    # role_arn: Amazon Resource Name (ARN) of the role that you want to assume
+    # role_session_name: Name for the session
+    # bucket_bundle: Object path from where you want to get the bundle
+    # bucket_manifest: Object path from where you want to get the manifest
+    # path_to_save_bundle: Path where you want to save the bundle at
+    # path_to_save_manifest: Path where you want to save the manifest at
```
setiah, line 19: Could you add these comments as pydoc 

"""
"""

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
...
+
+        try:
+            client.download_file(bucket_name, bucket_bundle, path_to_save_bundle)
+            client.download_file(bucket_name, bucket_manifest, path_to_save_manifest)
+        except:
+            print("An exception occurred. No file is present in the bucket.")
+
+    # put_S3Objects:Used for putting the files into S3
+    # file_name: The file that you want to put in the S3 bucket
+    # file_path: The Object path where you want to save the file
```
setiah, line 42: add these comments as pydocs

bundle-workflow/src/get_files_from_S3.py
```diff
@@ -0,0 +1,36 @@
+# !/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+
+import argparse
+
```
setiah, line 10: remove extra space

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
...
+    # role_session_name: Name for the session
+    # bucket_bundle: Object path from where you want to get the bundle
+    # bucket_manifest: Object path from where you want to get the manifest
+    # path_to_save_bundle: Path where you want to save the bundle at
+    # path_to_save_manifest: Path where you want to save the manifest at
+    def __init__(self, role_arn, role_session_name):
+        self.role_arn = role_arn
+        self.role_session_name = role_session_name
+
+    def get_S3_objects(self, bucket_name, bucket_bundle, bucket_manifest, path_to_save_bundle, path_to_save_manifest):
```
VachaShah, line 24: Is getting the manifest mandatory for this method? What if I just have an S3 path and want to get the files that are in that bucket, for example for DependencyInstaller. Currently the code here is assuming that all the user passes in all the parameters.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
...
+            DurationSeconds=3600)['Credentials']
+
+        client = boto3.client('s3', aws_access_key_id=assumed_role['AccessKeyId'], aws_secret_access_key=assumed_role['SecretAccessKey'],
+                              aws_session_token=assumed_role['SessionToken'])
+
+        try:
+            client.download_file(bucket_name, bucket_bundle, path_to_save_bundle)
+            client.download_file(bucket_name, bucket_manifest, path_to_save_manifest)
+        except:
+            print("An exception occurred. No file is present in the bucket.")
```
VachaShah, line 38: There might be other valid exceptions which need to be raised, currently for all the exceptions, this is printing a "no file present" error message.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
...
+            RoleSessionName=self.role_session_name,
+            DurationSeconds=3600)['Credentials']
+
+        client = boto3.client('s3', aws_access_key_id=assumed_role['AccessKeyId'], aws_secret_access_key=assumed_role['SecretAccessKey'],
+                              aws_session_token=assumed_role['SessionToken'])
+
+        try:
+            client.upload_file(file_name, bucket_name, file_path)
+        except:
+            print("An exception occurred. No file found for uploading.")
```
VachaShah, line 56: Please see my comment for exceptions above.

VachaShah (commented):
This needs tests and documentation to show how the functionality can be used.

bundle-workflow/src/read_write_s3.py
setiah, line 24: We need to make this more generic to achieve two use cases - 
1. Download a particular S3 object
2. Download all maven dependencies from a given s3 location

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,56 @@
...
+    # bucket_name: Name of the bucket you want to get the file from
+    # role_arn: Amazon Resource Name (ARN) of the role that you want to assume
+    # role_session_name: Name for the session
+    # bucket_bundle: Object path from where you want to get the bundle
+    # bucket_manifest: Object path from where you want to get the manifest
+    # path_to_save_bundle: Path where you want to save the bundle at
+    # path_to_save_manifest: Path where you want to save the manifest at
+    def __init__(self, role_arn, role_session_name):
+        self.role_arn = role_arn
+        self.role_session_name = role_session_name
```
setiah, line 22: Make the role parameters optional. Abstract the role passing part from caller, by using defaults from environment vars (for now) to read/write from s3. By keeping these optional, the caller can choose to pass these for specific use cases, while the defaults work for most

bundle-workflow/src/read_write_s3.py
meghasaik: I have added them in the 
"""
"""
pydoc block.

bundle-workflow/src/read_write_s3.py
meghasaik: I have modified the method to take only the bundle parameter where through the S3 path we can get the files that are in that bucket

bundle-workflow/src/read_write_s3.py
meghasaik: I have modified it by printing out the exception that has occurred as either of the three exceptions might occur which are 1.bucket name does not exist 2. Not a directory for the object path 3. local file path when not given properly

bundle-workflow/src/read_write_s3.py
meghasaik: I have modified it now, as the bundle_path arg will be used for downloading the entire file/folder.

bundle-workflow/src/read_write_s3.py
meghasaik: The role parameters  will be optional and the default values will there for these parameters to take their values from.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,71 @@
...
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import os
+
+import boto3
+
+
+class ReadWriteFiles:
```
dblock, line 12: This should represent a thing, e.g. `S3` or `S3Bucket` and take bucket as a parameter since it's required in both put/get.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,71 @@
...
+    role_session_name: Name for the session
+    Here, default value for session is taken. Can be changed through -s <session-name>
+    bucket_bundle: Object path from where you want to get the bundle
+    path_to_save_bundle: Path where you want to save the bundle at
+    """
+    def __init__(self, role_arn, role_session_name):
+        self.role_arn = role_arn
+        self.role_session_name = role_session_name
+
+    def get_S3_objects(self, bucket_name, bucket_bundle, path_to_save_bundle=None):
```
dblock, line 27: Rename to `get` or `download`.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,71 @@
...
+            for obj in objects:
+                target = obj.key if path_to_save_bundle is None \
+                    else os.path.join(path_to_save_bundle, os.path.relpath(obj.key, bucket_bundle))
+                if not os.path.exists(os.path.dirname(target)):
+                    os.makedirs(os.path.dirname(target))
+                if obj.key[-1] == '/':
+                    continue
+                my_bucket.download_file(obj.key, target)
+        except Exception as e:
+            print(e)
```
dblock, line 52: Don't swallow exception, let it get raised.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    put_S3Objects:Used for putting the files into S3
+    file_name: The file that you want to put in the S3 bucket
+    file_path: The Object path where you want to save the file
+    """
+    def put_S3_objects(self, bucket_name, file_name, file_path):
```
dblock, line 58: Rename to `put` or `upload`.

bundle-workflow/src/read_write_s3.py
dblock: The parameters `bucket_bundle` and `path_to_save_bundle` don't make sense in S3 context. Should be `prefix` and `dest`. 

What's a `None` dest? I would just require it.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,71 @@
...
+            RoleArn=self.role_arn,
+            RoleSessionName=self.role_session_name,
+            DurationSeconds=3600)['Credentials']
+
+        client = boto3.client('s3', aws_access_key_id=assumed_role['AccessKeyId'], aws_secret_access_key=assumed_role['SecretAccessKey'],
+                              aws_session_token=assumed_role['SessionToken'])
+
+        try:
+            client.upload_file(file_name, bucket_name, file_path)
+        except Exception as e:
```
dblock, line 70: Don't swallow exception.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,71 @@
...
+        self.role_session_name = role_session_name
+
+    def get_S3_objects(self, bucket_name, bucket_bundle, path_to_save_bundle=None):
+        sts_connection = boto3.client('sts')
+        assumed_role = sts_connection.assume_role(
+            RoleArn=self.role_arn,
+            RoleSessionName=self.role_session_name,
+            DurationSeconds=3600)['Credentials']
+
+        boto3.client('s3',
```
dblock, line 34: This code is the same between `get` and `put`. Extract into the constructor that creates a new `client` for a bucket.

dblock (changes requested):
+ needs tests

setiah:
This is replaced with #367

PR #318, opened 2021-08-25
Added DependencyInstaller to copy dependencies from s3 to maven local

### LINKED ISSUE #276: Setup maven and build dependencies using DependencyInstaller
Required for 
* Integ tests #327 
* BwC tests

VachaShah:
Signed-off-by: Vacha <vachshah@amazon.com>

### Description
This PR introduces a DependencyInstaller, requirements in [#276](https://github.com/opensearch-project/opensearch-build/issues/276#issuecomment-910800754). This will be used by integration tests, performance tests and bwc tests when run in the pipeline. It covers the following scenarios:
1. Install maven dependencies from S3 to maven local
2. Download build dependencies from S3 to a custom local path (to be used for scenarios like [this](https://github.com/opensearch-project/opensearch-build/blob/main/tools/standard-test/integtest_dependencies_opensearch.sh))
3. Clean up the downloaded contents

Note: I have merged the changes from #367 to use the S3 utility functions.
 
### Issues Resolved
#276 
 
### Check List


dblock (dismissed):

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,47 @@
+import os
+import shutil
+
+
+class DependencyInstaller:
+    """
+    Provides functionality to copy the maven dependencies from S3 to maven local.
+    """
+
+    def __init__(self, build_id, dependency_name, version, arch):
```
setiah, line 10: instead of "build_id, version, arch" -> use build_manfiest as input, which has all these and more. Also, you can skip dependency_name in __init__. DependencyInstaller should provide a method to sync  - all maven dependencies, or - a list of dependencies.

bundle-workflow/tests/test_workflow/helper/test_utils.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+
+class TestUtils:
+    def get_test_dependencies():
```
peternied, line 5: Could you declare this function inside of the test class, if you foresee its reuse elsewhere please add comments or additional configuration

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,47 @@
...
+        return [
+            file_name
+            for file_name in os.listdir(os.path.dirname(os.path.abspath(__file__)))
+            if os.path.isfile(
+                os.path.join(os.path.dirname(os.path.abspath(__file__)), file_name)
+            )
+        ]
+
+    def install(self):
+        # s3_path = f"/builds/{self.version}/{self.build_id}/{self.arch}/maven/{self.dependency_path()}"
```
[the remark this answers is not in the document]
VachaShah: This is for the TODO where the files will be downloaded from S3 once the functionality is ready in PR #316.

bundle-workflow/src/test_workflow/dependency_installer.py
VachaShah: Do you mean that if the user passes in the list of dependencies, then get the files for those dependencies and if nothing is passed in, get the files for all dependencies?

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+
+    # TODO: This is currently a stubbed function which returns files from the current directory,
+    # to be replaced after it is implemented
+    def download(self):
+        return [
+            file_name
+            for file_name in os.listdir(os.path.dirname(os.path.abspath(__file__)))
+            if os.path.isfile(
+                os.path.join(os.path.dirname(os.path.abspath(__file__)), file_name)
```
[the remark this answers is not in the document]
VachaShah: The `pipenv run black .` reformatted it this way since the `flake8` was failing with the if section being right indented. Definitely agree on the high bar! For the final code, the currently stubbed method for downloading the dependencies from s3 needs to be replaced once that functionality is merged and the tests corresponding to that stubbed method would change as well. I raised this PR early to get it out there and get feedback on the other parts and then can get a final review once the stubbed method is replaced.

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,47 @@
...
+    Provides functionality to copy the maven dependencies from S3 to maven local.
+    """
+
+    def __init__(self, build_id, dependency_name, version, arch):
+        self.build_id = build_id
+        self.dependency_name = dependency_name
+        self.version = version
+        self.arch = arch
+        self.dependency_path = f"org/opensearch/{self.dependency_name}/{self.version}/"
+        self.maven_local_path = os.path.join(
```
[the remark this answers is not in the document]
setiah: i'm not sure what you mean by workers reuse. What I'm saying here is the workers would sync the dependencies in ~/.m2/repository of jenkins host. If there are two concurrent test jobs running, one for 1.1 another for 1.0, since they filesystem is same, they might end up overwriting each others dependencies.

bundle-workflow/src/test_workflow/dependency_installer.py
setiah: Yeah 2 functions essentially - 
1. for syncing a list of maven dependencies say for a specific plugin.
2. for syncing all maven dependencies from build/ folder.

setiah: 2. can be treated optional for now. 1 is must have

bundle-workflow/src/test_workflow/dependency_installer.py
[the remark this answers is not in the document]
VachaShah: Added build-id in the path to keep the executions independent.

bundle-workflow/src/test_workflow/dependency_installer.py
VachaShah: Added functionality for 1.

bundle-workflow/tests/test_workflow/helper/test_utils.py
VachaShah: Removed the common test utils and added it in the test class itself.

dblock (changes requested):
This is good! I think there are some should/must have's below but it's close.

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -0,0 +1,65 @@
...
+        ]
+
+    class MavenLocalFileHandlerTests(unittest.TestCase):
+        def setUp(self):
+            self.maven_local_file_handler = DependencyInstaller.MavenLocalFileHandler()
+
+        def test_copy(self):
+            with tempfile.TemporaryFile() as maven_local_path:
+                test_files = DependencyInstallerTests.__get_test_dependencies()
+                self.assertTrue(test_files)
```
dblock, line 62: This only makes sure "something" is returned. Check that len(), the number of files returned is > 0.

dblock (approved):

dblock (approved):

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    def __init__(self, build):
+        self.build_id = build.id
+        self.version = build.version
+        self.arch = build.architecture
+
+    # TODO: This is currently a stubbed function which returns files from the current directory,
+    # to be replaced after it is implemented
+    def download(self):
```
setiah, line 17: we can use the `download_folder` method available in #367

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,47 @@
+import os
+import shutil
+
+
+class DependencyInstaller:
```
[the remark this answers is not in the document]
setiah: Dependency installer requirements - https://github.com/opensearch-project/opensearch-build/issues/276#issuecomment-910800754

peternied (approved):

VachaShah:
Hi @dblock, sorry for a review request again for this PR, just updated the installer with the requirements mentioned in [#276 (comment)](https://github.com/opensearch-project/opensearch-build/issues/276#issuecomment-910800754) to handle more use cases and merged it with Himanshu's PR for S3 utility to use that functionality. Would love your review on it!

bundle-workflow/src/s3_utility.py
```diff
@@ -0,0 +1,97 @@
+# SPDX-License-Identifier: Apache-2.0
```
dblock, line 1: I had made a ton of comments on this file in a different PR, I think in #367 but I am no longer sure. Those needs to be addressed first.

dblock (changes requested):
I had made a ton of comments on this file in a different PR, I think in #367 but I am no longer sure. Those needs to be addressed first.

bundle-workflow/src/s3_utility.py
VachaShah: Yes that is the PR that introduces the S3 utility. I merged it in my branch to call the download_folder method from #367, so it is showing up in my PR.

dblock: Ok, that code is not ready to be merged anywhere.

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+
+    def installMavenDependencies(self, dependency_list):
+        """
+        Downloads the maven dependencies from S3 and puts them on the maven local path
+        for each dependency in the dependency_list.
+
+        :param dependency_list: list of dependencies for which the maven artifacts need to be downloaded.
+        """
+        for dependency in dependency_list:
```
setiah, line 31: Could you help me understand how the dependency list is expected here? Seems like it expects the paths instead of dependency names. I think we should accept dependency names and generate paths to pull down from S3.

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,62 @@
...
+                f".m2/repository/org/opensearch/{dependency}/{self.version}/",
+            )
+            self.s3_utility.download_folder(s3_path, maven_local_path)
+
+    def installBuildDependencies(self, dependency_list, custom_local_path):
+        """
+        Downloads the build dependencies from S3 and puts them on the given custom path
+        for each dependency in the dependency_list.
+
+        :param dependency_list: list of dependencies for which the build artifacts need to be downloaded.
```
setiah, line 44: it should be list of dependency names

setiah: need to change the code to handle this

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,62 @@
...
+        :param custom_local_path: the path where the downloaded dependencies need to copied.
+        """
+        for dependency in dependency_list:
+            s3_path = f"{self.s3_build_location}/{dependency}-{self.version}.zip"
+            self.s3_utility.download_file(s3_path, custom_local_path)
+
+    def cleanUp(self, local_path):
```
setiah, line 51: The paths where these dependencies are installed, should be members of the DependencyInstaller object. The caller need not provide them, it can be optional though. DependencyInstaller should be able to cleanup everything it created by itself w/o input requirements from caller

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        for dependency in dependency_list:
+            s3_path = f"{self.s3_maven_location}/{dependency}"
+            maven_local_path = os.path.join(
+                os.path.expanduser("~"),
+                f".m2/repository/org/opensearch/{dependency}/{self.version}/",
+            )
+            self.s3_utility.download_folder(s3_path, maven_local_path)
+
+    def installBuildDependencies(self, dependency_list, custom_local_path):
```
setiah, line 39: save the custom_local_path to cleanup later in the cleanUp method

bundle-workflow/src/test_workflow/dependency_installer.py
VachaShah: It is the dependency name that is expected here. [Example tests](https://github.com/opensearch-project/opensearch-build/pull/318/files#diff-05f802ba74b1ca82bd5f6ef6241b4a5bc5a71035c756ff71b53ae963690475b1R22)

bundle-workflow/src/test_workflow/dependency_installer.py
VachaShah: I will edit the comment to dependency names.

bundle-workflow/src/s3_utility.py
```diff
@@ -0,0 +1,97 @@
...
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import os
```
setiah, line 7: You can untrack these files in your commit and still use them for local development. That way, they won't be raised as part of your PR.

bundle-workflow/src/test_workflow/dependency_installer.py
VachaShah: The custom path for build dependencies would have to be passed by the user, for maven dependencies I can try to store the location to use for cleanup.

VachaShah:
This PR is dependent on the S3 utility PR #367. I will raise the final version once that PR is merged in.

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -0,0 +1,83 @@
...
+        mock_s3_utility.assert_has_calls(
+            [
+                call(
+                    f"{self.dependency_installer.s3_build_location}/{dependencies[0]}-{self.dependency_installer.version}.zip",
+                    os.path.dirname(__file__),
+                )
+            ]
+        )
+
+    def test_cleanUpForDir(self):
```
dblock, line 62: Don't use camelCase in Python.

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,62 @@
...
+        self.build_id = build.id
+        self.version = build.version
+        self.arch = build.architecture
+        self.s3_utility = S3ReadWrite()
+        self.s3_maven_location = (
+            f"/builds/{self.version}/{self.build_id}/{self.arch}/maven"
+        )
+        self.s3_build_location = (
+            f"/builds/{self.version}/{self.build_id}/{self.arch}/plugins"
+        )
```
setiah, line 22: nit: this location provider logic is common across build and test workflows and should be read from a single place. Check if there's already a utility/or we can make it. It's okay if you create an issue and take it separately outside this PR.

bundle-workflow/src/test_workflow/dependency_installer.py
setiah: If that's the case, this path seems wrong ` s3_path = f"{self.s3_maven_location}/{dependency}"` It exists under org/opensearch...

bundle-workflow/src/test_workflow/dependency_installer.py
setiah: thx, that works.

bundle-workflow/Pipfile
```diff
@@ -12,6 +12,10 @@ isort = "~=5.9"
 flake8 = "~=3.9"
 mypy = "~=0.9"
 pytest = "*"
+boto3 = "1.18.31"
+boto3-stubs = "1.18.34"
+mypy-boto3-s3 = "1.18.29"
```
owaiskazi19, line 17: Instead of this, we can install it like `boto3-stubs[s3]` and then later import `from mypy_boto3_s3 import S3Client, S3ServiceResource`.  We won't need `boto3-stubs` then if we are using boto3 only for S3.
Ref: https://pypi.org/project/mypy-boto3-s3/

bundle-workflow/src/test_workflow/dependency_installer.py
VachaShah: For maven dependencies, each dependency would have a separate path, if we store them when the install_maven_dependencies is called, we would end up cleaning all of them in the cleanup method. Would it not be better if the user passes in the path that needs to be cleaned so that there is no incorrect cleanup? LMK what you think.

For all the other comments and requirements, I have made the changes accordingly.

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Provides functionality to clean up the downloaded contents in a maven local path or a custom path.
+        """
+        try:
+            if os.path.isfile(local_path) or os.path.islink(local_path):
+                os.unlink(local_path)
+            elif os.path.isdir(local_path):
+                shutil.rmtree(local_path)
+        except OSError as e:
+            print(f"Failed to clean {local_path}. Reason: {e}")
```
dblock, line 64: Use `logging.error`.

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,65 @@
...
+        :param dependency_list: list of dependency names with version for which the build artifacts need to be downloaded.
+        Example: {'opensearch-job-scheduler':'1.1.0.0'}
+        :param custom_local_path: the path where the downloaded dependencies need to copied.
+        """
+        s3_bucket = S3Bucket(self.s3_build_location)
+        for dependency, version in dependency_dict.items():
+            s3_path = f"{dependency}-{version}.zip"
+            s3_bucket.download_file(s3_path, custom_local_path)
+
+    def cleanup(self, local_path):
```
dblock, line 54: This deletes a folder, which is ... dangerous if we get it wrong. For maven dependencies I don't think we will ever want to be deleting them after downloading explicitly because they will be intermixed with other dependencies that may have been potentially dropped into `~m2`. We already are doing cleanup in https://github.com/opensearch-project/opensearch-build/pull/403. I think you should delete it from the PR and close any issues that say otherwise.

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -0,0 +1,93 @@
...
+                call(
+                    "opensearch-anomaly-detection/1.1.0.0",
+                    maven_local_paths[1],
+                ),
+            ]
+        )
+
+    @patch("test_workflow.dependency_installer.S3Bucket")
+    def test_install_build_dependencies(self, mock_s3_bucket):
+        s3_bucket = mock_s3_bucket.return_value
```
dblock, line 54: = S3Bucket() - because it's mocked you will get the mock object

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -0,0 +1,93 @@
...
+        self.maxDiff = None
+        self.manifest_filename = os.path.join(
+            os.path.dirname(__file__),
+            "../tests_assemble_workflow/data/opensearch-build-1.1.0.yml",
+        )
+        self.manifest = BuildManifest.from_path(self.manifest_filename)
+
+    @patch("test_workflow.dependency_installer.S3Bucket")
+    def test_install_maven_dependencies(self, mock_s3_bucket):
+        s3_bucket = mock_s3_bucket.return_value
```
dblock, line 21: = S3Bucket()

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -0,0 +1,93 @@
...
+        self.dependency_installer = DependencyInstaller(self.manifest.build)
```
dblock, line 22: This is a test, don't assign anything to `self.` other than in `setUp`. Looks like `dependency_installer` is needed in every test, so move this into `setUp`.

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -0,0 +1,93 @@
...
+        dependencies = dict(
+            {
+                "opensearch-job-scheduler": "1.1.0.0",
+                "opensearch-anomaly-detection": "1.1.0.0",
+            }
+        )
+        self.dependency_installer.install_maven_dependencies(dependencies)
+        self.assertEqual(s3_bucket.download_folder.call_count, 2)
+        maven_local_paths = []
```
dblock, line 31: Extract into a method in `DependencyInstaller` that gives you a maven local path for a dependency. You have similar code in DependencyInstaller itself.

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -0,0 +1,93 @@
...
+                os.path.join(
+                    os.path.expanduser("~"),
+                    f".m2/repository/org/opensearch/{dependency}/{version}/",
+                )
+            )
+        s3_bucket.download_folder.assert_has_calls(
+            [
+                call(
+                    "opensearch-job-scheduler/1.1.0.0",
+                    maven_local_paths[0],
```
dblock, line 43: This will become something like `dependency_installer.maven_local_path('opensearch-job-scheduler')`

dblock (approved):
I have some should have's, but won't hold the PR for them.

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,65 @@
...
+        Example: {'opensearch-job-scheduler':'1.1.0.0', 'opensearch-core':'1.1.0'}
+        """
+        s3_bucket = S3Bucket(self.s3_maven_location)
+        for dependency, version in dependency_dict.items():
+            s3_path = f"{dependency}/{version}"
+            maven_local_path = os.path.join(
+                os.path.expanduser("~"),
+                f".m2/repository/org/opensearch/{dependency}/{version}/",
+            )
+            s3_bucket.download_folder(s3_path, maven_local_path)
```
setiah, line 38: nit: this is s3_relative_path

bundle-workflow/src/test_workflow/dependency_installer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def install_maven_dependencies(self, dependency_dict):
+        """
+        Downloads the maven dependencies from S3 and puts them on the maven local path
+        for each dependency in the dependency_list.
+
+        :param dependency_dict: list of dependency names with version for which the maven artifacts need to be downloaded.
+        Example: {'opensearch-job-scheduler':'1.1.0.0', 'opensearch-core':'1.1.0'}
+        """
+        s3_bucket = S3Bucket(self.s3_maven_location)
```
setiah, line 31: This looks incorrect. S3Bucket needs to be initialized with bucket_name

bundle-workflow/src/test_workflow/dependency_installer.py
setiah: It should be s3_maven_location for the dependency  which is a relative path inside the bucket. An example -  `builds/1.1.0/24/x64/maven/org/opensearch/`

bundle-workflow/src/test_workflow/dependency_installer.py
setiah: @dblock - valid point. But I think we still need to cleanup dependencies which are synced as part of one test run, else we will unnecessarily bloat the local maven over a period of time. IMO we should switch back to the custom maven local path, based on a random tmp location, where we download dependencies and cleanup after the test run. This keeps the concurrent test runs isolated. We planned on doing it earlier but then switched to having docker provide that isolation by mounting on a randomized tmp path. I think we need it in the code as developers will run this code on local setup as well where docker mounting won't provide that sort of isolation. @VachaShah and I discussed about it yesterday.

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -0,0 +1,93 @@
...
+class DependencyInstallerTests(unittest.TestCase):
+    def setUp(self):
+        self.maxDiff = None
+        self.manifest_filename = os.path.join(
+            os.path.dirname(__file__),
+            "../tests_assemble_workflow/data/opensearch-build-1.1.0.yml",
+        )
+        self.manifest = BuildManifest.from_path(self.manifest_filename)
+
+    @patch("test_workflow.dependency_installer.S3Bucket")
```
setiah, line 19: you can replace this with S3Bucket that's checked into main

bundle-workflow/src/test_workflow/dependency_installer.py
peternied: >bloat the local maven over a period of time.

Jenkins Agent nodes will only exist for a single workflow then are torn down, how fast is this bloat?  Enough that we need to clean it out to complete a workflow?

dblock: For the record CI will be providing throw away containers with their own file system for these jobs.

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
VachaShah: I tried it but it tries to call the actual constructor instead of mocked.

bundle-workflow/tests/tests_test_workflow/test_dependency_installer.py
VachaShah: This is the mocked version of the `S3Bucket` checked in main.

bundle-workflow/src/test_workflow/dependency_installer.py
VachaShah: Removed the cleanup logic.

VachaShah:
Hi @dblock and @setiah, I have addressed the comments. LMK if this looks good to be merged.

dblock (approved):

peternied (approved):

setiah (approved):
Thanks @VachaShah. LGTM!

PR #338, opened 2021-08-27
Integ tests changes 

setiah:
Integ tests working changes, includes changes for 
* code refactoring, 
* dependency management, 
* supporting test manifest
 

### Check List


bundle-workflow/src/run_integ_test.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
dblock: Not sure I follow.

1. If you need to `import` something it should include `src` via `__init__py`.
2. If the tool assumes locations based on where you're running from the code should change to get a relative location to `os.path.dirname(__file__)`, which doesn't move, similarly to how we do it in ScriptFinder.

bundle-workflow/tests/resources/build_manifest.yml
```diff
@@ -0,0 +1,2462 @@
+build:
```
dblock, line 1: 1. The file should include its version, to avoid future confusion, this is a 1.1.0 buld manifest.
2. We've started a pattern of putting these files under a `data` subfolder of the tests that use them, e.g. https://github.com/opensearch-project/opensearch-build/tree/main/bundle-workflow/tests/test_manifests/data, rather than having a bucket for resources, so we don't end up with one test modifying other tests' data.

dblock (commented):
Move `resources` into a `data` subfolder, version manifests.

bundle-workflow/src/run_integ_test.py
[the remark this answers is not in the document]
setiah: Hey @dblock - sure let me explain. What I meant was, if we have the structure as `test_suites/perf/runner.py`, the runner.py resides in `bundle-workflow/src/test_workflow/test_suites/perf/`. The sys.path (PYTHONPATH) in this case points to '../bundle-workflow/src/test_workflow/test_suites/perf/' and thus fails to read all other modules like git and manifest which are defined in `bundle_workflow/src/git` or `bundle_workflow/src/manifest`. 

To have runner.py import these modules, we need to modify the PYTHONPATH to include upper level directories. This needs to be done right before these modules are imported. like below - 

```
import argparse
import os
import subprocess
import sys

relative_location_find = <find the parent dir location>
sys.path.insert(relative_location_find) <------

from git.git_repository import GitRepository
from manifests.build_manifest import BuildManifest
from manifests.bundle_manifest import BundleManifest
```

For now I kept it simple by having run file in the parent module which knows system paths for modules down the hierarchy. I am not sure what is the right best practice for this but I can revisit this later as a separate PR when I learn more.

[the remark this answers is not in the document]
dblock: I think I get it. These shouldn't be tools. The tool IMO should be [test.sh](https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/test.sh) and the rest is `import`-ed and called and then all the paths would work. The test.py can support`--type=integtest`.

bundle-workflow/src/run_integ_test.py
```diff
@@ -0,0 +1,89 @@
...
+                   shell=True, check=True, capture_output=True)
+
+
+def main():
+    args = parse_arguments()
+    bundle_manifest = BundleManifest.from_file(args.bundle_manifest)
+    build_manifest = BuildManifest.from_file(args.build_manifest)
+    test_manifest = TestManifest.from_file(args.test_manifest)
+    integ_test_config = dict()
+    for component in test_manifest.components:
```
saratvemulapalli, line 72: Trying to understand: Test manifest will be populated by hand as and when new components are supported?
Inner question: How is the test manifest generated, where would it reside at?

bundle-workflow/src/test_workflow/integ/integ_test_suite.py
```diff
@@ -0,0 +1,76 @@
...
+    Kicks of integration tests for a component based on test configurations provided in
+    test_support_matrix.yml
+    """
+    def __init__(self, component, test_config, bundle_manifest, work_dir):
+        self.component = component
+        self.bundle_manifest = bundle_manifest
+        self.work_dir = work_dir
+        self.test_config = test_config
+        self.script_finder = ScriptFinder()
+        self.repo = GitRepository(self.component.repository, self.component.commit_id, os.path.join(self.work_dir, self.component.name))
```
saratvemulapalli, line 26: This could use test_component and directly passed to integ_test_suite: https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/src/test_workflow/test_component.py

bundle-workflow/src/test_workflow/integ/integ_test_suite.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def execute(self):
+        self._fetch_plugin_specific_dependencies()
+        for config in self.test_config.integ_test['test-configs']:
+            self._setup_cluster_and_execute_test_config(config)
+
+    # TODO: fetch pre-built dependencies from s3
+    def _fetch_plugin_specific_dependencies(self):
```
saratvemulapalli, line 34: This does not scale if new dependencies are added. 
Probably tied to the question above for test_manifest, is the plan to read these dependencies from the test_manifest?

bundle-workflow/src/run_integ_test.py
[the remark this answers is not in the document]
setiah: @dblock sure, i would take the wrapping logic as a separate PR when jenkifying these tools as test jobs.

bundle-workflow/src/run_integ_test.py
setiah: test manifest is like an allowlist file. you can think of it as test support matrix. It is not autogenerated. It will reside in the github repo and will keep updating based on the available test support coverage for a component.

bundle-workflow/src/run_integ_test.py
```diff
@@ -0,0 +1,89 @@
...
+COMMON_DEPENDENCIES = [
+    'OpenSearch',
+    'common-utils',
+    'job-scheduler',
+    'alerting'
+]
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(description="Test an OpenSearch Bundle")
```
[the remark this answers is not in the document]
setiah: sure, since this is a moving target, would like to get this merged and take up the refactoring with #357

bundle-workflow/src/test_workflow/integ/integ_test_suite.py
setiah: this is a stop gap solution. Next we will fetch all maven dependencies from s3 and later on we can replace this with a more sophisticated dependency management solution for plugins

bundle-workflow/tests/resources/build_manifest.yml
setiah: these were just place holders for referencing until we start reading manifest files from s3. Removed them

bundle-workflow/src/test_workflow/integ_test_suite.py
```diff
@@ -5,28 +5,72 @@
...
-from paths.tree_walker import walk
-from system.execute import execute
+from test_workflow.local_test_cluster import LocalTestCluster
 
 
 class IntegTestSuite:
-    def __init__(self, name, repo, test_recorder):
-        self.name = name
-        self.repo = repo
-        self.test_recorder = test_recorder
```
setiah, line 18: this component is not tested with the new IntegTestSuite, which is otherwise working. I will integrate this back outside of this PR. Just want to make sure it doesn't break the working integ test workflow.

bundle-workflow/src/run_integ_test.py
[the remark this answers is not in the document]
saratvemulapalli: Sounds good to me.

bundle-workflow/src/test_workflow/integ/integ_test_suite.py
saratvemulapalli: Yeah I agree. Simplest thing to do is to pull all the maven artifacts and eventually optimize it.

bundle-workflow/src/run_integ_test.py
```diff
@@ -0,0 +1,89 @@
...
+
+# TODO: 1. log test related logging into a log file. Output only workflow logs on shell.
+# TODO: 2. Move common functions to utils.py
+
+
+COMMON_DEPENDENCIES = [
+    'OpenSearch',
+    'common-utils',
+    'job-scheduler',
+    'alerting'
```
saratvemulapalli, line 28: As as I could recall notifications is a common dependency and Alerting depends on it.
Anyway way this code should be going away with DependencyInstaller coming in.

bundle-workflow/src/test_workflow/config/test_manifest.yml
```diff
@@ -0,0 +1,16 @@
+components:
+  - name: index-management
+    integ-test:
```
saratvemulapalli, line 3: Just a thought: IntegTests and BwcTests will have the same dependencies for each plugin because they are compile time dependencies and use the same framework. 
Would it be any different for PerformanceTests?

saratvemulapalli (approved):
Overall the changes look good to me. 
I understand we are trying to move fast, but please add unit tests in a follow up PR. 

Couple of other minor comments. Thanks @setiah for the change!

bundle-workflow/src/test_workflow/integ_test_suite.py
```diff
@@ -5,20 +5,73 @@
...
+        subprocess.run('mv alerting notifications', shell=True, check=True)
+        os.chdir(self.work_dir + '/' + '/notifications')
+        subprocess.run(self.work_dir +
+                       '/opensearch-build/tools/standard-test/integtest_dependencies_opensearch.sh alerting ' +
+                       self.bundle_manifest.build.version, shell=True, check=True, capture_output=True)
+
+    # TODO: revisit this once the test_manifest.yml is finalized
+    def _is_security_enabled(self, config):
+        # TODO: Separate this logic in function once we have test-configs defined
+        if config == 'with-security':
```
owaiskazi19, line 55: Can be written as `return True if config == 'with-security' else False`

bundle-workflow/src/run_integ_test.py
```diff
@@ -0,0 +1,89 @@
...
+    for component in build_manifest.components:
+        if component.name in COMMON_DEPENDENCIES:
+            print("Pulling " + component.name)
+            GitRepository(component.repository, component.commit_id, os.path.join(work_dir, component.name))
+
+
+# TODO: replace with DependencyProvider - https://github.com/opensearch-project/opensearch-build/issues/283
+def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
+    os.chdir(work_dir + "/OpenSearch")
+    subprocess.run(work_dir + '/opensearch-build/tools/standard-test/integtest_dependencies_opensearch.sh opensearch ' + manifest_build_ver,
```
owaiskazi19, line 59: We can pull the command as a common variable and then append `opensearch ` and `common-utils` to it. Just to make the change and avoid mirroring

bundle-workflow/src/test_workflow/integ_test_suite.py
```diff
@@ -5,20 +5,73 @@
...
+        self._fetch_plugin_specific_dependencies()
+        for config in self.test_config.integ_test['test-configs']:
+            self._setup_cluster_and_execute_test_config(config)
+
+    # TODO: fetch pre-built dependencies from s3
+    def _fetch_plugin_specific_dependencies(self):
+        os.chdir(self.work_dir)
+        subprocess.run('mv -v job-scheduler ' + self.component.name, shell=True, check=True)
+        os.chdir(self.work_dir + '/' + self.component.name + '/job-scheduler')
+        subprocess.run(self.work_dir +
```
owaiskazi19, line 42: Same as above for the command

owaiskazi19:
Few minor changes. Can pick those up in the next PR.

bundle-workflow/src/run_integ_test.py
setiah: ack. this logic will go away very soon with s3 provided dependencies. just a stop gap for now.

bundle-workflow/src/test_workflow/config/test_manifest.yml
setiah: yeah, perf test run differently than integ tests and bwc tests which are gradle based.

bundle-workflow/src/run_integ_test.py
setiah: Yes, UTs will be added once the integration is complete. The problem here is that it relies on alot of dependent components - DependencyInstaller, S3ReadWrite, build manifests, Integration test setup of plugins. It sticks together these components to run end-to-end workflow for integration testing. These changes have been helpful in catching problems with integration tests early on and callout the requirements for dependency teams. This has helped in parallelizing the effort and articulating the interfaces required for e2e interaction and given we didn't have a design, this has been super helpful in clearing out the ambiguous pieces. Thus you can see that part of it is still written in a poc way. As we productionize things and come up with right interfaces for interactions between individual components, these pieces will be structured in the right way, with UTs. Ideally we should have continued this development in a feature branch, but since we are past that stage, we can continue with main branch and iterate here and since none of this is running in production at this point (productionizing is another task), it works okay.

bundle-workflow/src/test_workflow/config/test_manifest.yml
dblock: FWIW, we typically wrap that kind of stuff with custom scripts ala build.sh or install.sh.

owaiskazi19 (approved):
Few minor changes. Can pick those up in the next PR.

setiah:
Ignoring the "outdated" pipenv flake failures as the standards have changed as W504 is adopted over W503, as per https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator. The code is compliant with W504 styling

dblock:
This broke the build.

Issue #130, opened 2021-08-02
Publish human-readable test run reports as a simple HTML document


gaiksaya:
Publish the performance testing results

camerski:
All test results should be stored in the same location as the bundle. The testing workflows should be given a path to a bundle manifest, for example `s3://bucket/path/to/bundles/1/manifest.yml`, load all the information they need from that manifest, and write results back to a sub-folder (e.g. `s3://bucket/path/to/bundles/1/test_results/performance_tests`).

This should work for, at minimum, S3 locations and local filesystem locations (i.e. `file://` URLs).

dblock:
@gaiksaya looking at https://github.com/opensearch-project/opensearch-build/pull/366 can you please elaborate on what we're trying to do?

I am told the perf test API only outputs JSON? Is this true? How did we use those results in the past? Who looks at these results? 

It strikes me that it's unlikely that the tool doesn't already produce humanly readable results, or that we aren't adding the feature to the tool instead of doing JSON to HTML conversion in our build infrastructure.

peternied:
I believe surfacing the json is more than enough for right now.  Performance reporting which can including tracking and comparisons is a significant amount of work

saratvemulapalli:
Closing this as the change is merged.

PR #366, opened 2021-09-01
Convert the output from the perf test into a HTML doc, to put it into S3 Bucket.

meghasaik:
### Description
The output generated in the console from a perf test is put into a file and taken as an input through command line args. This file is either converted to a HTML displaying either the result in tabular formats or directly as an error response. Then, this file is put into the S3 bucket.
 
### Issues Resolved
Issue No.130 
https://github.com/opensearch-project/opensearch-build/issues/130
 
### Check List


bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,69 @@
...
+# compatible open source license.
+import os
+
+import boto3
+import botocore
+
+
+class S3Bucket:
+    """
+    get_S3Objects:Used for getting the files from S3
```
dblock, line 14: - Documentation doesn't match the code
- File name doesn't match the class

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,69 @@
...
+            DurationSeconds=3600)['Credentials']
+        self.client = boto3.client('s3',
+                                   aws_access_key_id=assumed_role['AccessKeyId'],
+                                   aws_secret_access_key=assumed_role['SecretAccessKey'],
+                                   aws_session_token=assumed_role['SessionToken'])
+
+    def download(self, prefix, dest, file):
+
+        try:
+            if (file == "folder"):
```
dblock, line 40: If the method to download an entire folder is different from downloading a single file, and the caller knows which one is which, create two methods: `download_folder` and `download_file` instead of asking the caller to pass a "folder" parameter.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,69 @@
...
+        assumed_role = sts_connection.assume_role(
+            RoleArn=self.role_arn,
+            RoleSessionName=self.role_session_name,
+            DurationSeconds=3600)['Credentials']
+        self.client = boto3.client('s3',
+                                   aws_access_key_id=assumed_role['AccessKeyId'],
+                                   aws_secret_access_key=assumed_role['SecretAccessKey'],
+                                   aws_session_token=assumed_role['SessionToken'])
+
+    def download(self, prefix, dest, file):
```
dblock, line 37: Is this supposed to download subfolders too? Does it work?

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,69 @@
...
+            if (file == "folder"):
+                s3_resource = boto3.resource('s3')
+                my_bucket = s3_resource.Bucket(self.bucket_name)
+                objects = my_bucket.objects.filter(Prefix=prefix)
+                for obj in objects:
+                    target = obj.key if dest is None \
+                        else os.path.join(dest, os.path.relpath(obj.key, prefix))
+                    if not os.path.exists(os.path.dirname(target)):
+                        os.makedirs(os.path.dirname(target))
+                    if obj.key[-1] == '/':
```
dblock, line 49: When does this happen?

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,69 @@
...
+                        continue
+                    my_bucket.download_file(obj.key, target)
+            else:
+                self.client.download_file(self.bucket_name, prefix, dest)
+
+        except botocore.exceptions.ClientError as e:
```
dblock, line 55: These errors are hiding the actual problem. Just delete all the `except` blocks, let the original error be raised.

bundle-workflow/src/read_write_s3.py
```diff
@@ -0,0 +1,69 @@
...
+            raise ValueError("The error is either: Wrong file or file path provided.")
+    """
+    put_S3Objects:Used for putting the files into S3
+    file_name: The file that you want to put in the S3 bucket
+    file_path: The Object path where you want to save the file
+    """
+    def upload(self, file_name, file_path):
+        try:
+            self.client.upload_file(file_name, self.bucket_name, file_path)
+        except Exception as e:
```
dblock, line 68: Remove the except block, it doesn't add anything useful that I cannot get from the original exception, and shouldn't be `ValueError` either.

bundle-workflow/src/test_workflow/put_file_into_S3.py
```diff
@@ -0,0 +1,361 @@
...
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+import argparse
+import json
+import os
+
+from src.read_write_s3 import S3Bucket
+
+
+class ConvertWrite:
```
dblock, line 14: I am not sure what we're trying to do here, but parsing HTML output to gather test results sounds like not a great idea. Test output must be able to produce structured output that we can examine (json/xml) or a log would suffice.

dblock (commented):
Code comments:

dblock:
@meghasaik What generates the test output? Why are we trying to parse HTML?

dblock:
You should separate the tasks in this PR into 2: 1) conversion from JSON results to HTML 2) reading and writing from S3.

bundle-workflow/src/test_workflow/put_file_into_S3.py
peternied: 💯   A human readable report while laudable is an incredibly complex requirement, and I'm not sure that it serve us to include one at this time.  The json response from mensor while domain specific is not so unreadable that it cannot be used.  Reporting performance data that is broadly useful is a complex task that is out of scope for near term deliverable.

No matter what decision you make, we should include the mensor json output in the artifacts.

bundle-workflow/src/test_workflow/put_file_into_S3.py
```diff
@@ -0,0 +1,361 @@
+# !/usr/bin/env python3
```
peternied, line 1: This file has too many jobs, its parsing input, transformation json into html, and writing out to a file.  These should all be seperate tasks, and I would advocate for the file uploading to be 1 pull request, and the transformation to be in another.

bundle-workflow/src/test_workflow/put_file_into_S3.py
```diff
@@ -0,0 +1,361 @@
...
+
+        return args
+
+    def conversion_to_html(self, args):
+        with open(args.file_path + args.result_file) as json_data:
+            try:
+                data = json.load(json_data)
+                testExecutionId = data['testExecutionId']
+                filename = "%s.html" % testExecutionId
+                values = {1: 'opCount', 2: 'opErrorCount', 3: 'opErrorRate', 4: 'p0', 5: 'p50', 6: 'p100', 7: 'p50',
```
peternied, line 42: Hardcoding the model characteristics into these objects strongly couples their ordering.  If there is an update to add additional fields this will be terrible to maintain.  There are several paradigms for this in software development such as [Model View Controller](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) that create contracts between the components to ease their reuse.

bundle-workflow/src/test_workflow/put_file_into_S3.py
```diff
@@ -0,0 +1,361 @@
...
+                          8: 'p90',
+                          9: 'p99', 10: 'p100'}
+
+                stored = {1: '1', 2: '2', 3: '3', 4: '4', 5: '5', 6: '6', 7: '7', 8: '8', 9: '9', 10: '10'}
```
peternied, line 46: This should be a loop, and why is it 10 items long?

bundle-workflow/src/test_workflow/put_file_into_S3.py
```diff
@@ -0,0 +1,361 @@
...
+                work_config = {1: 'dataset', 2: 'warmupIterations', 3: 'testIterations'}
+
+                def testResults_OS(arg1):
+                    if arg1 == 'index':
+                        arg1 = 'index'
+                    else:
+                        arg1 = 'query'
+
+                    for i in range(1, 11):
+                        if i == 4 or i == 5 or i == 6:
```
peternied, line 64: `Code Smell: In a loop you change behavior based on index`
This code is fragile as the ordering of the incoming data shapes your behavior, if a field is added not only do you need to update this line, but line 68 as well.

bundle-workflow/src/test_workflow/put_file_into_S3.py
```diff
@@ -0,0 +1,361 @@
...
+                            stored[i] = val
+                            print(stored[i])
+                        else:
+                            val = data['testResults']['operationsSummary'][arg1][values[i]]
+                            stored[i] = val
+                            print(stored[i])
+                    return stored
+
+                def testResults_Operations(arg1):
+                    if arg1 == 'auto-date-histogram':
```
peternied, line 79: `Code Smell: Many If/Else statements that repeat a the value matches allow assignment`
Valid values should be in a enum or array that is easily modified as this list will likely expand over time.  That said why does `index-append` need to be treated differently?

Even better rather than depending on detection of a value to change behavior can you check for a type of data and then shape the behavior according to the data?

bundle-workflow/src/test_workflow/put_file_into_S3.py
```diff
@@ -0,0 +1,361 @@
...
+                def testResults_Garbage():
+                    place_holder = {1: '1', 2: '2'}
+                    old = data['testResults']['garbageCollection']['overall']['oldGCTimeMillis']
+                    new = data['testResults']['garbageCollection']['overall']['youngGCTimeMillis']
+                    place_holder[1] = old
+                    place_holder[2] = new
+                    print(stored[i])
+                    return stored
+
+                text = """
```
peternied, line 126: `Code Smell: Long chunks of imperative code`
Create functions that encapsulate parts of the view e.g. GarbageCollection, Cpu, Operations so that they can be rearranged, it would reshape this section of the code to read more like an outline of the output which would be useful to see.  And as always smaller code snippets are easier to review and write unit tests for

bundle-workflow/src/test_workflow/put_file_into_S3.py
```diff
@@ -0,0 +1,361 @@
...
+                it = testResults_Operations('auto-date-histogram-with-tz')
+                for i in range(1, 11):
+                    text = text + "<td>" + str(it.get(i)) + "</td>"
+                text = text + "</tr>" + \
+                    """
+                                   <tr>
+                                       <th scope="row">index-append</th>
+                    """
+                it = testResults_Operations('index-append')
+                for i in range(1, 11):
```
peternied, line 248: Instead iterate over your data, not a constant value.

 If the number of elements changes consider how many lines of code need to be updated, yikes is how many!

bundle-workflow/src/test_workflow/put_file_into_S3.py
```diff
@@ -0,0 +1,361 @@
...
+        parser.add_argument("-s", "--session", nargs="?", default="get-files-session", help="Session name")
+        parser.add_argument("-b", "--bucket_name", help="Bucket name", required=True)
+        parser.add_argument("-o", "--object_path", help="ObjectPath in Bucket", required=True)
+        parser.add_argument("-f", "--file_path", help="Filepath of the result file", required=True)
+
+        args = parser.parse_args()
+
+        return args
+
+    def conversion_to_html(self, args):
```
peternied, line 36: Change this function to accept the json of mensor and return a string of the html - leave the file system management to another file

peternied (changes requested):
I think we could minimize this PR to just upload the mensor json results and consider the other code seperately

PR #367, opened 2021-09-01
S3 utility methods for reading and writing manifests, bundles, and logs in Jenkins

### LINKED ISSUE #260: Copy manifest and bundle from S3

setiah:
### Description
This PR aims to add utility methods for reading/writing to s3. It provides below methods 
* download_folder()
* download_file()
* upload_file() 

These methods will serve as base for
1. Downloading built maven dependencies from s3
2. Downloading the min and full distribution bundles from s3 for spinning up local test cluster
3. Uploading the test results and test logs to s3.
 
This PR overrides #316 . 

### Issues Resolved
#260 
 
### Check List


bundle-workflow/src/s3_utility.py
```diff
@@ -5,31 +5,35 @@
...
+from urllib.parse import urlparse
 
 import boto3
-
 from botocore.exceptions import ClientError
-from pathlib import Path
-from urllib.parse import urlparse
 
 
 def file_download_helper(bucket, key, path):
```
dblock, line 15: Refactor into a private method.

bundle-workflow/src/s3_utility.py
```diff
@@ -5,31 +5,35 @@
...
-    except ClientError as e: # TODO: handle the right exception
-        if e.response['Error']['Code'] == "404":
+    except ClientError as e:  # TODO: handle the right exception
+        if e.response["Error"]["Code"] == "404":
             print("The object does not exist.")
         else:
             raise
 
 
 class S3ReadWrite:
```
dblock, line 25: Rename class. Is this a bucket?

bundle-workflow/src/s3_utility.py
```diff
@@ -5,31 +5,35 @@
...
     def __init__(self, role_arn=None, role_session_name=None):
-        self.role_arn = role_arn if role_arn is not None else os.environ.get('JENKINS_S3_ROLE_ARN')
-        self.role_session_name = role_session_name \
-            if role_session_name is not None else os.environ.get('JENKINS_S3_ROLE_SESSION_NAME')
-        self.s3_resource = boto3.resource('s3')
-        self.s3_client = boto3.client('s3')
+        self.role_arn = (
+            role_arn if role_arn is not None else os.environ.get("JENKINS_S3_ROLE_ARN")
```
dblock, line 28: The ENV var having "JENKINS" it it seems wrong.

bundle-workflow/src/s3_utility.py
```diff
@@ -0,0 +1,96 @@
...
+import boto3
+from botocore.exceptions import ClientError
+
+
+def file_download_helper(bucket, key, path):
+    try:
+        bucket.download_file(key, path)
+    except ClientError as e:  # TODO: handle the right exception
+        if e.response["Error"]["Code"] == "404":
+            print("The object does not exist.")
```
dblock, line 20: Raise a FileNotFoundError

bundle-workflow/src/s3_utility.py
setiah: It's a utility class that provides methods to read/write files/folders to/from s3. Sure, any suggestions? I could think of S3Wrapper, what else?

bundle-workflow/src/s3_utility.py
setiah: renamed it to `TEST_ORCHESTRATOR_ROLE_ARN`

dblock: And why not `S3_ROLE_ARN` or `IAM_ROLE_ARN` or `ROLE_ARN`?

bundle-workflow/src/s3_utility.py
dblock: It represents an S3 bucket, but reads and writes in a certain peculiar way (assumes role). I would put this in a new namespace called `aws.s3` and call it a `Bucket`.

setiah: It does not represent an S3 bucket. It is like a wrapper over s3 boto apis - that wraps GetObject, PutObject apis to provide more specific abstractions for the purpose of test-infrastructure - download_folder, download_file, upload_file. You can provide uri related to any S3 bucket

bundle-workflow/src/s3_utility.py
setiah: Because it has specific permissions that are required by test-orchestrator workflow only. 

Read access for - build/ and bundles/
Write access for - test/

in the s3 bucket

bundle-workflow/src/s3_utility.py
dblock: Then it should be all static methods and called, dunno, `boto4`? I think abstracting this away in OO is cleaner.

bundle-workflow/src/s3_utility.py
dblock: If the class is generic, environment variables should not be specific.

setiah: I see your point now. Sure, will rename it to `ROLE_ARN`. Thx!

VachaShah:
You can use [`mypy-boto3-s3`](https://mypy-boto3.readthedocs.io/en/latest/) to resolve the mypy errors and then use the specific client and resource:

```
self.s3_resource: S3ServiceResource = boto3.resource("s3")
self.s3_client: S3Client = boto3.client("s3")
```

owaiskazi19:
> You can use [`mypy-boto3-s3`](https://mypy-boto3.readthedocs.io/en/latest/) to resolve the mypy errors and then use the specific client and resource:
> 
> ```
> self.s3_resource: S3ServiceResource = boto3.resource("s3")
> self.s3_client: S3Client = boto3.client("s3")
> ```

mypy requires stubbed version of the library to check for type. This should be present in Pipfile https://pypi.org/project/boto3-stubs/ to resolve errors related to boto3

VachaShah:
> > You can use [`mypy-boto3-s3`](https://mypy-boto3.readthedocs.io/en/latest/) to resolve the mypy errors and then use the specific client and resource:
> > ```
> > self.s3_resource: S3ServiceResource = boto3.resource("s3")
> > self.s3_client: S3Client = boto3.client("s3")
> > ```
> 
> mypy requires stubbed version of the library to check for type. This should be present in Pipfile https://pypi.org/project/boto3-stubs/ to resolve errors related to boto3

Yeah, will require changes to Pipfile as well: I have made changes here: https://github.com/opensearch-project/opensearch-build/pull/318/commits/a8a58c2bacc65ac5fbcd44c5b38babbb428ead9c to resolve the errors.

bundle-workflow/src/s3_utility.py
```diff
@@ -0,0 +1,137 @@
...
+            if role_arn is not None
+            else os.environ.get("AWS_ROLE_ARN")
+        )
+        self.role_session_name = (
+            role_session_name
+            if role_session_name is not None
+            else os.environ.get("AWS_ROLE_SESSION_NAME")
+        )
+        assumed_role_cred = self.__assume_role()
+        self.__s3_client = boto3.client(
```
dblock, line 36: Nit: I would save `__ assumed_role_cred` and create `__s3_client` and `__s3_resource ` on demand the first time you need it.

bundle-workflow/src/s3_utility.py
```diff
@@ -0,0 +1,137 @@
...
+        :param role_session_name: the aws role session name
+        """
+        s3bucket = cls(bucket_name, role_arn, role_session_name)
+        bucket = s3bucket.__s3_resource.Bucket(bucket_name)
+        local_dir = Path(dest)
+        file_name = key.split("/")[-1]
+        target = Path(local_dir) / Path(file_name)
+        return s3bucket.__file_download_helper(bucket, key, str(target))
+
+    @staticmethod
```
dblock, line 106: You have this to wrap `ClientError` into `S3DownloadFailureException`. Since I can also call `bucket.download_file` externally myself, I would get a `ClientError` there. I think it's surprising to get different exceptions depending on whether I call a static helper or the implementation in it, so I think you should remove this code and do the `except` inside `download_file`.

bundle-workflow/src/s3_utility.py
```diff
@@ -0,0 +1,137 @@
...
+
+        :param bucket_name: The s3 bucket name
+        :param key: The s3 key for the uploaded object
+        :param source: local path of the file
+        :param role_arn: the arn of the role that has permissions to access S3
+        :param role_session_name: the aws role session name
+        """
+        s3bucket = cls(bucket_name, role_arn, role_session_name)
+        try:
+            s3bucket.__s3_client.upload_file(source, bucket_name, key)
```
dblock, line 127: You shouldn't be reaching into a private method of `s3bucket`. Add a `s3bucket#upload_file` method.

bundle-workflow/src/s3_utility.py
```diff
@@ -0,0 +1,137 @@
...
+        except ClientError as e:
+            raise S3UploadFailureException(e)
+
+
+class S3DownloadFailureException(Exception):
```
dblock, line 132: I think the Python convention is to call these things `S3DownloadFailureError` vs. `Exception`, but they all derive from the `Exception` class.

I would create a `S3FileError` as a base class and store the key in it. Then `S3UploadError` and `S3DownloadError` would switch the error message.

bundle-workflow/src/s3_utility.py
```diff
@@ -0,0 +1,137 @@
...
+    def __assume_role(self):
+        try:
+            sts_connection = boto3.client("sts")
+            return sts_connection.assume_role(
+                RoleArn=self.role_arn,
+                RoleSessionName=self.role_session_name,
+                DurationSeconds=3600,
+            )["Credentials"]
+        except Exception as e:
+            print("Assume role failed due to ", str(e.__repr__()))
```
dblock, line 58: Wrap in a specialized `STSError`.

dblock (commented):
This is a lot cleaner and a much more sane interface, don't you think?

Some comments/improvements + tests and this is good to go.

bundle-workflow/src/s3_utility.py
setiah: `__file_download_helper` exists to wrap the logic common to both `download_folder` and `download_file`. Not sure how you want to replace this?

Also, the S3DownloadFailureException wraps the`ClientError`. The caller can see it in the stack trace. The reason for having the wrapper is to have a high level exception class catch all S3 download failures. We can revisit this later to add things like retryability, and make this more robust.

setiah:
thanks @dblock for reviewing the PR. If this looks good, can we merge it?

bundle-workflow/tests/test_aws/test_s3_bucket.py
```diff
@@ -0,0 +1,189 @@
...
+from unittest.mock import MagicMock, call, patch
+
+from botocore.exceptions import ClientError
+
+from aws.s3_bucket import S3Bucket, S3DownloadError, STSError
+
+mock_sts = MagicMock()
+mock_s3_resource = MagicMock()
+mock_s3_client = MagicMock()
+bucket_name = "unitTestBucket"
```
dblock, line 17: Any reason to reuse these globals instead of just doing `MagicMock()` in the side effect code?

dblock (approved):

bundle-workflow/tests/test_aws/test_s3_bucket.py
setiah: Yeah, somehow the `mock_s3_resource` wasn't picking the same mock object with MagicMock inside side effect. Will revisit this later and fix.

owaiskazi19 (approved):
LGTM!

PR #371, opened 2021-09-01
Test orchestrator pipeline jenkins DSL

### LINKED ISSUE #262: Make pipeline parameterized to take S3 manifest as input parameter
Tasks
- [x] Make pipeline parameterized to accept inputs from build workflow
- [x] Orchestrate the integ, bwc and perf test jobs. These jobs are already parameterized but need input from pipeline

setiah:
### Description
This PR is work in progress. We are collaborating on these changes so this PR is expected to stay in draft mode until they are ready. 

The code here defines test orchestrator pipeline and its test stages. The pipeline accepts a bunch of parameters from build workflow and uses those to kick off downstream test workflows. 
 
### Issues Resolved
#262
 
cc: @saratvemulapalli 

### Check List

### DIFF
--- a/jenkins/test_orchestrator/JenkinsFile
+++ b/jenkins/test_orchestrator/JenkinsFile
@@ -1,0 +1,49 @@
+pipeline {
+    agent any
+
+    stages {
+        stage('Read Manifest') {
+            steps {
+                echo "todo: read the manifest file and download the bundle"
+            }
+        }
+        stage('Execute Tests') {
+            steps {
+                // Freestyle build trigger calls a list of jobs
+                // Pipeline build() step only calls one job
+                // To run all three jobs in parallel, we use "parallel" step
+                // https://jenkins.io/doc/pipeline/examples/#jobs-in-parallel
+                parallel (
+                    IntegTests: {
+                        // build job: 'opensearch_bundle_integtest'
+                        build job: 'opensearch_bundle_integtest',
+                        parameters: [
+                            string(name: 'PLUGIN_REPO_REFERENCE', value: 'main'),
+                            string(name: 'PLUGIN_VERSION', value: '1.0.0.0'),
+                            string(name: 'VERSION', value: '1.0.0'),
+                            string(name: 'ENABLE_SECURITY', value: 'true'),
+                            string(name: 'PLUGIN_TYPE', value: 'opensearch'),
+                            string(name: 'PLUGIN_NAME', value: 'anomaly-detection'),
+                            string(name: 'ARCHITECTURE', value: 'x64'),
+                            string(name: 'CLUSTER_PORT', value: '9200'),
+                            string(name: 'REVISION', value: '1.0.0'),
+                            string(name: 'QUALIFIER', value: '')
+                            ]
+                    },
+                    PerfTests: {
+                        build job: 'perf-tests'
+                    },
+                    BwcTests: {
+                        build job: 'bwc-tests'
+                    },
+                    failFast: true)
+            }
+        }
+        stage('Notify') {
+            failFast false
+            steps {
+                echo "todo: Notify slack channel for successful build"
+            }
+        }
+    }
+}
--- a/jenkins/test_orchestrator/JenkinsFile
+++ b/jenkins/test_orchestrator/JenkinsFile
@@ -5,52 +5,41 @@ pipeline {
         string(name: 's3_bucket'),
         string(name: 'opensearch_version'),
         string(name: 'build_id'),
-        string(name: 'platform_architecture')
+        string(name: 'architecture')
     }
 
     stages {
-        stage('Read Manifest') {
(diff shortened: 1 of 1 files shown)


jenkins/test_orchestrator/JenkinsFile
```diff
@@ -0,0 +1,62 @@
...
+        string(name: 'architecture')
+    }
+
+    stages {
+        stage('Execute Tests') {
+            steps {
+                // To run all three test jobs in parallel, we use "parallel" step
+                // https://jenkins.io/doc/pipeline/examples/#jobs-in-parallel
+                parallel (
+                    IntegTests: {
```
peternied, line 17: Does this spin up 3 agent nodes?

setiah: no, this doesn't create any infra. It should be created separately via cdk.

peternied: Per #399 we should not use parallel, and instead run these stages serially until we can invest in this improvement.  As much as I know it would be a big perf boost, its not worth the stability trade-off.

setiah: If the parallel infrastructure is not available, this will automatically kick these stages off in serial order.

jenkins/test_orchestrator/JenkinsFile
```diff
@@ -0,0 +1,62 @@
...
+                            string(name: 'opensearch_version', value: ${params.opensearch_version}),
+                            string(name: 'build_id', value: ${params.build_id}),
+                            string(name: 'architecture', value: ${params.architecture})
+                        ]
+                    },
+                    failFast: false)
+            }
+        }
+
+        stage('ResultPublisher') {
```
setiah, line 48: will remove this stage as it is already handled inside the python tool at per test case level

jenkins/test_orchestrator/JenkinsFile
```diff
@@ -0,0 +1,62 @@
...
+    stages {
+        stage('Execute Tests') {
+            steps {
+                // To run all three test jobs in parallel, we use "parallel" step
+                // https://jenkins.io/doc/pipeline/examples/#jobs-in-parallel
+                parallel (
+                    IntegTests: {
+                        build job: 'integ-tests'
+                        parameters: [
+                            string(name: 's3_bucket', value: ${params.s3_bucket}),
```
saratvemulapalli, line 20: Artifact S3 bucket could be fetched via Jenkins env variable `ARTIFACT_BUCKET_NAME`

jenkins/test_orchestrator/JenkinsFile
```diff
@@ -0,0 +1,62 @@
...
+                            string(name: 'opensearch_version', value: ${params.opensearch_version}),
+                            string(name: 'build_id', value: ${params.build_id}),
+                            string(name: 'architecture', value: ${params.architecture})
```
saratvemulapalli, line 23: These jobs need to take in a `test_run_id`.

jenkins/test_orchestrator/JenkinsFile
saratvemulapalli: We might want to eventually consolidate all the results across test suites so that its readable for a single run. But again that P1/P2.

jenkins/test_orchestrator/JenkinsFile
```diff
@@ -0,0 +1,62 @@
+pipeline {
+    agent any
+
+    parameters {
+        string(name: 's3_bucket'),
```
saratvemulapalli, line 5: As commented below we do not need the S3 bucket parameter to be fed for the orchestrator.

jenkins/test_orchestrator/JenkinsFile
```diff
@@ -0,0 +1,62 @@
+pipeline {
```
saratvemulapalli, line 1: Since these are workflows, I have structured it under bundle-workflow
Ref: https://github.com/opensearch-project/opensearch-build/pull/423/files#diff-500b943d0ad57b148b713ee7cfbca15515eede9d0e9614d1ef8d9f16f30a0001
`bundle-workflow/jenkins_workflow/integtest/Jenkinsfile`

You could do something similar, `bundle-workflow/jenkins_workflow/test_orchestrator/Jenkinsfile` and also add a Readme.md

dblock:
What do we want to do with this @saratvemulapalli @setiah ?

saratvemulapalli:
> What do we want to do with this @saratvemulapalli @setiah ?

@setiah will be working on this soon.

jenkins/test_orchestrator/JenkinsFile
setiah: sure, added in #523

setiah:
Closing this in favor of updated PR #523 which addresses all comments

PR #372, opened 2021-09-01
Check versions and file names in plugin and maven artifacts.

### LINKED ISSUE #137: Check versions of artifacts bundled in a distribution
Bundle being built for a release do not verify the versions of the artifacts or that "everything works together".
Artifacts name them version x but can still be of version y and built out of version z of OpenSearch/Dashboards 
We'd like to improve the release quality. We should an automated way to verify the artifacts provided with the version they claim to be and the version we'd like to build for release.

Whats the motivation behind this?
Anomaly Detection plugin was still using OpenSearch and its dependencies with 1.0.0-rc1 and claiming the plugin version to be 1.0.0.0. But really underneath they were running different versions and cannot be bundled for a release as the plugin should have been built with 1.0.0 of OpenSearch and 1.0.0.0 of other dependent plugins/libraries.  
Ref: 
AD 1.0 release branch: https://github.com/opensearch-project/anomaly-detection/blob/1.0.0.0/build.gradle#L33-L35
Issue to fix after the release is done: https://github.com/opensearch-project/anomaly-detection/issues/153

We cannot manually scale and should have an automated workflow to check these artifacts for every release to start with.
With the CI infrastructure we are building, we should be able to add another CI job to verify the artifacts just before bundling them for the release.

### LINKED ISSUE #354: Check version of plugins and JARs being packaged and distributed
Ensure that when we build bundles the packaged ZIPs all have the matching version and snapshot flag(s) in automation. 

Assembly should fail if there's a version mismatch in any of the components. 

**Additional context**

- check the filename contains version and ends with -SNAPSHOT for snapshot builds
- check zip contains manifest that has a matching version
- check JAR version

#### Checking JAR Versions

```
$ unzip -p target/opensearch-security-1.1.0.0-SNAPSHOT.jar META-INF/MANIFEST.MF
Manifest-Version: 1.0
Created-By: Apache Maven 3.6.3
Build-Jdk: 14.0.2
Implementation-Title: OpenSearch Security
Implementation-Version: 1.1.0.0-SNAPSHOT
Implementation-Vendor-Id: org.opensearch
Implementation-URL: https://github.com/opensearch-project/security
Build-Time: 2021-08-30T22:05:33Z
Built-By: Amazon OpenDistro ElasticSearch Security Parent
git-sha1: 7d3518f77f7e614e7bec8523b2c867b34f1b67d6
```

dblock:

### Description

1. Adds component versions to the build manifest.
2. Checks versions and file names in plugin artifacts. Ensures that the plugin-descriptor.properties is built against the OpenSearch version being built, and that its own version matches the filename. 
3. Checks versions and file extensions in maven artifacts. Ensures that the MANIFEST.mf contains the correct `Implementation-Version`, if any. Some components like common-utils do not have this at all, will open an issue whether we need all of them to have it.

### Issues Resolved

Closes #137. 
 
### Check List

### DIFF
--- a/bundle-workflow/src/build_workflow/build_recorder.py
+++ b/bundle-workflow/src/build_workflow/build_recorder.py
@@ -5,7 +5,9 @@
 # compatible open source license.
 
 import os
+import re
 import shutil
+from zipfile import ZipFile
 
 import yaml
 
@@ -13,15 +15,56 @@ from manifests.build_manifest import BuildManifest
 
 
 class BuildRecorder:
+    class ArtifactInvalidError(Exception):
+        def __init__(self, path, message):
+            self.path = path
+            super().__init__(
+                f"Artifact {os.path.basename(path)} is invalid: {message}."
+            )
+
+    class PropertiesFile:
+        def __init__(self, filename, data):
+            self.filename = filename
+            self.data = data
+
+        def get_value(self, key):
+            result = re.findall(f"^{key}=(.*)$", self.data, re.MULTILINE)
+            return result[0] if result else None
+
+        def check_value(self, key, expected_value):
+            value = self.get_value(key)
+            if not value:
+                raise BuildRecorder.ArtifactInvalidError(
+                    self.filename,
+                    f"expected to have {key}={expected_value}, but none was found",
+                )
+            elif value != expected_value:
+                raise BuildRecorder.ArtifactInvalidError(
+                    self.filename,
+                    f"expected to have {key}={expected_value}, but was {value}",
+                )
+
     def __init__(self, build_id, output_dir, name, version, arch, snapshot):
         self.output_dir = output_dir
+        self.version = str(version)
+        self.opensearch_version = (
+            self.version + "-SNAPSHOT" if snapshot else self.version
+        )
+        # BUG: the 4th digit is dictated by the component, it's not .0, this will break for 1.1.0.1
+        self.component_version = (
+            self.version + ".0-SNAPSHOT" if snapshot else f"{self.version}.0"
+        )
         self.build_manifest = self.BuildManifestBuilder(
-            build_id, name, version, arch, snapshot
+            build_id, name, self.opensearch_version, arch, snapshot
         )
 
     def record_component(self, component_name, git_repo):
         self.build_manifest.append_component(
-            component_name, git_repo.url, git_repo.ref, git_repo.sha
+            component_name,
+            self.component_version,
+            git_repo.url,
+            git_repo.ref,
+            git_repo.sha,
         )
 
     def record_artifact(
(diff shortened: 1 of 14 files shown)


peternied (approved):

bundle-workflow/src/build_workflow/build_recorder.py
```diff
@@ -5,23 +5,66 @@
...
+        )
         self.build_manifest = self.BuildManifestBuilder(
-            build_id, name, version, arch, snapshot
+            build_id, name, self.opensearch_version, arch, snapshot
         )
 
     def record_component(self, component_name, git_repo):
         self.build_manifest.append_component(
-            component_name, git_repo.url, git_repo.ref, git_repo.sha
+            component_name,
```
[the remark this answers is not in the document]
dblock: I usually run _black_ that auto-formats, but we use _flake8_ to check. They aren't quite in sync, I think _black_ definitely complies to _flake8_.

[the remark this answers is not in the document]
dblock: Note that here a parameter was added (`self.component_version`).

peternied (approved):

bundle-workflow/src/build_workflow/build_recorder.py
```diff
@@ -6,22 +6,84 @@
...
+
     def __init__(self, build_id, output_dir, name, version, arch, snapshot):
         self.output_dir = output_dir
+        self.version = str(version)
+        self.opensearch_version = (
+            self.version + "-SNAPSHOT" if snapshot else self.version
+        )
+        # BUG: the 4th digit is dictated by the component, it's not .0, this will break for 1.1.0.1
+        self.component_version = (
+            self.version + ".0-SNAPSHOT" if snapshot else f"{self.version}.0"
```
setiah, line 74: can we replace SNAPSHOT with a variable QUALIFIER? It could be an "RC" candidate too in future

bundle-workflow/src/build_workflow/build_recorder.py
```diff
@@ -50,29 +114,79 @@ def write_manifest(self, dest_dir):
...
 
-        def append_component(self, name, repository_url, ref, commit_id):
+        def append_component(self, name, version, repository_url, ref, commit_id):
             component = {
                 "name": name,
                 "repository": repository_url,
                 "ref": ref,
                 "commit_id": commit_id,
                 "artifacts": {},
+                "version": version,
```
setiah, line 189: along similar lines as previous comment, does it make sense to have a `qualifier` (e.g. snapshot, rc, etc.) defined separately?

setiah (approved):
Minor comments. Mostly LGTM!

bundle-workflow/src/build_workflow/build_recorder.py
dblock: That won't be enough to support build with a qualifier, I opened https://github.com/opensearch-project/opensearch-build/issues/385 for that.

bundle-workflow/src/build_workflow/build_recorder.py
dblock: This is the actual component version as written into the JAR. We can separate a qualifier as part of https://github.com/opensearch-project/opensearch-build/issues/385

PR #375, opened 2021-09-02
Upload the file into S3

meghasaik:
### Description
This file takes the parameters from the user and gives the JSON output file to the class where the conversion to HTML occurs. It gets a HTML string which is converted to a HTML file either containing an output or an error message. This file is given to the S3 class having the upload function. 
 
### Issues Resolved
This PR resolves a part of the conversion of JSON output to HTML file and uploading it to the S3 bucket.
 
### Check List
- Signed-off-by: Megha Sai Kavikondala <kavmegha@amazon.com>


dblock (changes requested):
I don't see any code that actually uploads to S3 here. I see a tool that calls methods that don't exist.

meghasaik:
The name has to be changed, it is a little misleading. This tool basically calls the  method which put the files into S3. This methods are basically being used in the code for the Issue no 260.

dblock:
Happy to re-review when there's end-to-end working code with tests.

dblock:
This was done as part of https://github.com/opensearch-project/opensearch-build/pull/367 I believe, closing.

PR #376, opened 2021-09-02
The branch where the HTML conversion occurs

meghasaik:
### Description
This file takes the parameters from the user and gives the JSON output file to the class where the conversion to HTML occurs. It gets a HTML string which is converted to a HTML file either containing an output or an error message. This file is given to the S3 class having the upload function.
 
### Issues Resolved
This PR resolves a part of the conversion of JSON output to HTML file and uploading it to the S3 bucket.
 
### Check List


bundle-workflow/src/test_workflow/transform_MensorResults.py
```diff
@@ -0,0 +1,25 @@
...
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+import json
+
+import json2table
+
+
+class JsonTransformation:
+
+    def convert_to_html(self, json_file, path_to_jsonfile):
```
peternied, line 13: This interface seems a little clunky - please add comment to clarify.

What about accepting only the data object and returning a str that is html, let callers write it to a file or not.

bundle-workflow/src/test_workflow/transform_MensorResults.py
```diff
@@ -0,0 +1,25 @@
...
+        with open(path_to_jsonfile + '/' + json_file) as file:
+            try:
+                infoFromJson = json.load(file)
+                testExecutionId = infoFromJson['testExecutionId']
+                build_direction = "TOP_TO_BOTTOM"
+                table_attributes = {"style": "width:100%", "border": "solid"}
```
peternied, line 19: Can you provide a screenshot of how this renders?

bundle-workflow/Pipfile
```diff
@@ -9,6 +9,9 @@ requests = "~=2.26"
 isort = "~=5.9"
 flake8 = "~=3.9"
 pytest = "*"
+boto3 = "1.18.31"
+botocore = "1.21.33"
+json2table = "1.1.5"
```
peternied, line 14: There are a couple of libraries that do this [json2html](https://pypi.org/project/json2html), is there a reason you picked this one over others?

peternied (changes requested):
The commit history of this pull request still needs to be cleaned up, please rebase your history to collapse the number of commits associated with this change.

bundle-workflow/src/test_workflow/transform_MensorResults.py
meghasaik: As we are going to store both the html file as well as the json file in the S3 bucket, correct me if I am wrong but won't it be much simpler to store the output generated into the file directly and then working on that file

bundle-workflow/Pipfile
meghasaik: json2table gives more options in terms of rotating the table from vertical to a hroizontal view as well as table attributes.

bundle-workflow/src/test_workflow/transform_MensorResults.py
meghasaik: Screen Shot 2021-09-02 at 9 56 48 AM

This is for the vertical view


Screen Shot 2021-09-02 at 10 00 04 AM

This is for the horizontal view

bundle-workflow/Pipfile
```diff
@@ -9,6 +9,9 @@ requests = "~=2.26"
 isort = "~=5.9"
 flake8 = "~=3.9"
 pytest = "*"
+boto3 = "1.18.31"
```
dblock, line 12: This is from another PR.

bundle-workflow/src/test_workflow/transform_MensorResults.py
```diff
@@ -0,0 +1,25 @@
...
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+import json
+
+import json2table
+
+
+class JsonTransformation:
```
dblock, line 11: The filename doesn't match class name.

bundle-workflow/src/test_workflow/transform_MensorResults.py
```diff
@@ -0,0 +1,25 @@
...
+        with open(path_to_jsonfile + '/' + json_file) as file:
+            try:
+                infoFromJson = json.load(file)
+                testExecutionId = infoFromJson['testExecutionId']
+                build_direction = "TOP_TO_BOTTOM"
+                table_attributes = {"style": "width:100%", "border": "solid"}
+                stored = json2table.convert(infoFromJson,
+                                            build_direction=build_direction,
+                                            table_attributes=table_attributes)
+                return stored, "successful", testExecutionId
```
dblock, line 23: Errors are swallowed and some strange structure is returned.

dblock (changes requested):
The is not quite object-oriented and needs some design.

I suggest this: the class represent performance tests output, so I would call it `TestResults`. It will have a static method called `from_path(path)`that takes the test results JSON filename and loads that returning an instance of `TestResults`. Then `TestResults` has an instance method called `to_html` and `save_to(path)` that writes HTML for those results. Optionally `save_to` can look at the extension of `path` to decide whether to write HTML or JSON.

```python
class TestResults():
   @staticmethod
   def from_path(cls, path):
   
   def to_html(self):
   def save_to(self, path);
```


This code will need tests and documentation to get merged.

dblock:
You have a ton of commit history here that's not useful. Start a new branch.

If you accidentally made commits on your `main`, first make sure it's in sync with upstream.

```
git checkout main
git reset HEAD~25 --hard
git pull upstream main
git push origin main
```

... then start a new branch.

bundle-workflow/src/test_workflow/transform_MensorResults.py
meghasaik: So, the idea was, as the result generated might either be a json structure or error message, I thought of using try and except block for differentiating between result and error for printing that particular message into the file.

dblock: If the result is an error, throw an exception. 

Otherwise return the result. In this case returning `successful` is unnecessary because the function will fail (exception will be raised) on error, and `testExecutionId` is already known to the caller (it's in the json output that was input to this code).

dblock:
I believe this moved to a different repo, can it be closed? (If so please do)

meghasaik:
Yes, this has been moved to a different repo, I will close it now.

PR #377, opened 2021-09-02
Adding bwc_test.py as a single entry point for bwc tests and cleanup

### LINKED ISSUE #275: Add support for running bwctest.sh

saratvemulapalli:
Signed-off-by: Sarat Vemulapalli <vemulapallisarat@gmail.com>

### Description
Adding bwc_test.py as a single entry point for bwc tests and cleanup
 
### Issues Resolved
Closes #275 
 
### Check List

### DIFF
--- a/bundle-workflow/src/bwc_test.py
+++ b/bundle-workflow/src/bwc_test.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import os
+import sys
+
+from manifests.bundle_manifest import BundleManifest
+from system.temporary_directory import TemporaryDirectory
+from test_workflow.bwc_test_suite import BwcTestSuite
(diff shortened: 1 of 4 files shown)


dblock:
So all these test suites are similar and take `TestArgs`? Why wouldn't we subclass suites from something generic and call `.execute()` on them instead of having `main()` that is also executed from the top?

bundle-workflow/src/bwc_test.py
```diff
@@ -0,0 +1,30 @@
...
+import os
+import sys
+
+from manifests.bundle_manifest import BundleManifest
+from system.temporary_directory import TemporaryDirectory
+from test_workflow.bwc_test_suite import BwcTestSuite
+from test_workflow.test_args import TestArgs
+
+
+def main():
```
dblock, line 17: I think this should be a BwcTestSuite class with an `.execute` method that takes `TestArgs`. The only way to call this would be `test.sh`.

saratvemulapalli:
Had a chat with @dblock, will close this PR for now and work on other priority items, will come back to this. 

Summary:
The test workflows need restructuring for all the test suites. Today each test suite is using its own model of invoking the tests via test.sh or python <test_suite>.py.

PR #386, opened 2021-09-03
Sanity check the OpenSearch bundle.

dblock:

### Description

Today it's possible to make a change to an OpenSearch bundle manifest and break everything. The simplest mistake is a typo in the GitHub URL. This PR adds a CI workflow and a GitHub action that runs it to ensure that the components are sane.  Future versions can get more crafty in terms of what's being checked (e.g. plugin versions). 

Originally I thought I'd add a dry run or a check option to the build workflow, but I think we're going to get more and more advanced with what we want to do in checks, so adding a new workflow for that.
 
### Check List

### DIFF
--- a/.github/workflows/manifests.yml
+++ b/.github/workflows/manifests.yml
@@ -1,0 +1,24 @@
+name: manifests
+
+on: [push, pull_request]
+
+jobs:
+  check:
+
+    runs-on: ubuntu-latest
+
+    env:
+      PYTHON_VERSION: 3.7
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ env.PYTHON_VERSION }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+      - name: Install Pipenv and Dependencies
+        run: |
+          python -m pip install --upgrade pipenv wheel
+      - name: OpenSearch 1.1
+        run: |
+          ./bundle-workflow/ci.sh manifests/opensearch-1.1.0.yml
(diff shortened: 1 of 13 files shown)


.github/workflows/manifests.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Set up Python ${{ env.PYTHON_VERSION }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+      - name: Install Pipenv and Dependencies
+        run: |
+          python -m pip install --upgrade pipenv wheel
+      - name: OpenSearch 1.1
+        run: |
+          ./bundle-workflow/ci.sh manifests/opensearch-1.1.0.yml
```
peternied, line 24: Rather than only run on the "current" manifest, could we have this enforced on all manifest files?

peternied (approved):
I've got a slight anxiety about this check since it is an integration tests between repositories and thus could break out of band of changes here... however this is tempered by the fact no one should delete branches they want to ship.

.github/workflows/manifests.yml
dblock: Good idea.

dblock: I figured out how to dynamically list manifest files and parallel run checks on them :)

Screen Shot 2021-09-03 at 6 28 28 PM

dblock:
> I've got a slight anxiety about this check since it is an integration tests between repositories and thus could break out of band of changes here... however this is tempered by the fact no one should delete branches they want to ship.

And this should allow us to quickly locate any such problems!

.github/workflows/manifests.yml
peternied: Amazing! 🚢

peternied (dismissed):

peternied (approved):

mch2 (approved):

saratvemulapalli (approved):
This is awesome :)!

PR #402, opened 2021-09-07
Added support for working_subdirectory.

dblock:

### Description

Introduces `GitRepository#working_subdirectory` into the constructor so that you can default script execution in a subdir. 

This is useful for the two repos where the action happens in a subdirectory, and for a future check that I am working on that will be running `gradlew`. For reports and notebooks the custom scripts that do `cd subdir` are thus no longer necessary.  

The change required quite a bit of refactoring in execution of commands, so appreciate a careful read.
  
### Check List


bundle-workflow/tests/tests_build_workflow/test_builder.py
```diff
@@ -73,17 +75,17 @@ def test_export_artifacts(self, mock_walk):
         mock_walk.side_effect = self.mock_os_walk
         self.builder.export_artifacts()
         self.assertEqual(self.builder.build_recorder.record_artifact.call_count, 2)
-        self.builder.build_recorder.record_artifact.has_calls(
-            [
+        self.builder.build_recorder.record_artifact.assert_has_calls([
```
dblock, line 78: This was an earlier mistake that didn't assert anything.

bundle-workflow/src/git/git_repository.py
```diff
@@ -29,29 +29,38 @@ def __init__(self, url, ref, directory=None):
...
-        self.execute(f"git remote add origin {self.url}", True)
-        self.execute(f"git fetch --depth 1 origin {self.ref}", True)
-        self.execute("git checkout FETCH_HEAD", True)
-        self.sha = (
-            subprocess.check_output(["git", "rev-parse", "HEAD"], cwd=self.dir)
-            .decode()
-            .strip()
-        )
+        self.working_subdirectory = working_subdirectory
+        self.execute_silent("git init", self.dir)
```
mch2, line 33: I think self.dir will  always be set to a value earlier in this constructor, so it will never be 'None' here.  In execute_silent you'd never use the working_directory?

`cwd = cwd or self.working_directory`

dblock: That’s what we want. The root dir is where we’re are checking files out (internal to the class), but the working directory is where we run commands from outside.

dblock: You can call execute_silent externally and it will use working directory by default like all execute calls.

mch2: ahh I see, thanks!

mch2 (approved):

VachaShah (approved):

VachaShah (approved):
LGTM!

Issue #215, opened 2021-08-13
Implement a `PerformanceTestSuite` class and wire it in to `test.py`


gaiksaya:
Implement a `PerformanceTestSuite` class and wire it in to `test.py`. This could be similar to `IntegTestSuite` if we need per-component performance tests, or it could be a bundle-level thing. In any case, the interface should be similar: Accept a test cluster endpoint, return when the tests are finished, report results in some sensible fashion.

See https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/python/test.py and https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/python/test_workflow/integ_test_suite.py

_Originally posted by @camerski in https://github.com/opensearch-project/opensearch-build/issues/129#issuecomment-898723720_

- [ ] Integrate the performance test cluster
- [ ] Run rally test for the cluster

PR #314, opened 2021-08-25
Adds PerformanceTestSuiteClass and PerformanceTestCluster for Performance Testing

owaiskazi19:
### Description
Implemented PerformanceTestSuiteClass and PerformanceTestCluster. 
1. Creates a single node cluster by deploying cdk and returns a private IP of the cluster
2. Pass the private IP to rally to start performance test
3. Unit tests for Perf Test

### Prerequisites 
1.  Manifest file containing the build information
2. Config file having constants required for the rally to run
3. Set up `PYTHONPATH`
4. Add AssumeRole policy to the IAM role

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/215
https://github.com/opensearch-project/opensearch-build/issues/282

### Testing
- Done on Jenkins

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #404, opened 2021-09-07
Cleanup for Performance Test

owaiskazi19:

### Description
Addressed comments of https://github.com/opensearch-project/opensearch-build/pull/314
 
### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-build/issues/215
 
### Check List


bundle-workflow/src/perf_test.py
```diff
@@ -39,8 +39,7 @@ def get_infra_repo_url():
         security = True
 
 with WorkingDirectory(current_workspace) as curdir:
-    with Cluster.create(manifest, config, args.stack, security) as test_cluster_endpoint:
-
+    with TestCluster.create('PERFORMANCE_TEST', manifest, config, args.stack, security) as test_cluster_endpoint:
```
owaiskazi19, line 42: https://github.com/opensearch-project/opensearch-build/pull/314#discussion_r701504074

dblock, line 42: This should be `PerfTestCluster.create`

bundle-workflow/src/test_workflow/test_cluster.py
```diff
@@ -5,20 +5,29 @@
...
 class TestCluster(abc.ABC):
     """
     Abstract base class for all types of test clusters.
     """
-
+    @staticmethod
     @abc.abstractmethod
-    def create(self):
+    @contextmanager
+    def create(cluster_type, *args):
```
dblock, line 19: TestCluster should not know about its child, `PerfTestCluster`.  inherits TestCluster, then this code becomes:

```pythin
def create(cls, *args)
  cluster = cls(*args)
```

bundle-workflow/src/test_workflow/perf_test_cluster.py
```diff
@@ -1,23 +1,9 @@
...
-        perf_test_cluster = PerfTestCluster(manifest, config, stack_name, security)
-        try:
-            perf_test_cluster.create()
-            yield perf_test_cluster.endpoint()
-        finally:
-            perf_test_cluster.destroy()
-
-
-class PerfTestCluster(TestCluster):
+class PerfTestCluster():
```
dblock, line 6: Isn't a perf test cluster a test cluster? Why is the inheritance removed?

bundle-workflow/src/test_workflow/test_cluster.py
owaiskazi19: When I am invoking `create` method like
`with PerfTestCluster.create(manifest, config, args.stack, security) as test_cluster_endpoint:`, it calls the `PerfTestCluster`'s `create` method.

owaiskazi19: When I am invoking `create` method like
`with PerfTestCluster.create(manifest, config, args.stack, security) as test_cluster_endpoint:`. It calls the `PerfTestCluster`'s `create` method. I am not sure whether I am following the right path. Feel free to correct me here. The `create` method of `TestCluster` looks like below but won't get invoked, may be calling with `super` might help:
```python
        cluster = cls(*args)
        try:
            cluster.create()
            yield cluster.endpoint()
        finally:
            cluster.destroy()`

owaiskazi19: When I am invoking `create` method like
`with PerfTestCluster.create(manifest, config, args.stack, security) as test_cluster_endpoint:`. It calls the `PerfTestCluster`'s `create` method. I am not sure whether I am following the right path. Feel free to correct me here. The `create` method of `TestCluster` looks like below but won't get invoked, may be calling with `super` might help:
```python
        cluster = cls(*args)
        try:
            cluster.create()
            yield cluster.endpoint()
        finally:
            cluster.destroy()`

dblock: It will because it's a class method called `create`, not the instance method with the same name.

bundle-workflow/src/test_workflow/perf_test_cluster.py
owaiskazi19: Removed it by mistake. Added back

bundle-workflow/src/test_workflow/test_cluster.py
owaiskazi19: I changed `PerfTestCluster`'s method to `create_cluster` since both methods of `TestCluster` and `PerfTestCluster` have the same name i.e. `create`. The class method gets overridden by the instance method, since it belongs to the `PerfTestCluster` class.

dblock (approved):

peternied (approved):

Issue #355, opened 2021-08-30
[BUG] Unit Test TestBuilder.test_export_artifacts fails on desktop


saratvemulapalli:
**Describe the bug**
Unit Test TestBuilder.test_export_artifacts fails on desktop, some how this test works on the CI workflow.

```
~> bundle-workflow % pipenv run pytest
..........................F..................................                                                                                                                                  [100%]
============================================================================================== FAILURES ==============================================================================================
_________________________________________________________________________________ TestBuilder.test_export_artifacts __________________________________________________________________________________

self = <tests_build_workflow.test_builder.TestBuilder testMethod=test_export_artifacts>, mock_walk = <MagicMock name='walk' id='4516530320'>

    @patch("os.walk")
    def test_export_artifacts(self, mock_walk):
        mock_walk.side_effect = self.mock_os_walk
        self.builder.export_artifacts()
>       self.assertEqual(self.builder.build_recorder.record_artifact.call_count, 2)
E       AssertionError: 0 != 2

tests/tests_build_workflow/test_builder.py:81: AssertionError
====================================================================================== short test summary info =======================================================================================
FAILED tests/tests_build_workflow/test_builder.py::TestBuilder::test_export_artifacts - AssertionError: 0 != 2
1 failed, 60 passed in 8.49s

```

**To Reproduce**
1. Clone the code
2. cd bundle-workflow
3. `pipenv install`
4. `pipenv run pytest`

**Expected behavior**
Test passes.

 - OS: Mac OS X Big Sur

**Additional context**

abhinavGupta16:
@bbarani - I can replicate the issue on MacOS BigSur
_____________________________________________________________________________________________________________ TestBuilder.test_export_artifacts ______________________________________________________________________________________________________________

self = <tests_build_workflow.test_builder.TestBuilder testMethod=test_export_artifacts>, mock_walk = <MagicMock name='walk' id='4401875120'>

    @patch("os.walk")
    def test_export_artifacts(self, mock_walk):
        mock_walk.side_effect = self.mock_os_walk
        self.builder.export_artifacts()
        self.assertEqual(self.builder.build_recorder.record_artifact.call_count, 2)
>       AssertionError: 0 != 2

AssertionError
===================== short test summary info ======================
FAILED tests/tests_build_workflow/test_builder.py::TestBuilder::test_export_artifacts - AssertionError: 0 != 2`

dblock:
[...]

PR #409, opened 2021-09-08
Fix: unit test failure on OSX.

dblock:

### Description
There was a hard-coded `/tmp` that is `/private/tmp` on OSX.
 
### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-build/issues/355.
 
### Check List

### DIFF
--- a/bundle-workflow/tests/tests_build_workflow/test_builder.py
+++ b/bundle-workflow/tests/tests_build_workflow/test_builder.py
@@ -60,10 +60,10 @@ class TestBuilder(unittest.TestCase):
             "component", self.builder.git_repo
         )
 
-    def mock_os_walk(self, artifact_type):
-        if artifact_type == "/tmp/checked-out-component/artifacts/core-plugins":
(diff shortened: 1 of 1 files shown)


peternied (approved):
Seems like the root cause is that this unit test is not mocking out its dependencies fully.  IMO we shouldn't see behavior changes on different platforms.  Approved to unblock, but I'd recommend adding more patches during setup if needed

gaiksaya (approved):

VachaShah (approved):

dblock:
> Seems like the root cause is that this unit test is not mocking out its dependencies fully. IMO we shouldn't see behavior changes on different platforms. Approved to unblock, but I'd recommend adding more patches of os.* components instead

I think it does the right thing. The code walks a directory and yields full paths. Then we have an implementation that looks at what's yielded and returns some fake files underneath as a side effect. The test assumed that what was yielded was `/tmp/...`, which is wrong on OSX.

Issue #393, opened 2021-09-07
Unify invoking the test workflows


saratvemulapalli:
The test workflows need restructuring for all the test suites. Today each test suite is using its own model of invoking the tests via test.sh or python <test_suite>.py.

Follow up of PR: https://github.com/opensearch-project/opensearch-build/pull/377

saratvemulapalli:
This takes care restructuring Integ/BWC workflows to follow the same pattern. 
Will create a follow up issue for perf tests as it needs few more params.

saratvemulapalli:
Created an issue: https://github.com/opensearch-project/opensearch-build/issues/506

PR #423, opened 2021-09-09
Adding support to run integration tests in Jenkins

### LINKED ISSUE #263: Make all test jobs parameterized
As part of this task, we need to do -
 * Setting up right environment on jenkins for running the tests
 * Pull Job shell scripts from SCM
 * Wrap respective test.py(s) to invoke as jenkins job

saratvemulapalli:
### Description
Adding support to run integration tests in Jenkins.

Parameters Integration Tests need:
1. Artifacts S3 Bucket - To pull build/bundle manifests, maven dependencies.
2. OpenSearch version - To pull build/bundle manifest.
3. Build ID - To map Build ID to a test run (Test Run ID) and to pull the right manifests.
4. CPU Architecture - To identify which agent node to run on.
5. Test Run ID - To publish results for a unique job ID

### Issues Resolved
Closes #263 
 
### Check List


bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+    }
+    stages {
+        stage('Parameters') {
+            steps {
+                script {
+                    properties([
+                            parameters([
+                                    string(
+                                            defaultValue: '',
+                                            name: 'bundle_manifest',
```
peternied, line 18: Why are both manifest needed?

dblock (changes requested):
1. It might not be possible, but a bundle defines the architecture for which it's built. This job lets me try and run a bundle for x86 on arm, which is not desirable. Can this be loaded from the bundle?
2. I'm worried that we're introducing `test_integration.sh`, which we have agreed to really be `test.sh --suite integration` or something like that. Can we do this, even if it's super hacky inside `test.sh`, in this PR so we don't have to go change it all over the place?

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+        }
+        stage('Choose Agent Node') {
+            steps {
+                script {
+                    if ("${architecture}" == 'x64') {
+                        agentLabel = "Jenkins-Agent-al2-x64-m5xlarge"
+                    } else if ("${architecture}" == 'arm64') {
+                        agentLabel = "Jenkins-Agent-al2-arm64-m6gxlarge"
+                    } else {
+                        agentLabel = ""
```
peternied, line 44: Can we have this fail if there is no match?

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+                node {
+                    label "${agentLabel}"
+                }
+            }
+            steps {
+                withAWS(role: 'opensearch-test', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+                    s3Download(file: 'tests/bundle_manifest.yml', bucket: "${ARTIFACT_BUCKET_NAME}", path: "$bundle_manifest", force:true)
+                    s3Download(file: 'tests/build_manifest.yml', bucket: "${ARTIFACT_BUCKET_NAME}", path: "$build_manifest", force:true)
+                }
+                sh "./bundle-workflow/test_integration.sh --bundle-manifest tests/bundle_manifest.yml --build-manifest tests/build_manifest.yml --test-manifest bundle-workflow/src/test_workflow/config/test_manifest.yml"
```
peternied, line 60: Could we reshape this workflow step to be more granular and avoid having to know about s3, here is what I was thinking it would looking like:
```
steps { 
   withAWS(role: "opensearch-integration-test") {
      s3Download(file: "tests/bundle_manifest.yml", bucket: "${ARTIFACT_BUCKET_NAME}", path: "$bundle_manifest", force: true)
      sh "./bundle-workflow/download_bundle --manifest $bundle_manifest --destination /bundle_under_test"
   }
   sh "./bundle-workflow/test_integration.sh --manifest $bundle_manifest --bundle_under_test /bundle_under_test" --test-artifacts /test_artifacts-$BUNDLE_TEST_ID
}
post() {
   always {
      withAWS(role: 'opensearch-integration-test', ...) {
         sh "./bundle-workflow/collect_additional_artifacts.sh --test-artifacts /test_artifacts-${BUNDLE_TEST_ID}"
         s3Upload(file: '/test_artifacts-${architecture}-${BUNDLE_TEST_ID}', bucket: '${test_results_bucket}')
      }
   }
}
```

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+            agent {
+                node {
+                    label "${agentLabel}"
+                }
+            }
+            steps {
+                withAWS(role: 'opensearch-test', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+                    s3Download(file: 'tests/bundle_manifest.yml', bucket: "${ARTIFACT_BUCKET_NAME}", path: "$bundle_manifest", force:true)
+                    s3Download(file: 'tests/build_manifest.yml', bucket: "${ARTIFACT_BUCKET_NAME}", path: "$build_manifest", force:true)
+                }
```
peternied, line 59: Why do we have a `test_manifest.yml` do we need this managed as config rather than as code?

peternied (changes requested):
I was expecting to see all the legs of the integration tests executed, `integration`, `bwc`, `perf` is that coming in a future PR?

saratvemulapalli:
> I was expecting to see all the legs of the integration tests executed, `integration`, `bwc`, `perf` is that coming in a future PR?

Sure. This is a middle piece i.e bundle workflow -> test orchestrator -> individual test job. 
This helps in keeping the jobs independent while the orchestrator combines all of them into one place and executes integ tests, bwc tests and perf tests.
This PR lays down the framework of how an individual test job can run.

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
saratvemulapalli: As covered in the description, integration tests needs to understand what maven artifacts it needs to pull and where to pull them from for a given component. 
That said, we agreed upon having a single manifest to feed to test system which will be the source of truth.
Opened up an issue to take care of it: https://github.com/opensearch-project/opensearch-build/issues/433

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
saratvemulapalli: As we discussed, eventually all the tests will consume a place to store results and pull dependencies. 
Unfortunately we have to live with it that these test systems need to understand for now.

Eventually we should be able to publish test results from the jenkins stage avoiding the test suites to manage it for them  and be agnostic of where they are being stored.

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
saratvemulapalli: Few tests need additional configuration explaining what kind of tests they would like to run for a given component. 
This manifest helps understand that.

saratvemulapalli:
> 1. It might not be possible, but a bundle defines the architecture for which it's built. This job lets me try and run a bundle for x86 on arm, which is not desirable. Can this be loaded from the bundle?
> 2. I'm worried that we're introducing `test_integration.sh`, which we have agreed to really be `test.sh --suite integration` or something like that. Can we do this, even if it's super hacky inside `test.sh`, in this PR so we don't have to go change it all over the place?

1. It should be able to. Just didnt want to complicate the logic in Jenkins. But I agree and see the value that we are giving room to have mis-matched states while the truth already exists. I'll take care of this.
2. Absolutely, we went back and forth on this. I wanted to clear this up as part of #393 while not holding this one up. I'll put something really simple for this PR and we could double down on the structure with #393.

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
peternied: > Unfortunately we have to live with it that these test systems need to understand for now.

We can reduce a large amount of complexity in the tests if we do this.  If we are seeing trouble with the existing system please consider this an option to migrate to rather than additional investment

setiah: @saratvemulapalli we don't need to download manifest files from s3 here. We are doing that as part of the individual test jobs. The jenkins job should supply - s3_bucket_name, opensearch_version, build_id, architecture and test_id to the underlying tool for the test run

setiah, line 60: test_manifest will be part of the opensearch-build repo, so not required as input parameter

saratvemulapalli: Sure, the script today has it as a required parameter. 
We could always change this.

saratvemulapalli: Absolutely, that was the discussion we had with @peternied. Agree that this should go to orchestrator and the jobs will just expect a path to these manifests.

peternied:
@saratvemulapalli During our discussion yesterday reading the manifest for architecture and reducing number of input parameters were mentioned - I'm waiting on those change

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
setiah: No, the orchestrator shouldn't be pulling these manifest files.

Since each of these jobs runs (can run) on a different host than orchestrator, the orchestrator (pipeline) shouldn't be pulling manifest files. Individual jobs should be pulling them when they run based in the input parameters - s3_bucket, opensearch_version, build_id, architecture. This logic is already added to Integ tests in #456

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
setiah: The test manifest will be present in opensearch-build repo and doesn't need to be provided as input to the job. See #456

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
saratvemulapalli: Makes sense. So we are refactoring pulling the manifests into the Python scripts and keeping the jenkins job independent of it. 
 So I would have to wait until #456 is merged. 
 
 This would mean:
1. The Jenkins job will simply read through all the parameters which integtest needs to run and these parameters could be fed from test orchestrator vs :
    a. s3 bucket
    b. opensearch version
    c. build_id
    d. test_id
    e. architecture
2. Jenkins job will not pull the manifest files as the python test suites are pulling the manifests from S3. The consequence of this is that the Jenkins job will need to pull an architecture parameter which should be passed and will not read from the manifest.

setiah: That's correct. Instead of orchestrator, we want individual test jobs to pull the manifest and other artifacts from S3 because the jobs can run on any host irrespective of where orchestrator logic runs. This keeps it decoupled.
 
For testing purpose, you could use feature branch of 456 to unblock this.

bundle-workflow/README.md
```diff
@@ -102,10 +102,8 @@ Tests the OpenSearch bundle.
 
 This workflow contains integration, backwards compatibility and performance tests.
 
-The following example kicks off all test suites for a distribution of OpenSearch 1.1.0.
-
 ```bash
-./bundle-workflow/test.sh manifests/opensearch-1.1.0.yml
+./bundle-workflow/test.sh integ-test
```
dblock, line 106: Which manifest?

dblock (approved):

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+        stage('Run Integration Tests') {
+            agent {
+                node {
+                    label "${agentLabel}"
+                }
+            }
+            steps {
+                sh "./bundle-workflow/test.sh --s3-bucket ${s3_bucket} --opensearch-version ${opensearch_version} --build-id ${build_id} --architecture ${architecture} --test-run-id ${test_run_id}"
```
dblock, line 63: Shouldn't this have "integ-test" in the arguments?

dblock (changes requested):

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
saratvemulapalli: It should :), thanks for that. I have updated the runner later.

bundle-workflow/README.md
saratvemulapalli: Not sure if the question was what manifest do we use for integ tests. 
I would like to have #456 update this readme to add the rest of the new parameters to this readme.

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+    stages {
+        stage('Parameters') {
+            steps {
+                script {
+                    properties([
+                            parameters([
+                                    string(
+                                            defaultValue: '',
+                                            name: 's3_bucket',
+                                            trim: true
```
setiah, line 16: This can be read from environment right? Does it need to be a parameter?

bundle-workflow/test.sh
```diff
@@ -9,4 +9,11 @@
 set -e
 
 DIR="$(dirname "$0")"
-"$DIR/run.sh" "$DIR/src/test.py" $@
+case $1 in
```
setiah, line 12: This change is also present in #456 now as requested. Can we remove it from here?

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
saratvemulapalli: Yeah, the idea is test orchestrator tells which bucket to run with. And jobs are agnostic of the ENV even though they have access to it. This helps when a user wants to submit a job and use a different S3 bucket.

bundle-workflow/test.sh
saratvemulapalli: Sure that works :)

bundle-workflow/jenkins_workflow/integtest/Jenkinsfile
setiah: Changing bucket may not work because the jenkins role won't have permissions to read from a different s3 bucket. The role is also configured in jenkins env variables.

saratvemulapalli: Thinking about it. This actually makes sense. May be lets not optimize for other use cases let one flow work concretely. I'll use the ENV variable.

setiah (dismissed):
LGTM. Thanks @saratvemulapalli

bundle-workflow/test.sh
```diff
@@ -9,4 +9,4 @@
 set -e
 
 DIR="$(dirname "$0")"
-"$DIR/run.sh" "$DIR/src/test.py" $@
+"$DIR/run.sh" "$DIR/src/test.py" $@
```
dblock, line 12: Did you delete src/test.py?

bundle-workflow/jenkins_workflow/integtest/README.md
```diff
@@ -0,0 +1,15 @@
+- [Integration Tests Jenkins Job](#integration-tests-jenkins-job)
+  - [Job Parameters](#job-parameters)
+### Integration Tests Jenkins Job
+
+This job runs integration tests for a bundle via `test_integration.sh` in Jenkins.
```
dblock, line 5: I don't think this is correct anymore.

bundle-workflow/test.sh
saratvemulapalli: Yes, @setiah will add the test.sh in #456 as part of his PR. 
Ref: https://github.com/opensearch-project/opensearch-build/pull/423#discussion_r708765944

bundle-workflow/jenkins_workflow/integtest/README.md
saratvemulapalli: You are right. Let me fix it.

dblock (approved):

setiah (approved):

Issue #158, opened 2021-08-06
Create OpenSearch Dashboards 1.1 Distribution


dblock:
**Is your feature request related to a problem? Please describe.**

Produce an OpenSearch Dashboards bundle.

**Describe the solution you'd like**

Extend https://github.com/opensearch-project/opensearch-build/tree/main/scripts/bundle-build to support OpenSearch Dashboards. Will need a default script for each component and overrides. Might need to move some things around in scripts into an OpenSearch subfolder to accommodate.

bbarani:
The current progress of Opensearch Dashboards automation can be tracked here - https://github.com/opensearch-project/opensearch-build/issues/158

dblock:
The leftover work is to add components to dashboards manifest and hook up the build job to Jenkins.

peternied:
> The leftover work is to add components to dashboards manifest and hook up the build job to Jenkins.

@kavilla Should we spin those tasks into additional issues and update https://github.com/opensearch-project/opensearch-build/issues/573 or wait for them on this issue?

dblock:
We released 1.2 OpenSearch Dashboards using automated build, closing.

PR #435, opened 2021-09-09
[Build] bundle OpenSearch Dashboards

kavilla:
### Description
Produce an OpenSearch Dashboards bundle with plugins provided by a manifest.

This requires the manifest to be in order with plugins first and then Dashboards
being last. The plugins will need to pulled down and placed in the plugins
folder. Then we need to bootstrap to generate the modules so that we can actually
install the plugins to the release build.

What is missing?
* Environment setup, this requires nvm and yarn to be already set up.
We will need to set that up.
* Signing of the artifact
* Diff between release and snapshot
* Default script for plugins, this will need more information on how we want this to
look like.
* All the other plugins included into the manifest.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/158
 
### Check List


bundle-workflow/scripts/components/OpenSearch-Dashboards/build.sh
```diff
@@ -0,0 +1,109 @@
...
+        echo "Unsupported architecture: ${ARCHITECTURE}"
+        exit 1
+        ;;
+esac
+
+echo "Building node modules for core"
+yarn osd bootstrap
+
+echo "Building artifact"
+yarn build --skip-os-packages --release
```
kavilla, line 78: We are working on actually working on de-coupling things here: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/473

For now this will build all the non-OS Packages artifacts and then it will just grab the one it needs.

bundle-workflow/scripts/components/alertingDashboards/build.sh
```diff
@@ -0,0 +1,62 @@
...
+if [ -z "$VERSION" ]; then
+    echo "Error: You must specify the OpenSearch Dashboards version"
+    usage
+    exit 1
+fi
+
+[[ "$SNAPSHOT" == "true" ]] && VERSION=$VERSION-SNAPSHOT
+[ -z "$OUTPUT" ] && OUTPUT=artifacts
+
+PLUGIN_NAME=$(basename "$PWD")
```
kavilla, line 58: This will most likely be what the default build script would look like (besides the repos that are hybrid repos). I was thinking that we can add a new argument that tells the default build script that we are building for Dashboards and just use this section.

bundle-workflow/scripts/components/OpenSearch-Dashboards/build.sh
```diff
@@ -0,0 +1,109 @@
...
+        QUALIFIER="linux-arm64"
+        ;;
+    *)
+        echo "Unsupported architecture: ${ARCHITECTURE}"
+        exit 1
+        ;;
+esac
+
+echo "Building node modules for core"
+yarn osd bootstrap
```
kavilla, line 75: In it's current state there is a huge assumption that you have the proper node version and yarn installed in the environment that this is being ran in. idea was that we would get the bundling working with this assumption and have an improvement to have a setup script (https://github.com/opensearch-project/OpenSearch-Dashboards/projects/4#card-67798579)

bundle-workflow/scripts/components/OpenSearch-Dashboards/build.sh
```diff
@@ -0,0 +1,109 @@
...
+for plugin in ./plugins/*; do
+    PLUGIN_NAME="$(basename "${plugin}")"
+    cd plugins/$PLUGIN_NAME
+    yarn plugin-helpers build
+    cd ../..
+    ./"${OUTPUT}"/bundle/opensearch-dashboards/bin/opensearch-dashboards-plugin --allow-root install file:./plugins/$PLUGIN_NAME/build/$PLUGIN_NAME-$VERSION.zip
+done
+
+
```
kavilla, line 105: TODO: recreate the bundled tarball.

manifests/opensearch-dashboards-1.1.0.yml
```diff
@@ -0,0 +1,43 @@
...
+  # - name: notebooksDashboards
+  #   repository: https://github.com/opensearch-project/dashboards-notebooks.git
+  #   ref: main
+  # - name: traceAnalyticsDashboards
+  #   repository: https://github.com/opensearch-project/trace-analytics.git
+  #   ref: main
+  # - name: ganttChartDashboards
+  #   repository: ????????
+  #   ref: main
+  - name: OpenSearch-Dashboards
```
kavilla, line 41: This needs to be last.

bundle-workflow/scripts/components/OpenSearch-Dashboards/build.sh
dblock: That's totally fine.

bundle-workflow/scripts/components/OpenSearch-Dashboards/build.sh
```diff
@@ -0,0 +1,109 @@
...
+mkdir -p "${OUTPUT}/bundle"
+
+# Release gets built to /target, regular non-tarball gets built to /build
+cp target/$ARTIFACT_BUILD_NAME $OUTPUT/bundle/$ARTIFACT_TARGET_NAME
+mkdir -p "${OUTPUT}/bundle/opensearch-dashboards"
+
+# Untar for the plugins
+tar -xzf "${OUTPUT}"/bundle/$ARTIFACT_TARGET_NAME --strip-components=1 --directory "${OUTPUT}"/bundle/opensearch-dashboards
+
+echo "Building node modules for plugins"
```
dblock, line 94: The OpenSearch build process has an "assemble" step that installs plugins. Move this out, add a default `install.sh` for each plugin.

bundle-workflow/scripts/components/alertingDashboards/build.sh
dblock: Move it to a default script location (will need minor refactor for OpenSearch default) earlier than later.

dblock (commented):
This is good. Let's turn this into something we can start committing?

bundle-workflow/scripts/components/OpenSearch-Dashboards/install.sh
```diff
@@ -0,0 +1,49 @@
...
+        ?)
+            echo "Invalid option: -${arg}"
+            exit 1
+            ;;
+    esac
+done
+
+mkdir -p "${OUTPUT}/bundle/opensearch-dashboards"
+
+# Untar for the plugins
```
dblock, line 38: You should not be installing plugins here, but let the assemble step take care of it one plugin at a time.

peterzhuamazon (changes requested):
Hi @kavilla thanks for the initial scripts.

Please also check our old script for reference, you missed some steps such as adding the tweaked dashboards yml file, etc.

https://github.com/opensearch-project/opensearch-build/blob/main/release/tar/linux/opensearch-dashboards-tar-build.sh#L142-L143

bundle-workflow/scripts/components/OpenSearch-Dashboards/build.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# Copy artifact to bundle output with -min in the name
+[[ "$SNAPSHOT" == "true" ]] && IDENTIFIER="-SNAPSHOT"
+#ARTIFACT_BUILD_NAME=opensearch-dashboards-$VERSION$IDENTIFIER-$QUALIFIER.tar.gz
+ARTIFACT_BUILD_NAME=opensearch-dashboards-$VERSION-$QUALIFIER.tar.gz
+ARTIFACT_TARGET_NAME=opensearch-dashboards-min-$VERSION$IDENTIFIER-$QUALIFIER.tar.gz
+mkdir -p "${OUTPUT}/bundle"
+mkdir -p "${OUTPUT}/plugins"
+
+# Release gets built to /target, regular non-tarball gets built to /build
```
dblock, line 88: This needs to be moved out into the assemble step.

manifests/opensearch-dashboards-1.1.0.yml
dblock: It shouldn't be last if each component only builds as part of build, and is installed as part of assemble.

bundle-workflow/scripts/default/build.sh
```diff
@@ -13,10 +13,11 @@ function usage() {
     echo -e "-s SNAPSHOT\t[Optional] Build a snapshot, default is 'false'."
```
dblock, line 13: Move `scripts/default/build.sh` into `scripts/OpenSearch/default` and `scripts/OpenSearch-Dashboards/default` and call the right one, same with install.

DEVELOPER_GUIDE.md
```diff
@@ -48,6 +48,21 @@ $ pipenv --version
 pipenv, version 19.0
 ```
 
+#### Install NVM and Node
```
dblock, line 51: Add TOC.

dblock (commented):
Does this work? If so can we get this to a merge-able PR please?

There's a bunch of refactoring to do avoid many of the `if`'s, I can take it after it's merged.

PR #443, opened 2021-09-10
Refactor targets to avoid passing name, version, arch and snapshots around separately

dblock:
### Description

Builds and CI runs have a "target", which is typically a version and whether to build or not a snapshot. Build targets also have name and architecture. Refactor into `BuildTarget` and `CiTarget` classes to make things simpler.
 
### Check List


bundle-workflow/src/ci_workflow/check.py
```diff
@@ -8,17 +8,10 @@
...
+    def __init__(self, component, git_repo, target):
         self.component = component
         self.git_repo = git_repo
-        self.version = version
-        self.arch = arch
-        self.snapshot = snapshot
-        self.opensearch_version = version + "-SNAPSHOT" if snapshot else version
-        self.component_version = version + ".0-SNAPSHOT" if snapshot else f"{version}.0"
-        if self.component.name == "OpenSearch":
-            # HACK: OpenSearch version is 3-digits
```
VachaShah, line 20: Would this part be needed in the `BuildTarget` implementation?

dblock: It was ported into the check itself where we treat OpenSearch itself as a component. The build target defines the version that a component _should_ be.

bundle-workflow/tests/tests_ci_workflow/test_check.py
```diff
@@ -5,62 +5,7 @@
...
-            ),
-            git_repo=MagicMock(),
-            version="1.1.0",
-            arch="x86",
-            snapshot=True,
-        )
-
-    def test_component_version(self):
-        self.assertEqual(self.check.component_version, "1.1.0-SNAPSHOT")
+    pass
```
VachaShah, line 11: Is this removed since there it directly takes in the target now?

dblock: Right. This class no longer has any properties, just serves as a common constructor that stores `target`. I didn't want to remove the test class though in case we add something useful to it in the future that's testable.

bundle-workflow/src/ci_workflow/check.py
dblock: Here: https://github.com/opensearch-project/opensearch-build/pull/443/files/4aa773ceaeea391b9451ec182c128f9a8740e5aa#diff-a8e1cc3a88bb0cf63acc8cec63acc8c5d1b232896f05f3ad02c12a4783613d31R56

mch2 (approved):

bundle-workflow/src/ci_workflow/ci_target.py
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+
+class CiTarget:
+    version: str
```
owaiskazi19, line 9: Just confirming, won't we need `arch` as a part of `CiTarget`?

dblock: Nope, it doesn't use arch anywhere.

owaiskazi19 (approved):

bundle-workflow/src/ci_workflow/check.py
VachaShah: Got it! Thanks!

PR #456, opened 2021-09-10
Integtest input arguments changes and integration with s3

### LINKED ISSUE #457: Integration tests to pull artifacts from S3Bucket

setiah:
### Description
This PR has 2 main changes -
1. Make the input arguments streamlined with what Jenkins pipeline(and the build workflow) would provide to the integ test job.
2. Read artifacts and manifests for a given build from S3 bucket.
 
### Issues Resolved
#457 
 
### Check List


bundle-workflow/src/test_workflow/utils/build_manifest_provider.py
```diff
@@ -0,0 +1,27 @@
...
+
+import os
+
+from aws.s3_bucket import S3Bucket
+from manifests.build_manifest import BuildManifest
+
+
+class BuildManifestProvider:
+    @staticmethod
+    def get_build_manifest_relative_location(build_id, opensearch_version, architecture):
```
setiah, line 18: i'm open to moving these methods to build_manifest.py if that makes more sense

bundle-workflow/src/test_workflow/utils/bundle_manifest_provider.py
```diff
@@ -0,0 +1,35 @@
...
+# Modifications Copyright OpenSearch Contributors. See
+# GitHub history for details.
+
+import os
+
+from aws.s3_bucket import S3Bucket
+from manifests.bundle_manifest import BundleManifest
+
+
+class BundleManifestProvider:
```
setiah, line 16: open to moving these methods to bundle_manifest.py if that makes more sense

bundle-workflow/src/run_integ_test.py
```diff
@@ -101,16 +104,21 @@ def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
 def main():
     args = parse_arguments()
     console.configure(level=args.logging_level)
-    bundle_manifest = BundleManifest.from_file(args.bundle_manifest)
-    build_manifest = BuildManifest.from_file(args.build_manifest)
-    test_manifest = TestManifest.from_file(args.test_manifest)
+    script_dir = str(os.getcwd())
```
dblock, line 107: This path changes depending on where you run from, use `__file__` as elsewhere

bundle-workflow/src/run_integ_test.py
```diff
@@ -101,16 +104,21 @@ def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
...
+    test_manifest_path = ("%s%s" % (script_dir, '/src/test_workflow/config/test_manifest.yml'))
+    with open(test_manifest_path, 'r') as file:
+        test_manifest = TestManifest.from_file(file)
```
dblock, line 110: Use or add `from_path` and no need to open the file here

bundle-workflow/src/test_workflow/utils/build_manifest_provider.py
dblock: Yes it does. This is not Java :)

bundle-workflow/src/test_workflow/utils/build_manifest_provider.py
```diff
@@ -0,0 +1,27 @@
...
+        return f"builds/{opensearch_version}/{build_id}/{architecture}/manifest.yml"
+
+    @staticmethod
+    def load_manifest(bucket_name, build_id, opensearch_version, architecture):
```
dblock, line 22: This can become `BuildManifest.from_s3`

bundle-workflow/src/test_workflow/utils/build_manifest_provider.py
```diff
@@ -0,0 +1,27 @@
...
+        local_path = str(os.getcwd())
```
dblock, line 23: This should be a temporary file name or if streaming is supported no file at all.

bundle-workflow/src/test_workflow/utils/bundle_manifest_provider.py
dblock: Yes.

bundle-workflow/README.md
```diff
@@ -120,6 +120,15 @@ The following options are available.
 
 This step runs integration tests invoking `integtest.sh` in each component from bundle manifest.
 
+To run integration tests locally, use below command. It pulls down the built bundle and its manifest file from S3,
+reads all components of the bundle and runs integration tests against each component.
+ 
+```
+cd opensearch-build/bundle-workflow
+pipenv install
```
dblock, line 128: It's worth making this into test.sh earlier than later even if it's a little hacky, because this is effectively the calling interface.

It should be `test.sh --suite ... --S3..` and would make it callable from everywhere and no need to remember to use pipenv.

dblock (changes requested):
The reliance on current directory and the leftover temp file should def be fixed before merging.

bundle-workflow/src/test_workflow/utils/build_manifest_provider.py
saratvemulapalli: +1 I think it makes it more cleaner to put these utility functions with BundleManifest and BuildManifest.

bundle-workflow/README.md
saratvemulapalli: Same comment as dB. The runner should install all the python dependencies. I did see problems when the Jenkins job was being setup.

bundle-workflow/src/run_integ_test.py
```diff
@@ -29,13 +29,16 @@
 def parse_arguments():
```
saratvemulapalli, line 29: I do see these args as common for all test suites we should unify them eventually as TestArgs. 
Doesn't have to be part of this PR.

bundle-workflow/src/run_integ_test.py
```diff
@@ -29,13 +29,16 @@
...
     parser = argparse.ArgumentParser(description="Test an OpenSearch Bundle")
     parser.add_argument(
-        "--bundle-manifest", type=argparse.FileType("r"), help="Bundle Manifest file."
```
saratvemulapalli, line 32: We need another parameter for test run id.

bundle-workflow/src/run_integ_test.py
```diff
@@ -29,13 +29,16 @@
...
+        "--s3-bucket", type=str, help="S3 bucket name"
```
saratvemulapalli, line 32: As we are taking S3 bucket name, I think we should also let users configure the role to pull these artifacts. And the default could be `opensearch-test-role`.

bundle-workflow/src/run_integ_test.py
gaiksaya: I added that in my PR. See #467

bundle-workflow/README.md
setiah: Agree. It will be taken as a separate PR after this.

bundle-workflow/src/run_integ_test.py
setiah: That's the plan. I avoided refactoring at this point as Bwc uses TestArgs with different arguments. After this is merged, and once bwc is refactored, this logic can be moved to TestArgs

bundle-workflow/src/run_integ_test.py
setiah: that's configurable via environment variables.

bundle-workflow/src/test_workflow/utils/build_manifest_provider.py
setiah: the `S3Bucket.download_file` method requires a local destination as input.

bundle-workflow/src/test_workflow/utils/build_manifest_provider.py
setiah: sure, thx for the inputs

bundle-workflow/README.md
dblock: Please consider doing it as part of this PR? It's a tiny amount of work. This is just not reliable.

bundle-workflow/src/manifests/build_manifest.py
```diff
@@ -54,6 +57,18 @@ def to_dict(self):
             ),
         }
 
+    @staticmethod
+    def get_build_manifest_relative_location(build_id, opensearch_version, architecture):
+        return f"builds/{opensearch_version}/{build_id}/{architecture}/manifest.yml"
+
+    @staticmethod
+    def from_s3(bucket_name, build_id, opensearch_version, architecture):
+        local_path = str(os.getcwd())
```
dblock, line 66: This is going to leave files all around (you can't rely on the caller to be in a temporary directory). Download into a temporary file/path and automatically get rid of the file after loading it.

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -85,7 +86,7 @@ def _execute_integtest_sh(self, endpoint, port, security):
             self.component.name, self.repo.dir
         )
         if os.path.exists(script):
-            cmd = f"sh {script} -b {endpoint} -p {port} -s {str(security).lower()}"
+            cmd = f"bash {script} -b {endpoint} -p {port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
```
dblock, line 89: You should be executing `script` with the correct shebang, that needs a `chmod 755`.

dblock (changes requested):
A couple that I think are must have's.

bundle-workflow/src/manifests/build_manifest.py
setiah: the entire test logic executes in a tmp directory, which is automatically cleaned up after the test success/failure. These files as well get downloaded in the same tmp directory and get automatically cleaned up even for local runs. So it shouldn't leave files all over the place. However, I do see value in adding a cleanup inside this method, to make it complete in itself.

bundle-workflow/src/run_integ_test.py
```diff
@@ -29,13 +29,19 @@
...
     )
     parser.add_argument(
-        "--build-manifest", type=argparse.FileType("r"), help="Build Manifest file."
+        "--opensearch-version", type=str, help="OpenSearch version to test"
```
saratvemulapalli, line 35: It would be nice to have the suite configurable to just run integration tests in a single component. May be take `--component` as a parameter.

bundle-workflow/src/manifests/build_manifest.py
dblock: You're making an assumption about something happening externally, which can change without this code ever noticing.

bundle-workflow/README.md
saratvemulapalli: I have these basic changes in my PR. Feel free to use them or I could push it as part of my PR.
https://github.com/opensearch-project/opensearch-build/pull/423/files#diff-df03dd40a41319c80f83f3d4c9cde02ba41aed2d9d102e04546e9b8c1c5dc5ae

setiah: We have to revisit this part when make the input arguments uniform across all test jobs. Currently, BwC accepts different input parameters and so does the perf test. The plan was to do this unification and shell-ify as part of that single task, as it is all related. https://github.com/opensearch-project/opensearch-build/pull/456#discussion_r707860346
This doesn't seem to be a blocker for the purpose of this PR. A separate issue to track this #472.

dblock: I disagree with punting shellifying to the future. It's a trivial change that ensures that runs are predictably reliable. No need to acquire more debt.

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
setiah: Sure, the scripts in plugin repo are already w/ 755 permissions. I have updated the default integtest.sh w/ 755 as well.

bundle-workflow/src/run_integ_test.py
setiah: +1 to that. Will take this up as an enhancement separately

bundle-workflow/README.md
setiah: Added this

bundle-workflow/README.md
```diff
@@ -120,6 +120,20 @@ The following options are available.
...
+export AWS_ROLE_ARN=arn:aws:iam::<AWS_JENKINS_ACCOUNT>:role/opensearch-test
+export AWS_ROLE_SESSION_NAME=dummy-session
+
+Next, configure temporary credentials in environment w/
+export AWS_SESSION_TOKEN=<value>
+export AWS_ACCESS_KEY_ID=<value>
+export AWS_SECRET_ACCESS_KEY=<value>
+
+cd bundle-workflow
+bash test.sh integ-test --test-run-id <execution-id> --s3-bucket <bucket_name> --opensearch-version <version> --build-id <id> --architecture <arch>
```
dblock, line 135: This should be ./test.sh ...

Make sure there's a shebang in test.sh and the correct chmod 755 permissions on the .sh and .py files.

bundle-workflow/scripts/default/integtest.sh
```diff
@@ -1,2 +1,89 @@
...
+            SECURITY_ENABLED=$OPTARG
+            ;;
+        c)
+            CREDENTIAL=$OPTARG
+            ;;
+        v)
+            OPENSEARCH_VERSION=$OPTARG
+            ;;
+        n)
+            IS_SNAPSHOT=$OPTARG
```
dblock, line 47: We use SNAPSHOT in other scripts so let's rename this.

bundle-workflow/src/run_integ_test.py
```diff
@@ -101,16 +107,20 @@ def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
 def main():
     args = parse_arguments()
     console.configure(level=args.logging_level)
-    bundle_manifest = BundleManifest.from_file(args.bundle_manifest)
-    build_manifest = BuildManifest.from_file(args.build_manifest)
-    test_manifest = TestManifest.from_file(args.test_manifest)
+    script_dir = os.path.realpath(os.path.dirname(__file__))
+    test_manifest_path = ("%s%s" % (script_dir, '/test_workflow/config/test_manifest.yml'))
```
dblock, line 111: Should it be is.path.join?

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
dblock: This line shouldn't start with bash. The shebang inside the .sh file defines what shell to use.

dblock (dismissed):
Mostly should have's but nothing blocking.

bundle-workflow/src/manifests/bundle_manifest.py
```diff
@@ -45,6 +48,28 @@ def to_dict(self):
...
+        os.remove(os.path.realpath(os.path.join(work_dir, 'manifest.yml')))
+        return bundle_manifest
+
+    @staticmethod
+    def get_tarball_relative_location(build_id, opensearch_version, architecture):
+        return f"bundles/{opensearch_version}/{build_id}/{architecture}/opensearch-{opensearch_version}-linux-{architecture}.tar.gz"
+
+    @staticmethod
+    def get_tarball_name(opensearch_version, architecture):
+        return f"opensearch-{opensearch_version}-linux-{architecture}.tar.gz"
```
saratvemulapalli, line 67: Do we need to construct the tarball name?
The bundle manifest already has the location for a given build, its something like:
```
build:
  architecture: x64
  id: '32'
  location: https://artifacts.opensearch.org/bundles/1.1.0/32/opensearch-1.1.0-linux-x64.tar.gz
  name: OpenSearch
  version: 1.1.0
```

setiah: Yeah, either way we will have to construct it from the location or the parameters.

saratvemulapalli: Yeah instead reading from the manifest avoids having this logic baked into our code incase the format changes or new platforms are supported etc. One less problem to worry about :)

bundle-workflow/src/manifests/build_manifest.py
setiah: >However, I do see value in adding a cleanup inside this method, to make it complete in itself.

Agree. i fixed it w/ cleanup logic in latest commit.

bundle-workflow/src/manifests/bundle_manifest.py
setiah: Hey @saratvemulapalli . right now the entire codebase including build workflow is hardcoded to "linux" as platform. I think these things will gradually evolve for better, down the line, when we have more platforms getting added.

saratvemulapalli (approved):
Thanks @setiah for the change.
Changes look good to me!

owaiskazi19 (approved):

Issue #388, opened 2021-09-03
Automatically generate manifests for future OpenSearch versions


dblock:
**Is your feature request related to a problem? Please describe.**

Creation of next version's manifest is manual.

**Proposed Solution**

Add a Github Actions workflow that automatically generates and PRs a manifest for the next version.

1. Fetch the version of each repo main, including opensearch-project/OpenSearch#main. 
2. If the manifest for that version, e.g. 2.0, does not exist, create it, empty of components.
3. [Auto-submit a PR](https://code.dblock.org/2021/08/13/automating-code-changes-with-github-actions-making-pull-requests.html).  
4. For every new manifest, create a GitHub issue to release that version. 
5. Auto create child issues in every other repo to increment the version and join the release manifest. That manifest doesn’t include all components because they would have the wrong version.
6. Components will be slow to increment versions. Until they do we can't have a working "next version" manifest with all components, so we'll have to add those components one-by-one. It should also be possible to add components, for each component in 1.1 check whether component#main has been incremented to 2.0, if so add to 2.0 manifest. However components can be removed manually in versions and we don't want the GHA to keep re-adding them. Can be solved with a hint of hint on which components need to be watched, but maybe overkill.

dblock:
Once a manifest is created, a "release version XYZ" issue can also be created, and child issues to create a branch for every dependent component from the previous release to PR into the manifest

PR #461, opened 2021-09-13
Auto-generate next version manifest(s)

dblock:

### Description

Automatically generate next version's manifest.

1. Locate the latest manifest.
2. Attempt to get gradle properties from each component.
3. Generate a manifest for any new version.
4. Make a pull request, e.g. https://github.com/dblock/opensearch-build/pull/2

Now generates OpenSearch 1.0.1, 1.2.0 and 2.0.0 manifests.

I'm still writing more tests, but otherwise this is ready.
 
### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/388.
 
### Check List


.github/workflows/versions.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+      - name: Install Pipenv and Dependencies
+        run: |
+          python -m pip install --upgrade pipenv wheel
+      - name: Update OpenSearch Manifests
+        run: |
+          ./bundle-workflow/manifests.sh update
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
```
peternied, line 32: Could this pile up duplicate pull requests?  Do we have any control over managing duplication?

peternied (approved):

.github/workflows/versions.yml
dblock: The plugin assumes a branch name (which here I didn't specify so it will just be called "patch") and if it sees a PR from a branch called "patch", it will update it instead of creating a new one. It's pretty smart like that.

bundle-workflow/src/manifests_workflow/component_opensearch_min.py
```diff
@@ -0,0 +1,56 @@
...
+
+class ComponentOpenSearchMin(Component):
+    def __init__(self, repo, snapshot=False):
+        super().__init__("OpenSearch", repo, snapshot)
+
+    @classmethod
+    def get_root_branches(self):
+        branches = ["main"]
+        remote_branches = (
+            subprocess.check_output(
```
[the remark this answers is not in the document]
dblock: Makes sense, will revisit in another PR.

.github/workflows/versions.yml
peternied: Awesome! Could we add a branch name that is based off of the version number?  This way if there is a patch release and a minor release updates at the same time separate pull requests are created.

peternied (approved):

.github/workflows/versions.yml
dblock: Not so easy to know what's being added, would require returning something meaningful from the build that can be parsed by the workflow. I did update the branch name to `manifests/auto` so that the auto-pr doesn't walk over other PRs.

dblock: Setting `branch:` [didn't work](https://github.com/dblock/opensearch-build/runs/3612091607?check_suite_focus=true) on my fork with a permissions error that doesn't happen with defaults, so I undid that.

peternied (approved):

bundle-workflow/README.md
```diff
@@ -165,3 +166,26 @@ The following options are available.
 | --component [name] | Test a single component by name, e.g. `--component common-utils`.       |
 | --keep             | Do not delete the temporary working directory on both success or error. |
 | -v, --verbose      | Show more verbose output.                                               |
+
+### Auto-Generate Manifests
+
+The [manifests workflow](src/manifests) reacts to version increments in OpenSearch and its components by extracting Gradle properties from project branches. When a new version is identified, it creates a new input manifest in [manifests](../manifests) and [opens a pull request](../.github/workflows/manifests.yml).
```
mch2, line 172: Can we specify which branches of core & components are scanned?

If I'm reading this correctly, we are only scanning branches of core (with the specific 'root' pattern)  and the 'main' branch of component repos.  Is that correct?

mch2 (approved):

bundle-workflow/src/manifests_workflow/component_opensearch_min.py
```diff
@@ -0,0 +1,57 @@
...
+from manifests_workflow.component import Component
+from system.properties_file import PropertiesFile
+
+
+class ComponentOpenSearchMin(Component):
+    def __init__(self, repo, snapshot=False):
+        super().__init__("OpenSearch", repo, snapshot)
+
+    @classmethod
+    def get_root_branches(self):
```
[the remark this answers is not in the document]
dblock: Will do.

bundle-workflow/src/manifests_workflow/component_opensearch.py
```diff
@@ -0,0 +1,52 @@
...
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+from git.git_repository import GitRepository
+from manifests_workflow.component import Component
+from system.properties_file import PropertiesFile
+
+
+class ComponentOpenSearch(Component):
```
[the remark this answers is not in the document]
dblock: I like that file names sort :)

bundle-workflow/README.md
dblock: Correct. We’re looking at anything that seems like a version number. It’s really not that important though, we could be examining all branches.

We should probably not actually be looking at branches like 1.0 to save time - in our branching strategy we always increment on main or 1.x first.

Issue #341, opened 2021-08-27
TestPublisher: Implement result publisher


gaiksaya:
Test Publisher will be responsible for publishing the test results to a specific location. eg: S3, local or long lived cluster.

owaiskazi19:
Closes https://github.com/opensearch-project/opensearch-build/pull/938

Issue #340, opened 2021-08-27
TestRecorder: Add implementations to record generic information from test suites.


gaiksaya:
TestRecorder will be called in all the 3 tests Perf, Integ and BWC. It'll collect the test results once the test is finished and then pass the information to TestPublisher to publish it in S3.

- [ ] Add UT

gaiksaya:
Seeing some tests for this in the code base: https://github.com/opensearch-project/opensearch-build/tree/main/tests/tests_test_workflow/test_recorder
Closing this issue.

Issue #207, opened 2021-08-12
Implement results publisher


camerski:
Subtask of https://github.com/opensearch-project/opensearch-build/issues/177

Implement a results publisher that writes test logs and results to the same location that the bundle manifest came from (local filesystem or S3).

saratvemulapalli:
### Test Reporting
Test Reporting consists of TestRecorder and TestPublisher.

P0
- Implement TestRecorder which will record all test outputs.
- Implement TestPublisher which will publish results to S3. 

Here is the low level design proposal for each of the pieces

TestRecorder: It is instantiated once per test suite run.
```
class TestRecorder:
    def __init__(self, location, test_type, test_run_id, bundle_manifest):
         self.location = location (Use temp directory if location is not specified)
         self.test_type = test_type
         self.test_run_id = test_run_id
         self.bundle_manifest = bundle_manifest

    def record_remote_cluster_logs(self, component_name, exit_status, log_files):
        """
        Record remote cluster logs for a given component
        :param results: A generator that yields tuples containing test cluster log files, in the form (absolute_path, relative_path)
        """
        ...

    def record_local_test_logs(
        self, component_name, exit_status, stdout, stderr, results):
        """
        Record the outcome of test run on a component in {self.location}/{self.test_run_id}/{self.test_type}/{component_name}/test_outcome
        :param component_name: The name of the component that ran tests.
        :param exit_code: One of SUCCESS, FAILED, ERROR. SUCCESS means the tests ran and all of them passed.
        FAILED means the tests ran and at least one failed. ERROR means the test suite did not run correctly.
        :param stdout: A string containing the stdout stream from the test process.
        :param stderr: A string containing the stderr stream from the test process.
        :param results: A generator that yields tuples containing test results files, in the form (absolute_path, relative_path).
        This method generates test_result_manifest.yml and stores it under {self.location}/{self.test_run_id}/{component_name}/{self.test_type}/test_outcome/
        """
       ...

    def clean_up(self):
    """
    Cleans up all files under {self.location}
    """
```

`test_result_manifest.yml` contains:
```
test_type:integ_test
component:sql
exit_status: success
test_run_id: 1
```

TestPublisher:
```
class TestPublisher:
    def __init__(self, s3_bucket, bundle_manifest, test_recorder):
        self.s3_bucket = s3_bucket
        self.bundle_manifest = bundle_manifest
        self.test_recorder = test_recorder
       
    def publish_test_results_to_s3(self):
    """
[...]

PR #464, opened 2021-09-13
Added Test Publisher

owaiskazi19:

### Description
Implemented Test Publisher to store result of the test to S3.
Test Publisher will be invoked by Test Recorder. Part of https://github.com/opensearch-project/opensearch-build/issues/340
 
### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-build/issues/341
 
### Check List

### DIFF
--- a/bundle-workflow/src/test_workflow/test_publisher.py
+++ b/bundle-workflow/src/test_workflow/test_publisher.py
@@ -1,32 +1,26 @@
 import os
-import sys
-sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../src"))
-from aws.s3_bucket import S3Bucket
-from manifests.bundle_manifest import BundleManifest
+
 
 class TestPublisher:
-    def __init__(self, s3_bucket=None, bundle_manifest=None, test_recorder=None):
-        self.bundle_manifest = BundleManifest.from_path('/Users/kazabdu/Downloads/manifest.yaml')
-        #self.bundle_manifest = bundle_manifest
+    def __init__(self, s3_bucket, bundle_manifest, test_recorder):
+        self.s3_bucket = s3_bucket
+        self.bundle_manifest = bundle_manifest
         self.test_recorder = test_recorder
-    
+
     def publish_test_results_to_s3(self):
         """
             Publishes tests results to S3 pulling information from {self.test_recorder}
             And cleans up all local storage after publishing ({self.test_recorder}.clean_up())
         """
-        s3_bucket = S3Bucket('test-publisher-upload', 'arn:aws:iam::821468782434:role/test-publisher', 'test-publisher')
         base_path = self._get_base_path()
 
-        for subdir,dirs,files in os.walk('/Users/kazabdu/tmp/tmp-dir/tests/1/component/integ'):
+        for subdir, dirs, files in os.walk(self.test_recorder.location):
             test_path = subdir[subdir.find('tests'):]
             for file_name in files:
                 file_path = os.path.join(subdir, file_name)
                 if not file_path.startswith('.'):
                     s3_path = os.path.join(base_path, test_path, file_name)
-                    print('FILE', file_path)
-                    print('S3', s3_path)
-                    s3_bucket.upload_file(s3_path, file_path)
+                    self.s3_bucket.upload_file(s3_path, file_path)
 
     def _get_base_path(self):
         """
@@ -38,6 +32,3 @@ class TestPublisher:
         arch = self.bundle_manifest.build.architecture
         s3_path = os.path.join(work_dir, bundle_version, build_id, arch)
         return s3_path
-
-test_publisher = TestPublisher()
-test_publisher.publish_test_results_to_s3()
--- a/bundle-workflow/src/test_workflow/test_publisher.py
+++ b/bundle-workflow/src/test_workflow/test_publisher.py
@@ -18,12 +18,14 @@ class TestPublisher:
         s3_bucket = S3Bucket('test-publisher-upload', 'arn:aws:iam::821468782434:role/test-publisher', 'test-publisher')
         base_path = self._get_base_path()
 
-        for subdir,dirs,files in os.walk('/Users/kazabdu/1/'):
-            for filename in files:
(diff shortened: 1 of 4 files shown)


bundle-workflow/src/test_workflow/test_publisher.py
```diff
@@ -0,0 +1,39 @@
+import os
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../src"))
```
owaiskazi19, line 4: Is there any better way to do this? Require this to import aws module below

dblock: This should not be required at all when run from the test.sh driver because it introduces a stable path. That's why we have those things.

bundle-workflow/src/test_workflow/test_publisher.py
```diff
@@ -0,0 +1,39 @@
...
+from aws.s3_bucket import S3Bucket
+
+
+class TestPublisher:
```
dblock, line 8: I suggest renaming this to `TestResults`, since it wraps test results, and rename `publish_test_results_to_s3` to `to_s3`, similar to other `to_file` or `to_dict` that we have on other classes.

I would add `TestRecorder.results` that returns an instance of test results.

Remove `s3_bucket` from the constructor since it never consumes S3 results, and make it a parameter in `to_s3`, so `TestResults().to_s3(bucket)`.

owaiskazi19: I will change the name in the proposal as well https://github.com/opensearch-project/opensearch-build/issues/207#issuecomment-911927824

bundle-workflow/tests/test_results/test_test_results.py
```diff
@@ -0,0 +1,40 @@
+import os
+import unittest
+from unittest.mock import patch
+
+from aws.s3_bucket import S3Bucket
+from manifests.bundle_manifest import BundleManifest
+from test_workflow.test_results import TestResults
+
+
+class TestTestResults(unittest.TestCase):
```
saratvemulapalli, line 10: I chuckled at this!

bundle-workflow/src/test_workflow/test_publisher.py
saratvemulapalli: `TestPublisher` feels right to me as this class really just publishes the results while wrapping them up in the right directory structure. But I would leave it to dB who is our master of renaming :).

bundle-workflow/src/test_workflow/test_results.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, bundle_manifest, test_recorder):
+        self.bundle_manifest = bundle_manifest
+        self.test_recorder = test_recorder
+
+    def to_s3(self, bucket):
+        """
+            Publishes tests results to S3 pulling information from {self.test_recorder}
+            And cleans up all local storage after publishing ({self.test_recorder}.clean_up())
+        """
+        s3_bucket = S3Bucket(bucket, '<role-arn>', 'test-publisher-session')
```
saratvemulapalli, line 16: I think we already have a role defined. @gaiksaya could help and chime in with the role we have already created in Jenkins?

saratvemulapalli (commented):
Overall the change look good.

bundle-workflow/src/test_workflow/test_results.py
gaiksaya: Yes. Asked @owaiskazi19 to remove this params except the s3bucket. They are not required.

bundle-workflow/src/test_workflow/test_results.py
```diff
@@ -0,0 +1,36 @@
...
+
+class TestResults:
+    def __init__(self, bundle_manifest, test_recorder):
+        self.bundle_manifest = bundle_manifest
+        self.test_recorder = test_recorder
+
+    def to_s3(self, bucket):
+        """
+            Publishes tests results to S3 pulling information from {self.test_recorder}
+            And cleans up all local storage after publishing ({self.test_recorder}.clean_up())
```
saratvemulapalli, line 14: Fix this comment as TestRecorder cleans it up by itself and TestPublisher doesn't have to be worried about it.

gaiksaya: TestRecorder is not cleaning up. I removed the clean_up block but it is creating the temporary directory which I assume will be cleaned up eventually by the system?

saratvemulapalli: Responded on the other PR so that its clear: https://github.com/opensearch-project/opensearch-build/pull/467/files#r708746645
I'll update the comment on #207

bundle-workflow/src/test_workflow/test_publisher.py
owaiskazi19: @dblock any thoughts on this?

dblock: No strong feelings.

owaiskazi19:
Can open this PR once https://github.com/opensearch-project/opensearch-build/pull/467 gets merged.

bundle-workflow/src/test_workflow/test_publisher.py
```diff
@@ -0,0 +1,36 @@
...
+
+from aws.s3_bucket import S3Bucket
+
+
+class TestPublisher:
+    def __init__(self, bundle_manifest, test_recorder):
+        self.bundle_manifest = bundle_manifest
+        self.test_recorder = test_recorder
+
+    def to_s3(self):
```
setiah, line 11: we also need a `to_local` method to publish logs locally for devs

dblock:
#467 was merged, what's next for this?

owaiskazi19:
> #467 was merged, what's next for this?

#467 is closed and not merged. https://github.com/opensearch-project/opensearch-build/pull/587 is open.

peternied:
@owaiskazi19 it looks like #587 has been merged, did you want to pick this back up?

dblock:
Let's close this until @gaiksaya re-raises a working PR.

PR #465, opened 2021-09-13
Refactor manifest writing.

dblock:

### Description

Extracted from #461.

- Added `Manifest#to_file`.
- Refactored `__to_dict__` into private methods and added a single `Manifest#to_dict`. 
- Added `Manifest#compact` that removes empty keys when writing manifests.
- Made sure `schema-version` is `str`.

### Check List

### DIFF
--- a/bundle-workflow/src/manifests/manifest.py
+++ b/bundle-workflow/src/manifests/manifest.py
@@ -21,15 +21,15 @@ class Manifest(ABC):
 
     @classmethod
     def compact(cls, d):
-        clean = {}
+        result = {}
         for k, v in d.items():
             if isinstance(v, dict):
                 nested = cls.compact(v)
-                if len(nested.keys()) > 0:
-                    clean[k] = nested
-            elif v is not None and v != []:
-                clean[k] = v
-        return clean
+                if nested:
+                    result[k] = nested
+            elif v and v != []:
+                result[k] = v
+        return result
 
     def __to_dict(self):
         return {}
--- a/bundle-workflow/src/manifests/manifest.py
+++ b/bundle-workflow/src/manifests/manifest.py
@@ -39,7 +39,7 @@ class Manifest(ABC):
 
     def to_file(self, path):
         with open(path, "w") as file:
-            yaml.dump(self.to_dict(), file)
+            yaml.safe_dump(self.to_dict(), file)
 
     @abstractmethod
     def __init__(self, data):
--- a/bundle-workflow/src/manifests/manifest.py
+++ b/bundle-workflow/src/manifests/manifest.py
@@ -19,6 +19,28 @@ class Manifest(ABC):
         with open(path, "r") as f:
             return cls.from_file(f)
 
+    @classmethod
+    def compact(cls, d):
+        clean = {}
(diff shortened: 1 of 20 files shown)


bundle-workflow/src/manifests/manifest.py
```diff
@@ -19,6 +19,28 @@ def from_path(cls, path):
         with open(path, "r") as f:
             return cls.from_file(f)
 
+    @classmethod
+    def compact(cls, d):
+        clean = {}
+        for k, v in d.items():
+            if isinstance(v, dict):
+                nested = cls.compact(v)
+                if len(nested.keys()) > 0:
```
owaiskazi19, line 28: If we are just checking whether the dict is empty or not. Can be check like `if nested:`

bundle-workflow/src/manifests/manifest.py
```diff
@@ -19,6 +19,28 @@ def from_path(cls, path):
...
+                    clean[k] = nested
+            elif v is not None and v != []:
```
owaiskazi19, line 30: Can be check like `if not nested[k]`

bundle-workflow/src/manifests/manifest.py
dblock: This will teach me copying code from SO :) Fixed.

bundle-workflow/src/manifests/manifest.py
dblock: Not sure I follow, LMK if https://github.com/opensearch-project/opensearch-build/pull/465#:~:text=Resolve%20conversation-,Addressed%20comments.,-%E2%80%A6 can be improved further.

bundle-workflow/src/manifests/manifest.py
```diff
@@ -19,6 +19,28 @@ def from_path(cls, path):
...
+
+    def __to_dict(self):
+        return {}
+
+    def to_dict(self):
+        return Manifest.compact(self.__to_dict__())
+
+    def to_file(self, path):
+        with open(path, "w") as file:
+            yaml.dump(self.to_dict(), file)
```
owaiskazi19, line 42: Not sure but is using `safe_dump` a better way here?

bundle-workflow/src/manifests/manifest.py
owaiskazi19: Since we are checking whether the key has any value in the dictionary. Why not to check it like `if not dict[key]` which will check for value mapped to the key? Let me know what you think

dblock: I don't think I understand ;) We already have a hold on `v`, the value through the iteration above. If that value is `None` or `[]` we don't want to write it. Could you please write the function as you suggest?

owaiskazi19: Ohh. I missed this https://github.com/opensearch-project/opensearch-build/commit/1664903d1850bce165e180efc023affcae4523db commit. That exactly does what I suggested above.

VachaShah (approved):

owaiskazi19 (approved):
LGTM!

Issue #211, opened 2021-08-13
Trigger notifications after test run


setiah:
The [test-orchestrator-pipeline](https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/jenkins_workflow/test/orchestrator/Jenkinsfile#L45-L50) has a notify job that is currently no-op. This issue is for adding logic to the job to trigger notifications after a successful/failed test-execute step.

peternied:
This was completed in https://github.com/opensearch-project/opensearch-build/pull/1105

PR #466, opened 2021-09-13
Add notification during bundle builds

peternied:
Wiring up a simple notification via Slack integration webhook to keep the team abreast of the builds.

image
### Related Issue
* #211

### Check List

### DIFF
--- a/bundle-workflow/Jenkinsfile
+++ b/bundle-workflow/Jenkinsfile
@@ -119,6 +119,6 @@ void build() {
 
 void publishNotification(icon, msg) {
     withCredentials([string(credentialsId: 'BUILD_NOTICE_WEBHOOK', variable: 'TOKEN')]) {
-        sh """curl -XPOST --header "Content-Type: application/json" --data '{"result_text": "$icon ${env.JOB_NAME} [${env.BUILD_NUMBER}] $msg ${env.BUILD_URL}"}' $TOKEN"""
+        sh """curl -XPOST --header "Content-Type: application/json" --data '{"result_text": "$icon ${env.JOB_NAME} [${env.BUILD_NUMBER}] $msg ${env.BUILD_URL}\nManifest: ${BUILD_MANIFEST}"}' $TOKEN"""
     }
 }
--- a/bundle-workflow/Jenkinsfile
+++ b/bundle-workflow/Jenkinsfile
@@ -24,6 +24,11 @@ pipeline {
             }
         }
         stage('build') {
+            agent {
+                node {
+                    label 'Jenkins-Agent-al2-x64-m5xlarge'
+                }
+            }
             stages {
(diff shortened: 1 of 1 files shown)


bundle-workflow/Jenkinsfile
```diff
@@ -24,6 +24,11 @@ pipeline {
             }
         }
         stage('build') {
+            agent {
+                node {
+                    label 'Jenkins-Agent-al2-x64-m5xlarge'
+                }
+            }
```
peterzhuamazon, line 31: Why do we need to add another node here?

gaiksaya (approved):

bundle-workflow/Jenkinsfile
peternied: Since the top agent is `none` it has to be specified for any usages, see [agent directive](https://www.jenkins.io/doc/book/pipeline/syntax/#agent). The curl command needs to execute somewhere - and it would be better that it wasn't the main host.

Note: a very tiny docker image would be great to switch to when available

bundle-workflow/Jenkinsfile
```diff
@@ -87,6 +92,14 @@ pipeline {
                     }
                 }
             }
+            post() {
+                success {
+                    publishNotificaiton(":white_check-mark:", "Successful Build")
```
mch2, line 97: Spelling of this method is off -  publishNotific**ation**

bundle-workflow/Jenkinsfile
dblock: Love that we're using a m5xlarge to send a notification :)

bundle-workflow/Jenkinsfile
```diff
@@ -87,6 +92,14 @@ pipeline {
                     }
                 }
             }
+            post() {
+                success {
+                    publishNotification(":white_check-mark:", "Successful Build")
+                }
+                failure {
+                    publishNotification(":warning:", "Failed Build")
```
dblock, line 100: warning? not error?

dblock (dismissed):

dblock:
Can this notification include the branch/version that was built?

bundle-workflow/Jenkinsfile
peternied: I didn't see an icon that I liked in slack for this, ⚠️  : warning: seems to work well enough.   Open to suggestions

dblock: It's A-OK

gaiksaya (approved):

bundle-workflow/Jenkinsfile
bbarani: I prefer image icon

zelinh (approved):

PR #467, opened 2021-09-13
Add test_recorder implementation and integration with integ_test_suite

### LINKED ISSUE #340: TestRecorder: Add implementations to record generic information from test suites.

gaiksaya:

### Description
Adds test_recorder implementation and integration with integ_test_suite. 
Test_recorder will be instantiated by each test_suite. The test_recorder object will then be passed to test_publisher to store the results in a specific location (s3, long-lived-cluster, etc).

**To-Do:**
1. Integrate with other test_suite i.e. bwc-test, perf-test
2. Add UT
 
### Issues Resolved
#340 

### Tests:
```
/tmp/tmp93i0ybyw/tests/2/integ-test/index-management/with-security ~  ls local_cluster_logs test_outcome
local_cluster_logs:
gc.log                                 opensearch_deprecation.json            opensearch_index_indexing_slowlog.log  opensearch_server.json
gc.log.00                              opensearch_deprecation.log             opensearch_index_search_slowlog.json   stderr.txt
opensearch.log                         opensearch_index_indexing_slowlog.json opensearch_index_search_slowlog.log    stdout.txt

test_outcome:
index-management.yml stderr.txt           stdout.txt


~ cat test_outcome/index-management.yml
component_name: index-management
log_file_location: S3
status: FAILED/ERROR
test_config: with-security
test_run_id: '2'
test_type: integ-test
```
 
### Check List

### DIFF
--- a/bundle-workflow/src/run_integ_test.py
+++ b/bundle-workflow/src/run_integ_test.py
@@ -104,7 +104,6 @@ def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
 
 def main():
     args = parse_arguments()
-    logging.getLogger('').handlers = []
     console.configure(level=args.logging_level)
     bundle_manifest = BundleManifest.from_file(args.bundle_manifest)
     build_manifest = BuildManifest.from_file(args.build_manifest)
--- a/bundle-workflow/src/run_integ_test.py
+++ b/bundle-workflow/src/run_integ_test.py
@@ -18,6 +18,7 @@ from manifests.test_manifest import TestManifest
 from system import console
 from system.temporary_directory import TemporaryDirectory
 from test_workflow.integ_test.integ_test_suite import IntegTestSuite
+from test_workflow.test_recorder import TestRecorder
 
 # TODO: 1. log test related logging into a log file. Output only workflow logs on shell.
 # TODO: 2. Move common functions to utils.py
@@ -38,6 +39,9 @@ def parse_arguments():
         "--test-manifest", type=argparse.FileType("r"), help="Test Manifest file."
     )
     parser.add_argument(
+        "--test-run-id", type=str, help="Test Run ID"
+    )
+    parser.add_argument(
         "--keep",
         dest="keep",
         action="store_true",
@@ -54,7 +58,6 @@ def parse_arguments():
     )
     args = parser.parse_args()
     return args
-
 
 # TODO: replace with DependencyProvider - https://github.com/opensearch-project/opensearch-build/issues/283
 def pull_common_dependencies(work_dir, build_manifest):
@@ -100,10 +103,12 @@ def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
 
 def main():
     args = parse_arguments()
+    logging.getLogger('').handlers = []
     console.configure(level=args.logging_level)
     bundle_manifest = BundleManifest.from_file(args.bundle_manifest)
     build_manifest = BuildManifest.from_file(args.build_manifest)
     test_manifest = TestManifest.from_file(args.test_manifest)
+    test_recorder = TestRecorder(args.test_run_id, "integ-test")
     integ_test_config = dict()
     for component in test_manifest.components:
         if component.integ_test is not None:
@@ -120,6 +125,7 @@ def main():
                     integ_test_config[component.name],
                     bundle_manifest,
                     work_dir,
+                    test_recorder
                 )
                 test_suite.execute()
             else:
--- a/bundle-workflow/src/run_integ_test.py
+++ b/bundle-workflow/src/run_integ_test.py
@@ -59,6 +59,7 @@ def parse_arguments():
     args = parser.parse_args()
     return args
 
+
 # TODO: replace with DependencyProvider - https://github.com/opensearch-project/opensearch-build/issues/283
 def pull_common_dependencies(work_dir, build_manifest):
     logging.info("Pulling common dependencies for integration tests")
--- a/bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
+++ b/bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
@@ -33,12 +33,10 @@ class IntegTestSuite:
             self.component.commit_id,
             os.path.join(self.work_dir, self.component.name),
         )
-        self.individual_config = None
 
     def execute(self):
         self._fetch_plugin_specific_dependencies()
         for config in self.test_config.integ_test["test-configs"]:
-            self.individual_config = config
             self._setup_cluster_and_execute_test_config(config)
 
     # TODO: fetch pre-built dependencies from s3
@@ -83,9 +81,9 @@ class IntegTestSuite:
             os.chdir(self.work_dir)
             # TODO: (Create issue) Since plugins don't have integtest.sh in version branch, hardcoded it to main
             security = self._is_security_enabled(config)
-            self._execute_integtest_sh(test_cluster_endpoint, test_cluster_port, security)
+            self._execute_integtest_sh(test_cluster_endpoint, test_cluster_port, security, config)
 
-    def _execute_integtest_sh(self, endpoint, port, security):
+    def _execute_integtest_sh(self, endpoint, port, security, config):
         script = self.script_finder.find_integ_test_script(
             self.component.name, self.repo.dir
         )
@@ -95,9 +93,13 @@ class IntegTestSuite:
             results_dir = os.path.join(
                 self.repo.dir, "integ-test", "build", "reports", "tests", "integTest"
             )
-            self.test_recorder.record_test_outcome(
-                self.component.name, self.individual_config, status, stdout, stderr, walk(results_dir)
-            )
+            self.test_recorder.record_test_outcome(self.component.name,
+                                                   config,
+                                                   status,
+                                                   stdout,
+                                                   stderr,
+                                                   walk(results_dir)
+                                                   )
         else:
             logging.info(
                 f"{script} does not exist. Skipping integ tests for {self.component.name}"
--- a/bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
+++ b/bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
@@ -10,6 +10,7 @@ import subprocess
 
 from git.git_repository import GitRepository
 from paths.script_finder import ScriptFinder
+from paths.tree_walker import walk
 from system.execute import execute
 from test_workflow.integ_test.local_test_cluster import LocalTestCluster
 
@@ -20,21 +21,24 @@ class IntegTestSuite:
     test_support_matrix.yml
     """
 
-    def __init__(self, component, test_config, bundle_manifest, work_dir):
+    def __init__(self, component, test_config, bundle_manifest, work_dir, test_recorder):
         self.component = component
         self.bundle_manifest = bundle_manifest
         self.work_dir = work_dir
         self.test_config = test_config
+        self.test_recorder = test_recorder
         self.script_finder = ScriptFinder()
         self.repo = GitRepository(
             self.component.repository,
             self.component.commit_id,
             os.path.join(self.work_dir, self.component.name),
         )
+        self.individual_config = None
 
     def execute(self):
         self._fetch_plugin_specific_dependencies()
         for config in self.test_config.integ_test["test-configs"]:
+            self.individual_config = config
             self._setup_cluster_and_execute_test_config(config)
 
     # TODO: fetch pre-built dependencies from s3
@@ -73,11 +77,12 @@ class IntegTestSuite:
             return False
 
     def _setup_cluster_and_execute_test_config(self, config):
-        security = self._is_security_enabled(config)
-        with LocalTestCluster.create(self.work_dir, self.bundle_manifest, security) as (test_cluster_endpoint, test_cluster_port):
+        with LocalTestCluster.create(self.work_dir, self.bundle_manifest, self.component.name, config,
+                                     self.test_recorder) as (test_cluster_endpoint, test_cluster_port):
             logging.info("component name: " + self.component.name)
             os.chdir(self.work_dir)
             # TODO: (Create issue) Since plugins don't have integtest.sh in version branch, hardcoded it to main
+            security = self._is_security_enabled(config)
             self._execute_integtest_sh(test_cluster_endpoint, test_cluster_port, security)
 
     def _execute_integtest_sh(self, endpoint, port, security):
@@ -87,7 +92,13 @@ class IntegTestSuite:
         if os.path.exists(script):
             cmd = f"sh {script} -b {endpoint} -p {port} -s {str(security).lower()}"
             (status, stdout, stderr) = execute(cmd, self.repo.dir, True, False)
+            results_dir = os.path.join(
+                self.repo.dir, "integ-test", "build", "reports", "tests", "integTest"
+            )
+            self.test_recorder.record_test_outcome(
+                self.component.name, self.individual_config, status, stdout, stderr, walk(results_dir)
+            )
         else:
             logging.info(
-                f"{script} does not exist. Skipping integ tests for {self.name}"
+                f"{script} does not exist. Skipping integ tests for {self.component.name}"
             )
--- a/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
+++ b/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
@@ -61,7 +61,9 @@ class LocalTestCluster(TestCluster):
             return
         self.terminate_process()
         logs_files = walk(os.path.join(self.work_dir, self.install_dir, "logs"))
-        self.test_recorder.record_local_cluster_logs(self.component_name, self.component_config, self.local_cluster_stdout,
+        self.test_recorder.record_local_cluster_logs(self.component_name,
+                                                     self.component_config,
+                                                     self.local_cluster_stdout,
                                                      self.local_cluster_stderr,
                                                      logs_files)
 
--- a/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
+++ b/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
@@ -12,6 +12,7 @@ import urllib.request
 
 import requests
 
+from paths.tree_walker import walk
 from test_workflow.test_cluster import ClusterCreationException, TestCluster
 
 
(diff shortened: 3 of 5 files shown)


bundle-workflow/src/run_integ_test.py
```diff
@@ -100,10 +104,12 @@ def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
 
 def main():
     args = parse_arguments()
+    logging.getLogger('').handlers = []
```
dblock, line 107: This shouldn't be needed.

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -73,11 +77,12 @@ def _is_security_enabled(self, config):
             return False
 
     def _setup_cluster_and_execute_test_config(self, config):
```
dblock, line 79: What's the convention for private methods? I think we've been doing two `__`

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -20,21 +21,24 @@ class IntegTestSuite:
...
             self.component.repository,
             self.component.commit_id,
             os.path.join(self.work_dir, self.component.name),
         )
+        self.individual_config = None
 
     def execute(self):
         self._fetch_plugin_specific_dependencies()
         for config in self.test_config.integ_test["test-configs"]:
+            self.individual_config = config
```
dblock, line 41: This should be passed down as a parameter, it keeps changing inside an iteration.

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -87,7 +92,13 @@ def _execute_integtest_sh(self, endpoint, port, security):
         if os.path.exists(script):
             cmd = f"sh {script} -b {endpoint} -p {port} -s {str(security).lower()}"
             (status, stdout, stderr) = execute(cmd, self.repo.dir, True, False)
+            results_dir = os.path.join(
+                self.repo.dir, "integ-test", "build", "reports", "tests", "integTest"
+            )
+            self.test_recorder.record_test_outcome(
+                self.component.name, self.individual_config, status, stdout, stderr, walk(results_dir)
```
dblock, line 99: `self.individual_config` should come from input, otherwise it's the last item in `self.test_config.integ_test["test-configs"]:`, which is probably not what's intended

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -55,21 +60,31 @@ def destroy(self):
             logging.info("Local test cluster is not started")
             return
         self.terminate_process()
+        logs_files = walk(os.path.join(self.work_dir, self.install_dir, "logs"))
+        self.test_recorder.record_local_cluster_logs(self.component_name, self.component_config, self.local_cluster_stdout,
```
dblock, line 64: For clarity, put every argument on its own line

```python
self.test_recorder.record_local_cluster_logs(
     self.component_name, 
```

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,142 @@
...
-        logging.info(f"TestRecorder storing results in {location}")
+    ACCEPTED_TEST_TYPES = ["integ-test", "bwc-test", "perf-test"]
+
+    def __init__(self, test_run_id, test_type, location=None):
+        self.test_type = test_type
+        self.test_run_id = test_run_id
 
-    def record_cluster_logs(self, log_files):
+        if self.test_type not in self.ACCEPTED_TEST_TYPES:
+            raise ValueError(f"test_type is invalid. Acceptable test_type are {self.ACCEPTED_TEST_TYPES}")
```
dblock, line 22: Include the type passed in.

```
raise ValueError(f"Invalid test type: {test_type}, expected one of ...")
```

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,142 @@
...
+        os.makedirs(dest_directory, exist_ok=True)
+        return os.path.realpath(dest_directory)
+
+    def __get_exit_status(self, exit_code):
+        if exit_code == 0:
+            return "SUCCESS"
+        else:
+            return "FAILED/ERROR"
+
+    def __generate_test_outcome_yml(self, component_name, component_test_config, exit_status, log_file_location):
```
dblock, line 118: log_file_location -> output_path

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,142 @@
...
+            "component_name": component_name,
+            "test_config": component_test_config,
+            "status": exit_status,
+            "log_file_location": log_file_location
+        }
+        with open("%s.yml" % component_name, "w") as file:
+            yaml.dump(dict_file, file)
+        return os.path.realpath("%s.yml" % component_name)
+
+    def __generate_std_files(self, stdout, stderr, location):
```
dblock, line 131: location -> output_path

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,142 @@
...
+        stdout_file = open(os.path.join(location, "stdout.txt"), "w")
```
dblock, line 132: Use `with`, so you won't need a try/finally

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,142 @@
...
+        stderr_file = open(os.path.join(location, "stderr.txt"), "w")
+        try:
+            stdout_file.write(stdout)
+            stderr_file.write(stderr)
+        finally:
+            stdout_file.close()
+            stderr_file.close()
+
+    def cleanup(self):
```
dblock, line 141: Delete this, too dangerous, use a TemporaryFile location if you need to cleanup after the process.

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,142 @@
...
+        return os.path.realpath(dest_directory)
+
+    def __get_exit_status(self, exit_code):
+        if exit_code == 0:
+            return "SUCCESS"
+        else:
+            return "FAILED/ERROR"
+
+    def __generate_test_outcome_yml(self, component_name, component_test_config, exit_status, log_file_location):
+        dict_file = {
```
dblock, line 119: outcome = ...

dblock (changes requested):
Mostly small stuff below.

There's a lot of passing 4-6 parameters around, consider some refactoring into configuration objects.

You should work off your own fork to avoid everyone bringing everybody's branches along.

bundle-workflow/src/test_workflow/test_recorder.py
gaiksaya: It is not bring used right now, but this location should be cleaned up once it is passed and used by `test_publisher`

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
gaiksaya: Not written by me but I can change if @setiah is okay with that or not making changes on his side

bundle-workflow/src/run_integ_test.py
gaiksaya: Yeah the handlers were not getting cleared on my local. will remove it from here

bundle-workflow/src/test_workflow/test_recorder.py
dblock: Not by calling `rm -rf` or variations of that code, that is bound to nuke someone's files. Use TemporaryDirectory, please. For now you should stub/delete this code.

bundle-workflow/src/run_integ_test.py
gaiksaya: Done

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
gaiksaya: done

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
gaiksaya: Done

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
gaiksaya: Done

bundle-workflow/src/test_workflow/test_recorder.py
gaiksaya: Done

bundle-workflow/src/test_workflow/test_recorder.py
gaiksaya: done

bundle-workflow/src/run_integ_test.py
```diff
@@ -37,6 +38,9 @@ def parse_arguments():
     parser.add_argument(
         "--test-manifest", type=argparse.FileType("r"), help="Test Manifest file."
     )
+    parser.add_argument(
+        "--test-run-id", type=str, help="Test Run ID"
```
saratvemulapalli, line 42: You might want to sync up with @setiah as he also added this param as part of #456

bundle-workflow/src/test_workflow/test_recorder.py
gaiksaya: Done

bundle-workflow/src/test_workflow/test_recorder.py
gaiksaya: Removed

bundle-workflow/src/test_workflow/test_recorder.py
gaiksaya: Done

bundle-workflow/src/run_integ_test.py
```diff
@@ -104,6 +108,7 @@ def main():
     bundle_manifest = BundleManifest.from_file(args.bundle_manifest)
     build_manifest = BuildManifest.from_file(args.build_manifest)
     test_manifest = TestManifest.from_file(args.test_manifest)
+    test_recorder = TestRecorder(args.test_run_id, "integ-test")
```
saratvemulapalli, line 111: You might want to define a standard constants which defines each of the test suites. 
We would want all the workflows to use same identifiers.

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,133 @@
...
-
 import logging
+import os
+import shutil
+import tempfile
+
+import yaml
 
 
 class TestRecorder:
```
saratvemulapalli, line 14: Make this class a singleton. With the design pattern we have, we'd like to have just one instance of TestRecorder for a given testsuite/jenkins job. 
This also prevents other instantiations, avoids passing the test recorder instance in multiple places within the code base.

```suggestion
class TestRecorder:
   __instance = None
   @staticmethod 
   def getInstance(test_run_id, test_type, location=None):
      """ Static access method. """
      if TestRecorder.__instance == None:
         TestRecorder(test_run_id, test_type, location)
      return TestRecorder.__instance
```

Python by default doesn't have private constructors, if you'd like to block instantiation outside of the class, you could raise an error in `__init__` and use a different private method which could be called by `getInstance(...)`

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,133 @@
...
 
-    def record_integ_test_outcome(
-        self, component_name, exit_status, stdout, stderr, results
+        base = self.__create_base_folder_structure(component_name, component_test_config)
+        dest_directory = os.path.join(base, "remote_cluster_logs")
+        os.makedirs(dest_directory, exist_ok=False)
+        logging.info(f"Recording remote cluster logs for {component_name} in {os.path.realpath(dest_directory)}")
+        self.__generate_std_files(stdout, stderr, os.path.realpath(dest_directory))
+        exit_status = self.__get_exit_status(exit_code)
+        component_yml = self.__generate_test_outcome_yml(component_name, component_test_config, exit_status,
```
saratvemulapalli, line 71: nit:
```suggestion
        test_outcome_yml = self.__generate_test_outcome_yml(component_name, component_test_config, exit_status,
```

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,133 @@
...
+        base = self.__create_base_folder_structure(component_name, component_test_config)
+        dest_directory = os.path.join(base, "local_cluster_logs")
+        os.makedirs(dest_directory, exist_ok=False)
+        logging.info(
+            f"Recording local cluster logs for {component_name} with {component_test_config} config in {os.path.realpath(dest_directory)}")
+        self.__generate_std_files(stdout, stderr, os.path.realpath(dest_directory))
+        local_cluster_log_files = list(log_files)
+        for log_file in local_cluster_log_files:
+            dest_file = os.path.join(dest_directory, os.path.basename(log_file[0]))
+            shutil.copyfile(log_file[0], dest_file)
```
saratvemulapalli, line 50: Looks like this could be stubbed out into another method. This code is repeated in `record_remote_cluster_logs` and `record_test_outcome`

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,133 @@
...
+
+    def record_remote_cluster_logs(self, component_name, component_test_config, exit_code, stdout, stderr,
```
saratvemulapalli, line 52: The flow doesn't clean to me. Let me know if I didn't understand this correctly.

- For test suite which uses local cluster, it uses both `record_test_outcome` and `record_local_cluster_logs`.
   `record_test_outcome` generates the test outcome manifest.
- For test suite which uses remote cluster, it just uses `record_remote_cluster_logs `. `record_remote_cluster_logs` generates the test outcome manifest.

If that is the case why don't we merge both remote and local test cluster into one function `record_test_cluster` which takes in local/remote as a parameter. The only logic different between them is creating a different directory.

Let me know what you think.

bundle-workflow/src/run_integ_test.py
gaiksaya: do you mean this? https://github.com/opensearch-project/opensearch-build/pull/467/files#diff-7466053dcf4632ffc67632378636660e4b2a95bec30137fca501ab304e298779R15

bundle-workflow/src/test_workflow/test_recorder.py
gaiksaya: I agree. I'll work on this. One more logic that differs is remote cluster logs won't have log_files but just the location of the logs. Will see how to differentiate that.

bundle-workflow/src/test_workflow/test_recorder.py
dblock: Also rename getInstance to get_instance

bundle-workflow/src/run_integ_test.py
saratvemulapalli: Yeah that would a part of it. I was getting towards something in common which could be used by all workflows not something which TestRecorder can understand. But this is a good start.

bundle-workflow/src/test_workflow/test_recorder.py
```diff
@@ -3,34 +3,133 @@
...
+    ACCEPTED_TEST_TYPES = ["integ-test", "bwc-test", "perf-test"]
+
+    def __init__(self, test_run_id, test_type, location=None):
+        self.test_type = test_type
+        self.test_run_id = test_run_id
+
+        if self.test_type not in self.ACCEPTED_TEST_TYPES:
+            raise ValueError(f"test_type is invalid. Acceptable test_type are {self.ACCEPTED_TEST_TYPES}")
+
+        if location is None:
```
saratvemulapalli, line 24: We have to clean up the directory when TestRecorder goes out of scope. Tmp directory gives us a great feature which takes care of it. It's better to clean it up by ourself, we do not want to leave them on the system. 

Lets only use temporary directory as the only option for TestRecorder and when it goes out of scope let it destroy the directory and add a keep parameter if caller wants the directory to be preserved.
Anyway we don't have a use case yet for a custom directory, avoids the hassle of cleaning up.

gaiksaya: Makes sense. Regarding clean up. We cannot clean up the directory unless publisher is done publishing the result. So would it make sense that publisher deletes the directory when it is done publishing? or is there any other way?

dblock: The pattern I use is to have publisher create a temporary directory, tell the published to put files there via parameters, and never delete anything explicitly. Temporary directories are cleaned up automatically.

saratvemulapalli: lets leave the ownership to the test suite runners. They use test recorder, TestRecorder records them and deletes them if it goes out of scope. 
Any way TestPublisher(aka TestResults) consumes TestRecorder so it is implicitly taken care of.

dblock: Sarat suggests that the topmost caller creates a temporary dir and it's passed down to anyone that needs to write files.

gaiksaya: Yeah that is happening right now. Will see how to clean_up when the test_recorder object goes out of scope

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
setiah: You may skip this. This (and more) will be refactored in a separate PR which adds UTs as well

bundle-workflow/src/test_workflow/test_recorder.py
gaiksaya: `record_local_cluster_logs` also copies the local test cluster log files into the directory where remote cluster just gives the location of the logs. That will be an additional change but with right param it should be distinguishable

gaiksaya:
Closing this one. Please see: https://github.com/opensearch-project/opensearch-build/pull/587

PR #482, opened 2021-09-14
Generalize build arch for k-NN on x64

### LINKED ISSUE #481: k-NN library x64 build should use generic march
In ODFE, we had [run into issues](https://github.com/opendistro-for-elasticsearch/k-NN/pull/164) with library incompatibilities due to the k-NN library being compiled on a machine with support for SIMD instruction extensions and being run on a machine without support for SIMD instruction extensions. We added a small fix in ODFE CD to resolve this. 

ARM is not impacted.

We should have a similar fix in OpenSearch.

jmazanec15:

### Description
Adds generic x64 architecture option to k-NN library build.

The same thing was done in [ODFE](https://github.com/opendistro-for-elasticsearch/k-NN/blob/main/.github/workflows/CD.yml#L125).
 
### Issues Resolved
#481 
 
### Check List

### DIFF
--- a/bundle-workflow/scripts/components/k-NN/build.sh
+++ b/bundle-workflow/scripts/components/k-NN/build.sh
@@ -65,7 +65,7 @@ mkdir -p $OUTPUT/libs
 cd jni
 
 # For x64, generalize arch so library is compatible for processors without simd instruction extensions
-if ["$ARCHITECTURE" = "x64"]; then 
+if [ "$ARCHITECTURE" = "x64" ]; then 
     sed -i 's/-march=native/-march=x86-64/g' external/nmslib/similarity_search/CMakeLists.txt
 fi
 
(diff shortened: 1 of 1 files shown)


bundle-workflow/scripts/components/k-NN/build.sh
```diff
@@ -63,6 +63,12 @@ mkdir -p $OUTPUT/libs
 
 # Build knnlib and copy it to libs
 cd jni
+
+# For x64, generalize arch so library is compatible for processors without simd instruction extensions
+if ["$ARCHITECTURE" = "x64"]; then
```
peterzhuamazon, line 68: Nit: Makes it ` if [ "$ARCHITECTURE" = "x64" ]; then` for a better spacing?

jmazanec15: Updated.

peterzhuamazon (approved):

dblock (approved):
Let's go with this, but is there a way to modify the Makefile to do something different on x64 and not hack it?

peternied:
This PR has broken the build #483 - reverting to last known good state

Issue #470, opened 2021-09-14
Test that manifests don't contain float refs


dblock:
**Is your feature request related to a problem? Please describe.**

In #468 we wrote `ref: 1.1`. In Yaml that's not a string, but a Float.

**Describe the solution you'd like**
Check that manifest refs and versions are strings in CI.

dblock:
Closed via https://github.com/opensearch-project/opensearch-build/pull/509

PR #485, opened 2021-09-15
Check manifest fields for type errors.

dblock:

### Description

- Add type annotations and type checking to manifest fields, catching `ref` values that are floats instead of strings.
- Bring in dataclass type annotation checker from https://gist.github.com/svermeulen/8a6ad727113a6d07f78332152c1d33b2, this should really be PRed upstream to https://github.com/agronholm/typeguard/issues/161, I might take it up.
- Inherit `TestManifest` from `Manifest`, enabling `from_file` and `from_path`, add tests.
 
### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/470.

### Check List


bundle-workflow/src/manifests/typechecked.py
```diff
@@ -0,0 +1,111 @@
+# SPDX-License-Identifier: Apache-2.0
```
owaiskazi19, line 1: Not sure but can we have the name of the file as `typechecker` as `typechecked` is an internal keyword?

bundle-workflow/tests/tests_manifests/data/invalid-ref.yml
```diff
@@ -0,0 +1,12 @@
...
+build:
+  name: OpenSearch
+  version: 1.1.0
+components:
+  - name: OpenSearch
+    repository: https://github.com/opensearch-project/OpenSearch.git
+    ref: 1.1
+    checks:
+      - gradle:publish
+      - gradle:properties:version
```
owaiskazi19, line 12: New line here

owaiskazi19 (approved):
Few should have changes

peternied (changes requested):
This seems like a large/complex amount of work when all the types in the build manifest are strings. Is there a usecase where we need to support numeric values?

Could we do something simpler where we automatically convert all values to strings?

dblock:
> This seems like a large/complex amount of work when all the types in the build manifest are strings. Is there a usecase where we need to support numeric values?
> 
> Could we do something simpler where we automatically convert all values to strings?

I see the following options.

1. We could not merge this and `str()` the `ref` on a git repo. This would be a hack to work around the fact that 1.1 is of type float and not a string, and there's no guarantee that a branch name such as 1.01 doesn't suddenly convert to a string `1.099999` or something like that in the future, which will be hard to debug.
2. We could check explicitly that this one field is an `str` instead of adding a typechecker. Feels not very future-proof. Plus the check as is also catches issues such as not specifying a dictionary where it should be one - it's an actual validator for manifest schema.

WDYT?

bundle-workflow/tests/tests_manifests/data/invalid-ref.yml
dblock: Added.

bundle-workflow/src/manifests/typechecked.py
dblock: Renamed to match the class name, `dataclass_typechecked`.

dblock:
@peternied LMK what you think - this does fix a bug and prevents future bugs.

peternied:
What about defining a schema document and validating?  Implemented something close on my fork [Schematize and validate bundle manitests](https://github.com/peternied/opensearch-build/commit/3eecbdd74343457f02a25c80e5396f02bbc654b9)

peternied (approved):
Since python doesn't enforce typing I guess I am hesitant to rely on the current effort because while it works right now, a new untyped field could be added and it might be missed.

dblock:
I like the schema idea better. Do you want to finish that?

peternied:
@dblock Could you pick it up the schema change and make a new PR?

owaiskazi19 (approved):

dblock:
Closing in favor of https://github.com/opensearch-project/opensearch-build/pull/509.

Issue #453, opened 2021-09-10
Publish built artifacts publicly to `release-candidates/` folder 


setiah:
### Proposal

The build workflow today keeps the built tar artifacts in an internal S3 bucket. The test workflow pulls these artifacts from the internal S3 bucket to run all long running tests against them. There is no good reason to keep these artifacts internal. If instead of keeping these artifacts in internal bucket, we could push them to external `artifacts.opensearch.org` bucket under `release-candidates/` folder, we can simplify our release workflow by cutting one additional step (which is internal) while also making the release process more transparent to community. The tests results for a release candidates will be publicly viewable on the Jenkins dashboard. Some release candidates might pass all long running tests, and some might fail. The newest candidate that passes all long running tests, will always be tagged `latest` automatically as a post-testing step and can be used by community. At the time of release, the `latest` tagged release candidate will be released. 

There are also several benefits to this approach - 
1. It provides a *public* test audit history for all release candidates that were or will be promoted to release.
2. it keeps the snapshot and release-candidate workflows similar from testing perspective w/ only change being a different url to read from. A uniform approach keeps the cicd system simpler.

setiah:
cc: @mch2 @dblock

peternied:
Created https://github.com/opensearch-project/OpenSearch/issues/1240 to allow for ad-hoc locations for these artifacts instead of requiring the current hardcoded path

gaiksaya:
Issue is resolved by this PR: https://github.com/opensearch-project/opensearch-build/pull/1378

peternied:
I would like to verify this issue after we've had a build created

gaiksaya:
> I would like to verify this issue after we've had a build created

Hi @peternied, you can verify with recent build for `1.2.4` version. Let me know if you want the links to those.
Thanks!

Issue #448, opened 2021-09-10
Move bundle-workflow a directory up


dblock:
**Is your feature request related to a problem? Please describe.**

This repository is the entire bundle build top down, and mostly lives in bundle-workflow. Most developers live in `bundle-workflow/` from which everything runs. And starting with 1.1 the majority of other scripts from 1.0 will become obsolete. 

**Describe the solution you'd like**
- Move bundle-workflow one directory up to the root
- Remove/archive everything 1.0, #279 
- Rename bundle to distribution, #353 
- Reproducible Release Process #87
- End to end process documentation #448

PR #496, opened 2021-09-15
Move distribution build infra to CDK deployment

peternied:
### Description
Switching to CDK to support the additional aws tools needed to make build artifacts public
 
### Issues Resolved
* #453 
 
### Check List


bundle-workflow/infra/lib/distribution-build-stack.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export interface DistributionBuildStackProps extends StackProps {
+  name: string,
+  bucketName?: string,
+  buildRoleName?: string,
+  bundleRoleName?: string,
+  testRoleName?: string,
+}
+
+export class DistributionBuildStack extends Stack {
```
peternied, line 18: If you are looking for the 'meat' of this change as it were - this is it.  Everything else is support for the CDK build/deployment process.

bundle-workflow/infra/lib/distribution-build-stack.ts
```diff
@@ -0,0 +1,100 @@
...
+  ArnPrincipal, CanonicalUserPrincipal, IRole, PolicyStatement, Role,
+} from '@aws-cdk/aws-iam';
+import { Bucket } from '@aws-cdk/aws-s3';
+import {
+  Arn, CfnOutput, CfnParameter, Construct, Stack, StackProps,
+} from '@aws-cdk/core';
+
+export interface DistributionBuildStackProps extends StackProps {
+  name: string,
+  bucketName?: string,
```
peternied, line 12: These optional parameters are in place so we can deploy in an account that already has used the cloud formation template that I removed

bundle-workflow/infra/bin/distribution-build.ts
```diff
@@ -0,0 +1,27 @@
...
+// arn:aws:iam::336984078605:role/JenkinsStackDev-MainNodeInstanceRole6FA3C5C9-1PQFWOWDYN42U
+
+const stacks: DistributionBuildStackProps[] = [{
+  name: 'DistributionBuildStack-Staging',
+  bucketName: 'teststatus-buildbucket-1otmnpvrkhfji',
+  buildRoleName: 'opensearch-build',
+  bundleRoleName: 'opensearch-bundle',
+  testRoleName: 'opensearch-test',
+  env: {
+    account: '791023489099',
```
peternied, line 15: 99% sure we can include the account numbers in PR, but I need to confirm this.

bundle-workflow/infra/README.md
```diff
@@ -0,0 +1,12 @@
+# Bundle Workflow Infrastructure
```
peternied, line 1: Needs a real description of the design of this system

bundle-workflow/infra/test/distribution-build-stack.test.ts
```diff
@@ -0,0 +1,13 @@
+import { expect as expectCDK, matchTemplate, MatchStyle } from '@aws-cdk/assert';
+import * as cdk from '@aws-cdk/core';
+import { DistributionBuildStack } from '../lib/distribution-build-stack';
+
+test('Empty Stack', () => {
```
peternied, line 5: Needs actual tests

peternied (commented):
Here are some of the areas keeping this PR in draft.

dblock:
Given #448 do you want to think about putting it at the root?

bundle-workflow/infra/bin/distribution-build.ts
peternied: FYI - We should not, using our internal repository to store values - converting these over to parameters

peternied:
> Given #448 do you want to think about putting it at the root?

Moved to the root and ready for review!

dblock (dismissed):

peterzhuamazon:
What is the reason to have CDK here if we already have Jenkins.

peternied:
> What is the reason to have CDK here if we already have Jenkins.

These changes are specific to the build process, built off of the existing cloud formation template that was side-by-side with these changes.  This location will likely change when the new repo opens, but this is codifying the changes we need for the 1.1.0 release.

peterzhuamazon:
> existing cloud formation template that was side-by-side with these changes

Can you point me to those templates? In my understanding the bundle workflow is only running python and shell either locally or on CI.

peternied:
@peterzhuamazon The CFN template is in this PR, it is the [deleted](https://github.com/opensearch-project/opensearch-build/pull/496/files#diff-76b1f952750e04f87f6f17e543ce3770f37e92432568c440753d19f123821d82).

peternied:
@dblock @gaiksaya @peterzhuamazon Could you take a look?

dblock (approved):
I am not super versed in CDK so no detailed comments, YOLO.

peterzhuamazon (approved):

Issue #497, opened 2021-09-15
Remove `integtest.sh` from all plugin repos


setiah:
The [integtest.sh](https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/scripts/default/integtest.sh) tool contains the logic to run integration tests for a plugin. This logic is mostly common across most plugins, so it has been moved to `opensearch-build` repo. Thus it can be removed from the individual plugin repos.
However, if a plugin requires some custom logic to run integtests, which the standard tool doesn't provide, they can continue maintaining this integtest.sh in their own repo. In this case, when the integration tests are run, if a plugin has a integtest.sh tool in their repo, it gets precedence over the standard default integtest.sh in the `opensearch-build` repo. This precedence order logic is defined in ScriptFinder [here](https://github.com/opensearch-project/opensearch-build/blob/84f2fa1cf15abe314aee62dbd2cb39bf2c9bb65f/bundle-workflow/src/paths/script_finder.py#L65) 

Action items:

Raise PRs on all plugin repos and remove integtest.sh 
- [ ] index-management
- [ ] anomaly-detection,
- [ ] alerting
- [ ] asynchronous-search
- [ ] k-NN

Changes will need to be backported into 1.x branches if such exist, too.

dblock:
I took the liberty to make PRs in all plugin repos that had an integtest.sh file using meta, see https://github.com/opensearch-project/opensearch-plugins/pull/80/files.

dblock:
- https://github.com/opensearch-project/anomaly-detection/pull/220
- https://github.com/opensearch-project/sql/pull/213
- https://github.com/opensearch-project/alerting/pull/181
- https://github.com/opensearch-project/k-NN/pull/101

dblock:
Okay, this time it's ready.

I used #574: 

```
./bundle-workflow/checkout.sh manifests/1.1.0/opensearch-1.1.0.yml
cd /tmp/tmpco2jldpf
$ find . -name integtest.sh
```

PR #499, opened 2021-09-15
ScriptFinder logic temporary update to use correct integtest.sh 

### LINKED ISSUE #469: Make plugin integtest.sh run against non-snapshot build
The plugin integtest.sh picks up the opensearch version provided in build.gradle, which is 1.1.0-SNAPSHOT. Since the release candidates are non snapshot built artifacts, make this configurable in integ test job

setiah:
ScriptFinder logic temporary update to use correct integtest.sh  until #497 is fixed in plugin repos

Signed-off-by: Himanshu Setia <setiah@amazon.com>
 
### Issues Resolved
#469 
 
### Check List

### DIFF
--- a/bundle-workflow/src/paths/script_finder.py
+++ b/bundle-workflow/src/paths/script_finder.py
@@ -64,8 +64,9 @@ class ScriptFinder:
     @classmethod
     def find_integ_test_script(cls, component_name, git_dir):
         paths = [
-            os.path.realpath(os.path.join(git_dir, "integtest.sh")),
-            os.path.realpath(os.path.join(git_dir, "scripts/integtest.sh")),
+            # TODO: Uncomment this after the integtest.sh tool is removed from plugin repos. See issue #497
+            # os.path.realpath(os.path.join(git_dir, "integtest.sh")),
+            # os.path.realpath(os.path.join(git_dir, "scripts/integtest.sh")),
             os.path.realpath(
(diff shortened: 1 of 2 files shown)


bundle-workflow/src/paths/script_finder.py
```diff
@@ -64,8 +64,9 @@ def find_build_script(cls, component_name, git_dir):
     @classmethod
     def find_integ_test_script(cls, component_name, git_dir):
         paths = [
-            os.path.realpath(os.path.join(git_dir, "integtest.sh")),
-            os.path.realpath(os.path.join(git_dir, "scripts/integtest.sh")),
+            # TODO: Uncomment this after the integtest.sh tool is removed from plugin repos. See issue #497
+            # os.path.realpath(os.path.join(git_dir, "integtest.sh")),
+            # os.path.realpath(os.path.join(git_dir, "scripts/integtest.sh")),
```
dblock, line 69: Are there any repos doing this? If not I would uncomment it, this way repos can actually customize the integtest.sh script. And remove the option of having it at the root permanently.

dblock (approved):

saratvemulapalli (approved):

bundle-workflow/src/paths/script_finder.py
setiah: yeah, the 5 mentioned in issue #497  are doing this. The idea was to unblock integration testing, while we get these resolved in parallel, inside the mentioned plugin repos.

PR #509, opened 2021-09-16
Enforce manifest schema.

dblock:
### Description

Check schema for manifests. Replaces #470.  
 
### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-build/pull/485.
 
### Check List

### DIFF
--- a/bundle-workflow/src/build_workflow/build_recorder.py
+++ b/bundle-workflow/src/build_workflow/build_recorder.py
@@ -52,7 +52,7 @@ class BuildRecorder:
     def write_manifest(self):
         manifest_path = os.path.join(self.target.output_dir, "manifest.yml")
         self.get_manifest().to_file(manifest_path)
-        logging.info(f'Created build manifest {manifest_path}')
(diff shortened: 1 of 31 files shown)


bundle-workflow/src/build_workflow/build_recorder.py
```diff
@@ -68,11 +68,8 @@ def __init__(self, target):
             self.data["build"]["name"] = target.name
             self.data["build"]["version"] = target.opensearch_version
             self.data["build"]["architecture"] = target.arch
-            self.data["build"]["snapshot"] = str(target.snapshot).lower()
```
dblock, line 71: This is in none of our manifests, checked in the latest build. Not sure what's going on, I would expect the field to be written into them, but I didn't include it in the schema and we can put it back if we need it another time.

setiah:
LGTM!

setiah (approved):

saratvemulapalli (approved):
I like this! Thanks @dblock !

peternied (approved):

Issue #473, opened 2021-09-14
Fix integtests to work with multiple plugins 


setiah:
Currently, the integ test instrumentation works fine when running for one component and all possible configurations but it fails when multiple components are tested as part of the same workflow. The instrumentation logic needs to be fixed to handle that.

- [ ] Rework instrumentation logic
- [ ] Fix `alerting`, `reports-scheduler`, `notebooks` integ test instrumentation

Steps to reproduce -
1. Run integration tests on main branch
`bash test.sh integ-test --test-run-id <execution-id> --s3-bucket <bucket_name> --opensearch-version <version> --build-id <id> --architecture <arch>`

Issue signature -

minalsha:
@setiah  please add details with repro steps and what needs to be done. Thanks

owaiskazi19:
`reports-scheduler`, `notebooks` will be fixed as a part of https://github.com/opensearch-project/opensearch-build/issues/476

Issue #500, opened 2021-09-16
[BUG] Integration Tests are unable to start a local cluster 


saratvemulapalli:
**Describe the bug**
Trying to run integration tests on Jenkins ENV. The tests are unable to spin up a local cluster which is healthy.

**To Reproduce**
- Clone this respository
- `./bundle-workflow/test.sh integ-test --s3-bucket <s3_artifact_bucket> --opensearch-version 1.1.0 --build-id 32 --architecture x64 --test-run-id 1`

```
2021-09-16 00:04:50 INFO     Checked out https://github.com/opensearch-project/index-management.git@9450269ca1cf2757fec670f32d251c192c7d5808 into /tmp/tmp6izq2xxd/index-management at 9450269ca1cf2757fec670f32d251c192c7d5808
‘job-scheduler’ -> ‘index-management/job-scheduler’
2021-09-16 00:05:38 INFO     Creating local test cluster in /tmp/tmp6izq2xxd/local-test-cluster
2021-09-16 00:05:38 INFO     Downloading bundle from s3
2021-09-16 00:05:42 INFO     Downloaded bundle to /tmp/tmp6izq2xxd/local-test-cluster/opensearch-1.1.0-linux-x64.tar.gz
2021-09-16 00:05:42 INFO     Unpacking
2021-09-16 00:05:47 INFO     Unpacked
2021-09-16 00:05:47 INFO     Started OpenSearch with parent PID 18506
2021-09-16 00:05:47 INFO     Waiting for service to become available
2021-09-16 00:05:47 INFO     Pinging https://localhost:9200/_cluster/health attempt 0
2021-09-16 00:05:47 INFO     Service not available yet
2021-09-16 00:05:57 INFO     Pinging https://localhost:9200/_cluster/health attempt 1
2021-09-16 00:05:57 INFO     Service not available yet
2021-09-16 00:06:07 INFO     Pinging https://localhost:9200/_cluster/health attempt 2
2021-09-16 00:06:07 INFO     Service not available yet
2021-09-16 00:06:17 INFO     Pinging https://localhost:9200/_cluster/health attempt 3
2021-09-16 00:06:17 INFO     Service not available yet
2021-09-16 00:06:27 INFO     Pinging https://localhost:9200/_cluster/health attempt 4
2021-09-16 00:06:27 INFO     Service not available yet
2021-09-16 00:06:37 INFO     Pinging https://localhost:9200/_cluster/health attempt 5
[...]

PR #512, opened 2021-09-16
Fixed Integ Test To Work With Multiple Plugin and Cluster Status

owaiskazi19:
### Description
Fixed integ test to work for multiple plugin and removed the green status check for local test cluster
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/473 and https://github.com/opensearch-project/opensearch-build/issues/500
 
### Check List

### DIFF
--- a/bundle-workflow/src/run_integ_test.py
+++ b/bundle-workflow/src/run_integ_test.py
@@ -48,23 +48,23 @@ def pull_common_dependencies(work_dir, build_manifest):
 # TODO: replace with DependencyProvider - https://github.com/opensearch-project/opensearch-build/issues/283
 def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
     dependencies_dir = os.path.join(work_dir, 'dependencies')
-    os.chdir(dependencies_dir + "/OpenSearch")
+    opensearch_dir = os.path.join(dependencies_dir, 'OpenSearch')
+    os.chdir(opensearch_dir)
     deps_script = os.path.join(
         work_dir,
         "opensearch-build/tools/standard-test/integtest_dependencies_opensearch.sh",
     )
-    subprocess.run(
+    subprocess.check_output(
         f"{deps_script} opensearch {manifest_build_ver}",
-        shell=True,
-        check=True,
-        capture_output=True,
+        cwd=opensearch_dir,
+        shell=True
     )
-    os.chdir(dependencies_dir + "/common-utils")
-    subprocess.run(
+    common_utils_dir = os.path.join(dependencies_dir, 'common-utils')
+    os.chdir(common_utils_dir)
+    subprocess.check_output(
         f"{deps_script} common-utils {manifest_build_ver}",
-        shell=True,
-        check=True,
-        capture_output=True,
+        cwd=common_utils_dir,
+        shell=True
     )
 
 
--- a/bundle-workflow/src/run_integ_test.py
+++ b/bundle-workflow/src/run_integ_test.py
@@ -41,13 +41,14 @@ def pull_common_dependencies(work_dir, build_manifest):
             GitRepository(
                 component.repository,
                 component.commit_id,
-                os.path.join(work_dir, component.name),
+                os.path.join(work_dir, 'dependencies', component.name),
             )
 
 
 # TODO: replace with DependencyProvider - https://github.com/opensearch-project/opensearch-build/issues/283
 def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
-    os.chdir(work_dir + "/OpenSearch")
+    dependencies_dir = os.path.join(work_dir, 'dependencies')
+    os.chdir(dependencies_dir + "/OpenSearch")
     deps_script = os.path.join(
         work_dir,
         "opensearch-build/tools/standard-test/integtest_dependencies_opensearch.sh",
@@ -58,7 +59,7 @@ def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
         check=True,
         capture_output=True,
     )
-    os.chdir(work_dir + "/common-utils")
+    os.chdir(dependencies_dir + "/common-utils")
     subprocess.run(
         f"{deps_script} common-utils {manifest_build_ver}",
         shell=True,
--- a/bundle-workflow/src/test_workflow/config/test_manifest.yml
+++ b/bundle-workflow/src/test_workflow/config/test_manifest.yml
@@ -15,8 +15,6 @@ components:
   - name: anomaly-detection
     integ-test:
       dependencies:
-        - job-scheduler
-        - alerting
       test-configs:
         - with-security
         - without-security
@@ -28,8 +26,6 @@ components:
   - name: asynchronous-search
     integ-test:
       dependencies:
-        - job-scheduler
-        - alerting
       test-configs:
         - with-security
         - without-security
@@ -40,8 +36,6 @@ components:
   - name: sql
     integ-test:
       dependencies:
-        - job-scheduler
-        - alerting
       test-configs:
         - with-security
         - without-security
@@ -53,14 +47,12 @@ components:
   - name: k-NN
     integ-test:
       dependencies:
-        - job-scheduler
-        - alerting
       test-configs:
         - with-security
         - without-security
     bwc-test:
       dependencies:
       test-configs:
-       - with-security
+        - with-security
         - without-security
 schema-version: '1.0'
--- a/bundle-workflow/src/test_workflow/config/test_manifest.yml
+++ b/bundle-workflow/src/test_workflow/config/test_manifest.yml
@@ -1,9 +1,6 @@
 components:
   - name: index-management
     integ-test:
-      dependencies:
-        - job-scheduler
-        - alerting
       test-configs:
         - with-security
         - without-security
@@ -14,7 +11,6 @@ components:
         - without-security
   - name: anomaly-detection
     integ-test:
-      dependencies:
       test-configs:
         - with-security
         - without-security
@@ -25,7 +21,6 @@ components:
         - without-security
   - name: asynchronous-search
     integ-test:
-      dependencies:
       test-configs:
         - with-security
         - without-security
@@ -35,7 +30,6 @@ components:
         - without-security
   - name: sql
     integ-test:
-      dependencies:
       test-configs:
         - with-security
         - without-security
@@ -46,7 +40,6 @@ components:
         - without-security
   - name: k-NN
     integ-test:
-      dependencies:
       test-configs:
         - with-security
         - without-security
--- a/bundle-workflow/src/test_workflow/config/test_manifest.yml
+++ b/bundle-workflow/src/test_workflow/config/test_manifest.yml
@@ -1,17 +1,17 @@
 components:
-  # - name: index-management
-  #   integ-test:
-  #     dependencies:
-  #       - job-scheduler
-  #       - alerting
-  #     test-configs:
-  #       - with-security
-  #       - without-security
-  #   bwc-test:
-  #     dependencies:
-  #     test-configs:
-  #       - with-security
-  #       - without-security
+  - name: index-management
+    integ-test:
+      dependencies:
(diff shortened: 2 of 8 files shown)


bundle-workflow/src/test_workflow/config/test_manifest.yml
```diff
@@ -12,4 +12,56 @@ components:
...
+        - job-scheduler
+        - alerting
+      test-configs:
+        - with-security
+        - without-security
+    bwc-test:
+      dependencies:
+      test-configs:
+       - with-security
+        - without-security
```
saratvemulapalli, line 66: nit: indentation
```suggestion
        - with-security
        - without-security
```

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -92,7 +92,7 @@ def wait_for_service(self):
                 logging.info(f"Pinging {url} attempt {attempt}")
                 response = requests.get(url, verify=False, auth=("admin", "admin"))
                 logging.info(f"{response.status_code}: {response.text}")
-                if response.status_code == 200 and '"status":"green"' in response.text:
```
saratvemulapalli, line 95: Check if the cluster is yellow/green. 
We do not want to proceed if the cluster is red.

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -92,7 +92,7 @@ def wait_for_service(self):
...
+                if response.status_code == 200:
                     logging.info("Cluster is green")
```
saratvemulapalli, line 96: Please fix this comment and other places within this code where "green" is used.

owaiskazi19: That's right should fix this. Any suggestion you got for the comment?

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
owaiskazi19: I suppose the `status_code` will be 200 only when the cluster is green/yellow. Correct me if I am wrong here

bundle-workflow/src/test_workflow/config/test_manifest.yml
```diff
@@ -12,4 +12,56 @@ components:
       test-configs:
         - with-security
         - without-security
+  - name: anomaly-detection
+    integ-test:
+      dependencies:
+        - job-scheduler
+        - alerting
```
VachaShah, line 19: `job-scheduler` and `alerting` won't be dependencies for all plugins. Since we are not using this object yet, may be you could keep empty for all other plugins or add the correct dependencies for each plugin to avoid confusion.

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
setiah: it can be 200 and red

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
setiah: Cluster is green --> Service is available

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
saratvemulapalli: Correct status code will be 200, just verify `response.text` has green/yellow.

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
saratvemulapalli: May be "Cluster is healthy"/"Cluster is ready for the test".

bundle-workflow/src/test_workflow/config/test_manifest.yml
```diff
@@ -12,4 +12,56 @@ components:
...
+    bwc-test:
+      dependencies:
+      test-configs:
+        - with-security
+        - without-security
+  - name: sql
+    integ-test:
+      dependencies:
+        - job-scheduler
+        - alerting
```
setiah, line 45: (outside of scope of this PR) we should remove these `dependencies` in a separate PR and remove them from test_manifest schema. This will not be used.

bundle-workflow/src/test_workflow/config/test_manifest.yml
setiah: I think we should just remove them to keep things simple https://github.com/opensearch-project/opensearch-build/pull/512#discussion_r710581526. The dependency management logic needs to be handled via maven eventually.

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
owaiskazi19: Changed the message to `Service is available`

bundle-workflow/src/test_workflow/config/test_manifest.yml
owaiskazi19: Removed the `dependencies` and removed from schema as well

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
owaiskazi19: Changed the message to `Service is available`

bundle-workflow/src/test_workflow/config/test_manifest.yml
```diff
@@ -1,13 +1,51 @@
...
+      test-configs:
+        - with-security
+        - without-security
+  - name: asynchronous-search
+    integ-test:
+      test-configs:
+        - with-security
+        - without-security
+    bwc-test:
+      dependencies:
```
VachaShah, line 28: *Nit*: This section can be removed as well. You can do it as part of another PR though. Don't want to block the approval for it.

owaiskazi19: If bwc-test doesn't uses `dependencies`. I can remove it

VachaShah (dismissed):

bundle-workflow/src/run_integ_test.py
```diff
@@ -41,13 +41,14 @@ def pull_common_dependencies(work_dir, build_manifest):
...
-                os.path.join(work_dir, component.name),
+                os.path.join(work_dir, 'dependencies', component.name),
             )
 
 
 # TODO: replace with DependencyProvider - https://github.com/opensearch-project/opensearch-build/issues/283
 def sync_dependencies_to_maven_local(work_dir, manifest_build_ver):
-    os.chdir(work_dir + "/OpenSearch")
+    dependencies_dir = os.path.join(work_dir, 'dependencies')
+    os.chdir(dependencies_dir + "/OpenSearch")
```
dblock, line 51: Change `run` to `check_output`, which incidentally supports `cwd=` for the directory in which it runs.

dblock (commented):
Why are we removing `dependencies`? I thought that was a pretty good idea to keep things more data-driven/configurable. An alternative could be a custom script found via ScriptFinder that gets executed for a component that needs something. Either way hard-coding copying plugin files to hack our way to a working system doesn't seem like a great idea - we'll have to fix it again for OpenSearch Dashboards.

bundle-workflow/src/run_integ_test.py
setiah: +1

setiah:
> Why are we removing `dependencies`? I thought that was a pretty good idea to keep things more data-driven/configurable. An alternative could be a custom script found via ScriptFinder that gets executed for a component that needs something. Either way hard-coding copying plugin files to hack our way to a working system doesn't seem like a great idea - we'll have to fix it again for OpenSearch Dashboards.

Hey @dblock, the dependencies defined in the config file are not being used today. They were planned to be used by DependencyInstaller, which is wip. So it will be added back later, in the right format, w/ dependency installer once that is integrated.

setiah:
> > Why are we removing `dependencies`? I thought that was a pretty good idea to keep things more data-driven/configurable. An alternative could be a custom script found via ScriptFinder that gets executed for a component that needs something. Either way hard-coding copying plugin files to hack our way to a working system doesn't seem like a great idea - we'll have to fix it again for OpenSearch Dashboards.
> 
> Hey @dblock, the dependencies defined in the config file are not being used today. They were planned to be used by DependencyInstaller, which is wip. So it will be added back later, in the right format, w/ dependency installer once that is integrated.

A better way IMO would be if we can publish these maven artifacts to an internal s3 repo, the dependency management logic for plugins can be handled by gradle itself. This would be a much better solution given the plugin dependencies will keep evolving and the build repo will not be able to keep up. Having a consistent dependency management solution via maven should be the right way in the long run

dblock:
@setiah I would just keep the manifest that lists dependencies instead of hard-coding the list in code. The code itself can loop through dependencies and run whatever command it needs to publish these/build/whatever. This is what's done in manifest checks.

setiah:
> @setiah I would just keep the manifest that lists dependencies instead of hard-coding the list in code. The code itself can loop through dependencies and run whatever command it needs to publish these/build/whatever. This is what's done in manifest checks.

Had a quick chat with @dblock on this. The PR looks good to merge. In the subsequent revisions, we would change the dependency management logic to copy all maven dependencies from s3 to maven local, instead of copying plugin specific dependencies. This would simplify the logic further.

In future, a good optimization would be to make the maven path configurable in the integtest.sh. This would allow using a custom maven repo while running integ tests and automatically manage dependencies for a plugin with the help of gradle.

owaiskazi19:
@dblock @setiah I have added back the `dependencies` as it's not the part of this PR. If it's required we can create a separate issue and address it in the future PR.

owaiskazi19:
> > @setiah I would just keep the manifest that lists dependencies instead of hard-coding the list in code. The code itself can loop through dependencies and run whatever command it needs to publish these/build/whatever. This is what's done in manifest checks.
> 
> Had a quick chat with @dblock on this. The PR looks good to merge. In the subsequent revisions, we would change the dependency management logic to copy all maven dependencies from s3 to maven local, instead of copying plugin specific dependencies. This would simplify the logic further.
> 
> In future, a good optimization would be to make the maven path configurable in the integtest.sh. This would allow using a custom maven repo while running integ tests and automatically manage dependencies for a plugin with the help of gradle.

I missed this. I'll address it in the next PR

bundle-workflow/src/test_workflow/config/test_manifest.yml
```diff
@@ -8,6 +8,69 @@ components:
...
+        - job-scheduler
+        - alerting
+      test-configs:
+        - with-security
+        - without-security
+  - name: anomaly-detection
+    integ-test:
+      dependencies:
+        - job-scheduler
+        - alerting
```
saratvemulapalli, line 21: Why are all the plugins dependent on alerting and job-scheduler?

owaiskazi19: As per the discussion https://github.com/opensearch-project/opensearch-build/pull/512#issuecomment-921975871.  I'll address it in the next PR.

saratvemulapalli: Sure. I am fine addressing this in a quick follow up PR. If so please create an issue so we have a track.

But always, as best practices: 
- Don't commit code we are not planning to use.
- Commit incorrect configuration which is un-usable.

saratvemulapalli (approved):
Dropped a comment, other than that rest of the changes look good to me.
Thanks @owaiskazi19 !

bundle-workflow/src/test_workflow/config/test_manifest.yml
owaiskazi19: Yeah, that makes sense. Thanks for the new learning. I have create an issue to track this down: https://github.com/opensearch-project/opensearch-build/issues/521

VachaShah (approved):
LGTM!

PR #516, opened 2021-09-17
Handle child processes

### LINKED ISSUE #429: [BUG] LocalTestCluster fails during create and destroy
**Describe the bug**
The LocalTestCluster is not working with recent opensearch tarballs and fails to launch the OpenSearch process successfully. There are a couple of sub-issues here

1. The OpenSearch process is not setup properly.
2. The PID of the OpenSearch process seems to be reported incorrectly. It is the pid of shell subprocess that starts OpenSearch process. 
2. Missing logs to debug the opensearch process setup failures. The opensearch-tar-install.sh script logs are not available in stdout for debugging failure.

**To Reproduce**
1. Stub the download() method and provide a pre-downloaded and uncompressed tarball. Then use LocalTestCluster `create()` method to spin the cluster. It is the `create()` method where problem is seen.

**Expected behavior**
OpenSearch cluster should setup correctly and querying on port 9200 should respond. The opensearch logs should be reported on stdout for debugging.

**Plugins**
Full bundle - ran test with security.

**Screenshots**

**Host/Environment (please complete the following information):**
 - OS: Linux
 - Version = 1.1.0

gaiksaya:

### Description
The PID of the actual opensearch process was different from parent PID in ubuntu. 
This change checks for the child processes and kills them before terminating the actual PID.
In case of AL2 and MACOS, both actual process and PID are same. 
 
### Issues Resolved
#429 
 
### Check List

### DIFF
--- a/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
+++ b/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
@@ -93,7 +93,7 @@ class LocalTestCluster(TestCluster):
                 logging.info(f"Pinging {url} attempt {attempt}")
                 response = requests.get(url, verify=False, auth=("admin", "admin"))
                 logging.info(f"{response.status_code}: {response.text}")
-                if response.status_code == 200:
+                if response.status_code == 200 and '"status":"green"' in response.text:
                     logging.info("Cluster is green")
                     return
             except requests.exceptions.ConnectionError:
--- a/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
+++ b/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
@@ -9,7 +9,7 @@ import os
 import subprocess
 import time
 
-import psutil
+import psutil  # type: ignore
 import requests
 
 from aws.s3_bucket import S3Bucket
--- a/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
+++ b/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
@@ -103,12 +103,12 @@ class LocalTestCluster(TestCluster):
 
     def terminate_process(self):
         parent = psutil.Process(self.process.pid)
-        logging.info("Checking for child processes")
+        logging.debug("Checking for child processes")
         child_processes = parent.children(recursive=True)
         for child in child_processes:
-            logging.info(f"Found child process with pid {child.pid}")
+            logging.debug(f"Found child process with pid {child.pid}")
             if child.pid != self.process.pid:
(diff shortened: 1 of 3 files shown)


saratvemulapalli (dismissed):
Looks good to me. Thanks @gaiksaya !

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -9,6 +9,7 @@
 import subprocess
 import time
 
+import psutil  # type: ignore
```
saratvemulapalli, line 12: Curious why you are trying to ignore?

dblock: these modules come without types and pypy is complaining

dblock (dismissed):

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -101,6 +102,14 @@ def wait_for_service(self):
         raise ClusterCreationException("Cluster is not green after 10 attempts")
 
     def terminate_process(self):
+        parent = psutil.Process(self.process.pid)
+        logging.info("Checking for child processes")
```
dblock, line 106: Should probably be debug.

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -101,6 +102,14 @@ def wait_for_service(self):
...
+        child_processes = parent.children(recursive=True)
+        for child in child_processes:
+            logging.info(f"Found child process with pid {child.pid}")
+            if child.pid != self.process.pid:
+                logging.info(f"Sending SIGKILL to {child.pid} ")
```
dblock, line 111: Same, debug. These don't add much value in the output.

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
gaiksaya: Changed to debug

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
gaiksaya: Changed to debug

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
saratvemulapalli: Interesting..

saratvemulapalli (approved):

peternied (approved):

Issue #437, opened 2021-09-10
[BUG] Missing CI checks for various components


dblock:
**Describe the bug**

After https://github.com/opensearch-project/opensearch-build/pull/414 we're missing some CI checks for the following.

- [x] alerting is missing a dependencies check, need to support a custom gradle command to run to get those or teach alerting to respond to `gradle plugin:dependencies`
- [x] security is missing all checks, need maven versions of checks; it also can't properly support snapshot versions because we hack those in https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/scripts/components/security/build.sh#L66
- [x] performance-analyzer-rca is missing a dependencies check, need to support a custom gradle command to run to get those or teach alerting to respond to `gradle plugin:dependencies`

dblock:
Looks like neither performance-analyzer-rca, nor security depend on OpenSearch proper. For maven dependencies I started coding that in https://github.com/dblock/opensearch-build/commit/6d7a25e3ba8e2659b49053b731d60717a387f33c, but it's not needed ATM.

dblock:
Closing after https://github.com/opensearch-project/opensearch-build/pull/502

PR #519, opened 2021-09-17
Add checks to OpenSearch in auto-generated manifests.

dblock:

### Description

In https://github.com/opensearch-project/opensearch-build/pull/514 the generated manifests lacked basic checks. This adds version and build checks that ensures that the OpenSearch core builds successfully. 

I didn't add any automatic checks to the plugins because those aren't always the same (e.g. security doesn't have a version check because it uses maven that only builds one of the flavors of snapshot vs. non-snapshot and therefore we cannot check version without hacks), and because those would fail without first building OpenSearch.
  
### Check List

### DIFF
--- a/bundle-workflow/src/manifests_workflow/component.py
+++ b/bundle-workflow/src/manifests_workflow/component.py
@@ -4,15 +4,28 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+from manifests.manifest import Manifest
+
 
 class Component:
-    def __init__(self, name, repo, snapshot=False):
+    def __init__(self, name, repo, snapshot=False, checks=[]):
         self.name = name
         self.git_repo = repo
         self.snapshot = snapshot
+        self.checks = checks
 
     @classmethod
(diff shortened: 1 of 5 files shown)


saratvemulapalli (approved):
I understand we'd like to start with OpenSearch but we should add support to all plugins.
As called out in the description we could leave out a few to start with but most of the plugins should be do-able.

dblock:
> I understand we'd like to start with OpenSearch but we should add support to all plugins.
> As called out in the description we could leave out a few to start with but most of the plugins should be do-able.

We have support for all but 2 plugins (performance-analyzer-rca and security), see #437 for details.

This PR adds checks automatically. Because plugins require OpenSearch to first build and publish to maven local, the checks won't work unless OpenSearch is already present in the manifest. If a component increments a version, it will cause a new manifest to be generated without OpenSearch (it has not incremented its version yet). So the auto-PR will break 100% of the time, hence I didn't add checks to components.

VachaShah (approved):

PR #529, opened 2021-09-20
Added support to run tests for plugins with nested directory

### LINKED ISSUE #476: [META] Plugins Integration Testing for Test Automation Surge
Run the Integration Test cases in the OpenSearch distribution building workflow for the plugins

Successful:
|Plugin Name	|Result|
|---	|---	|
|AD Plugin	|<ul><li>- [x] With security</li><li>- [x] without security</li></ul>	|<ul><li>- [x] With security</li><li>- [x] without security</li></ul>	|<ul><li>- [x] </li></ul>	|
|SQL Plugin	|<ul><li>- [x] With security</li><li>- [x] without security</li></ul>	|<ul><li>- [x] With security</li><li>- [x] without security</li></ul>	|<ul><li>- [x] </li></ul>	|
|KNN Plugin	|<ul><li>- [x] With security</li><li>- [x] without security</li></ul>	|<ul><li>- [x] With security</li><li>- [x] without security</li></ul>	|<ul><li>- [x] </li></ul>	|
|Async-search	|<ul><li>- [x] With security</li><li>- [x] without security</li></ul>	|<ul><li>- [ ] With security</li><li>- [x] without security</li></ul>	|<ul><li>- [x] </li></ul>	|

Changes required:
|Plugin Name	|Result| Issue| Fix
|---	|---	|--- |---
|Alerting Plugin	|<ul><li>- [ ] With security</li><li>- [ ] without security</li></ul>	| Since we are downloading alerting plugin as a common dependency for other plugins, the logic to run integ test needs to be changed to run alerting as an individual plugin	| Fixed in https://github.com/opensearch-project/opensearch-build/pull/512
|Notebooks	|<ul><li>- [ ] With security</li><li>- [ ] without security</li></ul>	| Not able to find `/gradlew` to run integ test as the plugin resides in https://github.com/opensearch-project/dashboards-notebooks/tree/main/opensearch-notebooks	| Fixed in #529 
|Reports Scheduler	|<ul><li>- [ ] With security</li><li>- [ ] without security</li></ul>	| Not able to find `/gradlew` to run integ test as the plugin resides in https://github.com/opensearch-project/dashboards-reports/tree/main/reports-scheduler	| Fixed in #529

Reference for Plugins: https://github.com/opensearch-project/opensearch-build/issues/58

VachaShah:
Signed-off-by: Vacha <vachshah@amazon.com>

### Description
Adding support to run tests for plugins with nested directory like `dashboards-reports` and `dashboards-notebooks`.
 
### Issues Resolved
#476 
 
### Check List

### DIFF
--- a/bundle-workflow/scripts/default/integtest.sh
+++ b/bundle-workflow/scripts/default/integtest.sh
@@ -18,12 +18,11 @@ function usage() {
     echo -e "-c CREDENTIAL\t(usename:password), no defaults, effective when SECURITY_ENABLED=true."
     echo -e "-v OPENSEARCH_VERSION\t, no defaults"
     echo -e "-n SNAPSHOT\t, defaults to false"
-    echo -e "-d DIRECTORY\t, defaults to running directory. Some repos have more than 1 directory/component, specify the working directory to separate tests."
     echo -e "-h\tPrint this message."
     echo "--------------------------------------------------------------------------"
 }
 
-while getopts ":hb:p:s:c:v:n:d:" arg; do
+while getopts ":hb:p:s:c:v:n:" arg; do
     case $arg in
         h)
             usage
@@ -46,9 +45,6 @@ while getopts ":hb:p:s:c:v:n:d:" arg; do
             ;;
         n)
             SNAPSHOT=$OPTARG
-            ;;
-        d)
-            DIRECTORY=$OPTARG
             ;;
         :)
             echo "-${OPTARG} requires an argument"
@@ -90,8 +86,4 @@ then
   PASSWORD=`echo $CREDENTIAL | awk -F ':' '{print $2}'`
 fi
 
-if [ -n "$DIRECTORY" ]
-then
-  cd $DIRECTORY
-fi
 ./gradlew integTest -Dopensearch.version=$OPENSEARCH_VERSION -Dbuild.snapshot=$SNAPSHOT -Dtests.rest.cluster="$BIND_ADDRESS:$BIND_PORT" -Dtests.cluster="$BIND_ADDRESS:$BIND_PORT" -Dtests.clustername="opensearch-integrationtest" -Dhttps=$SECURITY_ENABLED -Duser=$USERNAME -Dpassword=$PASSWORD --console=plain
--- a/bundle-workflow/scripts/default/integtest.sh
+++ b/bundle-workflow/scripts/default/integtest.sh
@@ -18,11 +18,12 @@ function usage() {
     echo -e "-c CREDENTIAL\t(usename:password), no defaults, effective when SECURITY_ENABLED=true."
     echo -e "-v OPENSEARCH_VERSION\t, no defaults"
     echo -e "-n SNAPSHOT\t, defaults to false"
+    echo -e "-d DIRECTORY\t, defaults to running directory. Some repos have more than 1 directory/component, specify the working directory to separate tests."
     echo -e "-h\tPrint this message."
     echo "--------------------------------------------------------------------------"
 }
 
-while getopts ":hb:p:s:c:v:n" arg; do
+while getopts ":hb:p:s:c:v:n:d:" arg; do
     case $arg in
         h)
             usage
@@ -45,6 +46,9 @@ while getopts ":hb:p:s:c:v:n" arg; do
             ;;
         n)
             SNAPSHOT=$OPTARG
+            ;;
+        d)
+            DIRECTORY=$OPTARG
             ;;
         :)
             echo "-${OPTARG} requires an argument"
@@ -86,4 +90,8 @@ then
   PASSWORD=`echo $CREDENTIAL | awk -F ':' '{print $2}'`
 fi
 
+if [ -n "$DIRECTORY" ]
+then
+  cd $DIRECTORY
+fi
 ./gradlew integTest -Dopensearch.version=$OPENSEARCH_VERSION -Dbuild.snapshot=$SNAPSHOT -Dtests.rest.cluster="$BIND_ADDRESS:$BIND_PORT" -Dtests.cluster="$BIND_ADDRESS:$BIND_PORT" -Dtests.clustername="opensearch-integrationtest" -Dhttps=$SECURITY_ENABLED -Duser=$USERNAME -Dpassword=$PASSWORD --console=plain
--- a/bundle-workflow/src/manifests/test_manifest.py
+++ b/bundle-workflow/src/manifests/test_manifest.py
@@ -15,6 +15,7 @@ class TestManifest(Manifest):
         schema-version: '1.0'
         components:
           - name: index-management
+            working-directory: optional relative directory to run commands in
             integ-test:
               test-configs:
                 - with-security
@@ -34,6 +35,7 @@ class TestManifest(Manifest):
                 "type": "dict",
                 "schema": {
                     "name": {"required": True, "type": "string"},
+                    "working-directory": {"type": "string"},
                     "integ-test": {
                         "type": "dict",
                         "schema": {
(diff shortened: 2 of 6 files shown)


bundle-workflow/scripts/default/integtest.sh
```diff
@@ -86,4 +90,8 @@ then
   PASSWORD=`echo $CREDENTIAL | awk -F ':' '{print $2}'`
 fi
 
+if [ -n "$DIRECTORY" ]
+then
+  cd $DIRECTORY
+fi
```
setiah, line 96: we should not be changing directory inside integtest.sh. It can be handled outside of it. In the subprocess command to run shell script integtest.sh, you can specify what is the working directory to run the command from.

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -88,7 +88,11 @@ def _execute_integtest_sh(self, endpoint, port, security):
             self.component.name, self.repo.dir
         )
         if os.path.exists(script):
-            cmd = f"{script} -b {endpoint} -p {port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
+            cmd = (
+                f"{script} -b {endpoint} -p {port} -s {str(security).lower()} -v {self.bundle_manifest.build.version} -d {self.test_config.working_directory}"
```
owaiskazi19, line 92: We can have `cmd` pulled out as it's common around `if` and `else`. Append `-d` based on the check later

bundle-workflow/src/test_workflow/config/test_manifest.yml
```diff
@@ -44,4 +44,20 @@ components:
       test-configs:
         - with-security
         - without-security
+  - name: dashboards-reports
+    working-directory: reports-scheduler
```
owaiskazi19, line 48: Correct me here for the approach. Do we need to specify the working directory? Can't we check for `integtest.sh` file in the plugin repo if not present then change directory to the plugin directory where `integtest.sh` is present?

bundle-workflow/scripts/default/integtest.sh
```diff
@@ -18,11 +18,12 @@ function usage() {
     echo -e "-c CREDENTIAL\t(usename:password), no defaults, effective when SECURITY_ENABLED=true."
     echo -e "-v OPENSEARCH_VERSION\t, no defaults"
     echo -e "-n SNAPSHOT\t, defaults to false"
+    echo -e "-d DIRECTORY\t, defaults to running directory. Some repos have more than 1 directory/component, specify the working directory to separate tests."
```
owaiskazi19, line 21: Can we change the print statement to `Few repos have nested directory structure.  Need to specify the working directory to run integration tests.`

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -88,7 +88,11 @@ def _execute_integtest_sh(self, endpoint, port, security):
             self.component.name, self.repo.dir
         )
         if os.path.exists(script):
-            cmd = f"{script} -b {endpoint} -p {port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
+            cmd = (
```
setiah, line 91: you can avoid change in `cmd` by passing the right `cwd` in `execute()` method

bundle-workflow/tests/tests_manifests/test_test_manifest.py
```diff
@@ -14,6 +14,7 @@
 
 class TestTestManifest(unittest.TestCase):
     def setUp(self):
+        self.maxDiff = None
```
owaiskazi19, line 17: Noop?

bundle-workflow/src/manifests/test_manifest.py
```diff
@@ -15,6 +15,7 @@ class TestManifest(Manifest):
         schema-version: '1.0'
         components:
           - name: index-management
+            working-directory: optional relative directory to run commands in
```
setiah, line 18: can this be an optional parameter, so only outlier plugins have this in their configuration?

VachaShah: It is an optional plugin, only the outlier plugins have this specified, for example: https://github.com/opensearch-project/opensearch-build/pull/529/files#diff-7e054842e270f704c5ec5d3728ffce2dda6533adb3a626822b1ab586aa79acc8R55

bundle-workflow/src/test_workflow/config/test_manifest.yml
VachaShah: The `integtest.sh` are now removed from each plugin, we call the default `integtest.sh` which is present in the build repo. The working-directory is for it to understand where to run the `./gradlew integtest` command.

bundle-workflow/src/manifests/test_manifest.py
setiah: maybe rename this config param to `gradle-path` if that's what this represents.

bundle-workflow/src/test_workflow/config/test_manifest.yml
owaiskazi19: Makes sense. Thanks for the clearing it out.

dblock (changes requested):
Can you achieve the same result with `execute(cmd, os.path.join(self.repo.dir, working_directory), True, False)`, which would be a lot simpler, instead of having to pass it into the `integtest.sh`?

bundle-workflow/tests/tests_manifests/test_test_manifest.py
dblock: This shows a longer diff when tests fail, which is ok to leave IMO. Maybe there's a way to enable it globally ...

bundle-workflow/src/manifests/test_manifest.py
VachaShah: `gradle-path` would be specific to gradle plugins and we won't be able to re-use it for dashboards plugin. WDYT about `testing-directory`?

dblock: We already have `working-directory` for build, so naming-wise that matches and I personally like it.

VachaShah: Thats true, `testing-directory` would not represent it completely. I kept it `working-directory` to keep it consistent with the build side.

setiah: Sounds good, didn't know we use similar name for build. Makes sense to keep it consistent. Thanks!

dblock: It was added as part of https://github.com/opensearch-project/opensearch-build/pull/402.

dblock: If you are reading the input manifest anywhere here, you could just take it out of there.

VachaShah: I don't think we are reading the input manifest here, its just the bundle, build and test manifest.

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -89,7 +89,10 @@ def _execute_integtest_sh(self, endpoint, port, security):
         )
         if os.path.exists(script):
             cmd = f"{script} -b {endpoint} -p {port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
-            (status, stdout, stderr) = execute(cmd, self.repo.dir, True, False)
+            if self.test_config.working_directory is not None:
```
dblock, line 92: Nit: because these are the same operation functionally that differ in parameters I would:

```python
cmd = self.repo.dir
if self.test_config.working_directory:
   cmd = os.path.join(cmd, self.test_config.working_directory)
```

or

```
cmd = os.path.joimn(self.repo.dir, self.test_config.working_directory ) if self.test_config.working_directory else self.repo.dir
```

or, if `os.path.join(..., None)` works just `cmd = os.path.join(self.repo.dir, self.test_config.working_directory)`.

dblock (approved):
Looks clean!

setiah (approved):
LGTM. thanks @VachaShah

owaiskazi19 (approved):
Thanks for the changes @VachaShah !

PR #538, opened 2021-09-21
Support plugin specific config

### LINKED ISSUE #493: Support plugin specific opensearch.yml settings for local_test_cluster
The LocalTestCluster should support plugin specific opensearch settings during the local test cluster bootup. These are required for running tests on plugins. For example, this tool (which was used earlier) had a bunch of additional settings required for integration tests that were added as part of cluster setup https://github.com/opensearch-project/opensearch-build/blob/main/tools/standard-test/deploy_single_tar_cluster.sh.

gaiksaya:

### Description
Adds plugin specific configuration in `opensearch.yml` before the cluster starts
 
### Issues Resolved
#493 
 
### Check List

### DIFF
--- a/bundle-workflow/src/manifests/test_manifest.py
+++ b/bundle-workflow/src/manifests/test_manifest.py
@@ -22,7 +22,7 @@ class TestManifest(Manifest):
                 - without-security
                 - with-less-security
               additional-cluster-configs:
-                - "path.repo : /tmp/tmp_dir"
+                - key : value
             bwc-test:
               test-configs:
                 - with-security
@@ -50,7 +50,7 @@ class TestManifest(Manifest):
                                 ],
                             },
                             "additional-cluster-configs": {
-                                "type": "list",
+                                "type": "dict",
                             },
                         },
                     },
--- a/bundle-workflow/src/manifests/test_manifest.py
+++ b/bundle-workflow/src/manifests/test_manifest.py
@@ -21,6 +21,8 @@ class TestManifest(Manifest):
                 - with-security
                 - without-security
                 - with-less-security
+              additional-cluster-configs:
+                - "path.repo : /tmp/tmp_dir"
             bwc-test:
               test-configs:
                 - with-security
@@ -46,6 +48,9 @@ class TestManifest(Manifest):
                                     "without-security",
                                     "with-less-security",
                                 ],
+                            },
+                            "additional-cluster-configs": {
+                                "type": "list",
                             },
                         },
                     },
--- a/bundle-workflow/src/test_workflow/config/test_manifest.yml
+++ b/bundle-workflow/src/test_workflow/config/test_manifest.yml
@@ -4,6 +4,8 @@ components:
       test-configs:
         - with-security
         - without-security
+      additional-cluster-configs:
+        - "path.repo: /tmp/tmp_dir"
     bwc-test:
       test-configs:
         - with-security
@@ -31,6 +33,8 @@ components:
       test-configs:
         - with-security
         - without-security
+      additional-cluster-configs:
+        - "plugins.destination.host.deny_list: [10.0.0.0/8, 127.0.0.1]"
     bwc-test:
       test-configs:
         - with-security
@@ -40,6 +44,8 @@ components:
       test-configs:
         - with-security
         - without-security
+      additional-cluster-configs:
+        - "script.context.field.max_compilations_rate: 1000/1m"
     bwc-test:
       test-configs:
         - with-security
--- a/bundle-workflow/src/test_workflow/config/test_manifest.yml
+++ b/bundle-workflow/src/test_workflow/config/test_manifest.yml
@@ -4,8 +4,6 @@ components:
       test-configs:
         - with-security
         - without-security
-      additional-cluster-configs:
-        - "path.repo: /tmp/tmp_dir"
     bwc-test:
       test-configs:
         - with-security
@@ -34,7 +32,7 @@ components:
         - with-security
         - without-security
       additional-cluster-configs:
-        - "plugins.destination.host.deny_list: [10.0.0.0/8, 127.0.0.1]"
+        plugins.destination.host.deny_list : [10.0.0.0/8, 127.0.0.1]
     bwc-test:
       test-configs:
         - with-security
@@ -45,7 +43,7 @@ components:
         - with-security
         - without-security
       additional-cluster-configs:
-        - "script.context.field.max_compilations_rate: 1000/1m"
+        script.context.field.max_compilations_rate: 1000/1m
     bwc-test:
       test-configs:
         - with-security
--- a/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
+++ b/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
@@ -22,13 +22,21 @@ class LocalTestCluster(TestCluster):
     Represents an on-box test cluster. This class downloads a bundle (from a BundleManifest) and runs it as a background process.
     """
 
-    def __init__(self, work_dir, bundle_manifest, security_enabled, s3_bucket_name):
+    def __init__(self, work_dir, component_name, bundle_manifest, security_enabled, s3_bucket_name):
         self.manifest = bundle_manifest
         self.work_dir = os.path.join(work_dir, "local-test-cluster")
+        self.tmp_dir = os.path.join(self.work_dir, "tmp_dir")
         os.makedirs(self.work_dir, exist_ok=True)
+        os.makedirs(self.tmp_dir, exist_ok=True)
+        self.component_name = component_name
         self.security_enabled = security_enabled
         self.bucket_name = s3_bucket_name
         self.process = None
+        self.plugin_specific_config = {
+            "alerting": ["plugins.destination.host.deny_list: [10.0.0.0/8,  127.0.0.1]"],
+            "sql": ["script.context.field.max_compilations_rate: 1000/1m"],
+            "index-management": ["path.repo: %s" % os.path.realpath(self.tmp_dir)]
+        }
 
     def create_cluster(self):
         self.download()
@@ -37,6 +45,7 @@ class LocalTestCluster(TestCluster):
         self.install_dir = f"opensearch-{self.manifest.build.version}"
         if not self.security_enabled:
             self.disable_security(self.install_dir)
+        self.__add_plugin_specific_config(self.plugin_specific_config, os.path.join(self.install_dir, "config", "opensearch.yml"))
         self.process = subprocess.Popen(
             "./opensearch-tar-install.sh",
             cwd=self.install_dir,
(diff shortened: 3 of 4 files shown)


dblock (dismissed):
I am reluctantly approving this because it works.

It's a very tiny bit more effort to add a dictionary of config to the test manifest. Please do open an issue to refactor this out.

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -22,13 +22,21 @@ class LocalTestCluster(TestCluster):
...
+        os.makedirs(self.tmp_dir, exist_ok=True)
+        self.component_name = component_name
         self.security_enabled = security_enabled
         self.bucket_name = s3_bucket_name
         self.process = None
+        self.plugin_specific_config = {
+            "alerting": ["plugins.destination.host.deny_list: [10.0.0.0/8,  127.0.0.1]"],
+            "sql": ["script.context.field.max_compilations_rate: 1000/1m"],
+            "index-management": ["path.repo: %s" % os.path.realpath(self.tmp_dir)]
+        }
```
saratvemulapalli, line 39: This should go to test-manifest. Component owners should only have to modify in place and be done with it. 
Having config within the code brings in maintenance for us. This will eventually break when the plugin config changes.

gaiksaya: That's actually much better. Let me refactor it.

gaiksaya:
> It should be minor effort to refactor this. But I would leave it you.
> Also please add unit tests, we just have a new tiny function it should be quick.

Do we have unit tests for integ_test part of the code? I think @setiah is working on it. Will add it once that's done.

dblock:
> > It should be minor effort to refactor this. But I would leave it you.
> > Also please add unit tests, we just have a new tiny function it should be quick.
> 
> Do we have unit tests for integ_test part of the code? I think @setiah is working on it. Will add it once that's done.

That shouldn't stop you, but https://github.com/opensearch-project/opensearch-build/pull/539#pullrequestreview-760049118 needs another approval to get some.

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -84,6 +93,14 @@ def disable_security(self, dir):
             shell=True,
         )
 
+    def __add_plugin_specific_config(self, additional_config: dict, file):
```
setiah, line 96: these extra opensearch settings can be configuration dependent as well going forward, e.g. some settings applicable only for with-security config? I think it would be better if we have this logic outside of `local_test_cluster` and have the test_suite provide these additional settings for spinning up the cluster.

gaiksaya: How would you structure this in is test-manifest?

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
gaiksaya: Done please have a look. Thanks!

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
gaiksaya: Offline conversation: Added the `additional-cluster-configs` key under `integ-test`

bundle-workflow/src/manifests/test_manifest.py
```diff
@@ -21,6 +21,8 @@ class TestManifest(Manifest):
                 - with-security
                 - without-security
                 - with-less-security
+              additional-cluster-configs:
```
dblock, line 24: Maybe introduce a config: key with cluster: and tests: children?

bundle-workflow/src/test_workflow/config/test_manifest.yml
```diff
@@ -4,6 +4,8 @@ components:
       test-configs:
         - with-security
         - without-security
+      additional-cluster-configs:
+        - "path.repo: /tmp/tmp_dir"
```
dblock, line 8: Shouldn’t be relying on hardcoded location of temp. Maybe $TEMP and expand?

bundle-workflow/src/test_workflow/config/test_manifest.yml
```diff
@@ -31,6 +33,8 @@ components:
       test-configs:
         - with-security
         - without-security
+      additional-cluster-configs:
+        - "plugins.destination.host.deny_list: [10.0.0.0/8, 127.0.0.1]"
```
dblock, line 37: Why is this a string and not a dictionary?

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -22,12 +22,17 @@ class LocalTestCluster(TestCluster):
...
     """
 
-    def __init__(self, work_dir, bundle_manifest, security_enabled, s3_bucket_name):
+    def __init__(self, work_dir, component_name, additional_cluster_config, bundle_manifest, security_enabled,
+                 s3_bucket_name=None):
         self.manifest = bundle_manifest
         self.work_dir = os.path.join(work_dir, "local-test-cluster")
         os.makedirs(self.work_dir, exist_ok=True)
+        # Required by IM plugin as an additional config
+        os.makedirs("/tmp/tmp_dir", exist_ok=True)
```
dblock, line 31: This will fail in many ways. What does the IM plug-in actually need?

gaiksaya:
> > > It should be minor effort to refactor this. But I would leave it you.
> > > Also please add unit tests, we just have a new tiny function it should be quick.
> > 
> > 
> > Do we have unit tests for integ_test part of the code? I think @setiah is working on it. Will add it once that's done.
> 
> That shouldn't stop you, but [#539 (review)](https://github.com/opensearch-project/opensearch-build/pull/539#pullrequestreview-760049118) needs another approval to get some.

Talked you @VachaShah she is working on adding UT for entire `local_test_cluster.py` module. Will coordinate with her.

dblock (dismissed):
It’s better than the code we had before, but I have some concerns. These configs should be proper name: value y’all pairs and you can use a PropertyFile to write. And I don’t understand the temp thing.

Don’t want to hold a working system so approving.

bundle-workflow/src/test_workflow/config/test_manifest.yml
gaiksaya: Its a list of strings. Additional settings if required in future can be added. It returns as `['plugins.destination.host.deny_list: [10.0.0.0/8, 127.0.0.1]']` which is added in `opensearch.yml`

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
gaiksaya: They need a valid location [here](https://github.com/opensearch-project/opensearch-build/pull/538/files#diff-b6591330ab14f0b34306021b8744f1e9c93e87a43e90cabee0024d9acb95a61aR25) for one of their integration test

saratvemulapalli, line 31: We should figure out a longer term solution for this. 
This should come from the test-manifest instead of being hard-coding it.

We could do something like: 
```
components:
  - name: index-management
    integ-test:
      test-configs:
        - with-security
        - without-security
      additional-cluster-configs:
        - "path.repo: /tmp/tmp_dir"
      shell-commands:
        - "mkdir /tmp/tmp_dir"
```

saratvemulapalli (dismissed):
Thanks @gaiksaya for taking care of this.

Please open up issues to track unit tests (looks like that might come in another PR) and long term fix for executing custom shell commands for each test component.

bundle-workflow/src/test_workflow/config/test_manifest.yml
dblock: So these are YAML fragments that go into another yaml file. This should look like this:

```yaml
config:
   tests:
     - with-security
     - without-security
   cluster:
     - plugins.detination.host.deny_list: [ 10.0.0.0/8, 127.0.0.1 ]
  ```

File: bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py (hunk 6 above)
Comment (dblock) [reply bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py:31]: The `path.repo` option feels wrong to me altogether, we shouldn't be over-engineering around some strange test. The test itself can create any paths it wants. What does it expect inside `path.repo`?

Comment (gaiksaya) [reply bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py:31]: They write some stuff into that directory. @bowenlan-amzn might be able to answer in detail. I have opened an issue: https://github.com/opensearch-project/opensearch-build/issues/545

Comment (gaiksaya) [reply bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py:31]: Created the issue: https://github.com/opensearch-project/opensearch-build/issues/545

File: bundle-workflow/src/manifests/test_manifest.py (hunk 3 above)
Comment (dblock) [reply bundle-workflow/src/manifests/test_manifest.py:24]: I'm ok with issues being opened against this.

Comment (dblock) [reply bundle-workflow/src/manifests/test_manifest.py:24]: ... but while I am OK with moving forward adding this config _is_ the entire point of this PR. So we shouldn't be acquiring as much technical debt here.

File: bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py (hunk 6 above)
Comment (dblock) [reply bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py:31]: Please open an issue in index-management to remove the need for this.

File: bundle-workflow/src/test_workflow/config/test_manifest.yml (hunk 5 above)
Comment (setiah) [reply bundle-workflow/src/test_workflow/config/test_manifest.yml:37]: +1 for adding these settings as dict. It keeps it clear and avoid excessive parsing logic/errors in the code.

File: bundle-workflow/src/test_workflow/config/test_manifest.yml (hunk 4 above)
Comment (setiah) [reply bundle-workflow/src/test_workflow/config/test_manifest.yml:8]: i think we should clarify the context here why this is required. These additional configs should only be opensearch yaml settings IMO.

Comment (gaiksaya) [reply bundle-workflow/src/test_workflow/config/test_manifest.yml:8]: Agree. Like if we put this setting in `opensearch.yml` creation and deletion of that should be taken care by the tests. Should I remove this workaround from our end and let IM plugin fix it?

Comment (setiah) [reply bundle-workflow/src/test_workflow/config/test_manifest.yml:8]: yeah, let's do that and circle back on this requirement with index-management team.

File: bundle-workflow/src/test_workflow/config/test_manifest.yml (hunk 5 above)
Comment (gaiksaya) [reply bundle-workflow/src/test_workflow/config/test_manifest.yml:37]: FYI for it to be dict the test_manifest.yml will look like:
```yaml
  - name: alerting
    integ-test:
      test-configs:
        - with-security
        - without-security
      additional-cluster-configs:
        plugins.destination.host.deny_list : [10.0.0.0/8, 127.0.0.1]
```
Note the removal of `-`

bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py
gaiksaya: Created the [issue](https://github.com/opensearch-project/index-management/issues/151) with IM team

dblock (approved):
Much cleaner.

saratvemulapalli (approved):
This is nice, Thanks @gaiksaya !

Issue #532, opened 2021-09-20
Add unit test coverage for test workflow


setiah:
Improve the test_workflow UT coverage. 

Report as of 20 Sept. 

Module | statements | missing | excluded | coverage
-- | -- | -- | -- | --
Total | 330 | 198 | 0 | 40%
src/test_workflow/integ_test/integ_test_suite.py | 50 | 50 | 0 | 0%
src/test_workflow/integ_test/local_test_cluster.py | 95 | 95 | 0 | 0%
src/test_workflow/test_args.py | 30 | 30 | 0 | 0%
src/test_workflow/test_recorder.py | 9 | 9 | 0 | 0%
src/test_workflow/test_cluster.py | 18 | 8 | 0 | 56%
src/test_workflow/test_component.py | 9 | 3 | 0 | 67%
src/test_workflow/bwc_test/bwc_test_suite.py | 33 | 2 | 0 | 94%
src/test_workflow/perf_test/perf_test_suite.py | 21 | 1 | 0 | 95%
src/test_workflow/__init__.py | 0 | 0 | 0 | 100%
src/test_workflow/bwc_test/__init__.py | 0 | 0 | 0 | 100%
src/test_workflow/dependency_installer.py | 23 | 0 | 0 | 100%
src/test_workflow/integ_test/__init__.py | 0 | 0 | 0 | 100%
src/test_workflow/perf_test/__init__.py | 0 | 0 | 0 | 100%
src/test_workflow/perf_test/perf_test_cluster.py | 42 | 0 | 0 | 100%
src/test_workflow/utils/__init__.py | 0 | 0 | 0 | 100%

Issue #522, opened 2021-09-17
[META] Add input validation checks and debug logs for test jobs


setiah:
To improve debug-ability and prevent common mistakes, we should add some input validation checks and debug logs in the test script.

- [x] Add input parameter validation that checks the provided opensearch version, architecture and build id are in the right format: https://github.com/opensearch-project/opensearch-build/pull/539
- [x] Log the s3 key for debugging s3 download failures. 
- [x] Add logs for plugin tests on jenkins?
- [x] #525

minalsha:
@setiah what do you mean by "Add logs for plugin tests on jenkins?"  Is this if we dont have test recorder and test publisher done and hence we would need the logs for Plugins tests on Jenkins?

setiah:
Yeah the test recorder should publish all the logs to s3 eventually. In the mean time, we need some messaging on jenkins dashboard itself for an integration test run on a component to know if it failed or succeeded. #549 adds log messages if the component test failed.

PR #539, opened 2021-09-21
Added Unit Tests for test_args and test_component

owaiskazi19:
### Description
Added Unit Tests for test_args and test_component. 
Added validation for architecture and version in test_args
 
### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/532 and  https://github.com/opensearch-project/opensearch-build/issues/522
 
### Check List

### DIFF
--- a/bundle-workflow/src/test_workflow/test_args.py
+++ b/bundle-workflow/src/test_workflow/test_args.py
@@ -9,7 +9,7 @@
 
 import argparse
 import logging
-
+import semantic_version
 
 class TestArgs:
     s3_bucket: str
@@ -57,6 +57,12 @@ class TestArgs:
             dest="logging_level",
         )
         args = parser.parse_args()
+
+        if not semantic_version.validate(args.opensearch_version):
+            raise ValueError("Invalid version number")
+        if args.architecture not in ['x64', 'arm64']:
+            raise ValueError("Architecture is not supported. Supported architectures are arm64 and x64")
+        
         self.s3_bucket = args.s3_bucket
         self.opensearch_version = args.opensearch_version
         self.build_id = args.build_id
--- a/bundle-workflow/src/test_workflow/test_args.py
+++ b/bundle-workflow/src/test_workflow/test_args.py
@@ -13,14 +13,13 @@ import logging
 import semantic_version  # type:ignore
 
 
-class CheckOpenSearchVersion(argparse.Action):
-    def __call__(self, parser, namespace, values, option_string=None):
-        if not semantic_version.validate(values):
-            raise ValueError(f"Invalid version number: {values}")
-        setattr(namespace, self.dest, values)
-
-
 class TestArgs:
+    class CheckSemanticVersion(argparse.Action):
+        def __call__(self, parser, namespace, values, option_string=None):
+            if not semantic_version.validate(values):
+                raise ValueError(f"Invalid version number: {values}")
+            setattr(namespace, self.dest, values)
+
     s3_bucket: str
     opensearch_version: str
     build_id: int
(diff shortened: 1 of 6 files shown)


bundle-workflow/tests/tests_test_workflow/test_test_args.py
```diff
@@ -0,0 +1,188 @@
...
+            "--build-id",
+            "30",
+            "--architecture",
+            "x64",
+            "--test-run-id",
+            "6",
+        ],
+    )
+    def test_s3_bucket(self):
+        self.assertEqual(TestArgs().s3_bucket, "xyz")
```
dblock, line 32: Because the `@patch` calls are all the same, combine the tests into a single one that `self.assertEqual(...)` all the things, reduce copy-paste.

dblock (approved):

bundle-workflow/tests/tests_test_workflow/test_test_args.py
```diff
@@ -0,0 +1,114 @@
+import logging
+import os
+import unittest
+from unittest.mock import patch
+
+from test_workflow.test_args import TestArgs
+
+
+class TestTestArgs(unittest.TestCase):
```
setiah, line 9: Somewhat related to this - we need to add input validation on the test args to ensure version, architecture, etc. inputs are in desired format/options. We could probably address that as part of this PR

dblock (dismissed):

dblock (changes requested):
Validation is built into `argparse`, use it.

bundle-workflow/src/test_workflow/test_args.py
```diff
@@ -57,6 +59,12 @@ def __init__(self):
             dest="logging_level",
         )
         args = parser.parse_args()
+
+        if not semantic_version.validate(args.opensearch_version):
```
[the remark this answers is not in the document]
owaiskazi19: Thanks for this. Changed the approach

bundle-workflow/tests/tests_test_workflow/test_test_args.py
```diff
@@ -0,0 +1,154 @@
...
+            "1.1.0",
+            "--build-id",
+            "30",
+            "--architecture",
+            "x64",
+            "--test-run-id",
+            "6",
+        ],
+    )
+    def test_verbose_default(self):
```
dblock, line 133: assertEqual

bundle-workflow/tests/tests_test_workflow/test_test_args.py
```diff
@@ -0,0 +1,154 @@
...
+            "30",
+            "--architecture",
+            "x64",
+            "--test-run-id",
+            "6",
+        ],
+    )
+    def test_invalid_version(self):
+        with self.assertRaises(ValueError) as context:
+            self.assertEqual(TestArgs().opensearch_version, str(context.exception))
```
dblock, line 76: This looks suspicious as the error raised in code is not just the version. You actually never compare the error message, it bails with an exception earlier. The comparison should be done outside of asserRaises. Note that in python scope of things inside with isn’t limited to it, so you can check context error after the with statement.

dblock (changes requested):
There also a util directory, i think that’s leftover.

bundle-workflow/tests/tests_test_workflow/test_test_args.py
owaiskazi19: Makes sense. Changed it

owaiskazi19:
> There also a util directory, i think that’s leftover.

There was just an init file inside the utils directory and I have removed the directory as it was a leftover

dblock (approved):

setiah (approved):

PR #548, opened 2021-09-22
Add unittest for execute.py in system

### LINKED ISSUE #420: Add more unittest coverage for src in bundle-workflows/src/system
Currently we dont have a full test coverage for bundle-workflows/src/system

```
---------- coverage: platform darwin, python 3.7.7-final-0 -----------
Name                                Stmts   Miss  Cover
-------------------------------------------------------
src/system/__init__.py                  0      0   100%
src/system/arch.py                      8      0   100%
src/system/console.py                   4      4     0%
src/system/execute.py                   8      8     0%
src/system/properties_file.py          38      1    97%
src/system/temporary_directory.py      12      0   100%
src/system/working_directory.py         7      0   100%
-------------------------------------------------------
TOTAL                                  77     13    83%
```

We need to add more tests to the code.

Add more test coverage for src in bundle-workflows

**Additional context**

peterzhu1992:
Signed-off-by: Peter Zhu <peterzhu1992@gmail.com>

### Description
Add unittest for execute.py in system
 
### Issues Resolved
#420 
 
### Check List


peterzhu1992:
@mch2 @dblock please approve for workflow and check the code.
Thanks.

dblock (changes requested):
A unit test should have more predictable execution without branches. This one has 2:

- It may return one of x86, arm64, ... and the list will grow over time. 
- It won't work on windows that doesn't have `uname`.
- It checks whether the status is success/fail, which is not predictable

To solve this you can execute something more predictable like `echo xyz`, and compare the output.

Then, `execute` returns a tuple, so do this:

```
(status, stdout, stderr) = execute(...)
```

This lets you use `stdout.strip()` instead of `results[1].strip`.

Add a test for `stderr`, `>&2 echo "error"` for example will do that.

bundle-workflow/tests/tests_system/test_execute.py
```diff
@@ -0,0 +1,18 @@
...
+
+import subprocess
+import unittest
+from unittest.mock import patch
+
+from system.execute import execute
+
+
+class TestExecute(unittest.TestCase):
+    def test_current_arch(self):
```
dblock, line 15: The name of the function should be `test_execute`, because you're testing `system.execute`, not the fact that you're able to get the current arch.

peterzhuamazon:
> A unit test should have more predictable execution without branches. This one has 2:
> 
>     * It may return one of x86, arm64, ... and the list will grow over time.
> 
>     * It won't work on windows that doesn't have `uname`.
> 
>     * It checks whether the status is success/fail, which is not predictable
> 
> 
> To solve this you can execute something more predictable like `echo xyz`, and compare the output.
> 
> Then, `execute` returns a tuple, so do this:
> 
> ```
> (status, stdout, stderr) = execute(...)
> ```
> 
> This lets you use `stdout.strip()` instead of `results[1].strip`.
> 
> Add a test for `stderr`, `>&2 echo "error"` for example will do that.

Great advices @dblock will apply these and get better next time.

bundle-workflow/tests/tests_system/test_execute.py
peterzhuamazon: This is a miss due to copy paste from test_arch.py, will fix.

bundle-workflow/tests/tests_system/test_execute.py
```diff
@@ -0,0 +1,25 @@
...
+
+class TestExecute(unittest.TestCase):
+    def test_execute_stdout(self):
+        (status, stdout, stderr) = execute("echo success", "/")  # (0, 'success\n', '')
+        self.assertEqual(status, 0)
+        self.assertEqual(stdout.strip(), "success")
+        self.assertEqual(stderr.strip(), "")
+
+    def test_execute_stderr(self):
+        (status, stdout, stderr) = execute(">&2 echo error", "/")  # (0, '', 'error\n')
```
dblock, line 22: Add a check for the exit code, this could be a bug where you always get 0. Execute `exit 2` for example and check for 2 being returned.

dblock (approved):

bundle-workflow/tests/tests_system/test_execute.py
peterzhuamazon: That is a good improvement let me check

peterzhuamazon: fixed.

peterzhuamazon:
Hi @dblock @peternied please check again I have addressed all the concerns with more fine-grained test cases.

dblock (approved):

dblock:
Good work.

peternied (approved):

PR #550, opened 2021-09-22
Add cron schedule to the Jenkinsfile

### LINKED ISSUE #545: Test-manifest should take care of test related configurations specific to each plugin
Some plugins needs to have system related configurations (a directory for example) to run their integration tests. 
As of today we need to change and maintain our codebase to accommodate this requirements

We need to take care of this at test_manifest level where all plugin related configuration is at one place. 
Something like

```
components:
  - name: index-management
    integ-test:
      test-configs:
        - with-security
        - without-security
      additional-cluster-configs:
        - “path.repo: /tmp/tmp_dir”
      shell-commands:
        - “mkdir /tmp/tmp_dir”
 ```
The above shell command should run on the test environment and satisfy the requirements for `additional-cluster-configs`

### PR DESCRIPTION
**Pull request by peternied**
Resolves #545

### Check List

### DIFF
--- a/bundle-workflow/Jenkinsfile
+++ b/bundle-workflow/Jenkinsfile
@@ -7,6 +7,11 @@ pipeline {
         jdk "JDK14"
         maven "maven-3.8.2"
     }
+    triggers {
+      parameterizedCron '''
+        H */2 * * * %INPUT_MANIFEST=1.1.0/opensearch-1.1.0.yml
+      '''
+    }
     stages {
         stage('parameters') {
             steps {

### CODE REVIEW

File: bundle-workflow/Jenkinsfile [hunk 1]
```diff
@@ -7,6 +7,11 @@ pipeline {
         jdk "JDK14"
         maven "maven-3.8.2"
     }
+    triggers {
+      parameterizedCron '''
+        H */2 * * * %INPUT_MANIFEST=1.1.0/opensearch-1.1.0.yml
```
gaiksaya, line 12: What is `H` in this?

peternied: I do not understand cron much like regex, but there is this great tool [cron.help](https://cron.help/#*/2_*_*_*_*) that you could plug values into see what it might do.

Also - this is our existing schedule that runs every 2 hours, but is not in source control.

gaiksaya: For every 2 hours shouldn't it be `0 */2 * * *`?

bbarani: I think the question is "what is the value of 'H' here"?

bbarani: I figured it out - https://www.jenkins.io/doc/book/pipeline/syntax/#cron-syntax

gaiksaya: That explains it. Thanks @bbarani

gaiksaya (approved):

dblock (approved):

dblock:
So this gives me the ability to automagically start building other versions by editing this file?

Issue #553, opened 2021-09-22
Add support to fail test suite job if any of the components failed the test.


saratvemulapalli:
### Is your feature request related to a problem? Please describe

Integration Test/ BWC Tests run through all the components even though the component tests might fail. Which is ok but at the end of the test result the test suite exits with status 0 i.e everything went good. 

### Describe the solution you'd like

Collect all the test results and fail the job if any of the test component failed. 
In addition it would be nice to log what components failed.

PR #560, opened 2021-09-22
Added support to integ test failed job

owaiskazi19:

### Description
Log status code if any plugin fails to run integ test
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/553
 
### Check List

### DIFF
--- a/bundle-workflow/src/run_integ_test.py
+++ b/bundle-workflow/src/run_integ_test.py
@@ -59,11 +59,11 @@ def main():
                     work_dir,
                     args.s3_bucket
                 )
-                result = test_suite.execute()
-                if result[0] != 0:
-                    failed_components[component.name] = result
+                status, security = test_suite.execute()
+                if status != 0:
+                    failed_components[component.name] = [status, security]
                 else:
-                    passed_components[component.name] = result
+                    passed_components[component.name] = [status, security]
             else:
                 logging.info(
                     "Skipping tests for %s, as it is currently not supported"
@@ -72,14 +72,12 @@ def main():
 
         if passed_components:
             for component, result in passed_components.items():
-                logging.info(f'PASS: Integration Test for {component} with status code {result[0]}')
-                logging.info(result[1])
+                logging.info(f'PASS: Integration Test for {component} {result[1]} with status code {result[0]}')
 
         if failed_components:
             for component, result in failed_components.items():
-                logging.error(f'FAIL: Integration Test for {component} with status code {result[0]}')
-                logging.info(result[2])
-                sys.exit(1)
+                logging.error(f'FAIL: Integration Test for {component} {result[1]} with status code {result[0]}')
+            sys.exit(1)
 
 
 if __name__ == "__main__":
--- a/bundle-workflow/src/run_integ_test.py
+++ b/bundle-workflow/src/run_integ_test.py
@@ -47,6 +47,8 @@ def main():
             args.s3_bucket, args.build_id, args.opensearch_version, args.architecture, work_dir)
         pull_build_repo(work_dir)
         DependencyInstaller(build_manifest.build).install_all_maven_dependencies()
+        failed_components = dict()
+        passed_components = dict()
         for component in bundle_manifest.components:
             if component.name in integ_test_config.keys():
                 test_suite = IntegTestSuite(
@@ -57,12 +59,23 @@ def main():
                     work_dir,
                     args.s3_bucket
                 )
-                test_suite.execute()
+                status = test_suite.execute()
+                if status != 0:
+                    failed_components[component.name] = status
+                else:
+                    passed_components[component.name] = status
             else:
                 logging.info(
                     "Skipping tests for %s, as it is currently not supported"
                     % component.name
                 )
+
+        if failed_components:
+            for component, status in failed_components.items():
+                logging.error(f'FAIL: Integration Test for {component} with status code {status}')
+        if passed_components:
+            for component, status in passed_components.items():
+                logging.info(f'PASS: Integration Test for {component} with status code {status}')
 
 
 if __name__ == "__main__":
--- a/bundle-workflow/src/run_integ_test.py
+++ b/bundle-workflow/src/run_integ_test.py
@@ -59,23 +59,27 @@ def main():
                     work_dir,
                     args.s3_bucket
                 )
-                status = test_suite.execute()
-                if status != 0:
-                    failed_components[component.name] = status
+                result = test_suite.execute()
+                if result[0] != 0:
+                    failed_components[component.name] = result
                 else:
-                    passed_components[component.name] = status
+                    passed_components[component.name] = result
             else:
                 logging.info(
                     "Skipping tests for %s, as it is currently not supported"
                     % component.name
                 )
 
-        if failed_components:
-            for component, status in failed_components.items():
-                logging.error(f'FAIL: Integration Test for {component} with status code {status}')
         if passed_components:
-            for component, status in passed_components.items():
-                logging.info(f'PASS: Integration Test for {component} with status code {status}')
+            for component, result in passed_components.items():
+                logging.info(f'PASS: Integration Test for {component} with status code {result[0]}')
+                logging.info(result[1])
+
+        if failed_components:
+            for component, result in failed_components.items():
+                logging.error(f'FAIL: Integration Test for {component} with status code {result[0]}')
+                logging.info(result[2])
+                sys.exit(1)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


bundle-workflow/src/run_integ_test.py
```diff
@@ -86,6 +86,7 @@ def main():
             args.s3_bucket, args.build_id, args.opensearch_version, args.architecture, work_dir)
         pull_common_dependencies(work_dir, build_manifest)
         sync_dependencies_to_maven_local(work_dir, build_manifest.build.version)
+        failed_components = dict()
```
dblock, line 50: I would collect all results, then log both, not just failures. It's equally important to know what passed.

bundle-workflow/src/run_integ_test.py
```diff
@@ -95,13 +96,19 @@ def main():
...
+                    failed_components[component.name] = status
             else:
                 logging.info(
                     "Skipping tests for %s, as it is currently not supported"
                     % component.name
                 )
 
+        if failed_components:
+            for component, status in failed_components.items():
+                logging.info(f'Integration Test Failed for {component} with status code {status}')
```
dblock, line 72: Log passes and failures with something more visible, like `PASS:` and `FAIL`:

Use logging.error for FAILs.

dblock (changes requested):
Can we get some more information in the failures that look like actually what failed/errors/something useful?

bundle-workflow/src/run_integ_test.py
owaiskazi19: The error logs are already getting print here: https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py#L117.

`failed_components ` is just to avoid showing status code as 0 for any plugin failure

dblock: But I like the summary that I can easily scroll to at the end!

dblock (commented):
This might sound like over-engineering, but I feel like there's a lot of duplication starting here and I bet it's only a matter of time till we want to write this status to some kind of report.

I would make a new class called `TestResult` that hold a reference to the component and its run status and has `__str__` method to convert itself to string with PASS: or FAIL. Add a `TestResults` dictionary and `passed` and `failed` properties that use python iterators to filter out passed/failed results.

Then output is a 1-liner.

bundle-workflow/src/run_integ_test.py
```diff
@@ -57,13 +59,24 @@ def main():
...
                     % component.name
                 )
 
+        if failed_components:
+            for component, status in failed_components.items():
+                logging.error(f'FAIL: Integration Test for {component} with status code {status}')
+        if passed_components:
+            for component, status in passed_components.items():
+                logging.info(f'PASS: Integration Test for {component} with status code {status}')
+
```
saratvemulapalli, line 79: Also after printing out the summary, if there are failed components make sure the `run_integ_test.py` also fails with non-zero status code.

owaiskazi19:
> This might sound like over-engineering, but I feel like there's a lot of duplication starting here and I bet it's only a matter of time till we want to write this status to some kind of report.
> 
> I would make a new class called `TestResult` that hold a reference to the component and its run status and has `__str__` method to convert itself to string with PASS: or FAIL. Then add a `TestResults` class that inherits from dictionary and `passed` and `failed` properties that use python iterators to filter out passed/failed results.
> 
> Then output is a 1-liner.

This can be pick as an enhancement for TestRecorder and TestPublisher for the long term.

dblock:
> > This might sound like over-engineering, but I feel like there's a lot of duplication starting here and I bet it's only a matter of time till we want to write this status to some kind of report.
> > I would make a new class called `TestResult` that hold a reference to the component and its run status and has `__str__` method to convert itself to string with PASS: or FAIL. Then add a `TestResults` class that inherits from dictionary and `passed` and `failed` properties that use python iterators to filter out passed/failed results.
> > Then output is a 1-liner.
> 
> This can be pick as an enhancement for TestRecorder and TestPublisher for the long term.

OK. My 0.02c is that it's roughly the same amount of work with better design, so we're actively adding technical debt here.

dblock (approved):
I'm approving only to move forward. 

This implementation is brittle, uses two dictionaries, copy-pastes the same code, fetches results from things like `result[0]`, has no tests, and is very difficult to test. Implementing this change with an actual minimal design of two classes was roughly the same amount of work. I don't see why we'd be actively acquiring technical debt here.

saratvemulapalli:
I agree with @dblock, it would be more cleaner than to have some loose data structures with the code. 
I am ok with creating up a follow up issue and work along side with TestRecorder/TestPublisher.

bundle-workflow/src/run_integ_test.py
```diff
@@ -57,13 +59,28 @@ def main():
...
+        if passed_components:
+            for component, result in passed_components.items():
+                logging.info(f'PASS: Integration Test for {component} with status code {result[0]}')
+                logging.info(result[1])
+
+        if failed_components:
+            for component, result in failed_components.items():
+                logging.error(f'FAIL: Integration Test for {component} with status code {result[0]}')
+                logging.info(result[2])
+                sys.exit(1)
```
saratvemulapalli, line 82: This will exit the job after first component failure is reported. 
```suggestion
            sys.exit(1)
```

saratvemulapalli, line 82: This will exit the job after first component failure is reported.
We would only want to exit after reporting all the failures.
```suggestion
                logging.info(result[2])
            sys.exit(1)
```

owaiskazi19:
> I agree with @dblock, it would be more cleaner than to have some loose data structures within the code.
> I am ok with creating up a follow up issue and work along side with TestRecorder/TestPublisher.

Sure. I created an issue for the same https://github.com/opensearch-project/opensearch-build/issues/579

bundle-workflow/src/run_integ_test.py
```diff
@@ -57,13 +59,26 @@ def main():
...
                     % component.name
                 )
 
+        if passed_components:
+            for component, result in passed_components.items():
+                logging.info(f'PASS: Integration Test for {component} with status code {result[0]} and config is {result[1]}')
+
+        if failed_components:
+            for component, result in failed_components.items():
+                logging.error(f'FAIL: Integration Test for {component} with status code {result[0]} and config is {result[1]}')
```
saratvemulapalli, line 79: I would like, it feels more cleaner.
```suggestion
                logging.error(f'FAIL: Integration Test for {component} {result[1]} with status code {result[0]} ')
```

saratvemulapalli (approved):
Thanks @owaiskazi19 for the change!
Looks good to me.

saratvemulapalli (approved):

bundle-workflow/tests/test_integ_workflow/integ_test/test_integ_test_suite.py
```diff
@@ -58,13 +58,7 @@ def test_execute_with_multiple_test_configs(
...
                     True,
                     False,
-                ),
-                call(
-                    "integtest.sh -b localhost -p 9200 -s false -v 1.1.0",
-                    "/tmpdir/job-scheduler",
-                    True,
-                    False,
-                ),
+                )
```
setiah, line 61: This is incorrect. This test is to check all test-configs are executed as part of a component test suite. You are removing check for without-security config run here.

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -48,7 +48,8 @@ def execute(self):
         self.__install_build_dependencies()
         for config in self.test_config.integ_test["test-configs"]:
             security = self.__is_security_enabled(config)
-            self.__setup_cluster_and_execute_test_config(security)
+            status = self.__setup_cluster_and_execute_test_config(security)
+            return status, config
```
setiah, line 52: This is incorrect. It will return before the for loop is able to execute all configs.

dblock:
@setiah @owaiskazi19 It looks like we missed a bunch of things in the CR, I added notes to https://github.com/opensearch-project/opensearch-build/issues/579. Is someone taking this?

owaiskazi19:
> @setiah @owaiskazi19 It looks like we missed a bunch of things in the CR, I added notes to #579. Is someone taking this?

I am working on the fix for this CR. #579 can be pick later as fixing integ test will help us to create a meta issue for those plugins which are not running integ test as expected.

bundle-workflow/tests/test_integ_workflow/integ_test/test_integ_test_suite.py
owaiskazi19: Addressed in https://github.com/opensearch-project/opensearch-build/pull/630

bundle-workflow/src/test_workflow/integ_test/integ_test_suite.py
owaiskazi19: Addressed in https://github.com/opensearch-project/opensearch-build/pull/630

Issue #74, opened 2021-06-24
Lower CI Java Version to Java 11 


nknize:
Java 11 is an LTS release, Java 14 is not. Elastic really shouldn't have made the minimum java version 14 prematurely. Lucene doesn't even require minimum 14. The [community opened a PR](https://github.com/opensearch-project/OpenSearch/pull/636) to lower the minimum required java version back to 11 but it is stalled until CI can be lowered to match (otherwise CI will fail). The PR has stalled now for almost two months and the contributor has long given up. This is unfortunate and really quite unacceptable. Maybe we can get this taken care of to unblock that PR. 

**Update**: the remaining work is as follows. OpenSearch has CI defined in a private freestyle project that runs `gradle check` on OpenSearch#main, branches and PRs. Extract that into Jenkinsfile that is committed to the OpenSearch repo, similar to https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/Jenkinsfile.

peterzhuamazon:
Hi @nknize as of now the Infra Team did not add any JDK into the release artifact.
In fact, the JDK 14 is bundled with the core/min artifacts during the Gradle Assemble process.
From Infra Side, we only install the plugins and re-tar artifacts as bundle.

If this is something that can be resolved in gradle assemble that would be great.
If this is a hard target to fix in core, we can see what we can do in the bundle process.

Thanks.

dblock:
@peterzhuamazon Are you saying the version of Java that the build is run with isn't set to 14 in Jenkins?

nknize:
> @peterzhuamazon Are you saying the version of Java that the build is run with isn't set to 14 in Jenkins?

`JAVA{8-15}_HOME` is required to be set to build. So the jenkins build has to be setting those. Also, jdk is bundled and can be specified with: `-Druntime.java=14`. Is the jenkins runtime specifying this?

peterzhuamazon:
> @peterzhuamazon Are you saying the version of Java that the build is run with isn't set to 14 in Jenkins?

We set the build to 14 yes but that was purely due to JDK14 has been used for engine throughout ODFE.
I do not have knowledge on how JDK is being taken care during the assemble process.
Are you saying the gradle assemble take JDK and bundle the specific version based on the JDK env during build?

nknize:
> We set the build to 14 yes but that was purely due to JDK14 has been used for engine throughout ODFE.

[...]

Issue #209, opened 2021-08-13
Dockerize component build scripts


mch2:
Right now each component builds in a build.sh script that relies on the execution environment to set up the right versions of python, java, maven etc.  It would be better for each component to define a docker image that is needed to set up a repeatable build environment.

bbarani:
[Triage]: The scope should cover both the build and tests.

peterzhuamazon:
Link to #399

Issue #527, opened 2021-09-20
[Bundle Build Break] ARM Pip install failure psutil==5.8.0


peternied:
**To Reproduce**
1.  Start build on an ARM machine `./bundle-workflow/build.sh manifests/1.1.0/opensearch-1.1.0.yml`

```
Installing dependencies from Pipfile.lock (22e2de)...
[...]

PR #588, opened 2021-09-24
Add Jenkins Docker Parallel Support and Restructure CICD Environment for build/assemble/test

### LINKED ISSUE #399: Build process should use multiple agent nodes in parallel
Building is a critical path process, but if we are making multiple builds based on different architectures they are unrelated, and should be parallelized for speed.  With the existing setup for jenkins with cloud agents, when parallel is being used the same machine is being reused this is causing contention between staging directories, local maven, and other things.

Switch back to [parallel](https://www.jenkins.io/doc/book/pipeline/syntax/#parallel) for the stage execution, but come up with a method to make sure they are executed on different nodes.

There might be easier methods to signal/manage how this is done, open to alternative approaches that keep the isolation and would add speed to the existing processes.  Using docker for each agent could be a consideration

Isolation Diagram
[Img Source](https://gist.github.com/peternied/26d038d386b2e9119991bf3af61a3dd8)

**Additional context**
See #396

peterzhuamazon:
### Description
Add Jenkins Docker Parallel Support and Restructure CICD Environment for build/assemble/test

1. Modify Dockerfile to include necessary dependencies, include both x64 and arm64 build capability. Remove JDK and MAVEN as Jenkins dependencies, add to dockerfile
2. Added centos alongside the original al2 base image, as k-NN needs to be built on CentOS for older glibc version (2.26 vs 2.17) than the one on al2.
3. Upload docker images based on the dockerfiles to https://hub.docker.com/r/opensearchstaging/ci-runner.
4. Removed tools/standard-* folders as they are obsolete with the introduction of https://github.com/opensearch-project/opensearch-build/tree/main/bundle-workflow/src/test_workflow.
5. Move the standard deploy single node cluster script to scripts/ folder as it is still handy to use for testing on local.
6. Move all the dockerfiles to bundle-workflow/ folder as they are used for all bundle related actions.
7. Added entries to README and Documentations for usages.

Thanks.

 
### Issues Resolved
#399 #209 opensearch-project/opensearch-infra#109
 
### Check List


bundle-workflow/Jenkinsfile
```diff
@@ -112,6 +105,11 @@ pipeline {
 
 void build() {
     git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
+
+    // psutil does not support arm64 yet, and it is not needed for build and assamble on Jenkins
+    // https://github.com/giampaolo/psutil/issues/1966
+    sh "sed -i '/psutil/d' bundle-workflow/Pipfile"
```
peternied, line 111: We should revert the requirement for the library if it does not work in our supported platforms @gaiksaya

bundle-workflow/Jenkinsfile
```diff
@@ -57,14 +49,15 @@ pipeline {
...
                         }
                     }
                 }
                 stage('build-x86') {
                     agent {
-                        node {
-                            label 'Jenkins-Agent-al2-x64-m5xlarge'
+                        docker {
+                            label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+                            image 'opensearchstaging/ci-runner:al2-x64-arm64-jdk14-node10.24.1-20210924'
```
peternied, line 59: This naming schema is throwing me, this is a multi-arch image, can we just use noarch instead of x64/arm64?

bundle-workflow/Jenkinsfile
```diff
@@ -29,24 +25,20 @@ pipeline {
...
                 stage('build-snapshots') {
                     environment {
                         SNAPSHOT_REPO_URL = "https://aws.oss.sonatype.org/content/repositories/snapshots/"
                     }
                     agent {
-                        node {
-                            label 'Jenkins-Agent-al2-x64-m5xlarge'
+                        docker {
+                            label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+                            image 'opensearchstaging/ci-runner:al2-x64-arm64-jdk14-node10.24.1-20210924'
```
peternied, line 36: With this change, it seems like we are hardcoding the version of java even more firmly, do you have thoughts on after this change how we would resolve https://github.com/opensearch-project/opensearch-build/issues/74?

bundle-workflow/dockerfiles/ci-runner.centos8.dockerfile
```diff
@@ -0,0 +1,81 @@
...
+# Add AdoptOpenJDK Repo
+RUN echo -e "[AdoptOpenJDK]\nname=AdoptOpenJDK\nbaseurl=http://adoptopenjdk.jfrog.io/adoptopenjdk/rpm/centos/7/\$basearch\nenabled=1\ngpgcheck=1\ngpgkey=https://adoptopenjdk.jfrog.io/adoptopenjdk/api/gpg/key/public" > /etc/yum.repos.d/adoptopenjdk.repo
+
+# Add normal dependencies
+RUN yum clean all && \
+    yum update -y && \
+    yum install -y adoptopenjdk-14-hotspot which curl git tar net-tools procps-ng cmake python3 python3-pip && \
+    ln -sfn `which pip3` /usr/bin/pip && pip3 install pipenv && pipenv --version 
+
+# Add Dashboards dependencies
```
peternied, line 23: What are your thoughts on having multiple images vs a single image? Could we setup a base image for our CI systems and layer on top OpenSearch, dashboard, or other necessary components.

Is there much of  size/boot up time different with these dashboard/notebook specific dependences?

bundle-workflow/Jenkinsfile
```diff
@@ -75,14 +68,15 @@ pipeline {
...
                         }
                     }
                 }
                 stage('build-arm64') {
                     agent {
-                        node {
-                            label 'Jenkins-Agent-al2-arm64-m6gxlarge'
+                        docker {
+                            label 'Jenkins-Agent-al2-arm64-c6g4xlarge-Docker-Host'
+                            image 'opensearchstaging/ci-runner:al2-x64-arm64-jdk14-node10.24.1-20210924'
```
peternied, line 78: How do we version the docker image?  Do we need to keep updating the Jenkinsfile?

bundle-workflow/Jenkinsfile
```diff
@@ -57,14 +49,15 @@ pipeline {
...
-                            sh "rm -rf ~/.m2/"
                         }
                     }
                 }
                 stage('build-x86') {
                     agent {
-                        node {
-                            label 'Jenkins-Agent-al2-x64-m5xlarge'
+                        docker {
+                            label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
```
peternied, line 58: This is a sizeable host difference, does this build better/faster?

peternied (commented):
This is awesome!

bundle-workflow/Jenkinsfile
dblock: 💯 Let's open an issue at least.

bundle-workflow/Jenkinsfile
dblock: Looks like the plan is to have multiple images with different JDKs. This works for me, unless we want to build a single image with many different JDKs and be able to select which one is used by default.

dblock (commented):
This is a great step forward, I would 🚢 it.

bundle-workflow/Jenkinsfile
peterzhuamazon: Sure @peternied @dblock also psutil does not work on Windows with cygwin.

peterzhuamazon: https://github.com/opensearch-project/opensearch-build/issues/597

bundle-workflow/Jenkinsfile
peterzhuamazon: There are many other arch like ppcle64, Z, and else.

bundle-workflow/Jenkinsfile
gaiksaya: It works on arm64. See issue https://github.com/opensearch-project/opensearch-build/issues/527
Not sure about windows. If anyone has any work around I am all ears

owaiskazi19: Correct me if wrong @gaiksaya if we want parent pid here: https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py#L116. Can't we use `os.getppid()`?

gaiksaya: More than that we want children process id here: https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/src/test_workflow/integ_test/local_test_cluster.py#L118

owaiskazi19: For that something like: 
```shell
ps_output = run(['ps', '-opid', '--no-headers', '--ppid', str(os.getpid())],
                stdout=PIPE, encoding='utf8')
child_process_ids = [int(line) for line in ps_output.stdout.splitlines()]
``

Comment (dblock) [reply bundle-workflow/Jenkinsfile:111]: We're going to have a lot of other things broken on Windows, let's take that on separately?

File: bundle-workflow/Jenkinsfile (hunk 2 above)
Comment (peternied) [reply bundle-workflow/Jenkinsfile:59]: Is there documentation on the convention or common usages that I could see?

**Review by peternied — DISMISSED** [review]
Looking great - non of my questions are blocking, but I'd love more details

File: bundle-workflow/Jenkinsfile (hunk 5 above)
Comment (gaiksaya) [reply bundle-workflow/Jenkinsfile:78]: or add the latest tag to remain updated?

File: bundle-workflow/dockerfiles/ci-runner.centos8.dockerfile [hunk 7]
```diff
@@ -0,0 +1,81 @@
...
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+# This is a docker image specifically for standardize the ci/cd environment
+# for both developers and ci/cd tools in OpenSearch / OpenSearch-Dashboards
+# Please read the README.md file for all the information before using this dockerfile
+
+
+FROM centos:8
```
gaiksaya, line 12: Why do we need 2 images? Both Centos and AL2?

peternied (approved):

bundle-workflow/Jenkinsfile
peterzhuamazon: Already added python3-devel and close the issue.

bundle-workflow/Jenkinsfile
peterzhuamazon: I only see this as a improve item for later.
Right now I think it is ok to start with.

bundle-workflow/Jenkinsfile
peterzhuamazon: Exactly like @dblock says we are going to have images with multiple setups.
We can also have multiple Jenkinsfile as well.

bundle-workflow/dockerfiles/ci-runner.centos8.dockerfile
peterzhuamazon: Boot up is not a problem, and I am trying to reduce the number of images by adding all the dependencies.
There is no reason to have a base image again since we already based on minimal AL2/CentOS8.

bundle-workflow/Jenkinsfile
peterzhuamazon: I dont see that as an issue, just like we keep updating manifest with different branches.

bundle-workflow/Jenkinsfile
peterzhuamazon: This is a c54xlarge instance that specifically setup for docker, it can provision 10 docker containers per host and have enough power to run the builds.

bundle-workflow/dockerfiles/ci-runner.centos8.dockerfile
peterzhuamazon: Read description, KNN still needs CentOS8 to build.
@jmazanec15 would confirm this as the glibc version is different on both.

peterzhuamazon:
Hi @dblock @gaiksaya please check and approve, comment if you still have any concerns. Thanks.
I also add a new line of code to allow docker adding env var for JAVA_HOME.
This will unblock k-NN lib make issues as jni.ini is not found.

Thanks.

peternied (approved):

dblock (approved):
I do wonder whether we have something to learn from the likes of GitHub actions which don’t include Java or Maven in base images. This makes it easy to test matrixes of things.

bundle-workflow/dockerfiles/ci-runner.centos8.dockerfile
gaiksaya: Missed it. Thanks!

bundle-workflow/Jenkinsfile
gaiksaya: Wouldn't that be a maintainence nightmare for so many images with multiple combinations?
Maybe we should have something that would allow installation of matrix specific versions.

gaiksaya (approved):

bundle-workflow/Jenkinsfile
peterzhuamazon: You choose between convenience vs preserve integrity of the env across all platforms.
I like the approach of having multiple versions.
Welcome more opinions.

PR #590, opened 2021-09-24
[Build] bundle OpenSearch Dashboards

kavilla:
### Description
Produce an OpenSearch Dashboards bundle with plugins provided by a manifest.

This requires the manifest to be in order with plugins first and then Dashboards
being last. The plugins will need to pulled down and placed in the plugins
folder. Then we need to bootstrap to generate the modules so that we can actually
install the plugins to the release build.

What is missing or not verified
* Environment setup, this requires nvm and yarn to be already set up.
We will need to set that up.
* Dropping the custom config into the bundle
* Signing of the artifact
* Tests

### Issues Resolved
Partial https://github.com/opensearch-project/opensearch-build/issues/158

### Check List

### DIFF
--- a/bundle-workflow/scripts/default/build.sh
+++ b/bundle-workflow/scripts/default/build.sh
@@ -16,10 +16,11 @@ function usage() {
     echo -e "-s SNAPSHOT\t[Optional] Build a snapshot, default is 'false'."
     echo -e "-a ARCHITECTURE\t[Optional] Build architecture, ignored."
     echo -e "-o OUTPUT\t[Optional] Output path, default is 'artifacts'."
+    echo -e "-d DASHBOARDS\t[Optional] Build OpenSearch Dashboards, default is 'false'."
     echo -e "-h help"
 }
 
-while getopts ":h:v:s:o:a:" arg; do
+while getopts ":h:v:s:o:a:d:" arg; do
     case $arg in
         h)
             usage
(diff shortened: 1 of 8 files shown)


kavilla:
Previous PR: https://github.com/opensearch-project/opensearch-build/pull/435

dblock (approved):

dblock:
Good. I can take the refactoring of the dashboards `if` on top of this.

bundle-workflow/scripts/default/build.sh
```diff
@@ -55,15 +59,32 @@ if [ -z "$VERSION" ]; then
...
+else
+    ./gradlew assemble --no-daemon --refresh-dependencies -DskipTests=true -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT
+
+    zipPath=$(find . -path \*build/distributions/*.zip)
+    distributions="$(dirname "${zipPath}")"
+
+    echo "COPY ${distributions}/*.zip"
+    mkdir -p $OUTPUT/plugins
+    cp ${distributions}/*.zip ./$OUTPUT/plugins
+fi
```
saratvemulapalli, line 88: We should test these scripts as part of our CI workflows.
A follow up for both OpenSearch and Dashboards.

saratvemulapalli (approved):
This is nice, love to see Dashboards coming along quickly.

bundle-workflow/scripts/default/build.sh
saratvemulapalli: Opened up an issue: https://github.com/opensearch-project/opensearch-build/issues/591

PR #627, opened 2021-09-28
Change Jenkins Docker Runner to CentOS7 so k-NN lib will not crash on OS with lower glibc version

### LINKED ISSUE #481: k-NN library x64 build should use generic march

peterzhuamazon:
### Description
Change Jenkins Docker Runner to CentOS7 so k-NN lib will not crash on OS with lower glibc version
 
### Issues Resolved
#481
 
### Check List

### DIFF
--- a/bundle-workflow/Jenkinsfile
+++ b/bundle-workflow/Jenkinsfile
@@ -33,7 +33,7 @@ pipeline {
                     agent {
                         docker {
                             label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
-                            image 'opensearchstaging/ci-runner:al2-x64-arm64-jdk14-node10.24.1-20210924'
+                            image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdk14-node10.24.1-20210924'
                             // Unlike freestyle docker, pipeline docker does not login to the container and run commands
                             // It use executes which does not source the docker container internal ENV VAR
                             args '-e JAVA_HOME=/usr/lib/jvm/adoptopenjdk-14-hotspot'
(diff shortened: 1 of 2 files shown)


peterzhuamazon:
Hi @dblock @peternied this PR is to add the centos7 docker image, remove centos8 (as it will end its LTS by end of 2021), and run the build process in centos7 so that k-NN lib can be built correctly for x64 without crashing when load on CentOS7.

As a reminder we list CentOS7 as a support OS here: https://opensearch.org/docs/opensearch/install/compatibility/

Glibc Versions:
```
AL2 - 2.26
Ubuntu 20 - 2.27
Ubuntu 18 - 2.27
Ubuntu 16 - 2.23
Centos 7 - 2.17
Centos 8 - 2.28
RHEL 7 - 2.17
RHEL 8 - 2.28
```

peternied (commented):
Why update the bundle workflow job image?  There were no build errors during the normal job.

bundle-workflow/Jenkinsfile
```diff
@@ -33,7 +33,7 @@ pipeline {
                     agent {
                         docker {
                             label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
-                            image 'opensearchstaging/ci-runner:al2-x64-arm64-jdk14-node10.24.1-20210924'
+                            image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdk14-node10.24.1-20210924'
```
peternied, line 36: It would be great if we could automate building these images, created #628

dblock (approved):

peternied (approved):

peterzhuamazon:
> > Why update the bundle workflow job image? There were no build errors during this job.
> 
> If I understand correctly the k-nn native code compiles but doesn't run?

knnlib will compile on AL2 but it will crash during runtime on CentOS7.

PR #656, opened 2021-09-30
Added meta project to manipulate all released components.

dblock:

### Description

Added meta project to manipulate all released components. Moving from https://github.com/opensearch-project/opensearch-plugins/blob/main/META.md because this is where we do the releases. The template thus belongs here and also includes OpenSearch and OpenSearch Dashboards core repos. 
  
### Check List


meta/.meta
```diff
@@ -0,0 +1,24 @@
+{
```
saratvemulapalli, line 1: Why do we want to duplicate our component data? 
I see we already have https://github.com/opensearch-project/project-meta which takes care of this. 
Why not use it from opensearch-build for releases ?

dblock: In meta we have all our repositories, not all components that are included in an OpenSearch release. I considered creating another folder in project-meta for just those, but effectively this list is the sum of OpenSearch + OpenSearch Dashboards manifests and is about what we release, which all lives in this repo.

peternied (approved):

meta/.meta
saratvemulapalli: Ok I would rather have one single source of truth and may be add parameters. 

Anyway, how do we plan on to keep this up to date?
I assume its manual and for a component owner this should be part of onboarding(ONBOARDING.md) to release workflow.

saratvemulapalli (approved):
Changes look good to me. 
Thanks @dblock !

Dropped a minor comment.

meta/.meta
dblock: I agree one source of truth is what we want, and the truth is here ;)

This is manual for now indeed. I'll think of some automation that could ensure that what's in manifests matches this list.

PR #672, opened 2021-10-01
Handle PUBLIC_ARTIFACT_URL without trailing slash

### LINKED ISSUE #661: [Bug]: The location specified in bundle-manifest (manifest.yml) are invalid
### Describe the bug

A build manifest is bundled inside the OpenSearch bundle tarball (manifest.yml) which contains all the information about the components used to build the bundle. The bundle manifest contains a key `location` whose value is invalid for all components.

```
build:
  architecture: x64
  id: '317'
  location: https://ci.opensearch.org/ci/bundles/1.1.0/317/opensearch-1.1.0-linux-x64.tar.gz
  name: OpenSearch
  version: 1.1.0
components:
- commit_id: 15e9f137622d878b79103df8f82d78d782b686a1
  location: https://ci.opensearch.org/ci/builds/1.1.0/317/bundle/opensearch-min-1.1.0-linux-x64.tar.gz
  name: OpenSearch
  ref: '1.1'
```

### To reproduce

Download the bundle. Untar it and see the manifest.yml

Try accessing the location url to download individual component. It will give `Access Denied` error

### Expected behavior

The URLs should be valid. Each component should be downloadable from the given location url

### Screenshots

_No response_

### Host / Environment

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

peternied:
### Description
In the build workflow the path was added as `https://ci.opensearch.org/ci/dbc`, and when `urljoin` was used on that value it unexpectly removed the last element on the url [[StackOverflow](https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin)].

Instead of 'altering' the value in jenkins and breaking the notification message, updated the build script to handle this case and added many unit tests.
 
### Issues Resolved
Resolves #661
 
### Check List

### DIFF
--- a/bundle-workflow/src/assemble_workflow/bundle_recorder.py
+++ b/bundle-workflow/src/assemble_workflow/bundle_recorder.py
@@ -31,7 +31,7 @@ class BundleRecorder:
         return "-".join(parts) + ".tar.gz"
 
     def __get_public_url_path(self, folder, rel_path):
-        path = "{}/{}/{}/{}".format(folder, self.version, self.build_id, rel_path)
+        path = "/".join((folder, self.version, self.build_id, rel_path))
         return urljoin(self.public_url + "/", path)
 
     def __get_location(self, folder_name, rel_path, abs_path):
--- a/bundle-workflow/src/assemble_workflow/bundle_recorder.py
+++ b/bundle-workflow/src/assemble_workflow/bundle_recorder.py
@@ -32,7 +32,7 @@ class BundleRecorder:
 
     def __get_public_url_path(self, folder, rel_path):
         path = "{}/{}/{}/{}".format(folder, self.version, self.build_id, rel_path)
-        return urljoin(self.public_url, path)
+        return urljoin(self.public_url + "/", path)
 
     def __get_location(self, folder_name, rel_path, abs_path):
         if self.public_url:
--- a/bundle-workflow/tests/tests_assemble_workflow/test_bundle_recorder.py
+++ b/bundle-workflow/tests/tests_assemble_workflow/test_bundle_recorder.py
@@ -89,7 +89,7 @@ class TestBundleRecorder(unittest.TestCase):
                 self.assertEqual(yaml.safe_load(f), data)
 
     def test_record_component_public(self):
-        self.bundle_recorder.public_url = 'https://ci.opensearch.org/ci/xyz'
+        self.bundle_recorder.public_url = 'https://ci.opensearch.org/ci/os-distro-prod'
         component = BuildManifest.Component(
             {
                 "name": "job_scheduler",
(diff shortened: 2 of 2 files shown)


bundle-workflow/tests/tests_assemble_workflow/test_bundle_recorder.py
```diff
@@ -87,3 +87,60 @@ def test_write_manifest(self):
             data = self.bundle_recorder.get_manifest().to_dict()
             with open(manifest_path) as f:
                 self.assertEqual(yaml.safe_load(f), data)
+
+    def test_record_component_public(self):
+        self.bundle_recorder.public_url = 'https://ci.opensearch.org/ci/xyz'
```
peterzhuamazon, line 92: Why is there a `xyz` or `dbc` prefix before meeting `builds` folder?

peternied: Its a non-sense identifier to which jenkins cluster actually has the data, see https://github.com/opensearch-project/opensearch-build/tree/main/deployment#overall-system

dblock: It feels like it should be `opensearch` or `openseach-dashboards` instead of `dbc`, possible?

dblock: It feels like it should be `opensearch` or `opensearch-dashboards` rather than `dbc`, possible?

dblock (approved):

bundle-workflow/src/assemble_workflow/bundle_recorder.py
```diff
@@ -32,7 +32,7 @@ def __get_tar_name(self, build):
 
     def __get_public_url_path(self, folder, rel_path):
         path = "{}/{}/{}/{}".format(folder, self.version, self.build_id, rel_path)
```
dblock, line 34: also known as `"/".join(folder, self.version, self.build_id, rel_path)`

bundle-workflow/tests/tests_assemble_workflow/test_bundle_recorder.py
peternied: Renamed, but we still need to update the schema but that will not require changes into this code as it should happen at the jenkinsfile level

saratvemulapalli:
Curious why `https://ci.opensearch.org/ci/dbc` is being used for build workflow?

peternied:
> Curious why `https://ci.opensearch.org/ci/dbc` is being used for build workflow?

Its an identifier to which jenkins cluster actually has the data, see https://github.com/opensearch-project/opensearch-build/tree/main/deployment#overall-system I've since renamed this value to make it more apparent what it could be used for.

All said, the url does not matter to this system as it is prefixing passed value on the location url and not attempting to interprete or understand it.

bundle-workflow/src/assemble_workflow/bundle_recorder.py
```diff
@@ -31,8 +31,8 @@ def __get_tar_name(self, build):
         return "-".join(parts) + ".tar.gz"
 
     def __get_public_url_path(self, folder, rel_path):
-        path = "{}/{}/{}/{}".format(folder, self.version, self.build_id, rel_path)
-        return urljoin(self.public_url, path)
+        path = "/".join((folder, self.version, self.build_id, rel_path))
+        return urljoin(self.public_url + "/", path)
```
dblock, line 35: Why don't you include `public_url` in the above list?

dblock (approved):

saratvemulapalli (approved):

PR #709, opened 2021-10-07
Added examples of how to check for tags, versions and releases.

dblock:

### Description

Added examples of how to check for tags, versions and releases.

cc: @mmakaria
 
### Check List

### DIFF
--- a/meta/README.md
+++ b/meta/README.md
@@ -6,6 +6,8 @@
   - [Add a New Component](#add-a-new-component)
   - [Create or Update Release Labels](#create-or-update-release-labels)
   - [Create a Release Issue](#create-a-release-issue)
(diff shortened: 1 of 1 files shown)


meta/README.md
```diff
@@ -56,3 +58,21 @@ We create release issues in all component repos that link back to a parent relea
...
+```
+
+The command will fail and show in red for every repository without a tag.
+
+### Check for Versions and Releases
+
+Check whether a version or a release exists.
+
+```
+meta exec "gh release list | grep -E '(1.1.0.*)$'"
```
peterzhuamazon, line 77: Are we using git for Release Tags and CLI for Versions/Release?
Should we use the same tool for both tasks?

Thanks.

Issue #620, opened 2021-09-27
Run OpenSearch Dashboards build in Jenkins


dblock:
### Is your feature request related to a problem? Please describe

Coming from https://github.com/opensearch-project/opensearch-build/issues/158, add the manifest to Jenkins and publish artifacts.

### Describe the solution you'd like

I want to see 1.1.0 dashboards in Jenkins and dropping the bundle.

Issue #699, opened 2021-10-05
[Bug] Maven 1.1.0 failed Central Sync due to multiple errors


peterzhuamazon:
### Describe the bug

[Bug] Maven 1.1.0 failed Central Sync due to multiple errors.

When we try to release Maven Artifacts for 1.1.0 OpenSearch to Maven Central through sonatype, we encounter these errors that block us from closing.

Here are all the issues we see:

<details><summary>Event: Failed: Javadoc Validation:</summary>
<p>

```
Event: Failed: Javadoc Validation
Tuesday, October 05, 2021 17:07:20 EDT (GMT-0400)
typeId	javadoc-staging
failureMessage	Missing: no javadoc jar found in folder '/org/opensearch/performanceanalyzer-rca/1.1.0.0'
```

</p>
</details>

<details><summary>Event: Failed: Checksum Validation:</summary>
<p>

```
Event: Failed: Checksum Validation
Tuesday, October 05, 2021 17:07:21 EDT (GMT-0400)
typeId	checksum-staging
failureMessage	Required SHA-1: '/org/opensearch/opensearch-performance-analyzer/1.1.0.0/opensearch-performance-analyzer-1.1.0.0-javadoc.jar.sha1'
failureMessage	Required MD5: '/org/opensearch/opensearch-performance-analyzer/1.1.0.0/opensearch-performance-analyzer-1.1.0.0-javadoc.jar.md5'
failureMessage	Required SHA-1: '/org/opensearch/opensearch-performance-analyzer/1.1.0.0/opensearch-performance-analyzer-1.1.0.0.pom.sha1'
failureMessage	Required MD5: '/org/opensearch/opensearch-performance-analyzer/1.1.0.0/opensearch-performance-analyzer-1.1.0.0.pom.md5'
failureMessage	Required SHA-1: '/org/opensearch/opensearch-performance-analyzer/1.1.0.0/opensearch-performance-analyzer-1.1.0.0-sources.jar.sha1'
failureMessage	Required MD5: '/org/opensearch/opensearch-performance-analyzer/1.1.0.0/opensearch-performance-analyzer-1.1.0.0-sources.jar.md5'
failureMessage	Required SHA-1: '/org/opensearch/opensearch-performance-analyzer/1.1.0.0/opensearch-performance-analyzer-1.1.0.0.module.sha1'
[...]

Issue #669, opened 2021-10-01
[Bug]: Distribution builds can have path colisitions


peternied:
Distribution builds are not namespace'd based on the type of artifact produced so an OpenSearch and OpenSearch Dashboard build could collide.

### Actual
The current path generated is `https://ci.opensearch.org/ci/dbc/builds/1.1.0/328/x64/manifest.yml`.

### Expected
It should have the distribution type in the path like `https://ci.opensearch.org/ci/dbc/opensearch/builds/1.1.0/328/x64/manifest.yml` and `https://ci.opensearch.org/ci/dbc/opensearch-dashboards/builds/1.1.0/328/x64/manifest.yml`

dblock:
What is `dbc`, btw?

dblock:
Let's make sure to address https://github.com/opensearch-project/opensearch-build/issues/661 when we change these.

PR #717, opened 2021-10-08
[Build] Add OpenSearch Dashboards to Jenkinsfile

kavilla:
### Description
Creates OpenSearch Dashboards Jenkinsfile basically copying what was
in the existing OpenSearch Jenkinsfile. But skipping snapshots since
it was not doing anything with snapshots and publishing to S3
bucket paths of `/builds/dashboards/` and `/bundles/dashboards/` which
has not been verified yet. It was done this way, while keeping OpenSearch
to build to `/builds` and `/bundles`, to reduce the impact of adding
OpenSearch Dashboards.

For consistency, moved the root lvl Jenkinsfile which was only  building
OpenSearch to `jenkins/opensearch`.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/620
 
### Check List


jenkins/opensearch_dashboards/Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+        stage('build') {
+            parallel {
+                stage('build-x86') {
+                    agent {
+                        docker {
+                            label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+                            image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdk14-node10.24.1-cypress6.9.1-20211005'
+                            // Unlike freestyle docker, pipeline docker does not login to the container and run commands
+                            // It use executes which does not source the docker container internal ENV VAR
+                            args '-e JAVA_HOME=/usr/lib/jvm/adoptopenjdk-14-hotspot'
```
dblock, line 36: Is Java required to build OpenSearch Dashboards?

jenkins/opensearch_dashboards/Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+    git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
+
+    sh "./build.sh manifests/$INPUT_MANIFEST"
+    sh './assemble.sh artifacts/manifest.yml'
+
+    script { manifest = readYaml(file: 'artifacts/manifest.yml') }
+    def artifactPath = "${manifest.build.version}/${OPENSEARCH_BUILD_ID}/${manifest.build.architecture}";
+
+    withAWS(role: 'opensearch-bundle', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+        s3Upload(file: 'artifacts', bucket: "${ARTIFACT_BUCKET_NAME}", path: "builds/dashboards/${artifactPath}")
```
dblock, line 100: is there any intermediate output for dashboards in builds that's useful? In opensearch this is artifacts being published to maven for example; so do we need to upload it?

jenkins/opensearch_dashboards/Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+    writeFile(file: "notifications/${STAGE_NAME}.msg", text: message)
+    stash(includes: "notifications/*" , name: "notifications-${STAGE_NAME}")
+}
+
+/** Load all message in the jenkins queue and append them with a leading newline into a mutli-line string */
+String getAllJenkinsMessages() {
+    script {
+        // Stages must be explicitly added to prevent overwriting
+        // See https://ryan.himmelwright.net/post/jenkins-parallel-stashing/
+        def stages = ['build-x86', 'build-arm64']
```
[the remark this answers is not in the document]
dblock: Let's fix it in both please.

jenkins/opensearch_dashboards/Jenkinsfile
```diff
@@ -0,0 +1,143 @@
+pipeline {
+    agent none
+    environment {
+        OPENSEARCH_BUILD_ID = "${BUILD_NUMBER}"
```
dblock, line 4: OPENSEARCH_DASHBOARDS_BUILD_ID?

peternied: Alternative proposal, keep it as BUILD_NUMBER

jenkins/opensearch_dashboards/Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+    }
+    triggers {
+      parameterizedCron '''
+        H */2 * * * %INPUT_MANIFEST=1.1.0/opensearch-dashboards-1.1.0.yml
```
peternied, line 8: This will trigger every 2 hours, is that enough/not enough for your use case, how long to the builds take?

jenkins/opensearch_dashboards/Jenkinsfile
kavilla: We can build the "snapshots" which would does not include the --release flag to the build [here](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/components/OpenSearch-Dashboards/build.sh#L79). Should publish that as well to the S3 bucket?

dblock:
> > rename underscore opensearch_dashboards to `opensearch-dashboards`
> 
> I copied the format of `src/build_workflow/opensearch_dashboards` should we modify that as well?

That one is a source code namespace, while this is a name of a product, albeit lowercase.

kavilla:
> cc: @zelinh, in this PR there's some convention introduced for separating OpenSearch vs. OpenSearch Dashboards build, I don't want to end up with 3 locations, where are we with #669?

so this issue blocked by https://github.com/opensearch-project/opensearch-build/issues/669, 669 implies the build path as well.

src/build_workflow/build_target.py
```diff
@@ -27,7 +27,7 @@ def __init__(
         build_id=None,
         output_dir="artifacts",
     ):
-        self.build_id = os.getenv("OPENSEARCH_BUILD_ID") or build_id or uuid.uuid4().hex
+        self.build_id = os.getenv("OPENSEARCH_BUILD_ID") or os.getenv("OPENSEARCH_DASHBOARDS_BUILD_ID") or build_id or uuid.uuid4().hex
```
dblock, line 30: This definitely tells me that we should merge these into a single variable. Then I think there was a comment on why we rename `BUILD_NUMBER` to `OPENSEARCH_DASHBOARDS_BUILD_ID`. Feels like all these things could be collapsed into a single `BUILD_NUMBER`. Wouldn't it make things a lot simpler?

Plus we'll also refactor Jenkinsfile into a shared library at some point I'm sure to avoid all this code duplication.

dblock:
> so this issue blocked by #669, 669 implies the build path as well.

I am ok with merging this and doing #699 on top.

PR #718, opened 2021-10-11
Renaming bundle which is created as part of bundle build to dist.

### LINKED ISSUE #479: Rename bundle that is created as part of bundle build into bin dist or min

Bundle build produces:

```
bundle
core-plugins
libs
maven
plugins
```

The `bundle` name is confusing and contains only the -min artifact. We should rename it to maybe `bin`?

Rishikesh1159:
Signed-off-by: Rishikesh Pasham <rishireddy1159@gmail.com>

### Description
Bundle which is created as part of bundle build is confusing with it's name. So, it has been renamed from Bundle to dist.
 
### Issues Resolved
#479 
 
### Check List

### DIFF
--- a/src/assemble_workflow/bundle.py
+++ b/src/assemble_workflow/bundle.py
@@ -107,5 +107,5 @@ class Bundle(ABC):
             iter([c for c in build_components if "dist" in c.artifacts]), None
         )
         if min_bundle is None:
-            raise ValueError('Missing min "bundle" in input artifacts.')
+            raise ValueError('Missing min "dist" in input artifacts.')
         return min_bundle


src/manifests/manifest.py
```diff
@@ -12,7 +12,7 @@
 
 class Manifest(ABC):
     SCHEMA = {
-        "schema-version": {"required": True, "type": "string", "allowed": ["1.0"]}
+        "schema-version": {"required": True, "type": "string", "allowed": [""]}
```
dblock, line 15: This says that only a blank string value is allowed, and the corresponding test was updated as such. This isn't the case. Remove the `allowed` field altogether, which would allow anything, and write a test that checks that you can put 1.0, and that you cannot put no value.

tests/tests_manifests/data/min.yml
```diff
@@ -1 +1 @@
-schema-version: '1.0'
+schema-version: ''
```
dblock, line 1: This should be reverted, a minimal manifest is one that has a schema version.

I think you need 3 tests:

- schema-version: no value -> raises an exception
- schema-version: empty string -> raises an exception
- schema-version: something -> works

dblock (changes requested):
Almost there, see below.

tests/tests_manifests/test_manifest.py
```diff
@@ -34,12 +34,21 @@ def test_manifest_is_abstract(self):
         )
 
     def test_invalid_version(self):
```
dblock, line 36: Split this unit test into multiple individual tests for each invalid-schema-version- scenario.

Rishikesh1159: Sure @dblock I will make these changes. Thank you for your input.

src/assemble_workflow/bundle.py
```diff
@@ -104,7 +104,7 @@ def __get_plugins(self, build_components):
 
     def __get_min_bundle(self, build_components):
         min_bundle = next(
-            iter([c for c in build_components if "bundle" in c.artifacts]), None
+            iter([c for c in build_components if "dist" in c.artifacts]), None
         )
         if min_bundle is None:
             raise ValueError('Missing min "bundle" in input artifacts.')
```
VachaShah, line 110: Would the error message change to dist as well?

PR #729, opened 2021-10-12
Auto-generate OpenSearch Dashboards manifests.

dblock:

### Description

Added support for auto-generating OpenSearch Dashboards manifests. 

Running `./manifests.sh update` now would create the missing 1.0.1 and a new 1.2.0 OpenSearch Dashboards manifest.  
 
### Check List


src/manifests_workflow/component.py
```diff
@@ -15,10 +18,18 @@ def __init__(self, name, repo, snapshot=False, checks=[]):
...
+    def branches(self, uri):
+        branches = ["main"]
+        remote_branches = (
+            subprocess.check_output(
+                f"git ls-remote {uri} refs/heads/* | cut -f2 | cut -d/ -f3", shell=True
+            )
+            .decode()
+            .split("\n")
+        )
+        # return "main" and "x.y" branches only
```
peternied, line 30: This comment is misleading because main is included at line 22, I'd recommend splitting the comments or add a function comment that indicates what should be expected inside of branches instead of the inline comment

src/manifests_workflow/input_manifests_opensearch.py
```diff
@@ -0,0 +1,99 @@
...
+                        release_version = ".".join(component_version.split(".")[:3])
+                        if release_version not in main_versions.keys():
+                            main_versions[release_version] = []
+                        main_versions[release_version].append(component)
+                        logging.info(
+                            f"{component.name}#main is version {release_version} (from {component_version})"
+                        )
+                except subprocess.CalledProcessError as err:
+                    logging.warn(
+                        f"Error getting version of {component.name}: {str(err)}, ignored"
```
peternied, line 70: This happen enough that its worth swallowing? Seems like we would want to error out

src/manifests_workflow/input_manifests_opensearch.py
```diff
@@ -0,0 +1,99 @@
...
+import subprocess
+
+from manifests.input_manifest import InputManifest
+from manifests_workflow.component_opensearch import ComponentOpenSearch
+from manifests_workflow.component_opensearch_min import ComponentOpenSearchMin
+from manifests_workflow.input_manifests import InputManifests
+from system.temporary_directory import TemporaryDirectory
+
+
+class InputManifestsOpenSearch(InputManifests):
```
peternied, line 18: Why 2 different classes?  seems like this could be done with 2 classes that inherit or compose from a standard set of instructions

Issue #38, opened 2021-05-18
OpenSearch / OpenSearch Dashboards - macOS distribution (X64)


bbarani:
Is your feature request related to a problem? Please describe.
No

Describe the solution you'd like
Provide signed macOS (X64) artifact for OpenSearch / OpenSearch Dashboards

Describe alternatives you've considered
Generating the artifact using the source code

Additional context
Currently there are macOS (X64) artifacts for OpenSearch / OpenSearch Dashboards

skkosuri-amzn:
Please prioritize this to speed up development on mac.

dbbaughe:
@bbarani @saratvemulapalli @peternied @dblock 

Is this scheduled for 1.3 release?

This [comment](https://github.com/opensearch-project/OpenSearch/issues/1676#issuecomment-988404726) seems to imply it, but the roadmap for infra seemed to not have it on there or scheduled for 1.3 release.

Can we get some clarification on when this is going to be completed?

saratvemulapalli:
My comment is based of the roadmap: https://github.com/orgs/opensearch-project/projects/1#column-14092180. Assuming it's up to date.

dblock:
Someone contributed a formulae to Homebrew that builds and runs OpenSearch-min and OpenSearch Dashboards from source. 

https://github.com/Homebrew/homebrew-core/blob/master/Formula/opensearch.rb

See https://github.com/opensearch-project/opensearch-build/issues/1314#issuecomment-993789585

bbarani:
We are currently working on creating Linux based distributions (RPM and Deb) at this point in time. Please feel free to provide your inputs on this issue #741 to prioritize other non Linux distributions for us to pickup post that effort. As always, we welcome any contribution towards the development of additional distributions.

saratvemulapalli:
@bbarani with the launch of RPM/DEB, are we looking to add support for MacOS?

jmazanec15:
@bbarani is it possible to release an ARCHIVE testDistribution for Mac? Related to https://github.com/opensearch-project/OpenSearch/issues/1676.

setiah:
> @bbarani is it possible to release an ARCHIVE testDistribution for Mac? Related to [opensearch-project/OpenSearch#1676](https://github.com/opensearch-project/OpenSearch/issues/1676).

[...]

PR #740, opened 2021-10-13
WIP: Build on MacOS.

dblock:

### Description

With this change `build.sh manfests/1.1.0/opensarch-1.1.0.yml` works up until k-nn, which is #737.

Current scripts for OpenSearch hard-code `linux` when they see `-a x64`. Our options are:

1. This change. Quick but dirty.
2. Reuse `-a` (or replace it) and change it from "architecture" to "target", so `-a x64` to `-a darwin-x64` and `-a linux-x64`.
3. Add another option for passing on system, e.g. `-m darwin`.
4. Remove system and arch options and use current platform in the build scripts.

Thoughts? /cc:@peternied @bbarani 
 
### Issues Resolved

Part of #38 

### Check List

### DIFF
--- a/scripts/components/k-NN/build.sh
+++ b/scripts/components/k-NN/build.sh
@@ -79,6 +79,11 @@ if [ "$ARCHITECTURE" = "arm64" ]; then
     sed -i -e 's/-march=native/-march=armv8-a/g' external/nmslib/similarity_search/CMakeLists.txt
 fi
 
+if [ "$JAVA_HOME" = "" ]; then
+    export JAVA_HOME=`/usr/libexec/java_home`
+    echo "SET JAVA_HOME=$JAVA_HOME"
+fi
+
 cmake .
 make
 


peternied:
Reversing my stance after read up on those issues and quickly synced offline - Yup we need to support building on the mac, we have gaps around docker images and we will need to mitigate those so we can support top down building for linux, mac and windows.

peternied:
> Current scripts for OpenSearch hard-code linux when they see -a x64. Our options are:
> 
> 1. This change. Quick but dirty.
> 2. Reuse -a (or replace it) and change it from "architecture" to "target", so -a x64 to -a darwin-x64 and -a linux-x64.
> 3. Add another option for passing on system, e.g. -m darwin.
> 4. Remove system and arch options and use current platform in the build scripts.

In order of preference
1. Not blocking contributions is important, so I would weight this highly, however if we have cycles #3 is preferred

3. I think is this a good north star, it allows for flexibility in systems where we have common behavior or to diverge intelligently.  Might require a lot of param plumbing

2. Not a fan because it conflates architecture and platform into a single parameter, this will get merged/unmerged and when we have a noarch/multi-arch target it goes off the rails.

4. This defers the problem into the internals of the build system, which could work, but I feel like this would lead to some components building for arm, arm64, aarch64, etc.. if we have to take their word for it it means we can have deviation in terminology which doesn't make as much sense when we will be creating a distribution that should have a single platform/architecture associated with it.

PR #746, opened 2021-10-14
Add distribution type to url and change url location

zelinh:

### Description
Add distribution build to avoid path collision for OpenSearch and OpenSearch Dashboard. 
Create a new file to process the public_url to use it as a absolute path in `bundle_recorder.py` 
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/669
 
### Check List

### DIFF
--- a/src/assemble_workflow/bundle_recorder.py
+++ b/src/assemble_workflow/bundle_recorder.py
@@ -6,52 +6,50 @@
 
 import os
 from urllib.parse import urljoin
-
 from manifests.bundle_manifest import BundleManifest
-
+from assemble_workflow.public_urls import PublicURLpath
 
 class BundleRecorder:
     def __init__(self, build, output_dir, artifacts_dir):
         self.output_dir = output_dir
         self.build_id = build.id
-        self.public_url = os.getenv("PUBLIC_ARTIFACT_URL", None)
         self.version = build.version
         self.tar_name = self.__get_tar_name(build)
         self.artifacts_dir = artifacts_dir
         self.arch = build.architecture
+        self.urls = PublicURLpath(build).urls_path
         self.bundle_manifest = self.BundleManifestBuilder(
             build.id,
             build.name,
             build.version,
             build.architecture,
-            self.__get_tar_location(),
+            self.__get_tar_location(self.urls["bundles"]),
         )
 
     def __get_tar_name(self, build):
         parts = [build.name.lower().replace(" ", "-"), build.version, "linux", build.architecture]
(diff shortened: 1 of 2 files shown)


src/assemble_workflow/public_urls.py
```diff
@@ -0,0 +1,22 @@
...
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import os
+from urllib.parse import urljoin
+
+class PublicURLpath:
+    def __init__(self, build): 
+        self.public_url = os.getenv("PUBLIC_ARTIFACT_URL", None)
```
peternied, line 12: We should stop using the environment variable completely and move to having a base url passed in via the assembly.sh command which is then passed into the bundle recorder.

By using the environment variable this is effectively a hidden interface to this component, that makes it hard to test and debug.  If we are explicitly passing this value the behavior is much more transparent

src/assemble_workflow/bundle_recorder.py
```diff
@@ -6,60 +6,58 @@
...
-    def __get_public_url_path(self, folder, rel_path):
-        path = "/".join((folder, self.version, self.build_id, self.arch, rel_path))
-        return urljoin(self.public_url + "/", path)
+    def __get_public_url_path(self, url_path, rel_path):
+        return urljoin(url_path + "/", rel_path)
 
-    def __get_location(self, folder_name, rel_path, abs_path):
-        if self.public_url:
-            return self.__get_public_url_path(folder_name, rel_path)
+    def __get_location(self, url_path, rel_path, abs_path):
```
peternied, line 36: Don't use a parameter, have it as a class variable, the base url should be passed in during construction of the bundle recorder since it should be the same value for every bundled artifact

src/assemble_workflow/public_urls.py
zelinh: The thing is do we still support the case that no base url is assigned? Currently if the environment variable `PUBLIC_ARTIFACT_URL` doesn't exist, we will generate the location path based on local directory instead of an URL.

dblock: This URL is only used to decide what to write in the .yml output as a root so that various artifacts can be directly downloaded later. For local scenarios it's blank, and it's the local directory where we drop artifacts that needs to change to include product name, architecture, and platform.

zelinh:
I will close this draft PR and create a new one regarding of #669

PR #730, opened 2021-10-12
Jenkins misc cleanup

dblock:
### Description

- Stop building OpenSearch 1.1.0.
- Rename x86 to x64 in the OpenSearch Jenkins.
- Remove unnecessary ID varaibles.
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #748, opened 2021-10-14
Support -p platform in build scripts.

dblock:

### Description

Support for #740, adds `-p platform` to build scripts so we can pass `linux` or `darwin` around and default it to `uname -s`. This required re-overwriting common-utils, job-scheduler and alerting scripts. 

ScriptFinder is incorrectly picking the override from the repo with source code. The build repo should be the one always picked first because that's where we'd make a last minute fix, updated. 

Auto-resolving JAVA_HOME on MacOS.

### Check List

### DIFF
--- a/src/build_workflow/build_target.py
+++ b/src/build_workflow/build_target.py
@@ -7,13 +7,14 @@
 import os
 import uuid
 
-from system.arch import current_arch
+from system.os import current_arch, current_platform
 
 
 class BuildTarget:
     build_id: str
     name: str
     version: str
+    platform: str
     arch: str
     snapshot: bool
     output_dir: str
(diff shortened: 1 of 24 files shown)


src/build_workflow/build_target.py
```diff
@@ -7,31 +7,34 @@
...
         version,
+        platform=None,
         arch=None,
         name=None,
         snapshot=True,
         build_id=None,
         output_dir="artifacts",
     ):
-        self.build_id = os.getenv("OPENSEARCH_BUILD_ID") or os.getenv("OPENSEARCH_DASHBOARDS_BUILD_ID") or build_id or uuid.uuid4().hex
+        self.build_id = os.getenv("BUILD_NUMBER") or build_id or uuid.uuid4().hex
```
dblock, line 32: Side effect of https://github.com/opensearch-project/opensearch-build/pull/730 that we forgot to fix.

peternied: Yea, looks like those bundle numbers became `8c363d37f71349b1add343c0c48945a7`

peternied (approved):

VachaShah (approved):
LGTM!

Issue #739, opened 2021-10-13
[Bug]: Dashboards build agent can't build on ARM64 due to chromium missing


kavilla:
### Describe the bug

When running the build for OpenSearch Dashboards in Jenkins it will attempt to use a build agent that does not have chromium installed which is needed to even bootstrap OpenSearch Dashboards.

Might be more difficult than initially looked at because even if you try to just pass it to the Dashboards dependencies within the [Jenkinsfile](https://github.com/opensearch-project/opensearch-build/blob/main/docker/ci/dockerfiles/ci-runner.al2.dockerfile#L26) it will have no clue what the package is. Even tried to install the EPEL release package [[here](https://aws.amazon.com/premiumsupport/knowledge-center/ec2-enable-epel/)] and then install chromium as suggested online in some locations but any type of chromium is still not available after installing that package.

### To reproduce

I have tried using an AL2 instance with the Dockerfile for CentOS. Will try that but if you try to run it using the Dockerfile for AL2, you won't be able to install chromium still. Then if you try to build OpenSearch Dashboards you will the error message while running `yarn osd bootstrap`:

```
Directory: /tmp/tmpzvh8755p/OpenSearch-Dashboards/node_modules/puppeteer
Output:
The chromium binary is not available for arm64: 
If you are on Ubuntu, you can install with: 

 apt-get install chromium-browser

/tmp/tmpzvh8755p/OpenSearch-Dashboards/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:112
            throw new Error();
            ^

Error
    at /tmp/tmpzvh8755p/OpenSearch-Dashboards/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:112:19
    at FSReqWrap.oncomplete (fs.js:153:21)
```

### Expected behavior

Bootstrapping OpenSearch Dashboards builds successfully.

### Host / Environment

AL2 ARM64

PR #749, opened 2021-10-14
[Build] add arch arg to build workflow

kavilla:
### Description
Build target accepts an arch already so this enables passing the
arch while running the build workflow. This allows building target
operating systems that are different from the operating system of
the build agent.

Use case: chromium package not available on ARM64 but it is just
used for building so if we can just build on x64 then assemble
on ARM64 we can get a bundle that we are expecting.

Updated Jenkinsfile for OpenSearch Dashboards to utilize this
functionality.

### Issues Resolved
 
### Check List


jenkins/opensearch-dashboards/Jenkinsfile
```diff
@@ -42,22 +44,49 @@ pipeline {
...
+                                docker {
+                                    label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+                                    image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdk14-node10.24.1-cypress6.9.1-20211005'
+                                    alwaysPull true
+                                }
+                            }
+                            steps {
+                                script {
+                                    build("arm64")
+                                    stash includes : 'artifacts/**', name: 'buildArtifacts'
```
peternied, line 61: Stash's documentation says that it should be used for limited set of files at less than 100mb total and there are suggestions for other methods with https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#stash-stash-some-files-to-be-used-later-in-the-build

As is if this grows we could have strange failures

jenkins/opensearch-dashboards/Jenkinsfile
```diff
@@ -42,22 +44,49 @@ pipeline {
...
-                stage('build-arm64') {
-                    agent {
-                        docker {
-                            label 'Jenkins-Agent-al2-arm64-c6g4xlarge-Docker-Host'
-                            image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdk14-node10.24.1-cypress6.9.1-20211005'
-                            alwaysPull true
+                stage('arm64') {
+                    agent none
+                    stages {
+                        stage('build-arm64') {
```
peternied, line 50: This is pretty hacky - could we create an issue and resolve arm build process in a separate change?  I would be OK just removing it for the time being

Issue #732, opened 2021-10-12
Split, simplify and move usage of CI runners between OpenSearch and OpenSearch Dashboards


dblock:
### Is your feature request related to a problem? Please describe

CI runners are the kitchen sink of everything to support OpenSearch and OpenSearch Dashboards. We *will* run into conflicts sooner than later. We will also want to build multiple versions at the same time which will make it difficult to track what's what. For example, in `centos7-x64-arm64-jdk14-node10.24.1-cypress6.9.1-20211005`, OpenSearch doesn't need node or cypress, while OpenSearch Dashboards doesn't need Java. Then, we're going to lower the Java version for OpenSearch, so we'll need to use multiple images.

### Describe the solution you'd like

- [ ] Split images, e.g. `centos7-x64-arm64-jdk14` and `centos7-x64-arm64-node10.24.1-cypress6.9.1`.
- [ ] Move which build image to use into the build manifest so that it sticks to the version being built and not to the Jenkinsfile.

reta:
We certainly could do that, pre-baking the images has the benefits that we don't need to download and install JDKs every time (+ `gradle` as well), so in general - builds are faster and more stable.

peterzhuamazon:
> @peterzhuamazon thanks a lot for merging the PRs, could you please ping us when the new images are available so we could alter the build pipelines to use the per-branch JDK specifications, thank you!

Will ping.

peterzhuamazon:
https://hub.docker.com/r/opensearchstaging/ci-runner/tags?page=1&ordering=last_updated

gaiksaya:
How did we build these images? Manually or is that automated?

PR #754, opened 2021-10-15
Updated CI image to include JDK-8/11/17 distributions, kept JDK-14 as the default one

reta:
Signed-off-by: Andriy Redko <andriy.redko@aiven.io>

### Description
Updated CI image to include JDK-8/11/17 distributions, kept JDK-14 as the default one so the 1.x branches will continue to build properly.

```
declare -x JAVA11_HOME="/opt/java/openjdk-11"
declare -x JAVA14_HOME="/usr/lib/jvm/adoptopenjdk-14-hotspot"
declare -x JAVA17_HOME="/opt/java/openjdk-17"
declare -x JAVA8_HOME="/opt/java/openjdk-8"
declare -x JAVA_HOME="/usr/lib/jvm/adoptopenjdk-14-hotspot"
```
 
### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-build/issues/732, part of https://github.com/opensearch-project/OpenSearch/issues/1351
 
### Check List

### DIFF
--- a/docker/ci/dockerfiles/ci-runner.centos7.dockerfile
+++ b/docker/ci/dockerfiles/ci-runner.centos7.dockerfile
@@ -68,12 +68,12 @@ RUN set -eux; \
         regex="temurin([0-9]+)[-]"; \
         if [[ $jdk =~ $regex ]]; then \
             MAJOR=${BASH_REMATCH[1]}; \
-            curl -LfsSo /tmp/openjdk.tar.gz ${BINARY_URL}; \
-            echo "${ESUM} */tmp/openjdk.tar.gz" | sha256sum -c -; \
+            curl -LfsSo /tmp/openjdk-${MAJOR}.tar.gz ${BINARY_URL}; \
+            echo "${ESUM} */tmp/openjdk-${MAJOR}.tar.gz" | sha256sum -c -; \
             mkdir -p /opt/java/openjdk-${MAJOR}; \
             cd /opt/java/openjdk-${MAJOR}; \
-            tar -xf /tmp/openjdk.tar.gz --strip-components=1; \
-            rm -rf /tmp/openjdk.tar.gz; \
+            tar -xf /tmp/openjdk-${MAJOR}.tar.gz --strip-components=1; \
+            rm -rf /tmp/openjdk-${MAJOR}.tar.gz; \
             echo "export JAVA${MAJOR}_HOME=/opt/java/openjdk-${MAJOR}" >> /etc/profile.d/java_home.sh; \
         fi; \
     done;


reta:
@peterzhuamazon could you please take a look? Thank you! 

@dblock sadly we cannot cut off JDK-14 in one shot (we will break 1.x builds due to https://bugs.openjdk.java.net/browse/JDK-8209058) but with new CI images we could set JAVA_HOME from build manifests.

reta:
> Next question, why aren't we installing the other JDKs with yum install adoptopenjdk-... as well? Does it not support side-by-side installations?

A bit relates to previous comments, `Adoptium` at this moment does not have CentOS repositories (I didn't find any) and the recommended way to install distributions is to use archives.

dblock:
> > Next question, why aren't we installing the other JDKs with yum install adoptopenjdk-... as well? Does it not support side-by-side installations?
> 
> A bit relates to previous comments, `Adoptium` at this moment does not have CentOS repositories (I didn't find any) and the recommended way to install distributions is to use archives.

Thanks, makes sense.

reta:
@dblock just played with `sdk-man`, sadly there is no JDK-14 distributions anymore (AdoptOpenJDK is gone as a source) available. So we could only install JDK 8/11/17 from Adoptium. I will prepare the change and wait for @peterzhuamazon @bbarani feedback regarding their preferences.

dblock (approved):

docker/ci/dockerfiles/ci-runner.centos7.dockerfile
```diff
@@ -38,6 +38,46 @@ RUN yum install -y nss xorg-x11-fonts-100dpi xorg-x11-fonts-75dpi xorg-x11-utils
 # Add Yarn dependencies
 RUN yum groupinstall -y "Development Tools" && yum clean all && rm -rf /var/cache/yum/*
 
+# Downloads JDK-8, JDK-11 and JDK-17 distributions using Eclipse Adoptium project.
+# The distributions are extracted to /opt/java/ folder with environment variables JAVA8_HOME,
+# JAVA11_HOME and JAVA17_HOME exported and pointing at respective ones.
+RUN set -eux; \
+    ARCH="$(objdump="$(command -v objdump)" && objdump --file-headers "$objdump" | awk -F '[:,]+[[:space:]]+' '$1 == "architecture" { print $2 }')"; \
```
peterzhuamazon, line 45: Can you help me understand why we need objdump to check ARCH?

reta: :+1: Updated, thank you

docker/ci/dockerfiles/ci-runner.centos7.dockerfile
```diff
@@ -38,6 +38,46 @@ RUN yum install -y nss xorg-x11-fonts-100dpi xorg-x11-fonts-75dpi xorg-x11-utils
...
+         echo "Unsupported arch: ${ARCH}"; \
+         exit 1; \
+         ;; \
+    esac; \
+    for jdk in ${JDKS}; do \
+        ESUM=$(echo ${jdk} | cut -d '@' -f1); \
+        BINARY_URL=$(echo ${jdk} | cut -d '@' -f2); \
+        regex="temurin([0-9]+)[-]"; \
+        if [[ $jdk =~ $regex ]]; then \
+            MAJOR=${BASH_REMATCH[1]}; \
```
dblock, line 70: Is this necessary? Don’t we start with a well known set?

reta: Yes, basically we have URLs, with concrete version, like 11.0.12 fe., but we need only major, `11`. The way to extract that is to a) manually provide it b) extract from URL. So the regex in this case helps to automate the process, adding new JDK is just checksum + URL.

docker/ci/dockerfiles/ci-runner.centos7.dockerfile
```diff
@@ -38,6 +38,46 @@ RUN yum install -y nss xorg-x11-fonts-100dpi xorg-x11-fonts-75dpi xorg-x11-utils
...
+            curl -LfsSo /tmp/openjdk.tar.gz ${BINARY_URL}; \
```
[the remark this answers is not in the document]
reta: :+1: updated, thanks

peterzhuamazon (approved):
Thanks, great changes.
Wonder if you can do the same for AL2 image as well?

Issue #610, opened 2021-09-27
OpenSearch Dashboards default config


kavilla:
### Is your feature request related to a problem? Please describe

For the new build system for [Dashboards](https://github.com/opensearch-project/opensearch-build/issues/573), it is not building in the config like [previous build systems](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/legacy/tar/linux/opensearch-dashboards-tar-build.sh#L142-L143).

### Describe the solution you'd like

config/opensearch_dashboards.yml should be built from the repo into the bundle.

PR #758, opened 2021-10-16
[Bundle] include default files

kavilla:
### Description
Create the bundles with default files for OpenSearch and OpenSearch
Dashboards.

These files include:
* opensearch-tar-install.sh
* config/opensearch_dashboards.yml

However, we do not want to include opensearch-tar-install.sh into
the OpenSearch Dashboards bundle because the script will just fail
to run.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/610
 
### Check List

### DIFF
--- a/src/assemble_workflow/bundle.py
+++ b/src/assemble_workflow/bundle.py
@@ -54,6 +54,10 @@ class Bundle(ABC):
             f'{post_install_script} -a "{self.artifacts_dir}" -o "{self.archive_path}"'
         )
 
+    @abstractmethod
+    def copy_default_files(self):
+        logging.info('Copied default files')
+
     def build_tar(self, dest):
         tar_name = self.bundle_recorder.tar_name
         with tarfile.open(tar_name, "w:gz") as tar:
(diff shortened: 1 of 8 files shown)


src/assemble_workflow/bundle.py
```diff
@@ -109,3 +113,23 @@ def __get_min_bundle(self, build_components):
...
         return min_bundle
+
+    def copy_file(self, source, dest):
+        file = os.path.realpath(
+            os.path.join(
+                os.path.dirname(os.path.abspath(__file__)),
+                source,
+            )
+        )
+        if not os.path.isfile(file):
```
dblock, line 124: You can't exit in the middle of code. The next call to `shutil.copy2` will fail with an exception as needed.

dblock (changes requested):
This seems to replace a hard-coded copy by another hard-coded copy, so I think we're on the wrong track. 

During the assemble phase we iterate through all plugins and install them. Part of that we run a post-install script which can do things like copying additional files, e.g. [k-nn/install.sh](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/components/k-NN/install.sh#L42). Add an install step for the min artifact, and copy the custom .sh script for OpenSearch and the custom dashboards_config.yml in install.sh that will live under `scripts/.../OpenSearch` and `scripts/.../OpenSearch-Dashboards`. Then this entire code can be deleted.

PR #762, opened 2021-10-19
[Bundle] post-install for min applications

kavilla:
### Description
Adding install script for OpenSearch and OpenSearch Dashboards so
the bundles include default files.

These files include:
* opensearch-tar-install.sh
* config/opensearch_dashboards.yml

However, we did not want to include opensearch-tar-install.sh into
the OpenSearch Dashboards bundle because the script will fail to
execute as it is only expected to be an OpenSearch project.

Previous PR:
https://github.com/opensearch-project/opensearch-build/pull/758

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/610
 
### Check List

### DIFF
--- a/scripts/components/OpenSearch-Dashboards/install.sh
+++ b/scripts/components/OpenSearch-Dashboards/install.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+set -ex
+
+function usage() {
+    echo "Usage: $0 [args]"
+    echo ""
+    echo "Arguments:"
+    echo -e "-a ARTIFACTS\t[Required] Location of build artifacts."
+    echo -e "-o OUTPUT\t[Required] Output path."
+    echo -e "-h help"
+}
+
+while getopts ":h:a:o:" arg; do
+    case $arg in
+        o)
+            OUTPUT=$OPTARG
+            ;;
+        a)
+            ARTIFACTS=$OPTARG
+            ;;
+        h)
+            usage
+            exit 1
+            ;;
+        ?)
+            echo "Invalid option: -${arg}"
+            exit 1
+            ;;
+    esac
+done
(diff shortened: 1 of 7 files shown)


src/assemble_workflow/bundle.py
```diff
@@ -39,6 +40,12 @@ def __init__(self, build_manifest, artifacts_dir, bundle_recorder):
         self.min_tarball_path = self._copy_component(self.min_tarball, "dist")
         self.__unpack_min_tarball(self.tmp_dir.name)
 
+    def install_core(self):
```
kavilla, line 43: open to suggestions on what to call this function.

scripts/components/OpenSearch-Dashboards/install.sh
```diff
@@ -0,0 +1,40 @@
...
+            ;;
+        ?)
+            echo "Invalid option: -${arg}"
+            exit 1
+            ;;
+    esac
+done
+
+## Setup default config
+(DIR="$(dirname "$0")"; echo $DIR; cd $DIR; cp ../../../config/opensearch_dashboards.yml "$OUTPUT/config/")
```
dblock, line 40: This is a bash script, no need to save on lines, break this up!

src/assemble_workflow/bundle.py
dblock: install_min

src/assemble_workflow/bundle.py
```diff
@@ -39,6 +39,12 @@ def __init__(self, build_manifest, artifacts_dir, bundle_recorder):
...
+        post_install_script = ScriptFinder.find_install_script(self.min_tarball.name)
```
dblock, line 43: To be fair, it's the `install_script` or just `script`

dblock (approved):
Nits mostly, otherwise looks good.

scripts/components/OpenSearch/install.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
peternied, line 1: I would really prefer that this is python classes rather than shell scripts, its more testable and supports reuse

PR #763, opened 2021-10-19
[Build] support building targets for inputted platform and arch 

kavilla:
### Description
Adding `platform` and `arch` as a build argument to be passed to
the build target. This enables cross-platform builds, for example,
we can build linux x64 on a linux ARM64 machine.

We would want to do this because some of the libraries that are
required for building (not for runtime) are not available for
certain operating systems.

Also updated the OpenSearch Dashboards Jenkinsfile to pass the
platform and arch.

### Issues Resolved
 
### Check List

### DIFF
--- a/src/build_workflow/build_args.py
+++ b/src/build_workflow/build_args.py
@@ -14,6 +14,8 @@ class BuildArgs:
     snapshot: bool
     component: str
     keep: bool
+    platform: str
+    arch: str
 
     def __init__(self):
         parser = argparse.ArgumentParser(description="Build an OpenSearch Bundle")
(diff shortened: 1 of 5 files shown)


src/build_workflow/build_args.py
```diff
@@ -52,6 +60,8 @@ def __init__(self):
         self.snapshot = args.snapshot
         self.component = args.component
         self.keep = args.keep
+        self.platform = args.platform
+        self.arch = args.arch
```
dblock, line 64: You need defaults for these or it will be blank and break the existing OpenSearch build.

src/build_workflow/build_args.py
```diff
@@ -36,6 +38,12 @@ def __init__(self):
             action="store_true",
             help="Do not delete the working temporary directory.",
         )
+        parser.add_argument(
```
dblock, line 41: Limit values to the supported platforms and architectures.

PR #767, opened 2021-10-19
Build OpenSearch on Windows

dblock:
### Description

Fixes tests on Windows:
 
1. Use `os.path.join` vs. concatenating string paths or paths with a `/` separator.
2. Git on windows writes read-only files that cannot be unlinked, use workaround in https://stackoverflow.com/questions/1213706/what-user-do-python-scripts-run-as-in-windows.
3. Fixup load paths in `tests/tests_build_workflow`.
4. Make OpenSearch-min build script work on Windows.

### Issues Resolved

On top of #772. 

Part of #33. 

With this PR and fixes in https://github.com/opensearch-project/OpenSearch/pull/1412 you can successfully build OpenSearch 2.0.0 min with the following manifest.

```
build:
  name: OpenSearch
  version: 2.0.0
components:
  - name: OpenSearch
    repository: https://github.com/dblock/OpenSearch.git
    ref: "fix-windows-build"
    checks:
      - gradle:publish
      - gradle:properties:version
schema-version: '1.0'
```
  
```
$ ./build.sh manifests/2.0.0/opensearch-2.0.0.yml --snapshot

$ find artifacts/dist/
artifacts/dist/
artifacts/dist/opensearch-min-2.0.0-SNAPSHOT-windows-x64.zip
```

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #768, opened 2021-10-20
Align configurations of black and isort.

dblock:

### Description

Adds a pyproject.toml with black and isort configurations aligned, and on the same line length as flake8. With this change one does not attempt to auto-correct the other. Ran the auto-formatter on all of the source once which is all the other changes.
  
Extracted from https://github.com/opensearch-project/opensearch-build/pull/767.

### Check List

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,7 @@
+[tool.black]
+line-length = 160
+target-version = ['py37']
+
+[tool.isort]
+profile = "black"
+line_length = 160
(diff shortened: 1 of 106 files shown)


pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.black]
+line-length = 160
+target-version = ['py37']
+
+[tool.isort]
+profile = "black"
+line_length = 160
```
gaiksaya, line 7: What is the difference between line 7 and line 2

PR #776, opened 2021-10-21
[Build] support building targets for platform and arch inputs

kavilla:
### Description
Adding `platform` and `arch` as a build argument to be passed to
the build target. This enables cross-platform cross-arch builds,
for example, we can build linux x64 on a linux ARM64 machine.

We would want to do this because some of the libraries that are
required for building (not for runtime) are not available for
certain operating systems.

Also updated the OpenSearch Dashboards Jenkinsfile to pass the
platform and arch.

### Issues Resolved
 
### Check List


jenkins/opensearch-dashboards/Jenkinsfile
```diff
@@ -59,14 +112,21 @@ pipeline {
...
     git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
 
-    sh "./build.sh manifests/$INPUT_MANIFEST"
     sh './assemble.sh artifacts/manifest.yml'
+}
 
+void publish() {
     script { manifest = readYaml(file: 'artifacts/manifest.yml') }
-    def artifactPath = "${manifest.build.version}/${BUILD_NUMBER}/${manifest.build.architecture}";
+    def artifactPath = "${manifest.build.version}/${BUILD_NUMBER}/${manifest.build.platform}-${manifest.build.architecture}";
```
kavilla, line 129: highlighting updated s3 path.

src/build_workflow/build_args.py
```diff
@@ -10,10 +10,22 @@
 
 
 class BuildArgs:
+    SUPPORTED_PLATFORMS = [
+        "linux"
+    ]
+    SUPPORTED_ARCHITECTURES = [
```
kavilla, line 16: highlighting these proposed arches.

src/build_workflow/build_args.py
```diff
@@ -10,10 +10,22 @@
 
 
 class BuildArgs:
+    SUPPORTED_PLATFORMS = [
+        "linux"
```
dblock, line 14: We do also support `darwin` since https://github.com/opensearch-project/opensearch-build/pull/748, let's make sure we have a test so we don't remove it.

src/build_workflow/build_args.py
```diff
@@ -10,10 +10,22 @@
...
+    ]
+    SUPPORTED_ARCHITECTURES = [
+        "x64",
+        "x86_64",
```
dblock, line 18: Are x64 and x86_64 a dup here? I never know which one we pass from where, I think the two here should be `x64` and `arm64`, since this is what we return from `current_architecture`.

tests/test_run_build.py
```diff
@@ -88,3 +98,65 @@ def test_main(self, mock_temp, mock_repo, mock_recorder, mock_builder, *mocks):
 
         # the output manifest is written
         mock_recorder.return_value.write_manifest.assert_called()
+
+    OPENSEARCH_DASHBOARDS_MANIFEST = os.path.realpath(
+        os.path.join(
+            os.path.dirname(__file__), "../manifests/1.1.0/opensearch-dashboards-1.1.0.yml"
```
dblock, line 104: We haven't merged #767 yet, but start taking the habit of using `os.path.join` vs. `/`.

dblock (changes requested):
This LGTM! Let's fix the list of archictures.

dblock (approved):

jenkins/opensearch-dashboards/Jenkinsfile
peternied: You've got a race with https://github.com/opensearch-project/opensearch-build/pull/765 which cleans this all up

peternied (approved):

Issue #373, opened 2021-09-01
Add support for building patch releases


peterzhuamazon:
**Describe the bug**
As of now, bundle workflows scripts are using bundle version as core version in the build step.

https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/src/build.py#L56

This means if we have discrepancy between a bundle version and the core version it bundled, it would fail.
1.0.1 example:
https://github.com/opensearch-project/opensearch-build/blob/main/manifests/1.0.1/opensearch-1.0.1.yml#L4-L9

**To Reproduce**
Use above manifest of 1.0.1 and try to run build.sh within bundle workflow.

**Expected behavior**
Bundle 1.0.1 with 1.0.0 core would work.

**Plugins**
All Plugins component script, example here:
https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/scripts/components/alerting/build.sh#L71

**Screenshots**

**Host/Environment (please complete the following information):**
 - OS: [Linux]
 - Version [AL2]

**Additional context**

dblock:
We need to include component versions in the build manifest and possibly be able to pull binaries from a released location in assembly instead of rebuilding dependencies.

PR #789, opened 2021-10-22
Enable patch releases by fetching some components from existing distribution.

dblock:

### Description

- Adds a new component type that specifies "dist" to pull from an existing distribution instead of building from source.
- Adds "patches" into the input manifest that includes a list of existing versions this distribution patches.

This allows build to produce a mixed distribution with multiple versions of components, e.g. 1.1.0 + some components built from source = 1.1.1. 

In the provided example this build fails trying to build `anomaly-detection`, which expects OpenSearch core 1.1.1, but we're giving it 1.1.0. The builds for those patched components will need to pickup the released dependency from maven explicitly.
 
### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/373.
 
### Check List


dblock:
@saratvemulapalli @peternied Bump, this is ready for review please.

manifests/1.1.1/opensearch-1.1.1.yml
```diff
@@ -0,0 +1,17 @@
+---
+schema-version: "1.0"
+build:
+  name: OpenSearch
+  version: 1.1.1
+  patches: 
+    - 1.1.0
+components:
+  - name: OpenSearch
+    dist: https://ci.opensearch.org/ci/dbc/builds/1.1.0/405
```
peternied, line 10: The path will need to be inferred/discovered which will have been broken with the recent url restructuring that only impacted the 1.2.0+ builds.  Should we add parameters to help the build system construct the correct path to the manifest?

peternied (commented):
General feedback, deeper inspection later

manifests/1.1.1/opensearch-1.1.1.yml
dblock: Either in a separate PR or maybe we should just copy the manifest with bad URLs to a corrected copy in S3 and use that instead.

src/build_workflow/build_target.py
```diff
@@ -41,7 +43,22 @@ def __init__(
     def opensearch_version(self):
         return self.version + "-SNAPSHOT" if self.snapshot else self.version
```
peternied, line 44: Nearly all properties in this class have an `if` statement, code smells like we should create a subclass

src/build_workflow/builder_from_dist.py
```diff
@@ -0,0 +1,48 @@
...
+            self.ref = manifest.ref
+            self.sha = manifest.commit_id
+
+    def checkout(self, work_dir):
+        self.__download_build_manifest()
+
+    def build(self, build_recorder):
+        pass
+
+    def export_artifacts(self, build_recorder):
```
peternied, line 28: `export` doesn't seem to align with the action what about `import` or `download` + "_artifacts"?

manifests/1.1.1/opensearch-1.1.1.yml
peternied: The manifest is oriented around the components of with a platform / architecture, do we need another kind of manifest that is oriented around a releases artifacts.  Note, we can move this out to another PR

dblock: I don't see this as necessary. This is a bundle (assembly/distribution) manifest that can now include references to another bundle, and picks up components with matching platform/architecture.

src/build_workflow/build_target.py
dblock: https://github.com/opensearch-project/opensearch-build/issues/856

src/build_workflow/builder_from_dist.py
dblock: I thought about this. The "from source" code exports those artifacts, while the "from distribution" code downoads and exports those artifacts. They should really be spit up into `download_artifacts` and `export_artifacts`. I have another PR with more changes along those lines if that's OK?

peternied (approved):

saratvemulapalli (approved):
Lots of changes. 
Did a pass, overall it looks good to me.

Issue #798, opened 2021-10-22
[Bug]: opensearch bundle-build failed on jenkins


zelinh:
### Describe the bug

`+ /var/jenkins/workspace/bundle-build/publish/publish-snapshot.sh /var/jenkins/workspace/bundle-build/artifacts//maven`
`Invalid directory /var/jenkins/workspace/bundle-build/artifacts//maven does not exist`

Directory path seems to be wrong. 

### To reproduce

Run `bundle-build` job on jenkins

### Expected behavior

Should build successfully.

PR #802, opened 2021-10-25
Fix the bundle build with new path 

zelinh:
### Description
Path was updated from `artifacts` to `builds` and `bundles` to `dist`. 
This PR is updating some path leftover. 
 
### Issues Resolved
part of #798 
 
### Check List

### DIFF
--- a/deployment/lib/identities.ts
+++ b/deployment/lib/identities.ts
@@ -40,10 +40,10 @@ export class Identities {
     props.buildBucket.grantPut(buildRole, 'builds/*');
     props.buildBucket.grantRead(bundleRole, 'builds/*');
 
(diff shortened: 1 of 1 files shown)


gaiksaya:
> > > Does this also resolve the two `//` in `/var/jenkins/workspace/bundle-build/artifacts//maven` ?
> > 
> > 
> > No, I will look into it although I think it has no effect.
> 
> I think the value of $ARTIFACT_PATH was null hence we got the error for [line](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/Jenkinsfile#L75)

Checked the previous run. It has no affect. The value is not set even for successful build. Might need to debug and clean up. Approving now to unblock.

gaiksaya (approved):

PR #832, opened 2021-10-29
Change jenkins ci runner to the latest version (20211028)

zelinh:
### Description
Update the docker image to the latest version 
 
### Issues Resolved
 
### Check List

### DIFF
--- a/jenkins/opensearch/Jenkinsfile
+++ b/jenkins/opensearch/Jenkinsfile
@@ -25,7 +25,7 @@ pipeline {
             agent {
                 docker {
                     label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
-                    image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdk14-node10.24.1-cypress6.9.1-20211005'
+                    image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdk14-node10.24.1-cypress6.9.1-20211028'
                     alwaysPull true
                 }
             }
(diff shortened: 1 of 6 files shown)


jenkins/opensearch/Jenkinsfile
```diff
@@ -25,7 +25,7 @@ pipeline {
             agent {
                 docker {
                     label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
-                    image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdk14-node10.24.1-cypress6.9.1-20211005'
+                    image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028'
```
gaiksaya, line 28: Can we remove the dates from the image name? I don't see any use of it. We can use tags instead. So even external contributor can directly pull latest instead of wondering which one to use to build the artifacts.

peterzhuamazon: I disagree. The CI runner contains many runners for different purposes, a latest tag is useless as it cannot point to multiple images.

peterzhuamazon (changes requested):
Block the PR to wait for @jmazanec15 fix on centos7 cmake wheel on arm64, before merge it.

jenkins/opensearch/Jenkinsfile
gaiksaya: I see that `centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028` is itself is a tag and `opensearchstaging/ci-runner` is the repo. So no way of putting latest

peterzhuamazon (approved):
Image ready. Thanks.
https://hub.docker.com/r/opensearchstaging/ci-runner/tags?page=1&ordering=last_updated&name=centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028

Issue #822, opened 2021-10-27
Monitor 1.2.0 bundle builds and make sure they are passing


peternied:
Monitor on our slack channel that is posting build status periodically, if there is a failed build take action from this checklist:

![image](https://user-images.githubusercontent.com/2754967/139122112-538a3646-d3a8-4133-a817-50eca04cd351.png)

- Look at the build logs and determine if this is a new failure
   - If known, please reply to the post as a thread, and link the existing github issue - All done!
- Create a new issue on the component repository, or the OpenSearch-build repository.
  - Add `[Bundle Build Break]` to the title of the issue, make sure the title reference the problem source
  - Include the a log snippet and be sure that the .
  - Create a public gist of the full build log and attach it to the issue
  - Add the `v1.2.0` tag onto the issue.
  - See example issue https://github.com/opensearch-project/k-NN/issues/150
- Reply to thread with the new issue in the build status channel
- Create a new post in the release slack channel

Please also follow up with any questions, comments, or pull requests.

zelinh:
Bundle builds are working now.

peternied:
This is meant to be an ongoing tasks, we need to be sure that the 1.2.0 builds are working until release.

PR #836, opened 2021-10-29
Remove knn install script

jmazanec15:
Signed-off-by: John Mazanec <jmazane@amazon.com>

### Description
Currently, the k-NN install script copies the library into the the plugins folder. With https://github.com/opensearch-project/k-NN/pull/163 and https://github.com/opensearch-project/k-NN/pull/164, I added this step to the build phase. So, there is no longer any reason to have the install script here. It can fall back to default.
 
### Issues Resolved
#824 

### Check List
- [X] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #840, opened 2021-10-29
Remove k-NN component from manifest to unblock bundle-build workflow

zelinh:

### Description
Our bundle-build job on Jenkins has been failed for a while because of the k-NN plugin and its new docker image. We temporarily remove k-NN component from 1.2.0 manifest to unblock the bundle-build workflow on Jenkins. 
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/822
 
### Check List

### DIFF
--- a/manifests/1.2.0/opensearch-1.2.0.yml
+++ b/manifests/1.2.0/opensearch-1.2.0.yml
@@ -32,12 +32,3 @@ components:
     checks:
       - gradle:properties:version
       - gradle:dependencies:opensearch.version: alerting
-  - name: k-NN
-    repository: https://github.com/opensearch-project/k-NN.git
-    ref: "main"
-    platforms:
-      - darwin
-      - linux
-    checks:
-      - gradle:properties:version
-      - gradle:dependencies:opensearch.version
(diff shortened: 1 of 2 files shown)


gaiksaya:
I am worried about that we would be producing false positive builds. Like build will pass but the cluster wont start because of [knn config in install.sh](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/legacy/tar/linux/opensearch-tar-install.sh#L45)
What do you guys think @peternied @dblock ?

peterzhuamazon:
I think @jmazanec15  is nearly there with his fix, we have confirmed cmake is working properly this morning and he is checking the issues related to `make`. 

Let us see what he says about the progress.

Thanks.

peternied (approved):
I'd rather unblock the build and then we can address the issues with starting up the cluster separately.

gaiksaya (approved):

zelinh:
I think @peterzhuamazon is publishing the new staging docker image with the fix on k-NN component so we block this merge for few hours and see if it works.

peterzhuamazon:
Close this as Jack as fixed the issues in #836 #841.

Issue #852, opened 2021-11-02
Move process related logic under system/


tianleh:
### Is your feature request related to a problem? Please describe

part of #604

### Describe the solution you'd like

move code

tianleh:
PR has been approved and merged.

PR #845, opened 2021-11-01
Refactor LocalTestCluster into LocalTestClusterOpenSearch and LocalTestClusterOpenSearchDashboards.

tianleh:
### Description
Refactor LocalTestCluster into LocalTestClusterOpenSearch and LocalTestClusterOpenSearchDashboards.

To minimize the impact of this change, I still keep LocalTestCluster as it is which will be later replaced by LocalTestClusterOpenSearch. Thus there is no impact to the existing logic since the related change has not yet been used.

I also notice that the S3 path to build/bundle has been changed. However to keep the scope of this PR small, I will use the existing path for now.

### Test
The current 1.2.0 OpenSearch artifact has the issue `master_not_discovered_exception`. Thus my test strategy is to monitor whether the code execution can reach the health ping. I have been testing the LocalTestCluster class to use inside integ test suite to see the output. Below is the output snippet when trying to use ServiceOpenSearchDashboards and also comment out the logic to start OpenSearch inside it.

```
2021-10-31 23:50:21 INFO     Unpacking
2021-10-31 23:50:23 INFO     Unpacked
/tmp/tmp_oisyjff/local-test-cluster
opensearch-dashboards-1.2.0-linux-x64  opensearch-dashboards-1.2.0-linux-x64.tar.gz
2021-10-31 23:50:23 INFO     Started OpenSearch with parent PID 1238080
2021-10-31 23:50:23 INFO     Waiting for Dashboards service to become available
2021-10-31 23:50:23 INFO     Pinging http://localhost:5601/api/status attempt 0
2021-10-31 23:50:23 INFO     Service not available yet
2021-10-31 23:50:33 INFO     Pinging http://localhost:5601/api/status attempt 1
2021-10-31 23:50:33 INFO     Service not available yet
2021-10-31 23:50:43 INFO     Pinging http://localhost:5601/api/status attempt 2
```
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/837
 
### Check List
- [x ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #853, opened 2021-11-02
move process related logic to system dir

tianleh:

### Description
move process terminate logic to system dir. During the review of this PR https://github.com/opensearch-project/opensearch-build/pull/845, there is a comment to create a separate PR for moving process terminate logic. 
 
### Test
Since this PR just adds a utility function, we only test by adding a unit test and run the following test command.

```
opensearch@81f24e989d16 opensearch-build]$ pipenv run pytest tests/tests_system/test_process.py
.                                                                                                                                                  [100%]
1 passed in 0.02s

```

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/852
 
### Check List


.gitignore
```diff
@@ -17,4 +17,6 @@ out.txt
 /builds/
 /dist/
 
-/.vscode/
\ No newline at end of file
+/.vscode/
+
+unit_test*.txt
```
dblock, line 22: In your tests create temporary files so that the test cleans up after itself.

src/system/process.py
```diff
@@ -0,0 +1,51 @@
...
+        process.kill()
+        try:
+            logging.info("Waiting for process to terminate")
+            process.wait(10)
+        except subprocess.TimeoutExpired:
+            logging.info("Process failed to terminate even after SIGKILL")
+            raise
+    finally:
+        logging.info(f"Process terminated with exit code {process.returncode}")
+        if stdout:
```
dblock, line 39: This logic is specific to test cluster code and should be either moved out to the calling code or refactored to be generic. I think you want to wrap the process in a `Process` class that can deal with its starting and stopping, stderr and stdout.

tests/tests_system/test_process.py
```diff
@@ -0,0 +1,33 @@
...
+            stdout=stdout,
+            stderr=stderr,
+        )
+
+        self.assertIsNotNone(process)
+
+        terminated_process, local_cluster_stderr, local_cluster_stdout, return_code = terminate(process, work_dir, stdout, stderr)
+
+        self.assertIsNone(terminated_process)
+        self.assertIsNotNone(local_cluster_stdout)
```
dblock, line 33: If you refactor the process implementation all the way, then this code should look like this:

```python
with Process("ls", shell=True, cwd=".") as process:
    process.terminate()

self.assertIsEqual(process.stdout, ...)
```

unit_test/test_stdout.txt
```diff
@@ -0,0 +1,40 @@
+ADMINS.md
```
dblock, line 1: This will change with anything changing in this folder, so change the test command to be `echo something` to have predictable output.

dblock (changes requested):
You need to go further in refactoring this, and also change the code that uses it in the same PR.

.gitignore
tianleh: Added clean up logic

unit_test/test_stdout.txt
tianleh: Added removal logic. seems no longer applicable given the new logic.

tianleh:
> You need to go further in refactoring this, and also change the code that uses it in the same PR.

modified the local test cluster logic as well as its tests.

src/system/process.py
tianleh: Addressed

dblock (changes requested):

tests/tests_system/test_process.py
```diff
@@ -0,0 +1,30 @@
...
+
+from system.process import Process
+
+
+class TestProcess(unittest.TestCase):
+    def test(self):
+
+        process_handler = Process(".")
+
+        process_handler.start("ls", ".")
```
peterzhuamazon, line 18: What is this `ls` here for?

tests/tests_test_workflow/test_integ_workflow/integ_test/test_local_test_cluster.py
```diff
@@ -151,54 +147,7 @@ def test_wait_for_service_cluster_unavailable(self, mock_test_recorder, *mocks):
...
-        side_effect=__mock_process,
-    )
-    @patch("test_workflow.integ_test.local_test_cluster.subprocess.Popen.wait")
-    @patch("test_workflow.integ_test.local_test_cluster.subprocess.Popen.terminate")
-    @patch("test_workflow.integ_test.local_test_cluster.logging", return_value=MagicMock())
-    def test_terminate_process(self, mock_logging, mock_terminate, mock_wait, mock_process):
-        self.local_test_cluster.stdout = None
-        self.local_test_cluster.stderr = None
-        self.local_test_cluster.process = self.process
+    @patch.object(Process, 'terminate', return_value=("1", "2", "3"))
```
peterzhuamazon, line 150: Is this the number of termination or the seconds count for termination in return value?

src/system/process.py
```diff
@@ -0,0 +1,76 @@
...
+
+import psutil  # type: ignore
+
+
+class Process:
+    def __init__(
+        self,
+        work_dir
+    ):
+        os.system("pwd")
```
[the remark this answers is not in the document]
tianleh: addressed

src/system/process.py
```diff
@@ -0,0 +1,76 @@
...
+        self.work_dir = work_dir
+        self.process = None
+
+    def start(self, cmd, install_dir):
```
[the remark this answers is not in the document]
tianleh: addressed.

src/system/process.py
```diff
@@ -0,0 +1,76 @@
...
+        self.stdout = open("stdout.txt", "w")
+        self.stderr = open("stderr.txt", "w")
```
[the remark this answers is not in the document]
tianleh: Thanks. Looks much cleaner to use NamedTemporaryFile

src/system/process.py
```diff
@@ -0,0 +1,76 @@
...
+                with open(os.path.join(self.work_dir, self.stderr.name), "r") as stderr:
+                    self.stderr_data = stderr.read()
+                self.stderr.close()
+                self.stderr = None
+            self.return_code = self.process.returncode
+            self.process = None
+
+            return self.return_code, self.stdout_data, self.stderr_data
+
+    def get_pid(self):
```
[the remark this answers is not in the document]
tianleh: addressed

src/system/process.py
```diff
@@ -0,0 +1,76 @@
...
+        return self.process.pid if self.process is not None else None
```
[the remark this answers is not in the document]
tianleh: addressed

tests/tests_system/test_process.py
```diff
@@ -0,0 +1,30 @@
...
+
+        self.assertIsNotNone(process_handler.get_pid())
+
+        process_handler.terminate()
+
+        self.assertIsNone(process_handler.get_pid())
+
+        # clean up
+        os.remove("stdout.txt")
```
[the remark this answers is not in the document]
tianleh: Yes. Removed

tests/tests_system/test_process.py
tianleh: This is just for spinning up a simple test process. It needs some commands to run.

tests/tests_test_workflow/test_integ_workflow/integ_test/test_local_test_cluster.py
tianleh: Just provide some return values for the mock function in order for the test to continue without complaining.

src/system/process.py
```diff
@@ -0,0 +1,77 @@
...
+
+import psutil  # type: ignore
+
+
+class Process:
+    def __init__(
+        self,
+        work_dir
+    ):
+        self.work_dir = work_dir
```
dblock, line 20: work_dir is no longer used, remove

tests/tests_system/test_process.py
```diff
@@ -0,0 +1,25 @@
...
+
+        process_handler.start("ls", ".")
+
+        self.assertIsNotNone(process_handler.pid)
+
+        process_handler.terminate()
+
+        self.assertIsNone(process_handler.pid)
+
+    # TODO: add test for subprocess.TimeoutExpired
```
dblock, line 25: You need these tests and one for stdout and stderr

dblock (changes requested):

seraphjiang (approved):

src/system/process.py
tianleh: removed

seraphjiang (approved):

dblock (changes requested):
1. Re-implement tests for timeout expired, stderr and stdout. We had those and refactoring should be adding more tests, not less.
2. If I call `Process#start` twice I lose track of the process. Raise an exception if that happens. Add tests.
3. If I call `Process#terminate` on an unstarted process, I get errors. Raise an exception, add tests.
4. Add a `started()` method and use `if not process.started()` instead of `process.pid is None`.
5. In the test execute something that doesn't return, e.g. `bash`, so we can see that termination works.
6. Rebase.

seraphjiang:
could we ignore pyc file
refer to https://stackoverflow.com/questions/32110126/should-i-put-pyc-files-under-version-control

Tengda-He (approved):

src/system/process.py
```diff
@@ -0,0 +1,72 @@
...
+            if self.stdout:
+                self.stdout_data = self.stdout.read()
+                self.stdout.close()
+                self.stdout = None
+            if self.stderr:
+                self.stderr_data = self.stderr.read()
+                self.stderr.close()
+                self.stderr = None
+            self.return_code = self.process.returncode
+            self.process = None
```
seraphjiang, line 66: Don't see the need to use terminate first, if no response then try kill.

why not just kill()

btw. we should let original call stack being bubble up, instead of `raise` a new one.  

https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

seraphjiang, line 66: why do we need output direction? not quite convinced? let's remove and clean up this.

tianleh:
> could we ignore pyc file refer to https://stackoverflow.com/questions/32110126/should-i-put-pyc-files-under-version-control

Sure. removed them and also updated gitignore

src/system/process.py
tianleh: Let me simplify the logic to just kill()

src/system/process.py
```diff
@@ -0,0 +1,71 @@
...
+        self.return_code = self.process.returncode
+        self.process = None
+
+        return self.return_code, self.stdout_data, self.stderr_data
+
+    @property
+    def pid(self):
+        return self.process.pid if self.process else None
+
+    @property
```
dblock, line 65: Were these leftover from testing?

src/system/process.py
```diff
@@ -0,0 +1,71 @@
...
+            logging.debug(f"Found child process with pid {child.pid}")
+            if child.pid != self.process.pid:
+                logging.debug(f"Sending SIGKILL to {child.pid} ")
+                child.kill()
+        logging.info(f"Sending SIGKILL to PID {self.process.pid}")
+        
+        self.process.kill()
+    
+        logging.info(f"Process terminated with exit code {self.process.returncode}")
+        if self.stdout:
```
dblock, line 48: Does `self.stdout_data = self.stdout.read()` not work?

dblock (changes requested):
Tests still missing + rebase + some nits.

tianleh:
>     1. Re-implement tests for timeout expired, stderr and stdout. We had those and refactoring should be adding more tests, not less.
> 
>     2. If I call `Process#start` twice I lose track of the process. Raise an exception if that happens. Add tests.
> 
>     3. If I call `Process#terminate` on an unstarted process, I get errors. Raise an exception, add tests.
> 
>     4. Add a `started()` method and use `if not process.started()` instead of `process.pid is None`.
> 
>     5. In the test execute something that doesn't return, e.g. `bash`, so we can see that termination works.
> 
>     6. Rebase.

Thanks for the detailed instruction. Just try to think from the caller's perspective. The OpenSearch and OpenSearchDashboards will have their own instance of Process class. Thus we will not try to reuse the instance and call start twice. Also the exception handling about TimeoutExpired does not impact the caller's behavior. Thus to have a MVP feature working soon, I am proposing a tentative plan to use kill() directly and remove the exception handling, test the happy case, rebase and backlog all other items to issues.

A side question is for # 5, I do not quite get the idea. Could you help elaborate more?

src/system/process.py
tianleh: Just want to leverage the benefit of `with` statement to not worry about explicitly calling close. 

https://www.geeksforgeeks.org/with-statement-in-python/

dblock:
> Thanks for the detailed instruction. Just try to think from the caller's perspective. The OpenSearch and OpenSearchDashboards will have their own instance of Process class. Thus we will not try to reuse the instance and call start twice.

You can't rely on user to do the right thing for a class that should stand on its own. A developer can easily make a mistake in the future, so guard the code from it.

> Also the exception handling about TimeoutExpired does not impact the caller's behavior. Thus to have a MVP feature working soon, I am proposing a tentative plan to use kill() directly and remove the exception handling, test the happy case, rebase and backlog all other items to issues.

Let's see that PR.

> A side question is for # 5, I do not quite get the idea. Could you help elaborate more?

In your tests you call "ls". That process finishes before you know it, so you don't get to actually test termination. Instead, call `read -n1 ans` that blocks until someone presses a key, or `bash`, which will start a new shell and hang out there.

src/system/process.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class Process:
+    def __init__(
+        self
+    ):
+        self.process = None
+        self.stdout = None
+        self.stderr = None
+        self.has_started = False
```
dblock, line 21: You should call `self.started` and in the method check `self.process` thus avoiding another variable or potential race conditions.

src/system/process.py
```diff
@@ -0,0 +1,77 @@
...
+            with open(self.stderr.name) as stderr:
+                self.stderr_data = stderr.read()
+
+        self.return_code = self.process.returncode
+        self.process = None
+        self.has_started = False
+
+        return self.return_code, self.stdout_data, self.stderr_data
+
+    def started(self):
```
dblock, line 72: Make it a `@property`

src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -76,7 +69,7 @@ def port(self):
         return 9200
 
     def destroy(self):
-        if self.process is None:
+        if self.process_handler.pid is None:
```
dblock, line 72: `self.process_handler.started`

src/system/process.py
```diff
@@ -0,0 +1,77 @@
...
+        self
+    ):
+        self.process = None
+        self.stdout = None
+        self.stderr = None
+        self.has_started = False
+
+    def start(self, command, cwd):
+        if self.has_started:
+            raise AssertionError("Cannot start a started process!")
```
dblock, line 25: Add a class called `ProcessStartedError` and make its message to include the PID of the already started process.

The convention for messages is not to include any kind of exclamations, so something like `f"Process already started, pid: {pid}"`

dblock (changes requested):
There's a failing unit test + some should have's.

src/system/process.py
tianleh: I need to revert the change for using `with`. The windows check is failing.

```
        if self.stdout:
>           with open(self.stdout.name) as stdout:
E           PermissionError: [Errno 13] Permission denied: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpxl83o9tt'

src\system\process.py:59: PermissionError

----------- coverage: platform win32, python 3.7.9-final-0 -----------
Coverage XML written to file coverage.xml

=========================== short test summary info ===========================
FAILED tests/tests_system/test_process.py::TestProcess::test - PermissionErro...
1 failed, 355 passed in 61.81s (0:01:01)
```

I will just call read() and then close()

dblock: i don't think this is the problem.

tianleh:
> > Thanks for the detailed instruction. Just try to think from the caller's perspective. The OpenSearch and OpenSearchDashboards will have their own instance of Process class. Thus we will not try to reuse the instance and call start twice.
> 
> You can't rely on user to do the right thing for a class that should stand on its own. A developer can easily make a mistake in the future, so guard the code from it.
> 
> > Also the exception handling about TimeoutExpired does not impact the caller's behavior. Thus to have a MVP feature working soon, I am proposing a tentative plan to use kill() directly and remove the exception handling, test the happy case, rebase and backlog all other items to issues.
> 
> Let's see that PR.
> 
> > A side question is for # 5, I do not quite get the idea. Could you help elaborate more?
> 
> In your tests you call "ls". That process finishes before you know it, so you don't get to actually test termination. Instead, call `read -n1 ans` that blocks until someone presses a key, or `bash`, which will start a new shell and hang out there.

For # 5, I have experienced some issues about inconsistency behavior in my local AL2 machine and the Ubuntu check on PR. 
`process_handler.start("read -n1 ans", ".")` is working fine in AL2. However, when running from Github Action, it will fail the PR check "python-tests / test (ubuntu-latest) (pull_request) " with error about invalid option "-n1". I have tried multiple commits to fix it 
https://github.com/opensearch-project/opensearch-build/pull/853/commits/0b2d1ac5724c5131bb66d564a07cbf09c57a9d48

https://github.com/opensearch-project/opensearch-build/pull/853/commits/4f3b59a8732110a5578563ca9dc6100176393445

and none of these will pass the "python-tests / test (ubuntu-latest) (pull_request) ". 

Eventually the following way makes it work by running from a shell.

`process_handler.start("./tests/tests_system/data/wait_for_input.sh", ".")`

src/system/process.py
tianleh: After digging into the tempfile doc https://docs.python.org/3/library/tempfile.html, opening for a second time has different behaviors in different platforms.

```
Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later). 
```

tianleh:
All checks have passed. Addressing dblock@'s remaining comments.

dblock:
> All checks have passed. Addressing dblock@'s remaining comments.

You also need a test that shows that you're correctly reading stdout and stderr.

src/system/process.py
tianleh: expose output and error for testing.

tianleh:
> You also need a test that shows that you're correctly reading stdout and stderr.

Added the validation for these. Let me know if this is enough.

src/system/process.py
```diff
@@ -18,11 +18,10 @@ def __init__(
         self.process = None
         self.stdout = None
         self.stderr = None
-        self.has_started = False
 
     def start(self, command, cwd):
-        if self.has_started:
-            raise AssertionError("Cannot start a started process!")
+        if self.started:
+            raise ProcessStartedError(f"Process already started, pid: {self.pid}")
```
dblock, line 24: We typically sink the string message into the class and assign pid into it, so:

```
raise ProcessStartedError(self.pid)
```

```
class ProcessStartedError(Exception):
    def __init__(pid):
       self.pid = pid
       super().__init__(f"Process already started, pid: {pid}")
```

dblock (approved):
This is good. Do you want to take care of my suggestions below before/after the merge?

tianleh:
> This is good. Thanks for hanging in here. Do you want to take care of my suggestions below before/after the merge?

Let me quickly address your new comments as part of this PR.

tianleh:
> > This is good. Thanks for hanging in here. Do you want to take care of my suggestions below before/after the merge?
> 
> Let me quickly address your new comments as part of this PR.

Addressed the new comments about the Exception class.

PR #863, opened 2021-11-02
Refactor components into a subclass.

dblock:

### Description

Refactor a collection of components into a `Components` and `ComponentsManifest` base classes and use it everywhere. Makes it a lot easier to make a manifest with components that can be selected based on input options.
 
### Check List

### DIFF
--- a/src/manifests/manifest.py
+++ b/src/manifests/manifest.py
@@ -51,15 +51,17 @@ class Manifest(ABC):
 
     @classmethod
     def compact(cls, d):
-        result = {}
-        for k, v in d.items():
-            if isinstance(v, dict):
-                nested = cls.compact(v)
-                if nested:
-                    result[k] = nested
-            elif v and v != []:
-                result[k] = v
-        return result
+        if isinstance(d, list):
+            return list(map(lambda i: cls.compact(i), d))
+        elif isinstance(d, dict):
+            result = {}
+            for k, v in d.items():
+                v = cls.compact(v)
+                if v and v != []:
+                    result[k] = v
+            return result
+        else:
+            return d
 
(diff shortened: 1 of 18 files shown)


src/manifests/manifest.py
```diff
@@ -51,15 +51,17 @@ def from_path(cls, path):
...
-                if nested:
-                    result[k] = nested
-            elif v and v != []:
-                result[k] = v
-        return result
+        if isinstance(d, list):
+            return list(map(lambda i: cls.compact(i), d))
+        elif isinstance(d, dict):
+            result = {}
+            for k, v in d.items():
```
owaiskazi19, line 58: Can we be more specific with `k` and `v` here and for the rest below?

src/manifests/manifest.py
```diff
@@ -51,15 +51,17 @@ def from_path(cls, path):
...
+                v = cls.compact(v)
+                if v and v != []:
```
owaiskazi19, line 60:
```suggestion
                if v:
```
should do the work here

peternied (approved):

VachaShah (approved):

src/manifests/manifest.py
dblock: Well, they are key and value, but nothing specific :)

src/manifests/manifest.py
dblock: I wanted to capture intent. Is it true that an empty list returns false on []? Will check in a future PR. Weird.

dblock: https://github.com/opensearch-project/opensearch-build/pull/877

Issue #629, opened 2021-09-28
Add support for running tests in local desktop via build/bundle manifest


saratvemulapalli:
### Is your feature request related to a problem? Please describe

All test suites depend on pulling the manifest from S3. This experience is not great to run tests when manifests are present locally.

Coming from conversation of: https://github.com/opensearch-project/project-website/pull/293

### Describe the solution you'd like

Run test suites taking manifest as parameters and store test results in local.
```
./bundle-workflow/test.sh manifests/opensearch-1.1.0.yml
```

setiah:
The test-workflow depends on the directory layout for artifacts set by build workflow defines, to read below. 
* Maven artifacts
* Built zip dependencies like job-scheduler
* Bundles artifacts.

test-workflow is able to generate the download paths using `s3 bucket`, `build_id`, `architecture`. if we want to build and test locally, we might need to modify the inputs to pre-provided paths. These paths could be s3 or local or remote url. Doable but interface seems less cleaner than what it is. The interfacing job between the build workflow and test workflow will have to generate these paths for a build id to invoke test workflow. 

Another challenge with running test workflow locally is that the perf-test require vpc peering with rally setup, which is outside of the test setup process. This means we could enable only the integ and bwc test workflows to run locally

peterzhuamazon:
Please provide more information on how much work we need to put into this.

dblock:
I have implemented downloading the manifest and exporting all the files from a distribution using publicly available URLs in https://github.com/opensearch-project/opensearch-build/pull/789. After that's merged (cc: @saratvemulapalli and @peternied) I think we can replace all `from_s3` by `from_url`, followed by a new `.download()` that's a subset of the `export_artifacts()` method I added. I think there's a day's worth of work here.

dblock:
This is https://github.com/opensearch-project/opensearch-build/issues/797

PR #878, opened 2021-11-03
Remove S3 dependency and enable local integ testing  

dblock:

### Description

- Replaces dependency on S3 by downloading manifests and artifacts from public URLs.
- Enables running `./test.sh integ-test .` that uses a local build and distribution to execute tests.
 
Note that tests for 1.2.0 will continue failing because of DistributionDownloader, but all the dependencies are properly downloaded and installed from local or remote.

```
Cause 1: org.gradle.internal.resolve.ModuleVersionResolveException: Could not resolve opensearch-distribution:opensearch:1.2.0.
Caused by: org.gradle.api.resources.ResourceException: Could not get resource 'https://artifacts.opensearch.org/releases/core/opensearch/1.2.0/opensearch-min-1.2.0-linux-x64.tar.gz'.
```

### Issues Resolved

- Closes https://github.com/opensearch-project/opensearch-build/issues/629
 
### Check List


src/run_perf_test.py
```diff
@@ -11,23 +11,6 @@
 from test_workflow.perf_test.perf_test_cluster import PerfTestCluster
 from test_workflow.perf_test.perf_test_suite import PerfTestSuite
 
-"""
```
dblock, line 14: This made this implementation untestable and untested. Whenever you require the file it would run through this code.

owaiskazi19: Thanks for catching that.

Pipfile
```diff
@@ -13,17 +13,14 @@ isort = "~=5.9"
 flake8 = "~=3.9"
 mypy = "~=0.9"
 pytest = "*"
-boto3 = "1.18.31"
```
peternied, line 16: yay!

peternied (approved):
Nicely done

owaiskazi19 (approved):
LGTM!

PR #890, opened 2021-11-05
Lock 1.1.0 manifest with tags

dblock:
### Description

Since we incremented the version of OpenSearch on the 1.1 branch CI started failing because of a version mismatch, as expected. Lock the manifest of the released version to use tags.

Performance analyzer RCA is missing a tag, I opened https://github.com/opensearch-project/performance-analyzer-rca/issues/85. I am not sure what the right rev is as I cannot find it in the bundle manifest. This is odd. CI will pass once that's resolved.
 
### Check List

### DIFF
--- a/manifests/1.1.0/opensearch-1.1.0.yml
+++ b/manifests/1.1.0/opensearch-1.1.0.yml
@@ -6,51 +6,51 @@ build:
 components:
   - name: OpenSearch
     repository: https://github.com/opensearch-project/OpenSearch.git
-    ref: "1.1"
+    ref: 1.1.0
     checks:
       - gradle:publish
       - gradle:properties:version
   - name: common-utils
     repository: https://github.com/opensearch-project/common-utils.git
-    ref: "1.1"
+    ref: 1.1.0.0
     checks:
       - gradle:publish
       - gradle:properties:version
   - name: job-scheduler
     repository: https://github.com/opensearch-project/job-scheduler.git
-    ref: "1.1"
+    ref: 1.1.0.0
     checks:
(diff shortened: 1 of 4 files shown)


manifests/1.1.0/opensearch-1.1.0.yml
```diff
@@ -6,51 +6,51 @@ build:
 components:
   - name: OpenSearch
     repository: https://github.com/opensearch-project/OpenSearch.git
-    ref: "1.1"
+    ref: 1.1.0
```
gaiksaya, line 9: What do you think about updating them with the `commit_id` that we see in manifest.yml that is shipped with the bundle?

dblock: That could be a viable alternative, and we could eventually do it automatically, but I think the build's intent is better expressed with a tag here, would prefer leaving as is unless there's a compelling argument not to. This is also how we found that a tag was missing on performance-analyzer-rca.

dblock:
@kavilla care to do the same for the dashboards manifest?

manifests/1.1.0/opensearch-1.1.0.yml
gaiksaya: Sure. I think we have an issue to cut tags automatically after the release using the commit_id from manifest.yml so I think overall it the same.

gaiksaya (approved):

saratvemulapalli (approved):

peterzhuamazon:
#896 will lock the changes to tags.

PR #893, opened 2021-11-05
Added component manifest check.

dblock:

### Description

- Adds a check for components built from distribution. Ensures that the distribution is present and that the component is included in it.
- Updates 1.1.1 manifest to include OpenSearch core that incremented the version to 1.1.1. We can switch the components now as needed from dist to source.
 
### Check List


src/ci_workflow/ci_check_manifest_component.py
```diff
@@ -0,0 +1,25 @@
...
+from manifests.build_manifest import BuildManifest
+
+
+class CiCheckManifestComponent(CiCheckDist):
+    class MissingComponentError(Exception):
+        def __init__(self, component, url):
+            super().__init__(f"Missing {component} in {url}.")
+
+    def check(self):
+        for architecture in ["x64", "arm64"]:
```
peternied, line 19: Can we import from build args where this would be verified rather than duplicate? https://github.com/opensearch-project/opensearch-build/blob/fba6709809ebe0e51b6f856e0ebeaa1de52e655e/src/build_workflow/build_args.py#L14-L17

src/ci_workflow/ci_check_manifest_component.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class CiCheckManifestComponent(CiCheckDist):
+    class MissingComponentError(Exception):
+        def __init__(self, component, url):
+            super().__init__(f"Missing {component} in {url}.")
+
+    def check(self):
+        for architecture in BuildArgs.SUPPORTED_ARCHITECTURES:
+            url = "/".join([self.component.dist, architecture, "manifest.yml"])
```
owaiskazi19, line 21: Can we use `urllib` here to construct the url?

peternied (approved):

src/ci_workflow/ci_check_manifest_component.py
dblock: I had tried, but that gets confused with slashes (does relative path with two slashes), doesn't allow to join more than 2 arguments, etc. So left it alone.

owaiskazi19 (approved):

Issue #675, opened 2021-10-04
[Feature] Create Build lock file and use it during the build process


peternied:
_Conceptually this feature is aligned with npm's [package-lock.json](https://docs.npmjs.com/cli/v6/configuring-npm/package-locks) or python's [pipfile.lock](https://realpython.com/pipenv-guide/#the-pipfilelock)._

Reproducing build issues can be difficult if OpenSearch-build or one of the branches has changed since an error has first occurred.  Its also hard to identify if a build has already been made already - there is an cron job that rebuilds often and is typically rebuilding components that have already been built.

A new system should check the current repository, and the state of the remote repositories much like the current workflow
[run_manifests.py](https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/src/run_manifests.py).

The new manifest would look similar to the distribution manifest, with notable omissions of `location`, `architecture`, `id`, a new section called `build-system` would also be added to track the build system and version details.

By running `diff` against this output, the build process could process if there was an update, or end early if it is not necessary.  Additional changes could be made to use this file as a way to lock the built version into place so that only the associated commits are checked out rather than based on the point in time when the git clone take place during the build.

Example build-snapshot.yml
```
build:
  name: OpenSearch
  version: 1.1.0
build-system:
- commit_id: d5966e1cd1e1fa8914d6f6fda3ea9a80ef63d8e6
  name: OpenSearch-build
  ref: 'main'
  repository: https://github.com/opensearch-project/opensearch-build.git
components:
- commit_id: 15e9f137622d878b79103df8f82d78d782b686a1
  name: OpenSearch
  ref: '1.1'
  repository: https://github.com/opensearch-project/OpenSearch.git
- commit_id: 0a4ce10a812165e95801b77ede689064ffee2188
  name: job-scheduler
  ref: '1.1'
  repository: https://github.com/opensearch-project/job-scheduler.git
```

Example schema:
```
{
    "build": {
        "required": true,
        "type": "dict",
        "schema": {
            "name": {"required": true, "type": "string"},
            "version": {"required": true, "type": "string"}
        }
    },
    "schema-version": {"required": true, "type": "string", "allowed": ["1.0"]},
    "build-system": {
        "commit_id": {"required": true, "type": "string"},
        "name": {"required": true, "type": "string"},
        "ref": {"required": true, "type": "string"},
        "repository": {"required": true, "type": "string"}
    },
    "components": {
        "required": true,
        "type": "list",
        "schema": {
[...]

PR #904, opened 2021-11-08
Add support for .lock manifests to avoid building unnecessarily.

dblock:

### Description

Automatically generates a .lock file with stable refs. Subsequent builds will exit early if nothing has changed in either the manifest or any of the repos referenced in the manifest.

We would need to save those lock files across builds in order to take advantage of this system in Jenkins CI. One way would be to commit them. Thoughts @peternied? 

### Issues Resolved

Implements https://github.com/opensearch-project/opensearch-build/issues/675.
 
### Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ The following options are available in `build.sh`.
 |--------------------|-------------------------------------------------------------------------|
 | --snapshot         | Build a snapshot instead of a release artifact, default is `false`.     |
 | --component [name] | Rebuild a single component by name, e.g. `--component common-utils`.    |
-| --keep             | DDo not delete the temporary working directory on both success or error.|
+| --keep             | Do not delete the temporary working directory on both success or error. |
 | -l, --lock         | Generate a stable reference manifest.                                   |
 | -v, --verbose      | Show more verbose output.                                               |
 
@@ -187,22 +187,20 @@ Each component build relies on a `build.sh` script that is used to prepare bundl
 
 ##### Avoiding Rebuilds
 
-Builds can automatically generate a `manifest.lock` file with stable git references by specifying `--lock`. The output can then be reused as input manifest after checking against a collection of prior builds.
+Builds can automatically generate a `manifest.lock` file with stable git references (commit IDs) and build options (platform, architecture and snapshot) by specifying `--lock`. The output can then be reused as input manifest after checking against a collection of prior builds.
 
 ```bash
-PLATFORM=linux
-ARCHITECTURE=x64
 MANIFEST=manifests/1.1.0/opensearch-1.1.0.yml
-SHAS=shas/opensearch/$PLATFORM-$ARCHITECTURE
+SHAS=shas
 
-./build.sh --platform $PLATFORM --architecture $ARCHITECTURE --lock $MANIFEST # generates opensearch-1.1.0.yml.lock
+./build.sh --lock $MANIFEST # generates opensearch-1.1.0.yml.lock
 
 MANIFEST_SHA=$(sha1sum $MANIFEST.lock | cut -f1 -d' ') # generate a checksum of the stable manifest
 
 if test -f "$SHAS/$MANIFEST_SHA.lock"; then
   echo "Skipping $MANIFEST_SHA"
 else
-  ./build.sh --platform $PLATFORM --architecture $ARCHITECTURE $MANIFEST.lock # rebuild using stable references in .lock
+  ./build.sh $MANIFEST.lock # rebuild using stable references in .lock
   mkdir -p $SHAS
   cp $MANIFEST.lock $SHAS/$MANIFEST_SHA.lock # save the stable reference manifest
 fi
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@
     - [Check Out Source](#check-out-source)
     - [Build from Source](#build-from-source)
       - [Custom Build Scripts](#custom-build-scripts)
+      - [Avoiding Rebuilds](#avoiding-rebuilds)
       - [Patch Releases](#patch-releases)
     - [Assemble the Bundle](#assemble-the-bundle)
       - [Cross-Platform Builds](#cross-platform-builds)
@@ -176,12 +177,17 @@ The following options are available in `build.sh`.
 |--------------------|-------------------------------------------------------------------------|
 | --snapshot         | Build a snapshot instead of a release artifact, default is `false`.     |
 | --component [name] | Rebuild a single component by name, e.g. `--component common-utils`.    |
-| --keep             | Do not delete the temporary working directory on both success or error. |
+| --keep             | Do not generate or use a stable reference manifest.                     |
+| -l, --lock         | Generate and use a stable reference manifest.                           |
 | -v, --verbose      | Show more verbose output.                                               |
 
 ##### Custom Build Scripts
 
 Each component build relies on a `build.sh` script that is used to prepare bundle artifacts for a particular bundle version that takes two arguments: version and target architecture. By default the tool will look for a script in [scripts/components](scripts/components), then in the checked-out repository in `build/build.sh`, then default to a Gradle build implemented in [scripts/default/build.sh](scripts/default/build.sh).
+
+##### Avoiding Rebuilds
+
+Builds can automatically generate `manifest.lock` files with stable git references by specifying `--lock`. If no changes on any repo in the manifest have been made since the last `.lock` file was produced, the build process will exit successfully.
 
 ##### Patch Releases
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -175,13 +176,20 @@ The following options are available in `build.sh`.
 |--------------------|-------------------------------------------------------------------------|
 | --snapshot         | Build a snapshot instead of a release artifact, default is `false`.     |
 | --component [name] | Rebuild a single component by name, e.g. `--component common-utils`.    |
-| --keep             | Do not delete the temporary working directory on both success or error. |
+| --keep             | Do not generate or use a stable reference manifest.                     |
+| -n, --no-lock      | Show more verbose output.                                               |
```
peternied, line 180: Copypasta error

README.md
```diff
@@ -175,13 +176,20 @@ The following options are available in `build.sh`.
...
 
 ##### Custom Build Scripts
 
 Each component build relies on a `build.sh` script that is used to prepare bundle artifacts for a particular bundle version that takes two arguments: version and target architecture. By default the tool will look for a script in [scripts/components](scripts/components), then in the checked-out repository in `build/build.sh`, then default to a Gradle build implemented in [scripts/default/build.sh](scripts/default/build.sh).
 
+##### Avoiding Rebuilds
+
+Builds automatically generate `manifest.lock` files with stable git references. If no changes on any repo in the manifest have been made since the last .lock file was produced, the build process will exit successfully.
+
+You can turn off this behavior with `--no-lock`.
```
peternied, line 191: Can we have this behavior not on by default? Seems like something that could mess up a local build if you weren't expecting it.  I would rather have the CI systems (GHA/Jenkins) opt-in

peternied (commented):
I would prefer rebuilding on a dev box by default, otherwise you've got to add the flag, or delete the file from your PR.

peternied:
> We would need to save those lock files across builds in order to take advantage of this system in Jenkins CI. One way would be to commit them. Thoughts @peternied?

Great idea, the pathing that is used right now is inline with the manifest, I think it might make sense to break the lockfile location to a separate path that would be different for each system so GHA / Jenkins onboarding would be done separately which could be more granular to the task

README.md
dblock: Fixed.

README.md
dblock: Updated with `--lock` to enable this behavior instead of `--no-lock` to disable it.

dblock:
> > We would need to save those lock files across builds in order to take advantage of this system in Jenkins CI. One way would be to commit them. Thoughts @peternied?
> 
> Great idea, the pathing that is used right now is inline with the manifest, I think it might make sense to break the lockfile location to a separate path that would be different for each system so GHA / Jenkins onboarding would be done separately which could be more granular to the task

Updated to disable this behavior by default and pass `--lock` to enable it.

If we think of Rubygems or Pipenv as a model, then we probably never want to have to specify a location for a `.lock` file as it's directly related to the manifest itself - hence should live next to it. I'd prefer if we thought about this as "CI caching some output". Just don't know how best to implement it.

peternied:
> I'd prefer if we thought about this as "CI caching some output". 

I think you are conflating dependency management and building.  In NPM world (I imagine it aligns with ruby/python), `npm ci` uses the package-lock.json file to install exactly the same dependencies.  This installation process is repeatable as it will always ensure the same components are installed or fail because it cannot recover the same dependency tree.  It cannot confirm the existence of a build in a distributed system.

By depending on the local file that is left around after a build, its the opposite pattern of using docker images that are deleted when build is done.

If we want to if we generated a build if and only if one wasn't already created we need an inventories of lock files to check, scan those lock files for a match.  We could build that with the lock file you generate in this PR but we would need to build that scanning process.

dblock:
I guess if our goal is just to avoid rebuilds, this could work:

1. `build.sh --lock manifest.yml` # only produces manifest.yml.lock, doesn't build
2. check whether `S3/builds/opensearch/x64/linux/sha1(manifest.yml.lock)` exists, if so bail
3. `build.sh manifest.yml.lock`
4. upload `S3/builds/opensearch/x64/linux/sha1(manifest.yml.lock).yml`

WDYT? Feels like something is overlycomplicated ...

peternied:
I do not see a way we could make it simpler, works for me.

dblock:
I pushed an update that generates a `.lock` file and exits in all cases with `--lock` and provided an example of how to skip rebuilds in README.

I've also added architecture, platform, and snapshot options into the build portion of the input manifest that will be used as defaults. This way stable manifests are truly stable, not just for git refs but also for build options.

The next step would be to use this in a Jenkins pipeline. I think we want something like this:

```groovy
ifChanges("manifests/$INPUT_MANIFEST") {
    build()
}
```

Probably for another PR? Also maybe someone else can take this part?

src/manifests/input_manifest.py
```diff
@@ -49,6 +54,9 @@ class InputManifest(ComponentManifest):
                 "name": {"required": True, "type": "string"},
                 "version": {"required": True, "type": "string"},
                 "patches": {"type": "list", "schema": {"type": "string"}},
+                "platform": {"type": "string"},
+                "architecture": {"type": "string"},
+                "snapshot": {"type": "boolean"},
```
peternied, line 59: Rather than make these features of the build, these seem like part the CI section, maybe as command line args?

src/manifests/input_manifest.py
```diff
@@ -49,6 +54,9 @@ class InputManifest(ComponentManifest):
...
             },
         },
         "ci": {
```
peternied, line 62: What about the notion of the opensearch-build repositories state?  If there are changes to this repo the build might not be stable, if we had BWC tests I'd have more confidence, but I'm not sure we want to go to that length

peternied: Note; if we wanted to manage this less broadly, what about including the SHA of the source manifest file?

README.md
```diff
@@ -176,12 +177,34 @@ The following options are available in `build.sh`.
...
+SHAS=shas
+
+./build.sh --lock $MANIFEST # generates opensearch-1.1.0.yml.lock
+
+MANIFEST_SHA=$(sha1sum $MANIFEST.lock | cut -f1 -d' ') # generate a checksum of the stable manifest
+
+if test -f "$SHAS/$MANIFEST_SHA.lock"; then
+  echo "Skipping $MANIFEST_SHA"
+else
+  ./build.sh $MANIFEST.lock # rebuild using stable references in .lock
```
peternied, line 202: The lock check @199 aside, what happens if a build from x64 is attempted on an arm machine, will it error out?

peternied (commented):
Finished with more detailed review with some scope/design questions coming out of it.

README.md
```diff
@@ -176,12 +177,34 @@ The following options are available in `build.sh`.
...
+| -l, --lock         | Generate a stable reference manifest.                                   |
 | -v, --verbose      | Show more verbose output.                                               |
 
 ##### Custom Build Scripts
 
 Each component build relies on a `build.sh` script that is used to prepare bundle artifacts for a particular bundle version that takes two arguments: version and target architecture. By default the tool will look for a script in [scripts/components](scripts/components), then in the checked-out repository in `build/build.sh`, then default to a Gradle build implemented in [scripts/default/build.sh](scripts/default/build.sh).
 
+##### Avoiding Rebuilds
+
+Builds can automatically generate a `manifest.lock` file with stable git references (commit IDs) and build options (platform, architecture and snapshot) by specifying `--lock`. The output can then be reused as input manifest after checking against a collection of prior builds.
```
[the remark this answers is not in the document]
dblock: In here we don't reuse the **build** manifest, but use the **input** manifest. We _could_ transform a build manifest into a stable input manifest or reuse a build manifest, but if we were to use the build manifest we'd have to find out what the latest build was to retrieve it first.

src/manifests/input_manifest.py
dblock: There's 100% chance that this repo will have changed a month after, say, a 1.0 release. But that's just tooling and shouldn't need to rebuild 1.0.

dblock: We could include the SHA of the original manifest file, but how and why would we use that?

src/manifests/input_manifest.py
dblock: It's actually a feature of the build in the build manifest, so I think this is desirable. The fact that they are passed command line is not material IMO.

README.md
dblock: The manifest would be new here - it would have been re-generated above when we ran with `--lock`.

src/manifests/input_manifest.py
peternied: You've convinced me, I don't think we need to be overzealous in the reproducibility space, especially since jenkins isn't going to have a stable ref for a considerable time.

README.md
peternied: If we are storing manifests in an external location this could happen by accident and possibly go unnoticed

peternied (approved):
Remaining unresolved questions should not block this pull request.

Issue #731, opened 2021-10-12
Refactor common parts of OpenSearch and OpenSearch Dashboards Jenkins jobs into a Jenkins shared library


dblock:
### Is your feature request related to a problem? Please describe

We have numerous common parts in OpenSearch and OpenSearch Dashboards Jenkins jobs, such as bundle workflow and notifications. 

### Describe the solution you'd like

Common parts should be supplied from a jenkins workflow library instead of copy-pasting.

https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/Jenkinsfile
https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/Jenkinsfile

dblock:
Closed via #975

PR #921, opened 2021-11-09
Extract Jenkins messages into a shared library.

dblock:

### Description

Before writing more copy-paste code in Jenkinsfiles for https://github.com/opensearch-project/opensearch-build/pull/904 I thought I'd try to refactor some existing Jenkinsfile code. This PR extracts messages into two shared libraries.

Jenkins libraries live in `vars` by convention. Workarounds suggest doing `git init` in the subfolder and using that, but I couldn't get it to work. There also seems to be [hints at a `libraryPath` option](https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/8984698474fdfbef2cb921fb0febb55d22668ba7/src/main/java/org/jenkinsci/plugins/workflow/libs/SCMRetriever.java#L82), but I wasn't able to set it anywhere.

I have an example running in our private staging Jenkins instance under `job/dblock-refactor-jenkinsfile` if you want to take a look at a successful run.
 
### Issues Resolved

Beginning of https://github.com/opensearch-project/opensearch-build/issues/731

### Relevant Links

- https://stackoverflow.com/questions/46213913/load-jenkins-pipeline-shared-library-from-same-repository
- https://code-held.com/2020/01/22/jenkins-local-shared-library/

### Check List

### DIFF
--- a/jenkins/opensearch-dashboards/Jenkinsfile
+++ b/jenkins/opensearch-dashboards/Jenkinsfile
@@ -1,3 +1,6 @@
+library identifier: 'addJenkinsMessage@current', retriever: legacySCM(scm)
+library identifier: 'getJenkinsMessages@current', retriever: legacySCM(scm)
+
 pipeline {
     agent none
     triggers {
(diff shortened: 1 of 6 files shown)


peternied (approved):
Nice! And its good to know when I'm building Jenkinsfiles it takes more than 10 tries for others as well

dblock:
> Nice! And its good to know when I'm building Jenkinsfiles it takes more than 10 tries for others as well

When you force push, Jenkins still tracks the old ref. It took me 10 tries to figure that out because it wasn't picking up my changes. The more I use it the more I appreciate GHA.

kavilla (approved):
Awesome. Thanks!

gaiksaya (approved):

PR #927, opened 2021-11-09
Revert "Extract Jenkins messages into a shared library."

peternied:
Reverts opensearch-project/opensearch-build#921

This broken the CI build, going to roll back so we can get a RC build
```
WorkflowScript: 137: Expected a step @ line 137, column 25.
                           messages = getJenkinsMessages(['build-x64', 'build-arm64'])
```


dblock (approved):

dblock:
My bad, sorry. The problem is that the execution here needs to be inside a `script { }` block.

Issue #908, opened 2021-11-08
Create Docker images off of bundle build


peternied:
In order to make validating 1.2.0 easier, provide docker images that can be pulled for testing

- [ ] Add new docker references in the staging docker
- [ ] Update the build process to publish the image
  - [dockerfile](https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/dockerfiles/opensearch.al2.dockerfile), https://hub.docker.com/r/opensearchstaging/opensearch
  - [dockerfile](https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile),  https://hub.docker.com/r/opensearchstaging/opensearch-dashboards

- [ ] Update the notification to include a notice about the docker image

This should allow anyone to create a security/no-security build for OpenSearch and OpenSearch Dashboards

peterzhuamazon:
Just sync up with @zelinh and he will add a post task to bundle-build to trigger docker-build with these commands for now:

```
$ cd docker/release && curl -sSL $URL -o opensearch.tgz && bash build-image-multi-arch.sh -v 1.2.0 -f ./dockerfiles/opensearch.al2.dockerfile -p opensearch -a 'x64,arm64' -r opensearchstaging/opensearch -t ./opensearch.tgz
```

VERSION will be dynamically retrieved from bundle-build as well as the URL for tarball downloads.

Thanks.

peterzhuamazon:
We are able to push the 1st draft to dockerhub with Jenkins.
@zelinh is still improving the code.
We will be able to create a PR soon.

```

$ curl https://localhost:9200 -u admin:admin --insecure
{
  "name" : "4ef213db75b5",
  "cluster_name" : "docker-cluster",
  "cluster_uuid" : "gOYhllg8RTecEeAUizpdJw",
  "version" : {
    "distribution" : "opensearch",
    "number" : "1.2.0",
    "build_type" : "tar",
    "build_hash" : "d3101081aff62aa317c0b0334305f99f570a80ee",
    "build_date" : "2021-11-09T00:57:54.309054Z",
    "build_snapshot" : false,
    "lucene_version" : "8.10.1",
    "minimum_wire_compatibility_version" : "6.8.0",
    "minimum_index_compatibility_version" : "6.0.0-beta1"
  },
  "tagline" : "The OpenSearch Project: https://opensearch.org/"
}
```

[...]

PR #928, opened 2021-11-09
Add docker build trigger after bundle builds

peterzhuamazon:
Signed-off-by: Peter Zhu <zhujiaxi@amazon.com>

### Description
Add docker build trigger after bundle builds
 
### Issues Resolved
#908
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #929, opened 2021-11-09
Trigger docker image build after successful bundle build dashboards

zelinh:

### Description
Add docker build trigger after bundle builds dashboards
 
### Issues Resolved
part of #908 for dashboards
 
### Check List

### DIFF
--- a/jenkins/opensearch-dashboards/Jenkinsfile
+++ b/jenkins/opensearch-dashboards/Jenkinsfile
@@ -131,6 +131,19 @@ pipeline {
                 success {
                     node('Jenkins-Agent-al2-x64-c54xlarge-Docker-Host') {
                         publishNotification(":white_check_mark:", "Successful Build", "\n${getAllJenkinsMessages()}")
+                        def VERSION = sh(script: 'echo ${INPUT_MANIFEST} | grep -Po "[0-9.]+?(?=.yml)"', returnStdout: true).trim()
+                        sh "echo VERSION:${VERSION} BUILD_NUMBER:${env.BUILD_NUMBER}"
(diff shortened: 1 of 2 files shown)


peterzhuamazon (approved):
Thanks.

gaiksaya (approved):

jenkins/opensearch-dashboards/Jenkinsfile
```diff
@@ -131,6 +131,19 @@ pipeline {
                 success {
                     node('Jenkins-Agent-al2-x64-c54xlarge-Docker-Host') {
                         publishNotification(":white_check_mark:", "Successful Build", "\n${getAllJenkinsMessages()}")
```
dblock, line 133: Same as for #928, this notifies before it finishes, potentially hiding the error/failing the workflow in a success step.

Any reason why building a docker image isn't a regular step in the workflow?

zelinh: > Same as for #928, this notifies before it finishes, potentially hiding the error/failing the workflow in a success step.
> 
> Any reason why building a docker image isn't a regular step in the workflow?

Because I think we don't wish to build the docker image if the bundle build workflow fails. Furthermore, our approach to trigger the `docker-build` workflow will take the tarballs from successful bundle build.

peternied: In the pipeline workflow there are serial and parallel stages.  If in series, if a previous stage fails it will prevent the docker image from being created.  Creating docker image does not have to happen via job invocation as done here, it can be included in this pipeline as a stage, which this is a form of code sharing the code @dblock just created a code review for a much more testable mechanism.

We have written in these jenkinsfiles in a fragile way for haste, great bias for action.  Lets update this code to be more maintainable by creating an issue to clean up the following areas: 
* Use the generated manifest files to lookup the artifacts, removing duplication of url generation, stashing the manifest would be appropriate
* Move docker image generation into a new stage and add notifications after it has finished
* Move docker image generation code into shared library

I recommend we do these one at a time to ease troubleshooting and speed it up.

PR #938, opened 2021-11-10
Added Test Publisher

owaiskazi19:
### Description

- Implemented Test Publisher to store result of the test to S3. 
- Path to test-results `bundlebuild/${opensearch_version}/${build_id}/${platform}/${architecture}/test-results/test_run_id/test_type/component`. 
- Added `platform` as a build parameter.
- Fixed run command for Integ Test to execute without S3 dependencies

Ref: https://github.com/opensearch-project/opensearch-build/issues/207 for the discussion.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/341

### Check List

### DIFF
--- a/jenkins/test/testsuite/Jenkinsfile
+++ b/jenkins/test/testsuite/Jenkinsfile
@@ -59,7 +59,7 @@ pipeline {
                 }
             }
             steps {
-                sh "./test.sh ${JOB_NAME} --s3-bucket ${ARTIFACT_BUCKET_NAME} --opensearch-version ${opensearch_version} --build-id ${build_id} --architecture ${architecture} --test-run-id ${test_run_id}"
+                sh "./test.sh integ-test --s3-bucket ${ARTIFACT_BUCKET_NAME} --opensearch-version ${opensearch_version} --build-id ${build_id} --architecture ${architecture} --test-run-id ${test_run_id}"
             }
             post() {
                 always {
--- a/jenkins/test/testsuite/Jenkinsfile
+++ b/jenkins/test/testsuite/Jenkinsfile
@@ -59,6 +59,7 @@ pipeline {
                 }
             }
             steps {
+                sh "pip3 install --upgrade pip"
                 sh "./test.sh integ-test --s3-bucket ${ARTIFACT_BUCKET_NAME} --opensearch-version ${opensearch_version} --build-id ${build_id} --architecture ${architecture} --test-run-id ${test_run_id}"
             }
             post() {
--- a/jenkins/test/testsuite/Jenkinsfile
+++ b/jenkins/test/testsuite/Jenkinsfile
@@ -60,7 +60,7 @@ pipeline {
             }
             steps {
                 sh "pip3 install --upgrade pip"
-                sh "./test.sh integ-test --s3-bucket ${ARTIFACT_BUCKET_NAME} --opensearch-version ${opensearch_version} --build-id ${build_id} --architecture ${architecture} --test-run-id ${test_run_id}"
+                sh "./test.sh integ-test https://ci.opensearch.org/ci/dbc/bundle-build/${opensearch_version}/${build_id}/linux/${architecture}"
             }
             post() {
                 always {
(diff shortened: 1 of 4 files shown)


test.sh
```diff
@@ -9,6 +9,7 @@
 set -e
 
 DIR="$(dirname "$0")"
+mkdir -p 'test-results'
```
dblock, line 12: Let's do this in code before we write results like for all the other paths (we create directories for builds and dist).

dblock (changes requested):

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -34,6 +34,11 @@ pipeline {
                                             name: 'test_run_id',
```
gaiksaya, line 34: AFAIR, test_run_id is the jenkins `build_number`. Should that be set as default value for `test_run_id` param? That is not something a user will or should set

owaiskazi19: Coming from the code: https://github.com/opensearch-project/opensearch-build/blob/main/src/run_integ_test.py#L33. `test_run_id` is passed like this. It's safe to remove it from the param.

owaiskazi19: Pushed the change.

gaiksaya: It is being retrieved from TestArgs() https://github.com/opensearch-project/opensearch-build/blob/9bc0a41493d99b2c03f5193ca5d0437b6609907f/src/test_workflow/test_args.py#L34

gaiksaya: I dont think it should be removed. How will we know which test results are from which workflow run. Or is that information not relevant?

@peternied @dblock thoughts? Also don't know why we added `uuid.uuid4().hex` for test_run_id.

owaiskazi19: Are you suggesting a path like `bundlebuild/${opensearch_version}/${build_id}/${platform}/${architecture}/test_run_id/test-results/test_type/component`?

gaiksaya: Nope that is take care of in test_recorder. Ask is just set the `$BUILD_NUMBER` as default value of `test_run_id` in jenkins parameters.

owaiskazi19: I have removed `test_run_id` from Jenkinsfile as test_run_id is already being set in the code mentioned above.

dblock (approved):

jenkins/test/testsuite/Jenkinsfile
owaiskazi19: As discussed offline with @gaiksaya. I'll change the path to `bundlebuild/${opensearch_version}/${build_id}/${platform}/${architecture}/test_run_id/test-results/test_type/component` to record BUILD_NUM as the test_run_id

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -59,13 +60,28 @@ pipeline {
                 }
             }
             steps {
-                sh "./test.sh ${JOB_NAME} --s3-bucket ${ARTIFACT_BUCKET_NAME} --opensearch-version ${opensearch_version} --build-id ${build_id} --architecture ${architecture} --test-run-id ${test_run_id}"
+                script {
+                    basePath = "https://ci.opensearch.org/ci/dbc/bundle-build/${opensearch_version}/${build_id}/${platform}/${architecture}"
```
peternied, line 64: Can we make this a parameter to the job? We just had a build break yesterday night due to building these paths inconsistently

owaiskazi19: `https://ci.opensearch.org/ci/dbc/bundle-build/`should be a parameter to the job?

peternied: Even more - ` basePath = "https://ci.opensearch.org/ci/dbc/bundle-build/${opensearch_version}/${build_id}/${platform}/${architecture}"` could be passed in a parameter, only the architecture is used to determine the agent node type all the rest of the values are passed around to no effect.

...lets come back to this after you merge to build momentum

peternied (approved):

gaiksaya (approved):
LGTM. Thanks!

Issue #934, opened 2021-11-10
[Bug]: Reporting build missing chromium dependency


joshuali925:
### Describe the bug

dashboards-reports plugin uses chromium to generate reports, the binary should be included in the artifact like this:
https://github.com/opensearch-project/dashboards-reports/blob/main/.github/workflows/dashboards-reports-test-and-build-workflow.yml#L67-L91

But this part is missing in [build.sh](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/components/reportsDashboards/build.sh#L64-L77).

### To reproduce

Run opensearch and dashboards, reporting does not work because chromium binary `reportsDashboards/.chromium/headless_shell` does not exist

peternied:
>What are the possible values of PLATFORM and ARCHITECTURE in build.sh?

`PLATFORM` can be: 
* linux
* darwin
* windows

`ARCHITECTURE` can be:
* x64
* arm64

See [example usage](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/components/OpenSearch/build.sh#L81-L113)

joshuali925:
will `ARCHITECTURE` be supplied? 
```bash
[ -z "$ARCHITECTURE" ] && ARCHITECTURE=`uname -m`
```
this can also be `x86_64`, `amd64`, `aarch64`, ... but the case only accepts `x64` and `arm64`

peternied:
This a better source of truth, here are the values that will be allowed by the build system

```python
class BuildArgs:
    SUPPORTED_PLATFORMS = ["linux", "darwin", "windows"]
    SUPPORTED_ARCHITECTURES = [
        "x64",
        "arm64",
    ]
    ```
https://github.com/opensearch-project/opensearch-build/blob/main/src/build_workflow/build_args.py#L12-L17

**Comment by dblock** [comment]
> I see, thanks. What should I do if some combinations are not supported by reporting yet? We currently do not have chromium builds for darwin-arm64 and windows-arm64.

Manifests support limiting platforms like [here](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/1.2.0/opensearch-1.2.0.yml#L50). There's no support for architectures, we probably want to extend that code to be able to write:

```
platforms:
   - windows:x64
   - darwin:arm64
```

Feel free to open an issue/PR :)

> i think it's trying to do `PLATFORM=$(uname -s | awk '{print tolower($0)}')`

Please fix!

PR #939, opened 2021-11-10
Fix assigning variable for detecting platform

joshuali925:
Signed-off-by: Joshua Li <joshuali925@gmail.com>

### Description
Found in https://github.com/opensearch-project/opensearch-build/issues/934#issuecomment-965653497.
Seems like the original implementation will pipe `PLATFORM` assignment output to awk (not very sure what the syntax would do, but `$PLATFORM` will always be empty). Use `$(...)` for variable assignments.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/scripts/components/OpenSearch-Dashboards/build.sh
+++ b/scripts/components/OpenSearch-Dashboards/build.sh
@@ -57,7 +57,7 @@ if [ -z "$VERSION" ]; then
 fi
 
 [ -z "$OUTPUT" ] && OUTPUT=artifacts
-[ -z "$PLATFORM" ] && PLATFORM=`uname -s` | awk '{print tolower($0)}'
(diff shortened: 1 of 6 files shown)


peterzhuamazon (approved):
Thanks.

peterzhuamazon:
Using `` is ok the issue is the original implementation close the block before values pipe to awk, cause awk to have no input data.
Thanks for fixing.

dblock (approved):

joshuali925:
> Using `` is ok the issue is the original implementation close the block before values pipe to awk, cause awk to have no input data. Thanks for fixing.

yes it would also work if `` wraps the pipe. FYI on the difference from shellcheck https://github.com/koalaman/shellcheck/wiki/Sc2006

PR #946, opened 2021-11-11
Align bwctest finder on other scripts.

dblock:

### Description

The order of scripts for bwctests is different than for other scripts, align those, add tests.

### Check List

### DIFF
--- a/scripts/components/OpenSearch/bwctest.sh
+++ b/scripts/components/OpenSearch/bwctest.sh
@@ -6,4 +6,4 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
-# TODO: This is used by ScriptFinder tests, needs a working integtest.sh for OpenSearch.
+# TODO: This is used by ScriptFinder tests, needs a working bwctest.sh for OpenSearch.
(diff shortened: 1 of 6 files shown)


tests/tests_test_workflow/test_bwc_workflow/bwc_test/test_bwc_suite.py
```diff
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 #
-# The OpenSearch Contributors require contributions made to
+# The job-scheduler Contributors require contributions made to
```
VachaShah, line 3: Does this need to change?

scripts/components/OpenSearch/bwctest.sh
```diff
@@ -0,0 +1 @@
+# TODO: This is used by ScriptFinder tests, needs a working bwctest.sh for OpenSearch.
```
owaiskazi19, line 1: Do we need to have a header for this file as well?

tests/tests_test_workflow/test_bwc_workflow/bwc_test/test_bwc_suite.py
dblock: 🤦

scripts/components/OpenSearch/bwctest.sh
dblock: Added.

VachaShah (approved):
LGTM!

owaiskazi19 (approved):
Thanks for the changes!

peternied (approved):

PR #962, opened 2021-11-15
Moved docker build into a stage.

dblock:

### Description

Followup on https://github.com/opensearch-project/opensearch-build/pull/928 and https://github.com/opensearch-project/opensearch-build/pull/929/files#r746192714.

Jenkins will abort execution of subsequent pipelines on failure AFAIK, unless [you tell it not to](https://newbedev.com/scripted-jenkins-pipeline-continue-on-fail), so there's no need to have the docker build stage on success. Moved the docker build stage after the parallel build stage and moved the `post` stage to the pipeline level from the build stages. 

### Check List

### DIFF
--- a/jenkins/opensearch-dashboards/Jenkinsfile
+++ b/jenkins/opensearch-dashboards/Jenkinsfile
@@ -129,34 +129,41 @@ pipeline {
                     }
                 }
             }
-            post() {
-                success {
-                    node('Jenkins-Agent-al2-x64-c54xlarge-Docker-Host') {
-                        script {
-                            def VERSION = sh(script: 'echo ${INPUT_MANIFEST} | grep -Po "[0-9.]+?(?=.yml)"', returnStdout: true).trim()
-                            sh "echo VERSION:${VERSION} BUILD_NUMBER:${env.BUILD_NUMBER}"
-                            def URL_x64 = "${PUBLIC_ARTIFACT_URL}/${env.JOB_NAME}/${VERSION}/${env.BUILD_NUMBER}/linux/x64/dist/opensearch-dashboards-${VERSION}-linux-x64.tar.gz"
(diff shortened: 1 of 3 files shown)


peternied (approved):

gaiksaya:
So this stage is at the same level of `build` right? Not inside build? Just confirming

dblock:
> So this stage is at the same level of `build` right? Not inside build? Just confirming

It is. You can't add a `stage` next to a `parallel` construct anyway, that fails with an error.

Screen Shot 2021-11-15 at 12 03 38 PM

gaiksaya (approved):

peterzhuamazon (changes requested):
Block this as I dont think it is necessary to still trigger a docker build if the bundle build failed. I put it to the post/success section for exactly this purpose.

gaiksaya:
> Block this as I dont think it is necessary to still trigger a docker build if the bundle build failed. I put it to the post/success section for exactly this purpose.

It dont trigger docker build if the bundle build failed. Its a serial job. Not parallel one.

dblock:
> > Block this as I dont think it is necessary to still trigger a docker build if the bundle build failed. I put it to the post/success section for exactly this purpose.
> 
> It won't trigger docker build if the bundle build failed. Its a serial job. Not parallel one.

@peterzhuamazon ^

zelinh (approved):
This makes senses to me.

peterzhuamazon:
> > > Block this as I dont think it is necessary to still trigger a docker build if the bundle build failed. I put it to the post/success section for exactly this purpose.
> > 
> > 
> > It won't trigger docker build if the bundle build failed. Its a serial job. Not parallel one.
> 
> @peterzhuamazon ^

I see it now. Thanks for explanation.

peterzhuamazon (approved):

PR #966, opened 2021-11-16
Extracted artifact paths into input and build manifests.

dblock:

### Description

We do a lot of duplicate path work in Jenkinsfile. This PR extracts some of those concerns into class methods and has the benefit of adding tests for those. 

- Loads input and build manifests into `InputManifest` and `BuildManifest` classes, similar to what we have in Python.
- Extracts path and URL combinations into `InputManifest` and `BuildManifest` methods.
- Makes docker build command line generation readable by making an array and combining it with `&&`.

Note that it would make a lot of sense to pass the manifest file name and not the result of `readYaml` into `Manifest` constructors, but that doesn't work because of some kind of security sandbox in Jenkins. Passing `this` (steps) into the class and then calling `this.sh` works, but calling `this.readYaml` does not (aborts the script without error (?)). Trying to bring a third-party yaml loader library into Jenkins like `org.yaml.snakeyaml` used for mocks, seems even more tedious.  

Now that `build` is almost identical, and `assemble` is identical between `opensearch/Jenkinsfile` and `opensearch-dashboards/Jenkinsfile`, they can finally be moved to a shared implementation in a future PR.
 
### Check List


jenkins/opensearch-dashboards/Jenkinsfile
```diff
@@ -122,16 +122,20 @@ pipeline {
...
+                                string(name: 'DOCKER_BUILD_SCRIPT_WITH_COMMANDS', value: [
+                                        'id',
+                                        'pwd',
+                                        'cd docker/release',
+                                        "curl -sSL ${inputManifest.getPublicDistPath(env.PUBLIC_ARTIFACT_URL, env.BUILD_NUMBER, 'linux','x64')} -o opensearch-dashboards-x64.tgz",
+                                        "curl -sSL ${inputManifest.getPublicDistPath(env.PUBLIC_ARTIFACT_URL, env.BUILD_NUMBER, 'linux','arm64')} -o opensearch-dashboards-arm64.tgz",
+                                        "bash build-image-multi-arch.sh -v ${inputManifest.build.version} -f ./dockerfiles/opensearch-dashboards.al2.dockerfile -p opensearch-dashboards -a 'x64,arm64' -r opensearchstaging/opensearch-dashboards -t 'opensearch-dashboards-x64.tgz,opensearch-dashboards-arm64.tgz' -n ${env.BUILD_NUMBER}"
+                                    ].join(" && ")),
```
peterzhuamazon, line 138: Great changes.

peterzhuamazon (approved):
Thanks.

jenkins/opensearch-dashboards/Jenkinsfile
```diff
@@ -122,16 +122,20 @@ pipeline {
...
-                                string(name: 'DOCKER_BUILD_SCRIPT_WITH_COMMANDS', value: "id && pwd && cd docker/release && curl -sSL $URL_x64 -o opensearch-dashboards-x64.tgz && curl -sSL $URL_arm64 -o opensearch-dashboards-arm64.tgz && bash build-image-multi-arch.sh -v ${VERSION} -f ./dockerfiles/opensearch-dashboards.al2.dockerfile -p opensearch-dashboards -a 'x64,arm64' -r opensearchstaging/opensearch-dashboards -t 'opensearch-dashboards-x64.tgz,opensearch-dashboards-arm64.tgz' -n ${env.BUILD_NUMBER}"),
+                                string(name: 'DOCKER_BUILD_SCRIPT_WITH_COMMANDS', value: [
+                                        'id',
+                                        'pwd',
+                                        'cd docker/release',
+                                        "curl -sSL ${inputManifest.getPublicDistPath(env.PUBLIC_ARTIFACT_URL, env.BUILD_NUMBER, 'linux','x64')} -o opensearch-dashboards-x64.tgz",
+                                        "curl -sSL ${inputManifest.getPublicDistPath(env.PUBLIC_ARTIFACT_URL, env.BUILD_NUMBER, 'linux','arm64')} -o opensearch-dashboards-arm64.tgz",
```
zelinh, line 136: Are these supposed to call `getPublicDistUrl` method?

jenkins/opensearch/Jenkinsfile
```diff
@@ -123,16 +117,20 @@ pipeline {
...
                                 string(name: 'DOCKER_BUILD_GIT_REPOSITORY_REFERENCE', value: 'main'),
-                                string(name: 'DOCKER_BUILD_SCRIPT_WITH_COMMANDS', value: "id && pwd && cd docker/release && curl -sSL $URL_x64 -o opensearch-x64.tgz && curl -sSL $URL_arm64 -o opensearch-arm64.tgz && bash build-image-multi-arch.sh -v ${VERSION} -f ./dockerfiles/opensearch.al2.dockerfile -p opensearch -a 'x64,arm64' -r opensearchstaging/opensearch -t 'opensearch-x64.tgz,opensearch-arm64.tgz' -n ${env.BUILD_NUMBER}"),
+                                string(name: 'DOCKER_BUILD_SCRIPT_WITH_COMMANDS', value: [
+                                        'id',
+                                        'pwd',
+                                        'cd docker/release',
+                                        "curl -sSL ${inputManifest.getPublicDistPath(env.PUBLIC_ARTIFACT_URL, env.BUILD_NUMBER, 'linux','x64')} -o opensearch-x64.tgz",
+                                        "curl -sSL ${inputManifest.getPublicDistPath(env.PUBLIC_ARTIFACT_URL, env.BUILD_NUMBER, 'linux','arm64')} -o opensearch-arm64.tgz",
```
zelinh, line 131: Same with above ^^

jenkins/opensearch-dashboards/Jenkinsfile
dblock: Good catch!

jenkins/opensearch/Jenkinsfile
dblock: Yep, I removed the method altogether since it's not used.

src/jenkins/BuildManifest.groovy
```diff
@@ -0,0 +1,55 @@
...
+
+    BuildManifest(Map data) {
+        super(data)
+
+        this.build = new BuildManifest.Build(data.build)
+    }
+
+    public String getArtifactRoot(String buildNumber) {
+        return [
+            this.build.name.toLowerCase().replaceAll(' ', '-'),
```
zelinh, line 39: Are we changing the name here from `bundle-build/bundle-build-dashboards` to `opensearch/opensearch-dashboards`?

dblock: In the build manifest ([example](https://github.com/opensearch-project/opensearch-build/blob/main/tests/tests_ci_workflow/data/opensearch-1.1.0-x64-build-manifest.yml#L4)) this is "OpenSearch" or "OpenSearch Dashboards", so it becomes 'opensearch' and 'opensearch-dashboards'.

zelinh: I think our URL used to use the jenkins job names in the path which are `bundle-build` and `bundle-build-dashboards`,  and they help us distinguish between `OpenSearch` and `OpenSearch Dashboards`. I am good with these names change because they work for the same purpose. Just make sure we are aware of this.

dblock: Oh I see! I didn't realize we're using the job name for this purpose, this is my bad. Let's not change this - I added a change that uses `env.JOB_NAME` instead, please re-review and thanks for catching this!

zelinh (approved):
Thanks!

peternied (approved):

PR #970, opened 2021-11-16
Use concurrency when copy/download files

peternied:

### Description
Speed up downloads / copying by using a threadpool
 
### Tested
Add profiling code to the area where the concurrency is getting added, run ` ./test.sh integ-test https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/1048/linux/x64` from my cloud desktop, download only

- Existing code `Elapsed time: 260.8771 seconds`
- PR code `Elapsed time: 22.4884 seconds`

That is an order of magnitude faster, clear winner to me.

### Check List

### DIFF
--- a/src/test_workflow/dependency_installer.py
+++ b/src/test_workflow/dependency_installer.py
@@ -4,6 +4,7 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+import concurrent.futures
 import logging
 import os
 import shutil
@@ -52,17 +53,19 @@ class DependencyInstaller:
 
     def download(self, paths, category, dest):
         logging.info(f"Downloading to {dest} ...")
-        for path in paths:
-            url = "/".join([self.root_url, category, path])
-            # paths are prefixed by category, remove
-            local_path = os.path.join(dest, "/".join(path.split("/")[1:]))
-            self.__download_or_copy(url, local_path)
+        with concurrent.futures.ThreadPoolExecutor(max_workers=4) as executor:
+            for path in paths:
+                url = "/".join([self.root_url, category, path])
+                # paths are prefixed by category, remove
+                local_path = os.path.join(dest, "/".join(path.split("/")[1:]))
+                executor.submit(self.__download_or_copy, url, local_path)
 
     def __download_or_copy(self, source, dest):
         dest = os.path.realpath(dest)
         os.makedirs(os.path.dirname(dest), exist_ok=True)
         if validators.url(source):
             logging.info(f"Downloading {source} into {dest} ...")
+
             urllib.request.urlretrieve(source, dest)
         else:
             logging.info(f"Copying {source} into {dest} ...")
--- a/tests/tests_test_workflow/test_dependency_installer.py
+++ b/tests/tests_test_workflow/test_dependency_installer.py
@@ -1,6 +1,6 @@
 import os
 import unittest
-from unittest.mock import call, patch
+from unittest.mock import MagicMock, patch
 
 from manifests.build_manifest import BuildManifest
 from manifests.bundle_manifest import BundleManifest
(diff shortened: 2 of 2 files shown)


tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -60,13 +75,9 @@ def test_install_maven_dependencies_remote(self, mock_request, mock_copyfile, mo
...
         )
         mock_copyfile.assert_not_called()
-        mock_request.assert_has_calls(
-            [
-                call(
-                    "https://ci.opensearch.org/x/y/builds/maven/org/opensearch/notification/alerting-notification-1.2.0.0.jar",
-                    os.path.join(dependency_installer.maven_local_path, "org", "opensearch", "notification", "alerting-notification-1.2.0.0.jar"),
-                )
-            ]
+        mock_request.assert_called_with(
```
peternied, line 78: When I run on my dev box with pre-commit or running pytest I see errors where `/local` is appended to the path.  Is this a bad setup thing?  Curious in case others have seen it
```
>           raise AssertionError(_error_message()) from cause
E           AssertionError: Expected call: urlretrieve('https://ci.opensearch.org/x/y/builds/maven/org/opensearch/notification/alerting-notification-1.2.0.0.jar', '/home/petern/.m2/repository/org/opensearch/notification/alerting-notification-1.2.0.0.jar')
E           Actual call: urlretrieve('https://ci.opensearch.org/x/y/builds/maven/org/opensearch/notification/alerting-notification-1.2.0.0.jar', '/local/home/petern/.m2/repository/org/opensearch/notification/alerting-notification-1.2.0.0.jar')
```

dblock: That's the different between `path.realpath()` and not I believe, something with symlinks, don't remember. See if you can fix this by comparing `path.realpaths`?

src/test_workflow/dependency_installer.py
```diff
@@ -52,17 +53,19 @@ def download_dist(self, dest):
 
     def download(self, paths, category, dest):
         logging.info(f"Downloading to {dest} ...")
-        for path in paths:
-            url = "/".join([self.root_url, category, path])
-            # paths are prefixed by category, remove
-            local_path = os.path.join(dest, "/".join(path.split("/")[1:]))
-            self.__download_or_copy(url, local_path)
+        with concurrent.futures.ThreadPoolExecutor(max_workers=4) as executor:
```
dblock, line 56: Nit: `max_workers` should default to number of CPUs.

dblock (approved):
Nice.

src/test_workflow/dependency_installer.py
peternied: Download/Copy isn't CPU bound (typically) I want to make sure docker image copy is accelerated as well.

tests/tests_test_workflow/test_dependency_installer.py
peternied: Yup, right on.  I made sure the calling patterns align now and it works properly on my machine.

tests/tests_test_workflow/test_dependency_installer.py
```diff
@@ -13,16 +13,26 @@ class DependencyInstallerTests(unittest.TestCase):
...
+            return None
+        mock_threadpool.return_value.__enter__().submit.side_effect = submit_and_run
         dependency_installer = DependencyInstaller(
             self.DATA,
             BuildManifest.from_path(self.BUILD_MANIFEST),
             BundleManifest.from_path(self.DIST_MANIFEST_LOCAL)
         )
         dependency_installer.install_maven_dependencies()
+        self.assertEqual(mock_threadpool.call_count, 4)
+        self.assertEqual(mock_threadpool().__enter__().submit.call_count, 2375)
```
gaiksaya, line 34: Can you explain how is submit call count 2375? Never used threadpools before

peternied: that `submit(...)` function is called 2375 times, for each file in the manifest.  This assert is making sure that we actually performed the action that number of times.  If something was wrong with the executor it could record more/less and we would want to know

gaiksaya: Got it. Thanks!

gaiksaya (approved):

Issue #960, opened 2021-11-12
Notification after Integration Test run on Jenkins


owaiskazi19:
### Is your feature request related to a problem? Please describe

Currently, `TestPublisher` is publishing test results to S3. To make the results more visible send a notification informing about the test run.

### Describe the solution you'd like

Notification can be send to either slack or through email.

gaiksaya:
[Triage] @owaiskazi19 Are you working on this? If yes can you assign it to yourself and add the approach here?
Thanks!

owaiskazi19:
> [Triage] @owaiskazi19 Are you working on this? If yes can you assign it to yourself and add the approach here? Thanks!

I have assigned it. I'll add the approach once we finalize it. Thanks.

owaiskazi19:
Followed 2 designs for the issue:

1. [Discarded] Sending a single notification after every bundle is build and tested. Draft PR for the same: https://github.com/opensearch-project/opensearch-build/pull/972. Based on the discussion as all 3 tests Integ, Perf and BWC combined takes around 10 hrs to run. It'll delay the notification for a longer time.

2. Creating a separate channel for sending notification of tests as when they are run. https://github.com/opensearch-project/opensearch-build/pull/1105

PR #972, opened 2021-11-16
Notification after Integ Test run

owaiskazi19:
### Description
Combined the JenkinsFile for build and test to publish a single notification of the build and test.
Notification of Integ Test Run is WIP.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/960
 
### Check List


jenkins/opensearch/Jenkinsfile
```diff
@@ -174,6 +176,21 @@ void build() {
...
+
+    echo "Uploading Test Results to s3://${ARTIFACT_BUCKET_NAME}/${artifactPath}/test-results"
+    withAWS(role: 'opensearch-test', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+        s3Upload(file: 'test-results', bucket: "${ARTIFACT_BUCKET_NAME}", path: "${artifactPath}/test-results")
+    }
+
+}
+
 /** Publishes a notification to a slack instance*/
 void publishNotification(icon, msg, extra) {
```
owaiskazi19, line 195: Need help: How can we handle Notification of the build and test run together here https://github.com/opensearch-project/opensearch-build/pull/972/files#diff-d2e59eb1985423c8e84c1b54324763f503febaea168a3fab1f44339b754cac0fR149?

owaiskazi19: For example: `OpenSearch 1.3.0 #1231 has built and integ test failed`

jenkins/opensearch/Jenkinsfile
```diff
@@ -103,6 +104,7 @@ pipeline {
                     steps {
                         script {
                             build()
+                            test()
```
peternied, line 107: Food for thought, adding the testing immediately after the build step reuses the same docker host (OK!).  We could break the testing into its own stage so we can make the message more granular, then you can create a `post() { failure }` section to deal with tests failing, but the build not failing.

jenkins/opensearch/Jenkinsfile
peternied: `lib.Messages.new(this).add("${STAGE_NAME}", "${baseUrl}/builds/manifest.yml\n${baseUrl}/dist/manifest.yml")` is how to append output to the notifications.

If you wanted to check for pass/failure of tests, you might want to add a message that you can read.  There are some limitations with the message passing, so you'll need to append a '-test' or something to the `"${STAGE_NAME}"` and update the message fetching logic

When you are in the success / failure nodes you can attempt to read the messages `def stashed = lib.Messages.new(this).get(['build-x64', 'build-arm64'])` and take additional action.

peternied (commented):
Great start, pretty soon we will have automated results!

dblock:
I’m a bit concerned that after this we’ll be adding bcw and performance tests. This isn’t very scalable. Definitely split up tests in at least another stage to start, but maybe it should be a separate job triggered by a successful build? Also feels like something i would want to rerun on a transient failure on the same build number and for a given architecture.

peternied:
> I’m a bit concerned that after this we’ll be adding bcw and performance tests. This isn’t very scalable

I'm not sure how long BCW tests take to execute, but performance tests would add 8 hours to the time - which I will agree does not scale.

The current build process takes ~30 minutes, I think doubling this time would be worthwhile to ensure minimal level of functionality for our teams and the community alike.  @owaiskazi19 the integ-test jobs have a number of different times, how long do the tests take to execute today?

owaiskazi19:
Looking at the discussion, if we add the tests in the same job will be difficult to scale after we have BWC and Perf Test integrated as well. Integ test takes around 45min - 1hr, BWC takes around 30min-1hr and Perf Test takes around 8 hrs. So for every build we have to wait for at least 10hrs. @peternied I think it'll be better if we follow the same approach as before of  adding just notification for integ test job. I'm not sure if we should use the same slack channel as build or create a new one for the tests notification. Let me know WDYT.

peternied:
@owaiskazi19 Lets create a new channel for these, and then you can create a new slack bot and we can setup the secret on the cluster

peternied:
Looks like you've already created and used `INTEG_TEST_WEBHOOK` on the jenkins service, looks like you should be all set

owaiskazi19:
Closing this as followed a new design here: https://github.com/opensearch-project/opensearch-build/pull/1105

owaiskazi19:
Closing this as followed a new design here: https://github.com/opensearch-project/opensearch-build/pull/1105

PR #975, opened 2021-11-17
Extract common code into a global library

dblock:

### Description

Extracts build, assemble, upload and notify into separate functions part of the global library, each now with tests.
 
### Check List


vars/uploadArtifacts.groovy
```diff
@@ -0,0 +1,26 @@
...
+            s3Upload(file: 'builds', bucket: ARTIFACT_BUCKET_NAME, path: "${artifactPath}/builds")
+            s3Upload(file: 'dist', bucket: ARTIFACT_BUCKET_NAME, path: "${artifactPath}/dist")
+        } else {
+            echo "s3Upload(file: 'builds', bucket: ${ARTIFACT_BUCKET_NAME}, path: ${artifactPath}/builds)"
+            echo "s3Upload(file: 'dist', bucket: ${ARTIFACT_BUCKET_NAME}, path: ${artifactPath}/dist)"
+        }
+    }
+
+    def baseUrl = buildManifest.getArtifactRootUrl(PUBLIC_ARTIFACT_URL, JOB_NAME, BUILD_NUMBER)
+    lib.jenkins.Messages.new(this).add("${STAGE_NAME}", [
```
peternied, line 21: What about passing the message name as an argument instead of inferring from the stage?  Right now we are limited, to only added messages once in a stage.  Being that we have to fill the value into the `get` call it would be more readable to see the same identifier used twice

peternied (commented):
Really like the simplification of our jobs, looking forward to the additional PRs

vars/uploadArtifacts.groovy
dblock: I think all these functions should take entire maps of arguments that make a bit more sense, but not the kitchen sink of them like stage name and build number. I'll think about how to extract the build info into an object that can be passed around.

dblock:
This is at a good stopping point worth a merge, deploy and test. It accomplishes the extraction into `vars` (this folder name is called this way by convention in Jenkins libraries and cannot be changed), effectively reusing `build`, `assemble` and `publish` without any other changes, so we can use this model for test actions.

vars/uploadArtifacts.groovy
```diff
@@ -0,0 +1,26 @@
...
+    lib = library(identifier: "jenkins@current", retriever: legacySCM(scm))
+
+    def manifestFilename = args.manifest ?: 'builds/manifest.yml'
+    def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: manifestFilename))
+
+    def artifactPath = buildManifest.getArtifactRoot("${JOB_NAME}", "${BUILD_NUMBER}")
+    echo "Uploading to s3://${ARTIFACT_BUCKET_NAME}/${artifactPath}"
+
+    withAWS(role: 'opensearch-bundle', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+        if (args.upload) {
```
peternied, line 11: upload->dryrun?

peternied (approved):

vars/uploadArtifacts.groovy
dblock: I kinda wanted it to be `false` by default as a general rule of thumb, but it's not a pattern (yet). WDYT?

zelinh (approved):
I do like these simplifications. Thanks.

PR #981, opened 2021-11-17
Lower Java version to 11 for OpenSearch 1.3.0.

dblock:

### Description

We've lowered the JDK for OpenSearch to 11 in https://github.com/opensearch-project/OpenSearch/pull/940. 

### Issues Resolved

Part of https://github.com/opensearch-project/opensearch-build/issues/74.
 
### Check List

### DIFF
--- a/manifests/1.3.0/opensearch-1.3.0.yml
+++ b/manifests/1.3.0/opensearch-1.3.0.yml
@@ -1,11 +1,15 @@
+schema-version: '1.0'
 build:
   name: OpenSearch
   version: 1.3.0
+ci:
+  image:
+     name: "opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028"
+     args: "-e JAVA_HOME=/opt/java/openjdk-11"
 components:
-- checks:
-  - gradle:publish
-  - gradle:properties:version
-  name: OpenSearch
-  ref: 1.x
+- name: OpenSearch
+  ref: '1.x'
   repository: https://github.com/opensearch-project/OpenSearch.git
-schema-version: '1.0'
+  checks:
+    - gradle:publish
+    - gradle:properties:version


manifests/1.3.0/opensearch-1.3.0.yml
```diff
@@ -1,11 +1,15 @@
+schema-version: '1.0'
 build:
   name: OpenSearch
   version: 1.3.0
+ci:
+  image:
+     name: "opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028"
+     args: "-e JAVA_HOME=/opt/java/openjdk-11"
```
gaiksaya, line 8: Does this change the integration test environment jdk too?

dblock: Does not today. Because we have never run any automation for integration tests currently it would use whatever JAVA_HOME you have on your machine. When we put that into Jenkins, we'll need to decide what to use. Note that it sort of doesn't matter because you're deploying an OpenSearch cluster with the JVM packaged inside OpenSearch, so if we wanted to test OpenSearch with multiple JVMs, you'd need to produce distributions with different versions of the bundled JVM or swap the JVM before starting any test cluster.

peternied: Note; this is a great example of why I like the integration tests that start up a cluster to be coupled to the build process.. if that is a good enough trade-off is still up for debate.

peternied (approved):

manifests/1.3.0/opensearch-1.3.0.yml
gaiksaya: Makes sense. One way is to have no-jdk distribution and a jdk matrix to run integration test against.

gaiksaya (approved):

dblock:
cc: @reta

reta:
This is awesome, thanks @dblock ! :+1: :+1: :+1:

PR #978, opened 2021-11-17
fix the issue of starting opensearch

tianleh:
Signed-off-by: Tianle Huang <tianleh@amazon.com>

### Description
fix the issue of starting opensearch. See the related bug issue for details.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/977
 
### Test
Run this command
```
./test.sh integ-test https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/998/linux/x64
```

The output is 

```

2021-11-17 09:46:37 INFO     Skipping integ-tests for OpenSearch, as it is currently not declared in the test manifest
2021-11-17 09:46:37 INFO     Skipping integ-tests for job-scheduler, as it is currently not declared in the test manifest
2021-11-17 09:46:37 INFO     Executing "git init" in /tmp/tmpaonj1nll/alerting
2021-11-17 09:46:37 INFO     Executing "git remote add origin https://github.com/opensearch-project/alerting.git" in /tmp/tmpaonj1nll/alerting
2021-11-17 09:46:37 INFO     Executing "git fetch --depth 1 origin 4374e50f70ab7e8b3fe2a85a3b7b098c9e44d171" in /tmp/tmpaonj1nll/alerting
2021-11-17 09:46:38 INFO     Executing "git checkout FETCH_HEAD" in /tmp/tmpaonj1nll/alerting
2021-11-17 09:46:38 INFO     Executing "git rev-parse HEAD" in /tmp/tmpaonj1nll/alerting
2021-11-17 09:46:38 INFO     Checked out https://github.com/opensearch-project/alerting.git@4374e50f70ab7e8b3fe2a85a3b7b098c9e44d171 into /tmp/tmpaonj1nll/alerting at 4374e50f70ab7e8b3fe2a85a3b7b098c9e44d171
2021-11-17 09:46:38 INFO     Additional config found: {'plugins.destination.host.deny_list': ['10.0.0.0/8', '127.0.0.1']}
2021-11-17 09:46:38 INFO     Creating local test cluster in /tmp/tmpaonj1nll/local-test-cluster
2021-11-17 09:46:38 INFO     Downloading bundle
2021-11-17 09:46:38 INFO     Downloading https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/998/linux/x64/dist/opensearch-1.2.0-linux-x64.tar.gz into /tmp/tmpaonj1nll/local-test-cluster/opensearch-1.2.0-linux-x64.tar.gz ...
[...]

PR #987, opened 2021-11-18
rename stdout_data and stderr_data

tianleh:

### Description
This is a follow up to address one comment from https://github.com/opensearch-project/opensearch-build/pull/978/files about variable name. 
 
### Check List

### DIFF
--- a/src/system/process.py
+++ b/src/system/process.py
@@ -53,19 +53,19 @@ class Process:
         logging.info(f"Process killed with exit code {self.process.returncode}")
 
         if self.stdout:
-            self.stdout_data = self.stdout.read()
+            self.stdout_termination_data = self.stdout.read()
             self.stdout.close()
             self.stdout = None
 
         if self.stderr:
-            self.stderr_data = self.stderr.read()
+            self.stderr_termination_data = self.stderr.read()
             self.stderr.close()
             self.stderr = None
 
         self.return_code = self.process.returncode
         self.process = None
 
-        return self.return_code, self.stdout_data, self.stderr_data
+        return self.return_code, self.stdout_termination_data, self.stderr_termination_data
 
     @property
     def started(self):
--- a/src/system/process.py
+++ b/src/system/process.py
@@ -18,6 +18,8 @@ class Process:
         self.process = None
         self.stdout = None
         self.stderr = None
+        self.__stdout_data__ = None
+        self.__stderr_data__ = None
 
     def start(self, command, cwd):
         if self.started:
@@ -53,19 +55,19 @@ class Process:
         logging.info(f"Process killed with exit code {self.process.returncode}")
 
         if self.stdout:
-            self.stdout_termination_data = self.stdout.read()
+            self.__stdout_data__ = self.stdout.read()
             self.stdout.close()
             self.stdout = None
 
         if self.stderr:
-            self.stderr_termination_data = self.stderr.read()
+            self.__stderr_data__ = self.stderr.read()
             self.stderr.close()
             self.stderr = None
 
         self.return_code = self.process.returncode
         self.process = None
 
-        return self.return_code, self.stdout_termination_data, self.stderr_termination_data
+        return self.return_code
 
     @property
     def started(self):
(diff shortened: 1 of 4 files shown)


seraphjiang (approved):

dblock:
Wasn’t this in the context of also renaming output and error to stdout_data and stderr_data?

tianleh:
> Wasn’t this in the context of also renaming output and error to stdout_data and stderr_data? So to distinguish partial and complete results?

Use output as example, do you mean renaming `output` to `stdout` or `stdout_data`? There is a difference here. 

Option 1: `stdout` means that it is a file handler which callers can call `read` function and also check its `mode` (used in unit test).

Option 2: `stdout_data` means it is string data. This will require the `read` call to be inside Process class. On one hand, it simplifies the callers to not make the `read` call. The downside is that the caller cannot easily get its `mode`. Well, I could add two new properties to the Process class. (`stdout_mode`, `stderr_mode`)

Let me know which option you mean. (or Option 3: leave it as it is)

dblock:
> > Wasn’t this in the context of also renaming output and error to stdout_data and stderr_data? So to distinguish partial and complete results?
> 
> Use output as example, do you mean renaming `output` to `stdout` or `stdout_data`? There is a difference here.
> 
> Option 1: `stdout` means that it is a file handler which callers can call `read` function and also check its `mode` (used in unit test).
> 
> Option 2: `stdout_data` means it is string data. This will require the `read` call to be inside Process class. On one hand, it simplifies the callers to not make the `read` call. The downside is that the caller cannot easily get its `mode`. Well, I could add two new properties to the Process class. (`stdout_mode`, `stderr_mode`)
> 
> Let me know which option you mean. (or Option 3: leave it as it is)

I think neither actually. Thinking from the caller point of view, you want the live data while the process is running and complete data after the process ended. Would that simplify the interface?

1. Remove `output` and `error`, never return the handle.
2. Read if process is running and return termination data otherwise.

```python
def stdout_data(self):
   return self.stdout.read() if self.stdout else self.stdout_data
```

WDYT?

tianleh:
> > > Wasn’t this in the context of also renaming output and error to stdout_data and stderr_data? So to distinguish partial and complete results?
> > 
> > 
> > Use output as example, do you mean renaming `output` to `stdout` or `stdout_data`? There is a difference here.
> > Option 1: `stdout` means that it is a file handler which callers can call `read` function and also check its `mode` (used in unit test).
> > Option 2: `stdout_data` means it is string data. This will require the `read` call to be inside Process class. On one hand, it simplifies the callers to not make the `read` call. The downside is that the caller cannot easily get its `mode`. Well, I could add two new properties to the Process class. (`stdout_mode`, `stderr_mode`)
> > Let me know which option you mean. (or Option 3: leave it as it is)
> 
> I think neither actually. Thinking from the caller point of view, you want the live data while the process is running and complete data after the process ended. Would that simplify the interface?
> 
>     1. Remove `output` and `error`, never return the handle.
> 
>     2. Read if process is running and return termination data otherwise.
> 
> 
> ```python
> def Process(self):
>     self.__stdout_data__ = null
> 
> @property
> def stdout_data(self):
>    return self.stdout.read() if self.stdout else self.__stdout_data__
> ```
> 
> WDYT?

Thanks for the detailed explanation. Now I get what you mean. 

Just one additional question, my unit test https://github.com/opensearch-project/opensearch-build/blob/main/tests/tests_system/test_process.py#L22 relies on the file handler to get its open mode. If we do not expose the handler, then I will need to add two more class properties, `stdout_mode` and `stderr_mode`

dblock:
> Just one additional question, my unit test https://github.com/opensearch-project/opensearch-build/blob/main/tests/tests_system/test_process.py#L22 relies on the file handler to get its open mode. If we do not expose the handler, then I will need to add two more class properties, `stdout_mode` and `stderr_mode`

Instead, write another test that mocks `NamedTemporaryFile` and expects a call to it with `mode="r+"`,

tianleh:
Encountered mocking issues for `stdout_data` since it is a property of class instead of a regular method call. Still researching how to properly mock a property.

tianleh:
> Encountered mocking issues for `stdout_data` since it is a property of class instead of a regular method call. Still researching how to properly mock a property.

Fixed with `PropertyMock`

dblock (approved):

PR #1078, opened 2021-11-18
Update the configuration and add unified agent config file for WhiteSource

zelinh:

### Description
change the `configMode` to local so that it will look for the new added `whitesource.config` unified agent file in our root directory. 
The `configMode` is updated according to this: https://whitesource.atlassian.net/wiki/spaces/WD/pages/697696422/WhiteSource+for+GitHub.com#Scan-Settings-(scanSettings)
WhiteSource configuration parameters are created based on this
https://whitesource.atlassian.net/wiki/spaces/WD/pages/1544880156
 
### Issues Resolved
N/A
 
### Check List


bbarani (approved):

gaiksaya:
[Triage] Please change the severity to MEDIUM here https://github.com/opensearch-project/opensearch-build/pull/1078/files#diff-437152f9907363ac2cb2060bfe19b2b2e87b1dd7b37b8944fdf644b716316757R13

gaiksaya (approved):

whitesource.config
```diff
@@ -0,0 +1,375 @@
...
+python.ignoreSourceFiles=true
+python.runPipenvPreStep=true
+python.pipenvDevDependencies=true
+python.requirementsFileIncludes=dev-requirements.txt
+python.installVirtualenv=true
+ruby.resolveDependencies=true
+ruby.ignoreSourceFiles=false
+ruby.installMissingGems=true
+ruby.runBundleInstall=true
+ruby.overwriteGemFile=true
```
gaiksaya, line 169: Can we move this section to the top? It is hard to read inbetween the other config. Also where is the `dev-requirements.txt` from line 163

zelinh: It's in other python repos like this one. https://github.com/opensearch-project/opensearch-py/blob/main/dev-requirements.txt

gaiksaya: But this configuration is specific to our repository right?

zelinh: Yeah right. I am still using the same unified agent configuration for all since we want to see the performance of local config for WhiteSource integration. We are now only using the default one.

abhinavGupta16: +1 @gaiksaya  It's hard to read the file to find the uncommented items. Apart from moving the uncommented portion to the top, another question I would ask is, is it necessary to keep the commented items? Can we get rid of it?

peternied:
@zelinh with https://github.com/opensearch-project/opensearch-build/pull/1268 merged is this pull request still needed?

zelinh:
> @zelinh with #1268 merged is this pull request still needed?

This PR is used for WhiteSource native integration app with Github.com. What we did in #1268 is automating our previous manually scans, which is different from using the native WhiteSource plugin. I would say this is still needed as part of the WhiteSource native integration on Github. I will update the configuration in this PR so that the scan would be customized to our build repo.

peternied:
> > @zelinh with #1268 merged is this pull request still needed?
> 
> This PR is used for WhiteSource native integration app with Github.com. What we did in #1268 is automating our previous manually scans, which is different from using the native WhiteSource plugin. I would say this is still needed as part of the WhiteSource native integration on Github. I will update the configuration in this PR so that the scan would be customized to our build repo.

I am having trouble finding the overall design issue, what is the consumption model for these configuration files when run with WhiteSource's native integration?

zelinh:
> > > @zelinh with #1268 merged is this pull request still needed?
> > 
> > 
> > This PR is used for WhiteSource native integration app with Github.com. What we did in #1268 is automating our previous manually scans, which is different from using the native WhiteSource plugin. I would say this is still needed as part of the WhiteSource native integration on Github. I will update the configuration in this PR so that the scan would be customized to our build repo.
> 
> I am having trouble finding the overall design issue could you make sure to link this item to it - what is the consumption model for these configuration files when run with WhiteSource's native integration?

I updated the link in this PR description. Thanks for pointing out!

whitesource.config
gaiksaya: Was wondering if we should separate the actual config with the commented one or just remove it. Reason being if we need to add something or debug we can look at the commented config and see if anything applied to our use case. @zelinh you can decide. I am okay with anything, Thanks!

gaiksaya (approved):

whitesource.config
zelinh: We will always have the general configuration with all commented part here to track. https://github.com/opensearch-project/opensearch-build/blob/main/tools/vulnerability-scan/wss-unified-agent.config. Therefore we could have this one clean and customized for our repo.

PR #1099, opened 2021-11-18
Move parameters out of a stage.

dblock:

### Description

I am not sure why parameters even work inside a stage, but reading https://www.jenkins.io/doc/book/pipeline/syntax/ parameters can be written in a simpler way at a `pipeline` level. 

```
parameters {
  param1
  param2
  ...
}
```

Looks the same from my test config.

Screen Shot 2021-11-18 at 2 22 03 PM

### Check List

### DIFF
--- a/jenkins/docker/Jenkinsfile
+++ b/jenkins/docker/Jenkinsfile
@@ -1,42 +1,34 @@
 pipeline {
     options {
-        timeout(time: 3, unit: 'HOURS') 
+        timeout(time: 3, unit: 'HOURS')
     }
     agent none
+    parameters {
+        string(
+            defaultValue: 'https://github.com/opensearch-project/opensearch-build',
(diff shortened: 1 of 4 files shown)


gaiksaya:
Will these parameters then be available for all stages just in case we want to reuse those and be careful with overwriting the param values?

dblock:
> Will these parameters then be available for all stages just in case we want to reuse those and be careful with overwriting the param values?

AFAIK yes and that was the original intent, but I could be wrong, @peterzhuamazon will know best?

peternied (approved):

gaiksaya (approved):
Hope this does not break anything, @peterzhuamazon if any objections let us know.
Thanks!

peterzhuamazon:
Golden.

PR #1104, opened 2021-11-18
Always cleanup, load a library with @date and fix dashboards.

dblock:

### Description

Another build failed because it was missing $TOKEN, which was changed previously in https://github.com/opensearch-project/opensearch-build/pull/1101. The logs also contain that commit. The only explanation is that the change was not picked up from `vars`, meaning the library was loaded and not replaced. 

Attempting to clean the workspace every time to fix that and to use a @date-based reference to force re-loading.

Forgot to use new build lib code in dashboards builds everywhere. 

### Check List


gaiksaya:
Even if we do that I see this log `[WS-CLEANUP] Deferred wipeout is disabled by the job configuration...` looks like we need to enable it from job config?

dblock:
> Deferred wipeout

Apparently.

jenkins/opensearch-dashboards/Jenkinsfile
```diff
@@ -1,4 +1,4 @@
-lib = library(identifier: "jenkins@current", retriever: legacySCM(scm))
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
```
gaiksaya, line 1: Where is this library retrieved from?

dblock: legacySCM means basically local checked out workspace, but it also caches things in-memory in the service instance and i don't know what the reloading rules are. I suspect it asks "has anything changed" to the "localSCM", which is git, or at least it should.

gaiksaya (approved):

owaiskazi19 (approved):

PR #1105, opened 2021-11-18
Refactored code and Added Notification after Integration Test Run

owaiskazi19:
### Description
- Refactored Test Jenkinsfile with global variables
- Integrated a webhook to notify through slack after running integration test.
- Added new global variable `uploadTestResults`.
- Added Groovy tests.

Screen Shot 2021-11-18 at 2
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/960
 
### Check List


jenkins/test/testsuite/Jenkinsfile
```diff
@@ -68,20 +70,46 @@ pipeline {
...
+                    publishNotification(":warning:", "Failed Build", "\n${stashed}")
+                }
+            }
+        }
+    }
 }
 
 /** Publishes test results to S3 **/
-void publish() {
+void publishResults() {
```
dblock, line 101: Take a look at https://github.com/opensearch-project/opensearch-build/pull/979.

We've introduced https://github.com/opensearch-project/opensearch-build/blob/main/vars/uploadArtifacts.groovy, which uploads `builds` and `dist` and is identical to this entire piece of code. Extract the list of folders to upload and pass in `test-results` here, and `["builds", "dist"]` in the other Jenkinsfile(s).

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -68,20 +70,46 @@ pipeline {
...
     def artifactPath = "bundle-build/${opensearch_version}/${build_id}/${platform}/${architecture}"
     withAWS(role: 'opensearch-test', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
         s3Upload(file: 'test-results', bucket: "${ARTIFACT_BUCKET_NAME}", path: "${artifactPath}/test-results")
     }
-    echo "Uploaded Test Results to ${artifactPath}/test-results"
+    echo "Uploaded Test Results to s3://${ARTIFACT_BUCKET_NAME}/${artifactPath}/test-results"
+    lib.jenkins.Messages.new(this).add("${STAGE_NAME}", "s3://${ARTIFACT_BUCKET_NAME}/${artifactPath}/test-results")
+}
+
+/** Publishes a notification to a slack instance*/
```
dblock, line 110: Same as above. We have https://github.com/opensearch-project/opensearch-build/blob/main/vars/publishNotification.groovy. Parameterize the `credentialsId` and pass in `INTEG_TEST_WEBHOOK` vs. `BUILD_NOTICE_WEBHOOK`.

dblock (changes requested):
You can make this a lot shorter by reusing the global library.

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -68,20 +70,46 @@ pipeline {
...
 
 /** Publishes test results to S3 **/
-void publish() {
+void publishResults() {
     def artifactPath = "bundle-build/${opensearch_version}/${build_id}/${platform}/${architecture}"
     withAWS(role: 'opensearch-test', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
         s3Upload(file: 'test-results', bucket: "${ARTIFACT_BUCKET_NAME}", path: "${artifactPath}/test-results")
     }
-    echo "Uploaded Test Results to ${artifactPath}/test-results"
+    echo "Uploaded Test Results to s3://${ARTIFACT_BUCKET_NAME}/${artifactPath}/test-results"
```
gaiksaya, line 106: Results wont be downloaded via S3. Its not a public bucket. We should add public url to download the results. 
Also can we instruct the component team how they can download the results?

For example (just an example please check actual paths and names):
```
Local cluster logs for your test run: $public_url/${artifactPath}/test-results/<component>/$config/$local-cluster-logs

Your integ-test run logs: $public_url/${artifactPath}/test-results/<component>/$config/$test-outcome

Your overall status: $public_url/${artifactPath}/test-results/<component>/$config/test-outcome.yml
```

No need to specify the component name but just the format so that component teams can download their own logs

owaiskazi19: `component` are not available in JenkinsFile. Is there any way you can think of to extract those?

gaiksaya: Do not specific the component. Just mention the format in the message so the respective compoenent owner will replace the `component` with say `sql` and can download the logs as they hit above url

gaiksaya: Wait I dont think even that is doable. We cannot list things in s3. So they can only download the overall status. For all others its a folder with many files. Not sure how to solve that part.  like local-test-cluster logs contains 5-6 files and not a tar.gz or zip that the component owner can download

dblock: All the buckets are exposed via ci.opensearch.org, so publish those links.

jenkins/test/testsuite/Jenkinsfile
owaiskazi19: Used the same functions for test JenkinsFile as well.

jenkins/opensearch-dashboards/Jenkinsfile
```diff
@@ -164,7 +174,9 @@ pipeline {
                     publishNotification(
                         icon: ':white_check_mark:',
                         message: 'Successful Build',
-                        extra: stashed
+                        extra: stashed,
+                        credentialsId: 'BUILD_NOTICE_WEBHOOK',
+                        variable: 'BUILD_NOTICE_WEBHOOK'
```
dblock, line 179: No need to expose `variable`, `withCredentials` takes a name of this variable to assign the value to, and then we use it inside its block. Rename it to something more common like `WEBHOOK_URL` inside `publishNotification`.

jenkins/opensearch-dashboards/Jenkinsfile
```diff
@@ -62,7 +62,12 @@ pipeline {
                                 architecture: "x64"
                             )
                             assembleManifest()
-                            uploadArtifacts(upload: true)
+                            uploadArtifacts(
+                                upload: true, 
+                                dirs: ['builds', 'dist'], 
+                                paths: ['builds/manifest.yml', 'dist/manifest.yml'],
```
dblock, line 68: Derive paths for the notification from `dirs`, just pass the name of the index file, e.g. `index_file: manifest.yml` or `index_file: index.html`.

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -14,6 +16,11 @@ pipeline {
                 script {
                     properties([
                             parameters([
```
dblock, line 18: Let's take the parameters block out of a stage just like in https://github.com/opensearch-project/opensearch-build/pull/1099

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -53,7 +60,7 @@ pipeline {
                 }
             }
         }
-        stage('Run Integration Tests') {
+        stage('Integ-Test') {
```
dblock, line 63: lowercase to be consistent with other Jenkinsfile

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -68,20 +75,60 @@ pipeline {
             post() {
                 always {
                     script {
-                        publish()
+                        uploadArtifacts(
+                            upload: true,
+                            build_number: "${build_id}",
```
dblock, line 80: Where's build_id coming from? Is it used? Should it be `$BUILD_ID`. 

Also `"${x}"` is the same as `$x`.

manifests/1.2.0/opensearch-1.2.0.yml
```diff
@@ -7,6 +7,8 @@ ci:
 build:
   name: OpenSearch
   version: 1.2.0
+  platform: linux
+  architecture: x64
```
dblock, line 11: The input manifest should not be locking the platform and architecture, remove.

vars/publishNotification.groovy
```diff
@@ -6,13 +6,13 @@ void call(Map args = [:]) {
         args.extra
     ] - null).join("\n")
 
-    withCredentials([string(credentialsId: 'BUILD_NOTICE_WEBHOOK', variable: 'BUILD_NOTICE_WEBHOOK')]) {
+    withCredentials([string(credentialsId: args.credentialsId, variable: args.variable)]) {
```
dblock, line 9: args.variable -> 'WEBHOOK_URL` and reuse below, remove from params

vars/uploadArtifacts.groovy
```diff
@@ -4,23 +4,29 @@ void call(Map args = [:]) {
     def manifestFilename = args.manifest ?: 'builds/manifest.yml'
     def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: manifestFilename))
 
-    def artifactPath = buildManifest.getArtifactRoot("${JOB_NAME}", "${BUILD_NUMBER}")
+    def buildNumber = args.build_number ?: "${BUILD_NUMBER}"
```
dblock, line 7: Aren't those always present in all jobs? If they are not, we should either add them everywhere or remove them from everywhere. I would make sure the test workflow has them and restore the old code here that assumes they exist.

vars/uploadArtifacts.groovy
```diff
@@ -4,23 +4,29 @@ void call(Map args = [:]) {
...
 
-    def baseUrl = buildManifest.getArtifactRootUrl("${PUBLIC_ARTIFACT_URL}", "${JOB_NAME}", "${BUILD_NUMBER}")
-    lib.jenkins.Messages.new(this).add("${STAGE_NAME}", [
-            "${baseUrl}/builds/manifest.yml",
-            "${baseUrl}/dist/manifest.yml"
-        ].join("\n")
-    )
+    def baseUrl = buildManifest.getArtifactRootUrl("${PUBLIC_ARTIFACT_URL}", jobName, buildNumber)
+
+    def paths = []
```
dblock, line 26: Derive from dirs and a new parameter, index_name.

dblock (changes requested):
This is close!

jenkins/opensearch-dashboards/Jenkinsfile
owaiskazi19: We have different webhooks for build as `BUILD_NOTICE_WEBHOOK` and for integ test as `INTEG_TEST_WEBHOOK`. The reason being having 2 different channel to send notification. Is there anything I'm missing here?

jenkins/opensearch-dashboards/Jenkinsfile
owaiskazi19: For build we have the path to the manifest file but for tests we have path till `test-results` directory.

jenkins/test/testsuite/Jenkinsfile
owaiskazi19: This is the build_id on which we want to run tests on.

vars/uploadArtifacts.groovy
owaiskazi19: build_number for test is the build on which we want to run tests on. That's why took it as a parameter to the job.

manifests/1.2.0/opensearch-1.2.0.yml
owaiskazi19: Is there any other way of doing this because platform and architecture are taken from builds/manifest which is not present in the input manifest https://github.com/opensearch-project/opensearch-build/blob/main/src/jenkins/BuildManifest.groovy#L21

jenkins/opensearch-dashboards/Jenkinsfile
dblock: Yes, you're missing the fact that the variable used is purely internal to inside of the `withCredential` block. The following produces the same output

```
withCredentials([string(credentialsId: 'THIS_IS_THE_ID_OF_CREDENTIALS', variable: 'FOO')]) {
  echo $FOO
}

withCredentials([string(credentialsId: 'THIS_IS_THE_ID_OF_CREDENTIALS', variable: 'BAR')]) {
  echo $BAR
}

```

owaiskazi19: Also, I'm not sure but should we create separate global files for test instead of passing so many parameters because of different use cases? Let me know WDYT

dblock: I could be wrong, but AFAIK the ID is different, but the name of the variable that receives the value can be anything.

Screen Shot 2021-11-19 at 1 06 15 PM

So 

```
withCredentials([string(credentialsId: 'ID', variable: 'FOO')]) {
   echo $FOO
}
```

produces the same output as 

```
withCredentials([string(credentialsId: 'ID', variable: 'BAR')]) {
   echo $BAR
}
```

jenkins/opensearch-dashboards/Jenkinsfile
dblock: I think the two arrays is throwing me off. Then maybe something like this?

```
uploadArtifacts(
   upload: true
   artifacts: [
        {
           dir: 'builds',
           path: 'builds/manifest.yml',
        },
        {
           dir: 'builds',
           path: 'dist/manifest.yml',
        },
   ]
```

jenkins/test/testsuite/Jenkinsfile
dblock: Shouldn't it be `BUILD_ID` or `buildId`?

manifests/1.2.0/opensearch-1.2.0.yml
dblock: Isn't the input to the job a build ID that gives me a URL to a build YAML?

jenkins/test/testsuite/Jenkinsfile
gaiksaya: If you say BUILD_ID thats jenkins build_id (run number) I think we are looking for opensearch/dashboards build-id (basically jenkins run number of bundle-build) so that we grab right artifacts to run test aganist

vars/uploadArtifacts.groovy
dblock: You're testing a build, so you should be taking a build manifest as input, not an input manifest.

jenkins/test/testsuite/Jenkinsfile
peternied: Note; Just because s3 listing isn't allow doesn't mean you can't provide a pretty index.html or manifest.yml with urls to all the good stuff.

vars/uploadArtifacts.groovy
dblock: ... or since you're taking a build ID, architecture and platform, construct the URL to the build manifest and fetch it from there

jenkins/test/testsuite/Jenkinsfile
dblock: I see. We should rename that. It's too confusing to have `BUILD_ID` and `build_id`, but I understand what's going on now.

Currently `uploadArtifacts` takes a manifest and uploads `builds` and `dist`.

I think it would be totally fine to have `uploadTestResults` if these two re sufficiently different (looks like we don't need to read a manifest for test results because we pass in architecture and platform into the actual test job?). Then we don't need to make it parameterized that much. You could maybe extract the part in `withAWS` into `uploadDirs` still.

owaiskazi19: Yes, that's the case. We can name it as `build-id`.

jenkins/opensearch-dashboards/Jenkinsfile
owaiskazi19: This worked for test notification. Trying it for build notification to see if the notification is triggered. Thanks for the input.

owaiskazi19: Update: Worked for build as well.

vars/publishNotification.groovy
```diff
@@ -1,18 +1,19 @@
...
-    withCredentials([string(credentialsId: 'BUILD_NOTICE_WEBHOOK', variable: 'BUILD_NOTICE_WEBHOOK')]) {
+    withCredentials([string(credentialsId: args.credentialsId, variable: 'WEBHOOK_URL')]) {
         sh ([
             args.script ?: 'curl',
             '-XPOST',
             '--header "Content-Type: application/json"',
             "--data '{\"result_text\":\"${text}\"}'",
-            "\"${BUILD_NOTICE_WEBHOOK}\""
+            "\$" + 'WEBHOOK_URL'
         ].join(' '))
```
dblock, line 17: Does this actually resolve $WEBHOOK_URL? Doesn’t look like it according to the test. Should just be $WEBHOOK_URL

dblock (changes requested):
I think the WEBHOOK url doesn’t work, but I could be wrong.

vars/publishNotification.groovy
owaiskazi19: Yes. Tested the notifications

dblock: The code as is is still not readable and the + unnecessary. You want the value of $WEBHOOK_URL so just write that?

owaiskazi19: Made the change.

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -62,26 +59,60 @@ pipeline {
             steps {
                 script {
                     basePath = "https://ci.opensearch.org/ci/dbc/bundle-build/${opensearch_version}/${build_id}/${platform}/${architecture}"
+                    sh "wget ${basePath}/builds/manifest.yml"
```
dblock, line 62: Is this for debugging?

dblock (approved):

dblock (approved):

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -62,26 +59,60 @@ pipeline {
...
     }
-}
+    post() {
+        success {
+            node('Jenkins-Agent-al2-x64-m5xlarge') {
+                script {
+                    def stashed = lib.jenkins.Messages.new(this).get(['integ-test'])
+                    publishNotification(
+                        icon: ':white_check_mark:',
+                        message: 'Successful Build',
```
gaiksaya, line 87: Should this be `Integration tests successful`? Also below `Failed integration tests`?

jenkins/test/testsuite/Jenkinsfile
owaiskazi19: It's for fetching the build manifest.

dblock: I mean why is it needed?

```
    test_manifest = TestManifest.from_path(test_manifest_path)
    bundle_manifest = BundleManifest.from_urlpath("/".join([args.path.rstrip("/"), "dist/manifest.yml"]))
    build_manifest = BuildManifest.from_urlpath("/".join([args.path.rstrip("/"), "builds/manifest.yml"]))
```

This fetches everything from the `basePath` URL.

dblock: Answered my own question, we have code in publishTestResults that loads this manifest. Similarly, it should be loading it from the remote URL. Given that `readYaml` doesn't support it I think the "right thing to do" (TM) would be that `./test.sh ...` produces a `results.yml` that gets loaded at the end of the tests and then the publication location is derived from those test results, not from the build YAML. Please open an issue for this.

owaiskazi19: Created https://github.com/opensearch-project/opensearch-build/issues/1113

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -62,26 +59,60 @@ pipeline {
...
+        }
+        failure {
+            node('Jenkins-Agent-al2-x64-m5xlarge') {
+                script  {
+                    def stashed = lib.jenkins.Messages.new(this).get(['integ-test'])
+                    publishNotification(
+                        icon: ':warning:',
+                        message: 'Failed Integration Tests',
+                        extra: stashed,
+                        credentialsId: 'INTEG_TEST_WEBHOOK',
```
gaiksaya, line 107: Should this be `BUILD_NOTICE_WEBHOOK` as stated [here](https://github.com/opensearch-project/opensearch-build/pull/1105/files#diff-15de42ee50e7efc22a48cb1133f6a05fd4472e5a18fb004259b3cb3b5d11264fR160)

owaiskazi19: No. We have created a separate channel for the tests and it has a different webhook. Check the discussion here https://github.com/opensearch-project/opensearch-build/pull/972#issuecomment-971814828.

gaiksaya: My bad got confused with `tests` and `test` folder. Cool

gaiksaya (approved):

tests/jenkins/jobs/Build_Jenkinsfile
```diff
@@ -112,6 +118,12 @@ pipeline {
                                 upload: false,
                                 manifest: 'tests/data/opensearch-build-1.1.0.yml'
                             )
+                            uploadTestResults(
+                                upload: false,
+                                manifest: 'tests/data/opensearch-build-1.1.0.yml',
+                                jobName: "bundle_build",
```
dblock, line 124: Should this have been `bundle-build`?

owaiskazi19: This is for S3. We have it there as `bundle_build`.

gaiksaya: Nope its `bundle-build` even in S3. There is also `bundle_build` which is incorrect and contains just 1 build

dblock: I fixed it in https://github.com/opensearch-project/opensearch-build/pull/1111 if you care to CR.

owaiskazi19: Missed that we have 2 buckets. Thanks @dblock for the fix.

dblock: @gaiksaya I think it's safe to delete `bundle_build` in S3 to avoid confusion - want to do it?

gaiksaya: Yes and few others which were created based on job-names :) Will delete the unnecessary ones

PR #1130, opened 2021-11-23
Fix issue with docker volume-mounted config file

robinboening:
### Description
Using `sed -i` was causing an issue when a custom opensearch.yml file was mounted as a volume.

```
sed: cannot rename /usr/share/opensearch/config/sedqdMb0d: Device or resource busy
```

The reason for the issue was found by @unhipzippo https://github.com/opensearch-project/OpenSearch/issues/768#issuecomment-968140380 :heart:

> The "sed -i" is an attempt to modify the opensearch.yml file "in place" -- But according to the GNU sed documentation (https://www.gnu.org/software/sed/manual/sed.html#Command_002dLine-Options), "in-place" actually "does this by creating a temporary file and sending output to this file rather than to the standard output. ... the temporary file is renamed to the output file’s original name".
>
> I believe this rename would require changing the inode of the original file -- something that Docker volume mounts don't permit.
 
### Issues Resolved
Potentially https://github.com/opensearch-project/OpenSearch/issues/768 and https://github.com/opensearch-project/OpenSearch/issues/1579
 
### Check List

### DIFF
--- a/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
+++ b/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
@@ -177,8 +177,8 @@ if [ -d "$OPENSEARCH_DASHBOARDS_HOME/plugins/$SECURITY_DASHBOARDS_PLUGIN" ]; the
     if [ "$DISABLE_SECURITY_DASHBOARDS_PLUGIN" = "true" ]; then
         echo "Disabling OpenSearch Security Dashboards Plugin"
         ./bin/opensearch-dashboards-plugin remove securityDashboards
-        sed -i /^opensearch_security/d $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
-        sed -i 's/https/http/' $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
+        sed "/^opensearch_security/d" $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
+        sed "s/https/http/" $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
     fi
 fi
 
--- a/docker/release/config/opensearch/opensearch-docker-entrypoint.sh
+++ b/docker/release/config/opensearch/opensearch-docker-entrypoint.sh
@@ -66,11 +66,10 @@ if [ -d "$OPENSEARCH_HOME/plugins/$SECURITY_PLUGIN" ]; then
 
     if [ "$DISABLE_SECURITY_PLUGIN" = "true" ]; then
         echo "Disabling OpenSearch Security Plugin"
-        sed -i '/plugins.security.disabled/d' $OPENSEARCH_HOME/config/opensearch.yml
-        echo "plugins.security.disabled: true" >> $OPENSEARCH_HOME/config/opensearch.yml
+        sed "s/plugins.security.disabled.*$/plugins.security.disabled: true" $OPENSEARCH_HOME/config/opensearch.yml | tee $OPENSEARCH_HOME/config/opensearch.yml
     else
         echo "Enabling OpenSearch Security Plugin"
-        sed -i '/plugins.security.disabled/d' $OPENSEARCH_HOME/config/opensearch.yml
+        sed "/plugins.security.disabled/d" $OPENSEARCH_HOME/config/opensearch.yml | tee $OPENSEARCH_HOME/config/opensearch.yml
     fi
 fi
 


dblock (approved):

scratchings:
Whilst this fixes the initial issue with regard to the behavior of sed -i, it still prevents Opensearch from starting if Docker Compose 'config's are used to bind mount the opensearch.yml file (the most sensible way of distributing in a swarm mode cluster) as these are read-only.

dblock:
> Whilst this fixes the initial issue with regard to the behavior of sed -i, it still prevents Opensearch from starting if Docker Compose 'config's are used to bind mount the opensearch.yml file (the most sensible way of distributing in a swarm mode cluster) as these are read-only.

So what's the suggested fix?

@robinboening you need to `--amend -s` your commits for DCO, please.

scratchings:
Wrapping it with an test -w on the opensearch.yml file would be a simple work-around - if the file is being bind mounted, then it seems likely to me that the end user does not expect their config file to be modified on startup.

```
   if [ -w "$OPENSEARCH_HOME/config/opensearch.yml" ]; then
        if [ "$DISABLE_SECURITY_PLUGIN" = "true" ]; then
            echo "Disabling OpenSearch Security Plugin"
            sed 's/plugins.security.disabled.*$/plugins.security.disabled: true' $OPENSEARCH_HOME/config/opensearch.yml | tee $OPENSEARCH_HOME/config/opensearch.yml
        else
           echo "Enabling OpenSearch Security Plugin"
           sed '/plugins.security.disabled/d' $OPENSEARCH_HOME/config/opensearch.yml | tee $OPENSEARCH_HOME/config/opensearch.yml 
            echo "plugins.security.disabled: true" >> $OPENSEARCH_HOME/config/opensearch.yml
        fi
    fi
```

peternied:
@peterzhuamazon Could you take a look?

peterzhuamazon:
Hi @robinboening can you check @scratchings comment https://github.com/opensearch-project/opensearch-build/pull/1130#issuecomment-976972694 and see if that is useful and update the PR?

Or I can approve the PR at the current state.

Thanks.

robinboening:
@peterzhuamazon I simply don't have enough knowledge about docker/swarm to judge if the proposed change in https://github.com/opensearch-project/opensearch-build/pull/1130#issuecomment-976972694 is good or not.

@scratchings I thought docker creates writeable bind mounts by default. Referring to the docs https://docs.docker.com/storage/bind-mounts/#use-a-read-only-bind-mount only if specifically declared as readonly they're treated as such. I have never used swarm, so maybe its different in that mode... 🤷 it seems you know more on this topic, would you mind to enlighten me? :)

peterzhuamazon:
@scratchings comment seems simply checking whether the file is writable before the changes are made.
Which seems simple enough for me.
But I also want his comment on @robinboening replies.

Thanks.

scratchings:
> @scratchings I thought docker creates writeable bind mounts by default. Referring to the docs https://docs.docker.com/storage/bind-mounts/#use-a-read-only-bind-mount only if specifically declared as readonly they're treated as such. I have never used swarm, so maybe its different in that mode... 🤷 it seems you know more on this topic, would you mind to enlighten me? :)

Docker swarm mode has 'configs' <https://docs.docker.com/engine/swarm/configs/> and 'secrets' <https://docs.docker.com/engine/swarm/secrets> objects (docker compose can implement these for non-swarm systems to a limited extent). Both of these objects types are stored in the Swarm database that is distributed amongst the nodes so that your container can be started anywhere and have access to these items. Secrets are mounted from a RAM file system into /run/secrets/<name of secret> and configs into a location of your choice. In a swarm you have full control over ownership and access permissions (in plain compose you don't as they are 'faked' with bind mounts) they are ALWAYS read-only. If you've gone to the effort of importing the opensearch.yml file into a swarm configs item you are unlikely to want automatic modification of the content, the block here seems to be largely to support 'demo' deployments.

peterzhuamazon:
> > @scratchings I thought docker creates writeable bind mounts by default. Referring to the docs https://docs.docker.com/storage/bind-mounts/#use-a-read-only-bind-mount only if specifically declared as readonly they're treated as such. I have never used swarm, so maybe its different in that mode... 🤷 it seems you know more on this topic, would you mind to enlighten me? :)
> 
> Docker swarm mode has 'configs' https://docs.docker.com/engine/swarm/configs/ and 'secrets' https://docs.docker.com/engine/swarm/secrets objects (docker compose can implement these for non-swarm systems to a limited extent). Both of these objects types are stored in the Swarm database that is distributed amongst the nodes so that your container can be started anywhere and have access to these items. Secrets are mounted from a RAM file system into /run/secrets/ and configs into a location of your choice. In a swarm you have full control over ownership and access permissions (in plain compose you don't as they are 'faked' with bind mounts) they are ALWAYS read-only. If you've gone to the effort of importing the opensearch.yml file into a swarm configs item you are unlikely to want automatic modification of the content, the block here seems to be largely to support 'demo' deployments.

Thanks for the detailed explanation here @scratchings.
On that note, I do agree that a simple `-w` fix would be enough for us here.
@robinboening could update the PR accordingly.

Thanks.

peterzhuamazon (changes requested):
Block this change as we will not introduce this to 1.1.1 patch release. We will unblock this for the next major/minor release.

Thanks.

peterzhuamazon:
Since 1.2.1 OpenSearch is out we can resume this, @robinboening any updates to my previous comment?
Thanks.

gijs007:
This issue was reported in may, over half a year ago: https://github.com/opensearch-project/OpenSearch/issues/768
I see a fix is ready, but not merged into the main build. When can we expect this?

Note: At this point we still cannot rollout OpenSearch for production, since this issue prevents us from using custom configs with the recommended docker installation...

dblock:
I think I'd prefer the fix as written in this PR. Using `-w` avoids touching the configuration, but it also makes the assumption that the configuration doesn't need to change, i.e. it would silently swallow an error.

@peterzhuamazon Can we merge as is?

peterzhuamazon:
> I think I'd prefer the fix as written in this PR. Using `-w` avoids touching the configuration, but it also makes the assumption that the configuration doesn't need to change, i.e. it would silently swallow an error.
> 
> @peterzhuamazon Can we merge as is?

I am ok to let the error shows without outputing no errors at all.

However, I have since identified some missed changes for dashboards.

@robinboening would you mind adding similar changes for dashboards entrypoint.sh as well?
https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh#L180-L181

```
        sed -i /^opensearch_security/d $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
        sed -i 's/https/http/' $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
```

Thanks.

dblock:
> However, I have since identified some missed changes for dashboards.

Good catch.

@gijs007 @scratchings feel free to PR the complete fix on top of this, will merge, apologies for missing this earlier

robinboening:
> > I think I'd prefer the fix as written in this PR. Using `-w` avoids touching the configuration, but it also makes the assumption that the configuration doesn't need to change, i.e. it would silently swallow an error.
> > @peterzhuamazon Can we merge as is?
> 
> I agree with DB on this one.
> 
> However, I have since identified some missed changes for dashboards.
> 
> @robinboening would you mind adding similar changes for dashboards entrypoint.sh as well? https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh#L180-L181
> 
> ```
>         sed -i /^opensearch_security/d $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
>         sed -i 's/https/http/' $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
> ```
> 
> Thanks.

I can look into this as well. Hopefully today / tomorrow.

robinboening:
> However, I have since identified some missed changes for dashboards.

@dblock I've covered the other part and amended the changes to the commit.

> I think I'd prefer the fix as written in this PR. Using -w avoids touching the configuration, but it also makes the assumption that the configuration doesn't need to change, i.e. it would silently swallow an error.

I understand your arguments on this and I don't want to make this PR bigger than it is right now, but I'd like to kick off a separate discussion around this topic (maybe to be moved in to a new thread):

I am questioning if the current approach of *silently* altering a custom config file is any better than swallowing a potential error. My argument is that the user deploying OpenSearch is making a conscious decision on how the config file is customised and how it finally looks. OpenSearch then silently does its own changes on top of that, possibly without being noticed.

Shouldn't OpenSearch detect the error at runtime and throw a meaningful error instead of silently altering the config to make it right?

dblock (approved):

dblock:
> > However, I have since identified some missed changes for dashboards.
> 
> @dblock I've covered the other part and amended the changes to the commit.

Looks good, thanks.

> I am questioning if the current approach of _silently_ altering a custom config file is any better than swallowing a potential error. My argument is that the user deploying OpenSearch is making a conscious decision on how the config file is customised and how it finally looks. OpenSearch then silently does its own changes on top of that, possibly without being noticed.
> 
> Shouldn't OpenSearch detect the error at runtime and throw a meaningful error instead of silently altering the config to make it right?

I agree. Feel free to open a new issue for this in OpenSearch proper. I'd want to be able to set DISABLE_SECURITY as an ENV variable and see it override the setting without having to alter a .yml config file.

gijs007:
Glad to see this issue has been fixed. Would it be possible to get the docker images updated directly, since this is a blocking issue for production deployments?

peterzhuamazon (approved):
Thanks.

kinseii:
Excuse me, can you please tell me when a new release with this fix will be available?

PR #1139, opened 2021-11-23
Update release_template with step to update next release's manifest.

### LINKED ISSUE #1140: Nightly bundle builds do not include full bundle until T-14 days to release.
This is a suggestion to improve the release template and release cycle.

While performance testing for the 1.2 release we found a [performance degradation](https://github.com/opensearch-project/OpenSearch/issues/1560) in the bundle.  To try and diagnose the issue we pulled older nightly builds at various times and re-ran perf tests to identify when a degradation may have been introduced.  Ideally we would have these performance tests running nightly.  However, even when that is completed we will unfortunately be running against different configurations of the bundle because components are added to the manifest only [T-14 days](https://github.com/opensearch-project/opensearch-build/blob/main/.github/ISSUE_TEMPLATE/release_template.md#cicd---ends-release-minus-14-days) before release.

For example - right now for [2.0](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.0.0/opensearch-2.0.0.yml) and [1.3](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/1.3.0/opensearch-1.3.0.yml) releases, we have no components being built nightly other than OpenSearch. 

After every release we would have a functioning manifest for the next major and minor release that includes the known components at that time.  This will help us identify when an issue was introduced and will further aid us when we are running nightly benchmarks against the full bundle.

I have opened #1139 as a suggestion. An alternative here is we cut the release issue for the next release immediately on release day and ensure the manifest is updated right away.

mch2:

### Description
Updates the release template with a step to update the next release's manifest.
 
### Issues Resolved
related - https://github.com/opensearch-project/OpenSearch/issues/1317
closes #1140 
 
### Check List

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_template.md
+++ b/.github/ISSUE_TEMPLATE/release_template.md
@@ -67,6 +67,7 @@ __Replace with OpenSearch wide initiatives to improve quality and consistency.__
 
 - [ ] Create [release tags](https://github.com/opensearch-project/.github/blob/main/RELEASING.md#tagging) for each component.
 - [ ] Replace refs in [manifests/{{ env.VERSION }}](/opensearch-project/opensearch-build/tree/main/manifests/{{ env.VERSION }}) with tags.
+- [ ] Update the next release's manifest with a ref to build for your component to ensure it is included in nightly builds. For example, if 1.2.0 is being released, update the [manifests/1.3.0](/opensearch-project/opensearch-build/tree/main/manifests/1.3.0).
 - [ ] Update [this template](./release_template.md) with any new or missed steps.
 - [ ] Conduct a postmortem, and publish its results.
 


gaiksaya:
Should the nightly's be pointing to main of component's ref when a new manifest is created? And then change to 1.3 branch at T-14 days?

mch2:
> Should the nightly's be pointing to main of component's ref when a new manifest is created? And then change to 1.3 branch at T-14 days?

That makes sense for major releases but not for minor.  If the plugin repo is following the correct branching strategy we could do this for minors as well and set them to (version.x) branch.
@gaiksaya I should have cut this initially as an issue - lets chat on #1140.

dblock (changes requested):
I don't think we need this. It's a step in the next release's checklist.

Issue #1134, opened 2021-11-23
Distinguish OpenSearch vs OpenSearchDashboards manifest file during build and assemble workflow


tianleh:
### Is your feature request related to a problem? Please describe

Currently both manifest files share the same name inside `builds` and `dist` folders. Thus they cannot co-exist. We want to have both together. This is motivated by #604

### Describe the solution you'd like

OpenSearch and OpenSearchDashboards manifest files could be distinguished and thus co-exist.

dblock:
I'd rather not mix outputs, so either we add `./build.sh --output-dir <path>` or default to `builds/opensearch` vs. `builds/opensearch-dashboards`?

tianleh:
I will follow "default to builds/opensearch vs. builds/opensearch-dashboards"

PR #1145, opened 2021-11-23
Distinguish OpenSearch vs OpenSearchDashboards manifest file during b…

tianleh:
…uild and assemble workflow

### Description
To support running both OpenSearch and OpenSearchDashboards, we need to have access to both manifest files.
 
### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-build/issues/1134

### Test

Verify that local test cluster can run.

```
2021-11-23 21:48:06 INFO     Pinging service attempt 2
2021-11-23 21:48:06 INFO     Pinging https://localhost:9200/_cluster/health
/usr/share/opensearch/.local/share/virtualenvs/opensearch-build-B4-P-5sk/lib/python3.7/site-packages/urllib3/connectionpool.py:1020: InsecureRequestWarning: Unverified HTTPS request is being made to host 'localhost'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  InsecureRequestWarning,
2021-11-23 21:48:07 INFO     200: {"cluster_name":"opensearch","status":"green","timed_out":false,"number_of_nodes":1,"number_of_data_nodes":1,"discovered_master":true,"active_primary_shards":1,"active_shards":1,"relocating_shards":0,"initializing_shards":0,"unassigned_shards":0,"delayed_unassigned_shards":0,"number_of_pending_tasks":0,"number_of_in_flight_fetch":0,"task_max_waiting_in_queue_millis":0,"active_shards_percent_as_number":100.0}
2021-11-23 21:48:07 INFO     Service is available
```

### Check List


src/run_assemble.py
```diff
@@ -25,7 +25,11 @@ def main():
     build_manifest = BuildManifest.from_file(args.manifest)
     build = build_manifest.build
     artifacts_dir = os.path.dirname(os.path.realpath(args.manifest.name))
-    output_dir = os.path.join(os.getcwd(), "dist")
+    output_dir = os.path.join(
+        os.getcwd(),
+        "dist",
+        "opensearch" if build.name == "OpenSearch" else "opensearch-dashboards"
```
dblock, line 31: Derive this from the manifest `build.name`,  add a method that does `name.lower().replace(" ", "-")` and use that in other places we do the same thing.

src/run_build.py
```diff
@@ -43,6 +42,12 @@ def main():
             manifest.to_file(args.ref_manifest)
         exit(0)
 
+    output_dir = os.path.join(
```
dblock, line 45: Consolidate into `args.output_dir` or another class so we don't duplicate this code.

dblock (changes requested):
- Several Jenkinsfile(s) assume the directory to be `builds/` and `dist/`.
- Documentation describes `builds/` and `dist/`.

tianleh:
Extracted the common logic to a separate class for reuse between build and assemble.

Still working in progress for Jenkinsfile and Documentation

tianleh:
The commit history is messed up after rebasing. Checking how to bring it back to clean history.

tianleh:
Ready for review.

tianleh:
> 

Added

seraphjiang (approved):
Ship it

README.md
```diff
@@ -241,7 +241,7 @@ You can perform cross-platform builds. For example, build and assemble a Windows
 ```bash
 export JAVA_HOME=$(/usr/libexec/java_home) # required by OpenSearch install-plugin during assemble
 ./build.sh manifests/1.1.0/opensearch-1.1.0.yml --snapshot --platform windows
```
seraphjiang, line 243: in this doc, some version path use 1.1.0, some use 1.2.0, could we keep consistency

vars/uploadArtifacts.groovy
```diff
@@ -1,7 +1,7 @@
 void call(Map args = [:]) {
     def lib = library(identifier: "jenkins@20211123", retriever: legacySCM(scm))
 
-    def manifestFilename = args.dryRun ? 'tests/data/opensearch-build-1.1.0.yml' : 'builds/manifest.yml'
+    def manifestFilename = args.dryRun ? 'tests/data/opensearch-build-1.1.0.yml' : 'builds/opensearch/manifest.yml'
```
seraphjiang, line 4: for my curiosity, when will version bump up to 1.2.0

dblock (changes requested):
This works, however it's not object-oriented.

Having a `utils` is a common anti-pattern - it quickly becomes a garbage can of unrelated functions. Seeing `get_folder_name_from_build_name` with its long name told me that it struggled to find a home, it's neither a property of the build, nor a property of a manifest, so it became a global. Here's how I'd clean this up.

Location of output can become a first class concern. I would add `BuildOutputDir` and `AssembleOutputDir` classes that both inherit from `OutputDir`. That would have a `dir` property that is currently `output_dir`. I would put that class into `src/paths` for simpllicity.

```
class OutputDir(ABS):
   def __init__(cls, parent_dir, name, cwd=None, makedirs=true):
       self.dir = os.path.join(
           cwd or os.getcwd(),
           parent_dir,
           name.lower().replace(' ', '-')
       )
       
       if makedirs:
           os.makedirs(...)           

class BuildOutputDir(Output):
    def __init__(cls, name,  cwd=None, makedirs=True):
        super().__init__("builds", name, makedirs)
```

Now as a developer I can write:

```
output = BuildOutput(build.name, makedirs=true/false)

output.dir # what we want
```

There are no lingering globals, and in the future we can keep adding to `OutputDir` features such as "fail if the output directory already exists" or "tell me what's already in output dir so I can do an incremental build" or "tell me how big this build is so I can track the size of my artifacts".

kavilla:
I think this might impact:
https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/Jenkinsfile

Where it bundles up the build folder. Will verify.

dblock:
> I think this might impact: https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/Jenkinsfile
> 
> Where it bundles up the build folder. Will verify.

Those are shared functions now, so I think it's taken care of. https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/Jenkinsfile#L80 is `builds/*` so that should work if I am not mistaken.

tianleh:
> This works, however it's not object-oriented.
> 
> Having a `utils` is a common anti-pattern - it quickly becomes a garbage can of unrelated functions. Seeing `get_folder_name_from_build_name` with its long name told me that it struggled to find a home, it's neither a property of the build, nor a property of a manifest, so it became a global. Here's how I'd clean this up.
> 
> Location of output can become a first class concern. I would add `BuildOutputDir` and `AssembleOutputDir` classes that both inherit from `OutputDir`. That would have a `dir` property that is currently `output_dir`. I would put that class into `src/paths` for simpllicity.
> 
> ```
> class OutputDir(ABS):
>    def __init__(cls, parent_dir, name, cwd=None, makedirs=true):
>        self.dir = os.path.join(
>            cwd or os.getcwd(),
>            parent_dir,
>            name.lower().replace(' ', '-')
>        )
>        
>        if makedirs:
>            os.makedirs(...)           
> 
> class BuildOutputDir(Output):
>     def __init__(cls, name,  cwd=None, makedirs=True):
>         super().__init__("builds", name, cwd, makedirs)
> ```
> 
> Now as a developer I can write:
> 
> ```
> output = BuildOutput(build.name, makedirs=true/false)
> 
> output.dir # what we want
> ```
> 
> There are no lingering globals, and in the future we can keep adding to `OutputDir` features such as "fail if the output directory already exists" or "tell me what's already in output dir so I can do an incremental build" or "tell me how big this build is so I can track the size of my artifacts".

Thanks for the detailed instruction

tianleh:
Run build and assemble for both OpenSearch and OpenSearch Dashboards successfully. Below are the output snippets.

```

--

021-11-25 00:33:17 INFO     Recording plugins artifact for opensearch-observability: plugins/opensearch-observability-1.2.0.0.zip (from /tmp/tmpuhd5kr96/opensearch-observability/opensearch-observability/builds/plugins/opensearch-observability-1.2.0.0.zip)
2021-11-25 00:33:17 INFO     Checked /tmp/tmpuhd5kr96/opensearch-observability/opensearch-observability/builds/plugins/opensearch-observability-1.2.0.0.zip (1.2.0.0)
2021-11-25 00:33:17 INFO     Created build manifest /usr/share/opensearch/workspace/opensearch-build/builds/opensearch/manifest.yml
2021-11-25 00:33:19 INFO     Done.

--

2021-11-25 00:35:18 INFO     Published /usr/share/opensearch/workspace/opensearch-build/dist/opensearch/opensearch-1.2.0-linux-x64.tar.gz.
2021-11-25 00:35:19 INFO     Done.


---

2021-11-25 00:48:57 INFO     Checked /tmp/tmpz037gs2b/anomalyDetectionDashboards/builds/plugins/anomalyDetectionDashboards-1.2.0.zip (1.2.0.0)
2021-11-25 00:48:57 INFO     Created build manifest /usr/share/opensearch/workspace/opensearch-build/builds/opensearch-dashboards/manifest.yml
2021-11-25 00:49:05 INFO     Done.

--
2021-11-25 00:52:16 INFO     Installed plugins: ['alertingDashboards', 'indexManagementDashboards', 'anomalyDetectionDashboards', 'reportsDashboards', 'observabilityDashboards', 'queryWorkbenchDashboards', 'ganttChartDashboards', 'securityDashboards']
2021-11-25 00:53:26 INFO     Published /usr/share/opensearch/workspace/opensearch-build/dist/opensearch-dashboards/opensearch-dashboards-1.2.0-linux-x64.tar.gz.
2021-11-25 00:53:27 INFO     Done.
```

seraphjiang (approved):

dblock (approved):
Good.

I'd like someone (you?) to fix https://github.com/opensearch-project/opensearch-build/issues/1154 before merging this to not introduce more variables into an already broken build.

tianleh:
> 

Sure. PR sent to fix the issue https://github.com/opensearch-project/opensearch-build/pull/1162

PR #1147, opened 2021-11-23
Jenkins incremental build

dblock:

### Description

Skips builds for OpenSearch if nothing has changed.

- Check against existing SHAs in S3 and skip rebuilding x64 and arm64 artifacts.
- Check whether artifacts for the current build exist, and skip building docker artifacts otherwise.

### Check List


dblock:
@peternied @gaiksaya Could I please get a thorough review of this? Let's get it in?

vars/buildAssembleUpload.groovy
```diff
@@ -12,29 +12,42 @@ void call(Map args = [:]) {
...
+        }
+    }
 
-    assembleManifest(
-        args + [
-            manifest: args.dryRun ? 'tests/data/opensearch-build-1.1.0.yml' : 'builds/opensearch/manifest.yml'
-        ]
-    )
+    if (shaExists) {
+        lib.jenkins.Messages.new(this).add("${STAGE_NAME}", "Skipped ${STAGE_NAME}, ${manifestShaPath} exists.")
```
peternied, line 28: This will show up in the slack notification message, seems like if that is the case we would sent a different notification or not send one at all

vars/buildDockerImage.groovy
```diff
@@ -9,36 +9,47 @@ void call(Map args = [:]) {
...
+    def arm64Response = args.dryRun ? [status: 200] : httpRequest(httpMode: 'HEAD', url: arm64TarGz)
 
-    if (! args.dryRun) {
-        dockerBuild: {
-            build job: 'docker-build',
-            parameters: parameters
-        }
+    if (x64Response.status != 200 || arm64Response.status != 200) {
+        echo([
+            'Skipping docker build.',
```
peternied, line 17: Seems like this should error out if the artifacts cannot be found.

vars/buildDockerImage.groovy
```diff
@@ -9,36 +9,47 @@ void call(Map args = [:]) {
...
-    ]
+    def x64Response = args.dryRun ? [status: 200] : httpRequest(httpMode: 'HEAD', url: x64TarGz)
+    def arm64Response = args.dryRun ? [status: 200] : httpRequest(httpMode: 'HEAD', url: arm64TarGz)
 
-    if (! args.dryRun) {
-        dockerBuild: {
-            build job: 'docker-build',
-            parameters: parameters
-        }
+    if (x64Response.status != 200 || arm64Response.status != 200) {
```
peternied, line 15: Looks like this is using these existence checks as a flow of control mechanism so non-published builds don't have docker images created.  I would rather we not call the buildDockerImage at all if the build was already created

peternied (changes requested):

vars/buildAssembleUpload.groovy
dblock: We collect messages in the end and I believe would fail if there's no message from this stage, so this way the status in Slack shows clearly that the build was skipped for this stage.

vars/buildDockerImage.groovy
dblock: You're right, but there's no other easy way to tell the docker build that the incremental build didn't build without introducing a global, and I couldn't get it to work in tests.

vars/buildDockerImage.groovy
dblock: I'd like to leave this as a future exercise. I couldn't find an easy way to pass on something that says "both x64 and arm64 builds were skipped, therefore skip the build docker other than by checking that the artifact exists. But I have some ideas.

vars/buildAssembleUpload.groovy
```diff
@@ -12,29 +12,42 @@ void call(Map args = [:]) {
...
+    String manifestShaPath = "${shasRoot}/${manifestSHA}.yml"
 
-    buildManifest(
-        args + [
-            manifest: manifestLock
-        ]
-    )
+    Boolean shaExists = false
+    withAWS(role: 'opensearch-bundle', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+        if (!args.dryRun && s3DoesObjectExist(bucket: "${ARTIFACT_BUCKET_NAME}", path: manifestShaPath)) {
```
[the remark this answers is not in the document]
dblock: Makes it a lot more difficult to test the combination of parameters.

vars/buildDockerImage.groovy
peternied: Could you create an issue for this?

vars/buildAssembleUpload.groovy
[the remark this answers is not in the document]
peternied: Gotcha, so also ignoring read-only actions too

peternied (approved):

vars/buildDockerImage.groovy
dblock: https://github.com/opensearch-project/opensearch-build/issues/1183

Issue #1117, opened 2021-11-22
Support the RPM distribution builds


gaiksaya:
- Assemble process will have new param --distribution to specify which distribution to build
- Uncouple distribution creation with min tarball/zip file extension
- RPM will be built based on min tar + plugin zip installation + fpm bundle pkg
- User should be able to use ./assemble builds/manifests.yml --distribution rpm to build rpm and get a rpm file in return for installation.
- We design it to run on a specific Docker Image with ruby, rpm, fpm installed, not able to run directly on macOS directly. We will make changes to the docker files as well
- Part of the pre/post scripts and sysvinit/systemd will be taken from ODFE.
- Also need these scripts specifically for pkg from PA: https://github.com/opensearch-project/performance-analyzer/tree/main/packaging

Acceptance criteria: 
- [x] The RPM package builds for both arm64 and x64.
- [x] able to install the package with no errors. 
- [x] able to start the service through systemd.
- [x] able to remove installed/uninstall package from host.

zylzulu:
[Update] The .rpm package is able to be built after run ./assemble.sh builds/manifest.yml --distribution rpm. Here is the log for it: 
Installing dependencies in . ...
Installing dependencies from Pipfile.lock (c3492c)...
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/0 — 00:00:00
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
Running ./src/run_assemble.py builds/manifest.yml --distribution rpm ...
2021-11-23 19:42:39 INFO     Bundling OpenSearch (x64) on linux into /usr/share/opensearch/opensearch-rpm-test-in-docker/dist ...
2021-11-23 19:42:39 INFO     Copied min bundle to /tmp/tmpnuezu6mb/opensearch-min-1.1.0-linux-x64.tar.gz.
2021-11-23 19:42:39 INFO     Extracting dist into /tmp/tmpnuezu6mb.
2021-11-23 19:42:43 INFO     Extracted dist into /tmp/tmpnuezu6mb.
[...]

PR #1156, opened 2021-11-24
[RFC] OpenSearch / OpenSearch Dashboards - Support for RPM distribution

zylzulu:
### Description

- Make changes to the existing assemble workflow to enable new param --distribution to specify which distribution to build
- Make changes to the existing assemble workflow in order to call fpm to build rpm and get a rpm file in return for installation.
- RPM Package scripts are coded by @peterzhuamazon and moved from peter's fork here: https://github.com/opensearch-project/opensearch-build/commit/1773a41a7c373ece7fbad52038afb106c1341f14

 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1117
 
### Check List


src/assemble_workflow/bundle_recorder.py
```diff
@@ -35,7 +36,13 @@ def __get_package_name(self, build):
             build.platform,
             build.architecture,
         ]
-        return "-".join(parts) + (".zip" if build.platform == "windows" else ".tar.gz")
+        if self.distribution == "rpm":
```
dblock, line 39: Add a property called `extension` that uses map and look up the value with that.

```
EXTENSIONS = {
   "rpm": ".rpm",
   "tar.gz": ...
}
```

src/assemble_workflow/dist.py
```diff
@@ -16,9 +17,10 @@
 
 
 class Dist(ABC):
-    def __init__(self, name, path):
+    def __init__(self, name, path, distribution):
         self.name = name
         self.path = path
+        self.distribution = distribution
```
dblock, line 23: This shouldn't be necessary since the class `DistRpm`, `DistZip` "knows" what kind of distribution it is. If you need the name, then add a property called `.name` and return it. If you need the extension, add a property called `.extension` and return ".rpm", ".tar.gz", etc.

src/assemble_workflow/dist.py
```diff
@@ -36,30 +38,32 @@ def extract(self, dest):
...
 
     @classmethod
-    def from_path(cls, name, path):
-        ext = os.path.splitext(path)[1]
-        if ext == ".gz":
-            return DistTar(name, path)
-        elif ext == ".zip":
-            return DistZip(name, path)
+    def create_dist(cls, name, path, distribution):
+        if distribution == "rpm":
```
dblock, line 50: Make and use a map.

src/assemble_workflow/dist.py
```diff
@@ -72,6 +76,97 @@ def __extract__(self, dest):
...
+            tar.extractall(dest)
+
+    def __build__(self, bundle_recorder, dest):
+        logging.info("build for rpm distribution.")
+        manifest_data = bundle_recorder.bundle_manifest.data["build"]
+        product_name = manifest_data["name"].lower()
+        if product_name == "opensearch dashboards":
+            product_name = "-".join(product_name.split())
+            product_name_alt = "_".join(product_name.split())
+        output_dir = os.getcwd()
```
dblock, line 96: Should come from the outside, don't assume current directory.

src/assemble_workflow/dist.py
```diff
@@ -72,6 +76,97 @@ def __extract__(self, dest):
...
+        output_package_name = output_dir + "/" + bundle_recorder.package_name
+        description = " ".join([product_name , self.distribution, manifest_data["version"]])
+        scripts_path = output_dir + "/scripts/pkg"
+        pre_install_path = scripts_path + "/scripts/pre_install.sh"
```
dblock, line 100: Many of these properties are used once, no need to assign them to variables.

src/assemble_workflow/dist.py
```diff
@@ -72,6 +76,97 @@ def __extract__(self, dest):
...
+        logging.info("build for rpm distribution.")
+        manifest_data = bundle_recorder.bundle_manifest.data["build"]
+        product_name = manifest_data["name"].lower()
+        if product_name == "opensearch dashboards":
+            product_name = "-".join(product_name.split())
+            product_name_alt = "_".join(product_name.split())
+        output_dir = os.getcwd()
+        output_package_name = output_dir + "/" + bundle_recorder.package_name
+        description = " ".join([product_name , self.distribution, manifest_data["version"]])
+        scripts_path = output_dir + "/scripts/pkg"
```
dblock, line 99: Use `path.join` everywhere, don't assume `/` as a separator.

src/assemble_workflow/dist.py
```diff
@@ -72,6 +76,97 @@ def __extract__(self, dest):
...
+                "--template-value",
+                "product=" + product_name,
+                "--template-value",
+                "user=" + product_name,
+                "--template-value",
+                "group=" + product_name,
+                "--template-value",
+                "homeDir=/usr/share/" + product_name,
+                "--template-value",
+                "configDir=/etc" + product_name,
```
dblock, line 156: os.path.join(os.path.sep + 'etc', product_name)

dblock (changes requested):
Bunch of comments, but overall this works well. Let's get this to a mergeable state!

src/assemble_workflow/dist.py
zylzulu: Hi @dblock Thank you so much for your comments! But I'm a little bit confused about how to implement it if we don't add self.distribution to the Dist class. In DistRpm class, we need the distribution of "rpm". Do you mean to simply add a variable called distribution_name = "rpm" in the DistRpm class and use this variable whenever we use self.distribution before?

dblock: `DistZip` knows it's a zip and `DistRpm` knows it's rpm, so something like this:

```
class DistZip
    @property
    def distribution(self):
        return "zip"
```

zylzulu: I see. We only need "rpm" in the DistRpm class

src/assemble_workflow/dist.py
zylzulu: I see. This is not the output directory. os.getcwd() need to be used to get the current working directory which is used  for other variables, like scripts_path, to refer. I will change the name from output_dir to root.

dblock: If you need paths relative to this file, they can be computed from `__file__`, like in `ScriptFinder`. Generally avoid `chdir` and `getcwd`.

src/assemble_workflow/assemble_args.py
```diff
@@ -31,9 +31,17 @@ def __init__(self):
             const=logging.DEBUG,
             dest="logging_level",
         )
+        parser.add_argument(
+            "-d",
+            "--distribution",
+            dest="distribution",
+            help="Choose distribution",
+            default="tar",
+        )
```
dblock, line 40: Add `choices=self.SUPPORTED_DISTRIBUTIONS` and define  `SUPPORTED_DISTRIBUTIONS` similar to https://github.com/opensearch-project/opensearch-build/blob/c31d6a04ddcdbdb66a2f7ba9bfabdcf8c52f6d51/src/build_workflow/build_args.py#L13

src/assemble_workflow/assemble_args.py
```diff
@@ -31,9 +31,17 @@ def __init__(self):
             const=logging.DEBUG,
             dest="logging_level",
         )
+        parser.add_argument(
+            "-d",
+            "--distribution",
+            dest="distribution",
+            help="Choose distribution",
```
dblock, line 38: Add a period.

src/assemble_workflow/bundle_recorder.py
```diff
@@ -35,7 +36,12 @@ def __get_package_name(self, build):
             build.platform,
             build.architecture,
         ]
-        return "-".join(parts) + (".zip" if build.platform == "windows" else ".tar.gz")
+        EXTENSIONS = {
```
dblock, line 39: Move `EXTENSIONS` constant out of the function up into the class.

src/assemble_workflow/dist.py
```diff
@@ -24,6 +25,10 @@ def __init__(self, name, path):
     def __extract__(self, dest):
         pass
 
+    @abstractmethod
+    def get_distribution(self):
```
dblock, line 29: Rename to `distribution` and make it a `@property`.

src/assemble_workflow/dist.py
```diff
@@ -36,42 +41,145 @@ def extract(self, dest):
...
 
     @classmethod
-    def from_path(cls, name, path):
-        ext = os.path.splitext(path)[1]
-        if ext == ".gz":
-            return DistTar(name, path)
-        elif ext == ".zip":
-            return DistZip(name, path)
+    def create_dist(cls, name, path, distribution):
+        DISTRIBUTIONS = {
```
dblock, line 53: Move the constant out of the method and make a map between the extension and the class name, so:

```python
"rpm":DistRpm
...
```

then `return DISTRIBUTIONS[distribution](name, path)`, avoids having to copy constructor args and keeps the table a constant.

src/assemble_workflow/dist.py
```diff
@@ -36,42 +41,145 @@ def extract(self, dest):
...
+    def __build__(self, bundle_recorder, dest):
+        with tarfile.open(bundle_recorder.package_name, "w:gz") as tar:
             tar.add(self.archive_path, arcname=os.path.basename(self.archive_path))
+
+    @property
+    def get_distribution(self):
+        return "tar"
+
+
+class DistRpm(Dist):
```
dblock, line 96: Given how long these classes are becoming, move them to their own files (dist_rpm.py, etc.), and have their own `test_...` files.

Then `create_dist` will have to go into a new class `Dists` (plural) that will have a `create` (just call it that) static method.

src/assemble_workflow/dist.py
```diff
@@ -36,42 +41,145 @@ def extract(self, dest):
...
+                "--template-value",
+                "logDir=" + os.path.join(os.sep, "var", "log", product_name),               
+                "--exclude",
+                os.path.join("usr", "share", product_name, "data"),
+                "--exclude",
+                os.path.join("usr", "share", product_name, "config"),
+                "--architecture",
+                architecture_alt,
+                os.path.join(real_archive_path, "") + "=" + os.path.join(os.sep, "usr", "share", product_name, ""),
+                os.path.join(real_archive_path, "config", "") + "=" + os.path.join(os.sep, "etc", product_name, ""),
```
dblock, line 180: why is there an empty "" here? Is it to add a `/`?

src/run_assemble.py
```diff
@@ -30,16 +30,17 @@ def main():
...
-    with Bundles.create(build_manifest, artifacts_dir, bundle_recorder, args.keep) as bundle:
+    with Bundles.create(build_manifest, artifacts_dir, bundle_recorder, args.distribution, args.keep) as bundle:
         bundle.install_min()
         bundle.install_plugins()
         logging.info(f"Installed plugins: {bundle.installed_plugins}")
 
         #  Save a copy of the manifest inside of the tar
         bundle_recorder.write_manifest(bundle.min_dist.archive_path)
         bundle.package(output_dir)
+        logging.info(f"Package for {args.distribution}")
```
dblock, line 43: Packaging {args.distribution}.

dblock (changes requested):
Close.

Make sure to write tests for the actual `DistTar`, `DistZip` and `DistRpm` classes.

peternied:
This PR is weighing at +700 lines of code, can we break this into smaller parts to make merging this easier?

dblock:
> This PR is weighing at +700 lines of code, can we break this into smaller parts to make merging this easier?

The stuff in scripts/pkg is coming from a previous codebase, isn't it?

src/assemble_workflow/dist.py
zylzulu: Thank you for your comments! @dblock We are thinking about extracting the whole fpm build process of DistRpm() into its parent class Dist() as a method called fpm_build() or something else, since we are going to add one more DistDeb() which shares the same build process as DistRpm() in the future. Then, these two subclasses could inherit fpm_build() and we only need to implement this method once to increase reusability of the code. Do you think this idea make sense?

src/assemble_workflow/dist.py
zylzulu: Yes, in order to add "/" at the end.

peternied:
Python changes look pretty good to me, I'm not sure that we validated those script before.  If this goes in how do we know if it works?  I'm happy if there is a follow up task, I already created one for the automated testing https://github.com/opensearch-project/opensearch-build/issues/1244

zylzulu:
> > This PR is weighing at +700 lines of code, can we break this into smaller parts to make merging this easier?
> 
> The stuff in scripts/pkg is coming from a previous codebase, isn't it?

Thank you for your comment! @peternied They are come from the fork of @peterzhuamazon. Please refer to this commit https://github.com/opensearch-project/opensearch-build/pull/1156/commits/1773a41a7c373ece7fbad52038afb106c1341f14

zylzulu:
> Python changes look pretty good to me, I'm not sure that we validated those script before. If this goes in how do we know if it works? I'm happy if there is a follow up task, I already created one for the automated testing #1244

Thanks, Peter! We are still trying to make those shell scripts work. Currently, the RPM package can be generated by the python code, and it can be installed on CentOS successfully but still bugs when starting the service.

peternied:
> We are still trying to make those shell scripts work.

What do you think about removing the scripts so we can merge the python changes, and then review the shell scripts?

src/assemble_workflow/dist.py
dblock: Do these parameters refer to actual paths, settings or something else? Do they have to be *nix paths always? 

If yes, then you don't want `os.path.join` because on Windows this will result in `\` instead of `/`.

src/assemble_workflow/dist.py
dblock: FPM is its own tool/process, so you likely will want a utility class called `FpmBuilder` or something like that.

zylzulu: Ok. That makes sense

src/assemble_workflow/dist.py
zylzulu: Yes, they are actual paths, e.g. `usr/share/opensearch/` Does it exactly what we want here? using `\` as seperator in Windows and `/` in *nix?

zylzulu:
> > We are still trying to make those shell scripts work.
> 
> What do you think about removing the scripts so we can merge the python changes, and then review the shell scripts?

Thank you @peternied for helping me extracting all Python code from the shell scripts and rebase to a new branch. Will close this PR, and raise another one including changes on the assemble workflow.

src/assemble_workflow/dist.py
zylzulu: In this way, the class name will not be defined. How about keep the dictionary in the method but remove all the `(name, path)` ?

dblock: The pattern we use is to use another file called `Dists` and have the method there. Otherwise you're introducing a circular dependency.

nerijus:
Could you please link the new PRs here?

zylzulu:
> Could you please link the new PRs here?

We haven't created the new PR yet. I will update here once it is raised.

Issue #1177, opened 2021-11-29
Update Test Suite to start Dashboards test cluster


tianleh:
### Is your feature request related to a problem? Please describe

#604

### Describe the solution you'd like

Test Suite to start Dashboards test cluster

PR #1192, opened 2021-11-30
refactor integ_test_suite to support both OpenSearch and OpenSearch Dashboards

tianleh:
### Description
refactor integ_test_suite to support both OpenSearch and OpenSearch Dashboards
 
### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/1177
 
### Check List

### Test

Unit Test is working in progress.

Manual test is showing integ tests running from component to component


tianleh:
send for early feedback on the functional part. Unit Test is still working in progress.

src/test_workflow/integ_test/integ_test_suite_opensearch.py
```diff
@@ -0,0 +1,83 @@
...
+            self.work_dir,
+            self.component.name,
+            self.additional_cluster_config,
+            self.bundle_manifest_opensearch,
+            security,
+            config,
+            self.test_recorder,
+        ) as (test_cluster_endpoint, test_cluster_port):
+            self.pretty_print_message("Running integration tests for " + self.component.name)
+            os.chdir(self.work_dir)
```
dblock, line 82: Future fix, we should be passing `self.work_dir` through and running the command in that directory rather than `chdir`.

dblock (commented):
This looks good.

I see a few instances specific to a product type, such as `dependency_installer_opensearch` and `bundle_manifest_opensearch`, passed into a generic `IntegTestSuite`. Most of these probably need the same hierarchy: a generic dependency installer that has a subclass for OpenSearch vs. OpenSearch Dashboards, and `IntegTestSuite` would only know about the generic version.

peternied (dismissed):
Great refactor, please add unit tests

src/test_workflow/integ_test/integ_test_suite_opensearch.py
```diff
@@ -0,0 +1,83 @@
...
+
+from test_workflow.integ_test.integ_test_suite import IntegTestSuite, InvalidTestConfigError
+from test_workflow.integ_test.local_test_cluster import LocalTestCluster
+from test_workflow.test_result.test_component_results import TestComponentResults
+from test_workflow.test_result.test_result import TestResult
+
+
+class IntegTestSuiteOpenSearch(IntegTestSuite):
+    """
+    Kicks of integration tests for a component based on test configurations provided in
```
sichend, line 19: Kicks off?

src/test_workflow/integ_test/integ_test_suite_opensearch.py
```diff
@@ -0,0 +1,83 @@
...
+            work_dir,
+            component,
+            test_config,
+            test_recorder,
+            dependency_installer_opensearch,
+            bundle_manifest_opensearch,
+            build_manifest_opensearch
+        )
+
+    def execute_tests(self):
```
sichend, line 43: TestSuite.execute is more concise than TestSuite.execute_tests, unless there are other executions in this suite. For my learning, is there a specific reason to change the function name?

src/test_workflow/integ_test/integ_test_suite_opensearch.py
tianleh: Good catch.

src/test_workflow/integ_test/integ_test_suite_opensearch.py
tianleh: It also imports system execute. https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/integ_test_suite.py#L14 To avoid confusion, I call it execute_tests

tianleh:
Manual integ test output:

```
2021-12-01 04:34:59 INFO     | alerting             | with-security        | PASS  |    0 |
2021-12-01 04:34:59 INFO     | alerting             | without-security     | PASS  |    0 |
2021-12-01 04:34:59 INFO     | anomaly-detection    | with-security        | PASS  |    0 |
2021-12-01 04:34:59 INFO     | anomaly-detection    | without-security     | PASS  |    0 |
2021-12-01 04:34:59 ERROR    | asynchronous-search  | with-security        | FAIL  |    1 |
2021-12-01 04:34:59 INFO     | asynchronous-search  | without-security     | PASS  |    0 |
2021-12-01 04:34:59 INFO     | dashboards-reports   | without-security     | PASS  |    0 |
2021-12-01 04:34:59 INFO     | index-management     | with-security        | PASS  |    0 |
2021-12-01 04:34:59 INFO     | index-management     | without-security     | PASS  |    0 |
2021-12-01 04:34:59 ERROR    | k-NN                 | with-security        | FAIL  |    1 |
2021-12-01 04:34:59 INFO     | k-NN                 | without-security     | PASS  |    0 |
2021-12-01 04:34:59 INFO     | sql                  | with-security        | PASS  |    0 |
2021-12-01 04:34:59 INFO     | sql                  | without-security     | PASS  |    0 |
```

dblock (approved):
Good. So next step is to wire this up to run either OpenSearch or. Dashboards?

sichend (approved):

seraphjiang (approved):

tianleh:
> Great refactor, please add unit tests

Added

tianleh:
> 

Yes. Next will be to support either OpenSearch or OpenSearch Dashboard integ test from the test runner/shell https://github.com/opensearch-project/opensearch-build/blob/main/src/run_integ_test.py#L42

That will be the last piece of the puzzle.

PR #1213, opened 2021-12-01
Breakup tests into clear scenarios, setup and recorded steps

dblock:
### Description

We have a bunch of tests that include various steps in Jenkinsfile that get recorded. This is not consistent and does not express intent of the test. In this PR tests get separated and for each scenario a separate .txt file is recorded, making it easier to grok what the setup, expectation and output is for each scenario. It also reduces relying on the combination of effects of `dryRun` and `s3ObjectExists` for incremental build tests.

On top of https://github.com/opensearch-project/opensearch-build/pull/1212 which already implements this pattern for SHA tests.

- Extract tests for `buildAssembleUpload` from incremental build tests and separate test runs with different setup and recorded .txt steps. 
- Extract tests for `postCleanup`, `buildDockerImage`, and `publishNotification` into their own tests similarly.
- Pass in manifest in `publishNotification` instead of checking whether this is a build notification or a test notification (in the latter no manifest is referenced).
  
### Check List


tianleh (approved):

peternied (approved):

dblock:
Can I please get two approves for https://github.com/opensearch-project/opensearch-build/pull/1212? Just want to preserve that as a separate commit in history.

PR #1219, opened 2021-12-01
adding perms to grant read and put to clients dir on s3

### LINKED ISSUE #872: Publish artifacts into rubygems.org repository

build and upload gems to rubygems.org repository in following order

https://github.com/opensearch-project/opensearch-ruby/tree/main/opensearch-api
https://github.com/opensearch-project/opensearch-ruby/tree/main/opensearch-transport
https://github.com/opensearch-project/opensearch-ruby/tree/main/opensearch
https://github.com/opensearch-project/opensearch-ruby/tree/main/opensearch-dsl

### Tasks
- [x] Add perms to upload folder on s3
- [x] Update ruby version on docker image used with jenkins
- [x] Create `JenkinsFile` to run the build, sign and upload to s3 pipeline

Should be available at rubygems.org

_No response_

### Additional context

_No response_

abhinavGupta16:

### Description
Adding perms so bundle role can upload to the clients dir on s3 bucket
 
### Issues Resolved
#872 - Part of this issue
 
### Check List

### DIFF
--- a/deployment/lib/identities.ts
+++ b/deployment/lib/identities.ts
@@ -42,6 +42,9 @@ export class Identities {
     props.buildBucket.grantRead(bundleRole, '*/builds/*');
     props.buildBucket.grantPut(bundleRole, '*/builds/*');
 
+    props.buildBucket.grantRead(bundleRole, '*/clients/*');
+    props.buildBucket.grantPut(bundleRole, '*/clients/*');
+
     props.buildBucket.grantRead(bundleRole, '*/shas/*');
     props.buildBucket.grantPut(bundleRole, '*/shas/*');
 


gaiksaya (approved):

peternied (changes requested):
Why are we adding a client bucket?

peternied:
> Why are we adding a client bucket?

It seems like we should be pushing the artifacts into the `build` folder

gaiksaya:
> Why are we adding a client bucket?

Don't we cant to keep bundle builds and clients build separate?
I think its much cleaner and less confusing.

zelinh (approved):
I think it would be better if we have builds & clients separate. Be sure we deploy this change after merge.

peternied:
> > Why are we adding a client bucket?
> 
> Don't we want to keep bundle builds and clients build separate? I think its much cleaner and less confusing. Inside clients we would have folders per client basis (is at all required) like `opensearch-ruby`

All jenkins jobs should have their own top level s3 bucket so there is no possible intersection, is there a reason this isolation should be removed?

abhinavGupta16:
> > > Why are we adding a client bucket?
> > 
> > 
> > Don't we want to keep bundle builds and clients build separate? I think its much cleaner and less confusing. Inside clients we would have folders per client basis (is at all required) like `opensearch-ruby`
> 
> All jenkins jobs should have their own top level s3 bucket so there is no possible intersection, is there a reason this isolation should be removed?

I believe eventually with the number of clients increasing, it will be easier to manage in the future to have all the clients in one place. Each job having the same top level bucket may become messy later on. Thoughts?

gaiksaya:
> > > Why are we adding a client bucket?
> > 
> > 
> > Don't we want to keep bundle builds and clients build separate? I think its much cleaner and less confusing. Inside clients we would have folders per client basis (is at all required) like `opensearch-ruby`
> 
> All jenkins jobs should have their own top level s3 bucket so there is no possible intersection, is there a reason this isolation should be removed?

Are you saying each client should have their own bucket? That sounds good too. In that case @abhinavGupta16 would need to make changes to cdk to accommodate this. Along with new role that has access only to that bucket.

peternied:
> Are you saying each client should have their own bucket? That sounds good too. In that case @abhinavGupta16 would need to make changes to cdk to accommodate this. Along with new role that has access only to that bucket.

Here is how it looks today:

- BundleBuild
  - https://ci.opensearch.org/ci/dbc/bundle-build/2.0.0/1240/linux/arm64/builds/...
- BundleBuildDashboards
  - https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/2.0.0/1240/linux/arm64/builds/...

I am proposing we do _not change anything_ with our current artifact management process, this change could be
- OpenSearch-Ruby
  - https://ci.opensearch.org/ci/dbc/opensearch-ruby/1.2.0/17/linux/arm64/builds/...

abhinavGupta16:
> > > > Why are we adding a client bucket?
> > > 
> > > 
> > > Don't we want to keep bundle builds and clients build separate? I think its much cleaner and less confusing. Inside clients we would have folders per client basis (is at all required) like `opensearch-ruby`
> > 
> > 
> > All jenkins jobs should have their own top level s3 bucket so there is no possible intersection, is there a reason this isolation should be removed?
> 
> Are you saying each client should have their own bucket? That sounds good too. In that case @abhinavGupta16 would need to make changes to cdk to accommodate this. Along with new role that has access only to that bucket.

Won't that become too many roles as clients increase? It would be great if it would restrict other clients to be able to use the role? The way I see it, people can simply use the role they want and Jenkins would have permission to use and upload it.

abhinavGupta16:
> > Are you saying each client should have their own bucket? That sounds good too. In that case @abhinavGupta16 would need to make changes to cdk to accommodate this. Along with new role that has access only to that bucket.
> 
> Here is how it looks today:
> 
> * BundleBuild
>   
>   * https://ci.opensearch.org/ci/dbc/bundle-build/2.0.0/1240/linux/arm64/builds/...
> * BundleBuildDashboards
>   
>   * https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/2.0.0/1240/linux/arm64/builds/...
> 
> I am proposing we do _not change anything_ with our current artifact management process, the new job would could have a path like this
> 
> * OpenSearch-Ruby
>   
>   * https://ci.opensearch.org/ci/dbc/opensearch-ruby/1.2.0/17/linux/arm64/builds/...
> 
> FYI @abhinavGupta16

I see your point. Looking in the future, what happens when the number of clients increase? Wouldn’t it become a problem to look them up since all will be present in the root dir?

peternied:
> However, looking into the future, what happens when the number of clients increase?

Each client should have their own deployment configuration, if this number is sufficiently larger we will need to find a way to shard them, assuming we know how to divide them up I think we can just deploy a fresh Jenkins stamp with the existing bucket management tools.

peternied:
> problem to look them up since all will be present in the root dir

We should not be 'looking' at a directory for these proceses, that is an error prone process.  We will need automation that takes the results of the build process and then publishes.  We have the very begining of this for the OpenSearch service but we will need to advance this to support the large number of clients we have and will continue to have

dblock (commented):
What's special about "clients"? Shouldn't that be `builds/<product name>/...` or `dist/<product name>/...`?

abhinavGupta16:
I believe It's a design question that needs to be visited and discussed properly to see which solution is scalable and which solution works best for our org. I have created an issue for this - #1223

Since this issue needs perms now for deployment and release for rubygems, what should we do in the meantime here? Can we simply go ahead with this? @peternied @gaiksaya @zelinh @peterzhuamazon @dblock

dblock:
> I believe It's a design question that needs to be visited and discussed properly to see which solution is scalable and which solution works best for our org as we grow in size. I have created an issue for this - #1223
> 
> Since this issue needs perms now for deployment and release for rubygems, what should we do in the meantime here? Can we simply go ahead with this? @peternied @gaiksaya @zelinh @peterzhuamazon @dblock

Adding `clients` is my least favorite option. We already have a pattern with `job-name/version/build-number/platform/architecture/builds|dist`. Why do we need to add `clients` to it?

abhinavGupta16:
> > I believe It's a design question that needs to be visited and discussed properly to see which solution is scalable and which solution works best for our org as we grow in size. I have created an issue for this - #1223
> > Since this issue needs perms now for deployment and release for rubygems, what should we do in the meantime here? Can we simply go ahead with this? @peternied @gaiksaya @zelinh @peterzhuamazon @dblock
> 
> Adding `clients` is my least favorite option. We already have a pattern with `job-name/version/build-number/platform/architecture/builds|dist/product-name`. Why do we need to add `clients` to it?

We can put it in `builds` as well, however, we don't have specific platform or architecture associated with these. Therefore, not sure why we need them.

dblock:
> We can put it in `builds` as well, however, we don't have specific platform or architecture associated with these. Therefore, not sure why we need them.

In general ruby gems can target irb vs. jruby, so double-check whether we have any kind of such variation.

abhinavGupta16:
`BUCKET/JOB_NAME/VERSION/build` was decided as the path for s3.

Issue #1211, opened 2021-12-01
Run either OpenSearch or OpenSearch Dashboards integ test from test runner/shell


tianleh:
### Is your feature request related to a problem? Please describe

#604

### Describe the solution you'd like

Run either OpenSearch or OpenSearch Dashboards integ test from test runner/shell

PR #1222, opened 2021-12-01
Support running OpenSearch or OpenSearch Dashboards integ test from t…

tianleh:
…est manifest args

### Description
Add a new args param `--test-manifest-path` to specify the location of the test manifest file. If not specified, the test runner will by default run OpenSearch integ tests.
 
### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/1211
 
### Test
Unit Test is working in progress.

Manual run the following commands and show the correct test suite is being invoked.

### Check List

### DIFF
--- a/src/run_integ_test.py
+++ b/src/run_integ_test.py
@@ -10,8 +10,7 @@ import sys
 
 from manifests.test_manifest import TestManifest
 from system import console
-from test_workflow.integ_test.integ_test_runner_opensearch import IntegTestRunnerOpenSearch
-from test_workflow.integ_test.integ_test_runner_opensearch_dashboards import IntegTestRunnerOpenSearchDashboards
+from test_workflow.integ_test.integ_test_runners import IntegTestRunners
 from test_workflow.test_args import TestArgs
 
 
@@ -22,12 +21,7 @@ def main():
 
     test_manifest = TestManifest.from_path(args.test_manifest_path)
 
-    all_results = None
-
-    if test_manifest.name == "OpenSearch Dashboards":
-        all_results = IntegTestRunnerOpenSearchDashboards(args, test_manifest).run()
-    else:
-        all_results = IntegTestRunnerOpenSearch(args, test_manifest).run()
+    all_results = IntegTestRunners.from_test_manifest(args, test_manifest).run()
 
     all_results.log()
 
--- a/src/run_integ_test.py
+++ b/src/run_integ_test.py
@@ -5,59 +5,34 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
-import logging
 import os
 import sys
 
-from manifests.build_manifest import BuildManifest
-from manifests.bundle_manifest import BundleManifest
 from manifests.test_manifest import TestManifest
 from system import console
-from system.temporary_directory import TemporaryDirectory
-from test_workflow.dependency_installer_opensearch import DependencyInstallerOpenSearch
-from test_workflow.integ_test.integ_test_suite_opensearch import IntegTestSuiteOpenSearch
+from test_workflow.integ_test.integ_test_runner_opensearch import IntegTestRunnerOpenSearch
+from test_workflow.integ_test.integ_test_runner_opensearch_dashboards import IntegTestRunnerOpenSearchDashboards
 from test_workflow.test_args import TestArgs
-from test_workflow.test_recorder.test_recorder import TestRecorder
-from test_workflow.test_result.test_suite_results import TestSuiteResults
 
 
 def main():
     args = TestArgs()
+
     console.configure(level=args.logging_level)
+
     test_manifest = TestManifest.from_path(args.test_manifest_path)
-    bundle_manifest = BundleManifest.from_urlpath("/".join([args.path.rstrip("/"), "dist/opensearch/manifest.yml"]))
-    build_manifest = BuildManifest.from_urlpath("/".join([args.path.rstrip("/"), "builds/opensearch/manifest.yml"]))
-    dependency_installer = DependencyInstallerOpenSearch(args.path, build_manifest, bundle_manifest)
-    tests_dir = os.path.join(os.getcwd(), "test-results")
-    os.makedirs(tests_dir, exist_ok=True)
-    with TemporaryDirectory(keep=args.keep, chdir=True) as work_dir:
-        test_recorder = TestRecorder(args.test_run_id, "integ-test", tests_dir)
-        dependency_installer.install_maven_dependencies()
-        all_results = TestSuiteResults()
-        for component in bundle_manifest.components.select(focus=args.component):
-            if component.name in test_manifest.components:
-                test_config = test_manifest.components[component.name]
-                if test_config.integ_test:
-                    test_suite = IntegTestSuiteOpenSearch(
-                        dependency_installer,
-                        component,
-                        test_config,
-                        bundle_manifest,
-                        build_manifest,
-                        work_dir.name,
-                        test_recorder
-                    )
-                    test_results = test_suite.execute_tests()
-                    all_results.append(component.name, test_results)
-                else:
-                    logging.info(f"Skipping integ-tests for {component.name}, as it is currently not supported")
-            else:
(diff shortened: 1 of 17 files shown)


tianleh:
Send for early feedback on the functional part. Unit test is in progress.

src/run_integ_test.py
```diff
@@ -5,60 +5,40 @@
...
-from test_workflow.test_result.test_suite_results import TestSuiteResults
 
 
 def main():
     args = TestArgs()
+
     console.configure(level=args.logging_level)
-    test_manifest_path = os.path.join(os.path.dirname(__file__), "test_workflow", "config", "opensearch", "test_manifest.yml")
+
+    test_manifest_path = os.path.join(
```
dblock, line 23: I think you don't need this if the caller always has to specify a test manifest, which is completely fine (make that change).

src/run_integ_test.py
```diff
@@ -5,60 +5,40 @@
...
-                    all_results.append(component.name, test_results)
-                else:
-                    logging.info(f"Skipping integ-tests for {component.name}, as it is currently not supported")
-            else:
-                logging.info(f"Skipping integ-tests for {component.name}, as it is currently not declared in the test manifest")
 
-        all_results.log()
+    all_results = None
+
+    if test_manifest.name == "OpenSearch Dashboards":
```
dblock, line 33: Replace with a `IntegTestRunner.from(manifest.build.name)` or something like that, example in [bundles.py.](https://github.com/opensearch-project/opensearch-build/blob/main/src/assemble_workflow/bundles.py)

src/test_workflow/integ_test/integ_test_runner.py
```diff
@@ -0,0 +1,25 @@
...
+from manifests.bundle_manifest import BundleManifest
+from test_workflow.dependency_installer_opensearch import DependencyInstallerOpenSearch
+
+
+class IntegTestRunner(abc.ABC):
+    def __init__(self, args, test_manifest):
+        self.args = args
+        self.test_manifest = test_manifest
+
+        self.bundle_manifest = BundleManifest.from_urlpath("/".join([self.args.path.rstrip("/"), "dist/opensearch/manifest.yml"]))
```
dblock, line 19: This is opensearch specific, shouldn't this be in the parent class(es)?

src/test_workflow/integ_test/integ_test_runner_opensearch_dashboards.py
```diff
@@ -0,0 +1,62 @@
...
+        os.makedirs(tests_dir, exist_ok=True)
+        with TemporaryDirectory(keep=self.args.keep, chdir=True) as work_dir:
+            test_recorder = TestRecorder(self.args.test_run_id, "integ-test", tests_dir)
+
+            all_results = TestSuiteResults()
+            for component in self.build_manifest_opensearch_dashboards.components.select(focus=self.args.component):
+                if component.name in self.test_manifest.components:
+                    test_config = self.test_manifest.components[component.name]
+                    if test_config.integ_test:
+                        test_suite = IntegTestSuiteOpenSearchDashboards(
```
dblock, line 43: The entire method `run` is the same in both OpenSearch / Dashboards, except for the part that creates a test suite. That should go into `__create_test_suite(component, ...)`, then you'd call `execte_tests` on that. Some parameters, e.g. `test_recorder`, should likely be taken by `execute_tests` and not the constructor of the suite.

src/test_workflow/test_args.py
```diff
@@ -37,5 +38,10 @@ def __init__(self):
         self.logging_level = args.logging_level
         self.path = args.path if validators.url(args.path) else os.path.realpath(args.path)
 
+        if args.test_manifest_path:
```
dblock, line 41: Just require this.

dblock (changes requested):
I think as an increment it goes almost in the right direction, but something still feels broken by design. The test configuration is version-specific, isn't it? We should fix it by creating test manifests, e.g. `manifests/1.3.0/opensearch-1.3.0-test.yml` and requiring that as input to running integration tests.

src/test_workflow/integ_test/integ_test_runner.py
tianleh: Both runners need OpenSearch. Thus I put it here to avoid duplicated code.

dblock: I think logically it doesn't make sense: for dashboards this is a dependency. Codewise, this means composition rather than inheritance. If you build another class called "SomethingSomethingOpenSearch", you would add it to "dependencies" and call it accordingly.

tianleh:
> I think as an increment it goes almost in the right direction, but something still feels broken by design. The test configuration is version-specific, isn't it? We should fix it by creating test manifests, e.g. `manifests/1.3.0/opensearch-1.3.0-test.yml` and requiring that as input to running integration tests.

Created a separate PR to address this https://github.com/opensearch-project/opensearch-build/pull/1229

src/test_workflow/integ_test/integ_test_runner.py
tianleh: Created a base class `ServiceStartProperties` to encapsulate such. Code looks much cleaner

src/run_integ_test.py
tianleh: Since `IntegTestRunner` is `abc`, we cannot call its class method directly. Thus if/else is used to decide which child IntegTestRunner to use. The Bundles is not `abc`.

src/test_workflow/integ_test/integ_test_runner_opensearch_dashboards.py
tianleh: Created an abstract method `create_test_suite`. `test_recorder` is not for a single component and thus I keep it inside the `runner` instead of `suite`.

tianleh:
Addressed the functional parts. Unit test is in progress.

src/run_integ_test.py
dblock: Plus `IntegTestRunner` doesn't know about its child classes.

I typically add a `IntegTestRunners` class that knows about both.

dblock (approved):
This is a step forward, but you really need to get rid of this `if test_manifest.name == "OpenSearch Dashboards":` in a future PR.

src/run_integ_test.py
tianleh: Leveraged the skills learnt in `CiManifests` and updated.

src/test_workflow/integ_test/integ_test_runners.py
```diff
@@ -0,0 +1,21 @@
...
+# compatible open source license.
+
+
+from test_workflow.integ_test.integ_test_runner_opensearch import IntegTestRunnerOpenSearch
+from test_workflow.integ_test.integ_test_runner_opensearch_dashboards import IntegTestRunnerOpenSearchDashboards
+
+
+class IntegTestRunners:
+    def __klass(name):
+        if name == "OpenSearch Dashboards":
```
dblock, line 14: Let's do even better and future proof this.

```python
RUNNERS = {
  "OpenSearch" : IntegTestRunnerOpenSearch,
  "OpenSearch Dashboards" : IntegTestRunnerOpenSearchDashboards
}
```

Then you can get rid of `__klass` and just `return RUNNERS[name]`. It will fail with a `KeyError`, and if you want to be fancy you can first check whether `name in RUNNERS` and throw a better exception.

We do this here: https://github.com/opensearch-project/opensearch-build/blob/5820881cc214cb6c0a70ba4fd93decbaea324b76/src/manifests/manifest.py#L46

tianleh:
Related PR https://github.com/opensearch-project/opensearch-build/pull/1229 has been merged. 

Rebasing to be sync with latest `main`

dblock (approved):
Good.

tianleh:
Focusing on unit tests

seraphjiang (approved):

tianleh:
@dblock Added unit tests and all checks are successful.

tianleh:
Could you please take a look again? @dblock

tianleh:
I will merge now. If there are additional comments, I will address as follow up.

PR #1224, opened 2021-12-01
Typechecking for manifests

peternied:
### Description
Adding typechecking onto all files under `src/manifest` as the manifests are modeled data we should be able to get dependency resolution on.  We can start with these and then expand over time.

Note; I wasn't able to properly reference inner class types for passing via generics, this affected `ComponentManifest` as its pattern was to have an inner Components and Component classes to provide additional functionality, I've extracted these into out of the inner scope, but left them in the same files.

### Check List

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -31,7 +31,7 @@ jobs:
           pipenv run flake8 .
       - name: Run Type Checker
         run: |
-          pipenv run mypy .
+          pipenv run mypy src/manifests tests/tests_manifests
       - name: Run Tests with Coverage
         run: |
           pipenv run coverage run -m pytest --cov=./src --cov-report=xml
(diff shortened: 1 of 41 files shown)


dblock (approved):

gaiksaya (approved):
Looks Great!

Just a question: What does it mean when we write `# type: ignore[assignment]` and `# type: ignore[return-value]`

peternied:
> What does it mean when we write # type: ignore[assignment] and # type: ignore[return-value]

The errors can be looked up on the mypy docs, [return-value](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-return-value-is-compatible-return-value), [assignment](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-types-in-assignment-statement-assignment)

More detailed note, previously we were sprinkling `# type: ignore` in the codebase for typing errors, we had limited type checking enabled.  Now that we have more thorough checking more can go wrong, and the previous ignore statement would capture any kind of error, by being specific if a new kind of error comes up we should still get warned.  See https://mypy.readthedocs.io/en/stable/error_codes.html#error-codes for more background on the individual errors

Issue #1225, opened 2021-12-02
Build workflow force patch versions plugins zips to have the same version as the patch version


peterzhuamazon:
### Describe the bug

In https://github.com/opensearch-project/opensearch-build/issues/373 patch release is successfully added so I test it with OpenSearch-Dashboards 1.1.1 by manually creating a folder on S3 and add all the 1.1.0 zips, with the same structure as needed.

The artifacts are downloaded successfully but seems like it is still trying to match the upcoming dashboards version 1.1.1 to the plugin zips name which has 1.1.0.

https://github.com/opensearch-project/opensearch-build/blob/60345d6794104ce71cd42c[…]d_workflow/opensearch_dashboards/build_artifact_check_plugin.py

```

$ ./build.sh manifests/1.1.1/opensearch-dashboards-1.1.1.yml --platform linux
Installing dependencies in . ...
Installing dependencies from Pipfile.lock (c3492c)...
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/0 — 00:00:00
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
Running ./src/run_build.py manifests/1.1.1/opensearch-dashboards-1.1.1.yml --platform linux ...
2021-12-01 17:08:54 INFO     Building in /var/folders/d6/19_cf3qd6jb715ckzv_2jsvdmr559j/T/tmpsqjq68ai
2021-12-01 17:08:54 INFO     Building OpenSearch Dashboards (x64) into /Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/opensearch-build/builds/opensearch-dashboards
2021-12-01 17:08:54 INFO     Building OpenSearch-Dashboards
2021-12-01 17:08:54 INFO     Downloading https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/1.1.0/20210930/linux/x64/builds/opensearch-dashboards/manifest.yml ...
2021-12-01 17:08:54 INFO     Loading https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/1.1.0/20210930/linux/x64/builds/opensearch-dashboards/manifest.yml
[...]

PR #1226, opened 2021-12-02
Add support for OpenSearch Dashboards patches

dblock:
### Description

In https://github.com/opensearch-project/opensearch-build/pull/789 we introduced a mechanism to combine artifacts from previous releases in a new distribution. The version checks for OpenSearch were relaxed to support patches. This extends OpenSearch Dashboards version checks to do the same.

During 1.2.0 we have fixed https://github.com/opensearch-project/opensearch-build/issues/669 to avoid path collisions. This adjusts paths accordingly for downloading existing distributions.
 
### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/1225.
 
### Check List

### DIFF
--- a/src/build_workflow/opensearch_dashboards/build_artifact_check_plugin.py
+++ b/src/build_workflow/opensearch_dashboards/build_artifact_check_plugin.py
@@ -17,21 +17,25 @@ class BuildArtifactOpenSearchDashboardsCheckPlugin(BuildArtifactCheck):
     def check(self, path):
         if os.path.splitext(path)[1] != ".zip":
             raise BuildArtifactCheck.BuildArtifactInvalidError(path, "Not a zip file.")
-        opensearch_dashboards_version = re.sub(r"-SNAPSHOT$", "", self.target.opensearch_version)
-        match = re.search(rf"^(\w+)-{opensearch_dashboards_version}.zip$", os.path.basename(path))
+
+        match = re.search(r"^(\w+)-[\d\.]*.*.zip$", os.path.basename(path))
         if not match:
-            raise BuildArtifactCheck.BuildArtifactInvalidError(
-                path,
-                f"Expected filename to be in the format of pluginName-{opensearch_dashboards_version}.zip.",
-            )
+            raise BuildArtifactCheck.BuildArtifactInvalidError(path, "Expected filename to be in the format of pluginName-1.1.0.zip.")
+
         plugin_name = match.group(1)
+        valid_filenames = self.__valid_paths(plugin_name)
+        if not os.path.basename(path) in valid_filenames:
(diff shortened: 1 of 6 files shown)


src/build_workflow/opensearch_dashboards/build_artifact_check_plugin.py
```diff
@@ -17,21 +17,25 @@ class BuildArtifactOpenSearchDashboardsCheckPlugin(BuildArtifactCheck):
     def check(self, path):
         if os.path.splitext(path)[1] != ".zip":
             raise BuildArtifactCheck.BuildArtifactInvalidError(path, "Not a zip file.")
-        opensearch_dashboards_version = re.sub(r"-SNAPSHOT$", "", self.target.opensearch_version)
-        match = re.search(rf"^(\w+)-{opensearch_dashboards_version}.zip$", os.path.basename(path))
+
+        match = re.search(r"^(\w+)-[\d\.]*.*.zip$", os.path.basename(path))
```
peterzhuamazon, line 21: Will single digit be a problem?
We used to have things like 1.13.0 in ODFE, which is 2 digits after all.

tests/tests_build_workflow/opensearch_dashboards/test_build_artifact_opensearch_dashboards_check_plugin.py
```diff
@@ -24,12 +24,13 @@ def __mock(self, props={}, snapshot=True):
                     output_dir="output_dir",
                     name="OpenSearch",
                     version="1.1.0",
+                    patches=["1.0.0"],
```
peterzhuamazon, line 27: A bit confused here is this saying 1.1.0 can be a patch of 1.0.0?

dblock: it's just an example for tests, what this says is indeed that 1.1.0 is just 1.0.0 + some changes.

src/build_workflow/opensearch_dashboards/build_artifact_check_plugin.py
dblock: It shouldn't. This regex just says any combination of numbers and dots.

peterzhuamazon: Thanks, just confirmed by running it through a tester.

tests/tests_build_workflow/opensearch_dashboards/test_build_artifact_opensearch_dashboards_check_plugin.py
peterzhuamazon: Thanks.

peterzhuamazon (approved):
Thanks.

peternied (approved):

Issue #1228, opened 2021-12-02
Require test manifest path for running tests


tianleh:
### Is your feature request related to a problem? Please describe

#604

### Describe the solution you'd like

require test manifest path for running tests

tianleh:
Also move the test manifest files from `src` to `manifests`

dblock:
@tianleh No need to make new issues for things you're already working on :) less work for you

tianleh:
> 

Thanks for the suggestion. I originally created this issue because the effort seemed non-trivial. Also I am keeping practicing my ability to realize such phased changes in advance. With such tracking (or "backfilling") exercise, I can always look back later as self retrospective of task breakdown when I need to create tasks for a group of engineers.

If it is the label "untriaged" which may attract the team' attention to groom the issue, I can quickly remove such label post creation.

tianleh:
I have been maintaining the overall picture of the task breakdown in this issue https://github.com/opensearch-project/opensearch-build/issues/604 (scroll down a bit from the top)

PR #1229, opened 2021-12-02
require test manifest path for running tests

tianleh:

### Description
Move test manifest to the `manifests` dir and require `--test-manifest-path`. This will address parts of @dblock 's comments on https://github.com/opensearch-project/opensearch-build/pull/1222
 
### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/1228

### Test
local tests

### Check List


seraphjiang (approved):

tianleh:
Introduced an attribute `type` to `TestManifest` so that CI can have different validation logic for `InputManifest` vs `TestManifest`

dblock (changes requested):
1. <strike>I don't think I understand what purpose `type` serves when it can only be "test"? What other type will be there? When does it need to be distinguished from an input manifest?</strike> Oh I see, CI is failing on these manifests because it doesn't know that it's a test manifest. For now I'd rather not add a type but just hard-code skipping files that are called `-test.yml`.


2. It seems like all test workflows require the test manifest. If that's the case it should be a required positional parameter, e.g.`./test.sh [kind of test] [test manifest] [build location]` to be consistent with the other runners.

tianleh:
> 

For 1, sure. I can switch to a simple file name check.

For 2. This is already achieved by this PR. If not specified, this will fail 
`        self.test_manifest_path = args.test_manifest_path if validators.url(args.test_manifest_path) else os.path.realpath(args.test_manifest_path)
`

tests/test_run_ci.py
```diff
@@ -38,3 +38,11 @@ def test_main(self, mock_lists, mock_temp, *mocks):
...
         self.assertEqual(mock_lists.return_value.checkout.call_count, mock_lists.call_count)
         self.assertEqual(mock_lists.return_value.check.call_count, mock_lists.call_count)
+
+    OPENSEARCH_TEST_MANIFEST = os.path.realpath(os.path.join(os.path.dirname(__file__), "../manifests/1.3.0/opensearch-1.3.0-test.yml"))
+
+    @patch("argparse._sys.argv", ["run_ci.py", OPENSEARCH_TEST_MANIFEST])
+    @patch("run_ci.TemporaryDirectory")
+    def test_main_test_manifest(self, mock_temp, *mocks):
+        mock_temp.return_value.__enter__.return_value.name = tempfile.gettempdir()
+        main()
```
peternied, line 48: This unit test should have assertions on what happened

dblock:
> For 2. This is already achieved by this PR. If not specified, this will fail ` self.test_manifest_path = args.test_manifest_path if validators.url(args.test_manifest_path) else os.path.realpath(args.test_manifest_path)`

As a side effect - you get an error that the path specified doesn't exist or not a URL vs. you forgot to specify it. Make it required in test args. Then I would really change --arg to a positional operator, because it's always required to be consistent with all other runners.

src/run_ci.py
```diff
@@ -9,18 +9,31 @@
...
+    manifest_type = __get_manifest_type(args.manifest)
+
+    if manifest_type == "Test":
+        TestManifest.from_path(args.manifest.name)
+
+        logging.info("TestManifest schema validation succeeded")
+        logging.info("Done.")
+        return
+
+    manifest = InputManifest.from_path(args.manifest.name)
```
peternied, line 36: Shouldn't CI call have always expected a TestManifest type?

tests/test_run_ci.py
tianleh: Added

src/run_ci.py
tianleh: The existing manifests https://github.com/opensearch-project/opensearch-build/tree/main/manifests/1.3.0 are `InputManifest` which is different from the `TestManifest` newly added in this PR.

src/run_ci.py
```diff
@@ -24,9 +23,7 @@ def main():
     args = CiArgs()
     console.configure(level=args.logging_level)
 
-    manifest_type = __get_manifest_type(args.manifest)
-
-    if manifest_type == "Test":
+    if __is_test_manifest(args.manifest.name):
```
dblock, line 26: Can we please do a little bit better than an early return? I'll settle for an `if/else`, but ideally we make two classes `CiTestManifest` and `CiInputManifest` that inherit `CiManifest` and have `CiManifest.from_file(path)` that creates one or the other, then call a base `.check()` method.

src/test_workflow/test_args.py
```diff
@@ -24,7 +24,7 @@ class TestArgs:
     def __init__(self):
         parser = argparse.ArgumentParser(description="Test an OpenSearch Bundle")
         parser.add_argument("path", type=str, help="Location of build and bundle manifests.", default=".")
-        parser.add_argument("--test-manifest-path", type=str, help="Specify a test manifest path.")
+        parser.add_argument("test-manifest-path", type=str, help="Specify a test manifest path.")
```
dblock, line 27: Should this be the first argument, then the location of build and bundle manifest path already has a default, so you won't have to do `./test.sh . test_manifest`, can just do `./test.sh test_manifest`.

src/run_ci.py
tianleh: Added the new classes

src/test_workflow/test_args.py
tianleh: moved to first.

jenkins/test/testsuite/Jenkinsfile
```diff
@@ -60,7 +60,7 @@ pipeline {
                 script {
                     basePath = "https://ci.opensearch.org/ci/dbc/bundle-build/${opensearch_version}/${build_id}/${platform}/${architecture}"
                     sh "wget ${basePath}/builds/opensearch/manifest.yml"
-                    sh "./test.sh ${JOB_NAME} ${basePath} --test-run-id ${env.BUILD_NUMBER}"
+                    sh "./test.sh ${JOB_NAME} manifests/${opensearch_version}/opensearch-${opensearch_version}-test.yml ${basePath} --test-run-id ${env.BUILD_NUMBER}"
                 }
             }
             post() {
```
dblock, line 66: For later: we should probably move this file under `opensearch/` since we'll need one of these for Dashboards.

src/ci_workflow/ci_test_manifest.py
```diff
@@ -0,0 +1,28 @@
...
+from ci_workflow.ci_manifest import CiManifest
+from manifests.test_manifest import TestManifest
+
+
+class CiTestManifest(CiManifest):
+    def __init__(self, file):
+        self.file = file
+
+    def __from_file(self):
+        self.manifest = TestManifest.from_file(self.file)
```
dblock, line 19: Move this into the constructor.

dblock (changes requested):
You're almost there!

Look at what's common between `CiInputManifest` and `CiTestManifest`: they both create an instance of a Manifest by calling `.from_file` on it and both take `args`. Add that into `CiManifest` constructor.

```python
class CiManifest:
  def __init__(self, manifest, args):
    self.manifest = manifest
    self.args = args
```

Your child class becomes:

```python
class CiInputManifest(CiManifest):
  def __init__(self, filename, args):
    super().__init__(self, InputManifest.from_file(filename), args)
```

The `check` method is on both classes and should be implemented on `CiManifest` for all the common parts:

```python
def check(self):
    try:
       self.__check__()
    except:
       ...
       raise
```

Finally, `__is_test_manifest` is glaring "not object oriented", but now that your `CiManifest` takes the same arguments it becomes easier, add a `CiManifests` class.

```python
class CiManifests:
    def __klass(filename):
        if re.search(...):
            return CiTestManifest
        else:
            return CiInputManifest
   
    def from_file(filename, args):
        return self.__klass(filename)(filename, args)
 ```

tianleh:
> 

Thanks for the detailed instructions! The code now looks super elegant

jenkins/test/testsuite/Jenkinsfile
tianleh: Dashboards will have a test manifest with "opensearch-dashboards". Thus sub dir is no needed due to different names.

tianleh:
Noticed "This branch has conflicts that must be resolved". Taking a look

tianleh:
Resolved merge conflict. Please help review. @dblock

dblock (approved):
Nice and clean!

PR #1236, opened 2021-12-02
Jenkins remove dry run

dblock:
### Description

This is on top of #1233, so only review the [last commit](https://github.com/opensearch-project/opensearch-build/pull/1236/commits/5e0b9923ecda656ee7842f16c1cf3b0fb6c1aa01).

I was trying to use the Build_Jenkinsfile job that includes a bunch of `dryRun: true` to test incremental build and gave up. It's a faster build, but it's confusing because we constantly have to fake input and output. Instead, I ended up creating a real build job that doesn't have notifications or cron settings, and run that in Jenkins. That also seems to be what everybody else is doing or trying to do.

Unit tests already do a good job covering actual execution, too.

This removes `dryRun` altogether. 

### Check List


peternied (approved):
Thanks for updating the documentation too!

gaiksaya (approved):

Issue #1132, opened 2021-11-23
Remove the word 'bundle' from artifact downloads link


bbarani:
### Is your feature request related to a problem? Please describe

The link to download OpenSearch artifacts contains the word 'bundle' in it instead of distribution

https://artifacts.opensearch.org/releases/bundle/opensearch/1.2.0/... 

### Describe the solution you'd like

Replace the word 'bundle' with 'distribution' and update the dependencies accordingly

https://artifacts.opensearch.org/releases/bundle/opensearch/1.2.0/...  to https://artifacts.opensearch.org/releases/distributions/opensearch/1.2.0/...

abhinavGupta16:
[triaged] This issue needs to be groomed to checkout the scope of this task.

bbarani:
Closing this issue as its more of a suggestion and not a must have feature. We can re-open it if any changes is needed on the artifact url.

Issue #814, opened 2021-10-26
[Enhancement] Use default alpine docker image to run docker image picker


peterzhuamazon:
### Is your feature request related to a problem? Please describe

https://github.com/opensearch-project/opensearch-build/commit/6bcdf1650565efadba33209f284b3806114115ce#
@peternied unblock this by using the ci-runner image to trigger docker image picker.
It works but sort of defeat the purpose of a docker image picker.

Not only the above section has this issue, in the same Jenkinsfile, the post steps are directly running on LINUX HOST not in docker container.

I suggest replace both sections with a light weight image like Alpine since this is just a few steps, no need to pull a full stack packaged 2G image for that.

Thanks.

### Describe the solution you'd like

See Above.

bbarani:
@peterzhuamazon Can you provide an update for this issue?

peterzhuamazon:
Hi @bbarani already done right here:
* https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L76

PR #1254, opened 2021-12-06
Separate check for build from distribution build workflow

peternied:
Create a new job that will only check if the SHA for the current build file has been created, if so it will trigger the distribution build job.  This should allow for build issues to be discovered more quickly and reduce the amount of no-op runs seen in the distribution-build job in jenkins.

Note; this does rename the job files and so Jenkins updates will need to be make when this is merged.

### Testing
 - Created a new build job `petern-check-for-build` and `petern-distribution-build-opensearch`, played with different states include concurrent executing, completed builds and canceled builds.

### Check List


jenkins/opensearch/check-for-build.jenkinsfile
```diff
@@ -0,0 +1,86 @@
...
+        '''
+    }
+    parameters {
+        string(
+            name: 'INPUT_MANIFEST',
+            description: 'Input manifest under the manifests folder, e.g. 2.0.0/opensearch-2.0.0.yml.',
+            trim: true
+        )
+        string(
+            name: 'JOB_NAME',
```
gaiksaya, line 21: Jenkins has an env variable called `JOB_NAME` for the current job that is running. Will this overwrite that or how will it work? Should we change it to something little less generic?

peternied: Good idea, lets go with `TARGET_JOB_NAME`

jenkins/opensearch/check-for-build.jenkinsfile
```diff
@@ -0,0 +1,86 @@
...
+            description: 'Job to trigger if build has changed',
+            trim: true
+        )
+    }
+    stages {
+        stage('detect docker image + args') {
+            agent {
+                docker {
+                    label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+                    image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028'
```
peterzhuamazon, line 31: Why is this one specifically pointing to a image but `check-sha-exists` stage doesnt?

jenkins/opensearch/check-for-build.jenkinsfile
```diff
@@ -0,0 +1,86 @@
...
+            post {
+                always {
+                    postCleanup()
+                }
+            }
+        }
+    }
+    post {
+        failure {
+            node('Jenkins-Agent-al2-x64-c54xlarge-Docker-Host') {
```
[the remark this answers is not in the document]
peternied: Lets track this separately, can you create a PR for this or file an issue?

dblock:
What’s the plan for opensearch dashboards and can it be made part of this single job or duplicate this for OSD?

jenkins/opensearch/check-for-build.jenkinsfile
peternied: This job selects the docker image for the build process, this job needs to load a specific image, this is following the standard practices for our jenkinsfiles

peternied:
> What’s the plan for opensearch dashboards and can it be made part of this single job or duplicate this for OSD?

I've created https://github.com/opensearch-project/opensearch-build/issues/1260 to unified the workflows

jenkins/opensearch/check-for-build.jenkinsfile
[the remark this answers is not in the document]
peterzhuamazon: The issue is already file a while back: [#alpine](https://github.com/opensearch-project/opensearch-build/issues/814)

peterzhuamazon:
I have doubts on whether this would run correctly.
We are running on separate docker containers and the shas will be on a new build.
This is not the same as running on your local.

dblock:
So what advantage does this have over checking for SHA within the OpenSearch/Dashboards build?

dblock:
> I have doubts on whether this would run correctly. We are running on separate docker containers and the shas will be on a new build. This is not the same as running on your local.

*If* you specify the platform and architecture in the parameters (and don't pick them up from the current arch/platform) then the SHA will match.

peternied:
> So what advantage does this have over checking for SHA within the OpenSearch/Dashboards build? Seems like now we're just doing it twice. Is it to avoid the notification?

Avoiding the notification is useful as then we can dial up the speed so breaking changes are discovered more closely in time to their introduction. 


More importantly, we are avoiding the _appearance_ of builds when nothing is occurring.  This makes the build history of the distribution build much cleaner, while the pipeline view isn't public today, we hope to make it public shortly and its hard to understand why the build time went from 12m -> 25s unless you know that the job is 'skipping' earlier in the workflow.

If you would prefer not to have the secondary job then I would rather abort or cancel the build if the checksums comeback matching.  I would also want to fix snapshot builds so they are not run when there is a build checksum match

image

dblock:
I support separating checks, but given the above conversation I think this PR needs to:

a) Iterate over desired platforms, produce locks and check all the SHAs, if any doesn't exist, kick off the build.
b) Modify the mechanism to create a lock that is not tried to a platform and architecture.

I am not sure what I would choose.

I would then also remove the lock checking from the downstream job and always build + publish a SHA.

peternied:
> > I have doubts on whether this would run correctly. We are running on separate docker containers and the shas will be on a new build. This is not the same as running on your local.
> 
> _If_ you specify the platform and architecture in the parameters (and don't pick them up from the current arch/platform) then the SHA will match.
> 
> @peterzhuamazon is right: you get different SHAs for platform/arch, so there's no "unified" SHA, there are 2 separate checks for x64/arm64 in today's build
> 
> ```
> Skipped build-arm64, bundle-build/1.3.0/shas/linux/arm64/002b5c7833efc7b6a3cfe2dc40245d913afdc9f1.yml exists.
> Skipped build-x64, bundle-build/1.3.0/shas/linux/x64/7c3e7e3ae596ffc6bc7c8c4446eaf8432e3a1d91.yml exists.
> ```

If we have a matching SHA for x64 - we can assume that if we build the arm64 we would also have a matching SHA.  Its large reduction in the complexity of the check workflow to verify one platform instead of 2

dblock:
> If we have a matching SHA for x64 - we can assume that if we build the arm64 we would also have a matching SHA. Its large reduction in the complexity of the check workflow to verify one platform instead of 2

That's just not the right assumption, x64 could have failed and arm64 could have succeeded.

vars/getManifestSHA.groovy
```diff
@@ -14,7 +15,7 @@ Map call(Map args = [:]) {
 
     def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
     def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: manifestLock))
-    String shasRoot = inputManifest.getSHAsRoot("${JOB_NAME}", args.platform, args.architecture)
+    String shasRoot = inputManifest.getSHAsRoot(jobName, args.platform, args.architecture)
```
dblock, line 18: This needs to be a new location that doesn't consider platform and architecture.

dblock (changes requested):
Check where we include platform and architecture in Jenkins shared jobs, update docs, including https://github.com/opensearch-project/opensearch-build/tree/main/vars#all-in-one-build and https://github.com/opensearch-project/opensearch-build/blob/main/src/build_workflow/README.md#avoiding-rebuilds

dblock:
I would also get rid of the incremental checks in https://github.com/opensearch-project/opensearch-build/blob/67381d35c3df9b0bd26d20cd554dde32ee9e9aef/vars/buildAssembleUpload.groovy#L3 and same for the cross-platform version.

peternied:
Still need to remove many increment vs no-increment test

build.gradle
```diff
@@ -75,11 +75,6 @@ test {
         systemProperty "pipeline.stack.write", project.getProperty("pipeline.stack.write")
     }
 
-    jacoco {
-        classDumpDir = file("$buildDir/jacoco/classpathdumps")
-    }
-
-    finalizedBy jacocoTestReport
```
dblock, line 82: Why you no like Jacoco?

dblock (changes requested):
Looks like you're having trouble with test coverage. If you `./gradlew clean` first it always works. Either way revert the change below and this is good work!

build.gradle
peternied: Doh, debugging at its finest bleeding over.

peternied:
@gaiksaya @peterzhuamazon @dblock No more outstanding work AFAIK, please take another look at your leisure

jenkins/README.md
```diff
@@ -8,6 +10,18 @@
 * [Docker](./docker)
 * [Test](./test)
 
+
+### Check for Build
+
+To ensure that builds are trigger as changes come in the [Check-for-build](check-for-build.jenkinsfile) job is setup with a cron schedule to check manifests and run them at the appropriate build job.  From in a manifest a stable build version can be created and reference against uploaded versions.  If there is an exist match the build will be skipped, if not the job will be triggered in its own build workflow.
+
+For concurrent builds the Check-for-build job has an internal locks to prevent interruption for the same manifest and target job name, any requests that match an existing lock will be skipped to prevent a backup of queued build jobs
```
dblock, line 18: Missing period.

jenkins/README.md
```diff
@@ -8,6 +10,18 @@
...
+
+#### Add jobs to run frequently
+
+* For jobs that are in active development they should be set to `H/10 * * * *` which will check every 10 minutes
```
dblock, line 22: Set cron to .... for jobs that are, otherwise it's not very English.

Missing periods.

jenkins/README.md
```diff
@@ -8,6 +10,18 @@
...
 * [Test](./test)
 
+
+### Check for Build
+
+To ensure that builds are trigger as changes come in the [Check-for-build](check-for-build.jenkinsfile) job is setup with a cron schedule to check manifests and run them at the appropriate build job.  From in a manifest a stable build version can be created and reference against uploaded versions.  If there is an exist match the build will be skipped, if not the job will be triggered in its own build workflow.
+
+For concurrent builds the Check-for-build job has an internal locks to prevent interruption for the same manifest and target job name, any requests that match an existing lock will be skipped to prevent a backup of queued build jobs
+
+#### Add jobs to run frequently
```
dblock, line 20: frequently -> periodically

dblock (approved):

dblock (approved):

jenkins/check-for-build.jenkinsfile
```diff
@@ -0,0 +1,89 @@
+lib = library(identifier: "jenkins@20211123", retriever: legacySCM(scm))
+
+pipeline {
+    agent none
+    triggers {
+        parameterizedCron '''
+            H/10 * * * * %INPUT_MANIFEST=1.3.0/opensearch-1.3.0.yml;TARGET_JOB_NAME=distribution-build-opensearch
+            H/10 * * * * %INPUT_MANIFEST=1.3.0/opensearch-dashboards-1.3.0.yml;TARGET_JOB_NAME=distribution-build-opensearch-dashboards
```
gaiksaya, line 8: Did we change `bundle-build` job name to `distribution-build-opensearch`?

jenkins/check-for-build.jenkinsfile
```diff
@@ -0,0 +1,89 @@
...
+                script {
+                    lock(resource: "CheckForBuild-${INPUT_MANIFEST}-${TARGET_JOB_NAME}", skipIfLocked: true) {
+                        def sha = getManifestSHA(jobName: "${TARGET_JOB_NAME}")
+                        if (sha.exists) {
+                            echo "Skipping, ${sha.path} already exists."
+                        } else {
+                            build job: "${TARGET_JOB_NAME}", parameters: [
+                                string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}")
+                            ], wait: true
+                            echo "Build succeeded, uploading build SHA for that job"
```
gaiksaya, line 59: Don't we want a notification message when the actual build is successful?

src/build_workflow/README.md
```diff
@@ -96,4 +96,4 @@ else
 fi
 ```
 
-The [Jenkins workflows](../../jenkins) in this repository use this mechanism to avoid rebuilding all of OpenSearch and OpenSearch Dashboards unnecessarily. 
+The [Jenkins workflows](../../jenkins) in this repository can this mechanism to avoid rebuilding all of OpenSearch and OpenSearch Dashboards unnecessarily.
```
gaiksaya, line 99: can use :)

vars/assembleUpload.groovy
```diff
@@ -5,11 +5,7 @@ void call(Map args = [:]) {
     assembleManifest(args)
     uploadArtifacts(args)
 
-    withAWS(role: 'opensearch-bundle', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
-        s3Upload(bucket: "${ARTIFACT_BUCKET_NAME}", file: args.sha.lock, path: args.sha.path)
-    }
-
-    def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: args.sha.lock))
+    def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: args.manifest))
```
gaiksaya, line 8: If we are just uploading here can we change the file/lib name to `assemble.groovy`?

jenkins/check-for-build.jenkinsfile
peternied: Seems like as good a time as any to fix the names, this will need to be 2PR'ed on the Jenkins server to re-align all of the names, I am open to other names schemes if you have thoughts.

jenkins/check-for-build.jenkinsfile
peternied: I've delegated detailed notifications for the build jobs themselves as they will have all the relevant context,  The failure-only notification on this job is to make sure that if we have a failure checking/trigger for a build job it doesn't go unnoticed which could prevent the downstream error message from appearing

jenkins/check-for-build.jenkinsfile
gaiksaya: I am okay with this name. With that said, do we need to make changes to the tests here: https://github.com/opensearch-project/opensearch-build/pull/1254/files#diff-fecdf689e56f633fe0618499df897827d0c7c8426fbc7a7fd0b9e9a656760fb7R24
Many tests contain `bundle-build` as their job name param

jenkins/check-for-build.jenkinsfile
gaiksaya: Ahh! We have notification for build job. Forgot that, Thanks for reminding!

jenkins/check-for-build.jenkinsfile
peterzhuamazon: A lot of the hardcoding and even the current 1.1.1 are linked to bundle-build as a name on S3.
Adding this change would cause a lot of build failures.

peternied: Darn, lets follow up on a seperate PR to that removes the word bundle entirely

gaiksaya: Agreed! Thanks

vars/assembleUpload.groovy
peternied: This file stills `uploads` see line 6, so the name shouldn't change.  It just doesn't upload the lockfile anymore.

peternied: This file stills `uploads` see line 6, so the name shouldn't change.  It no longer uploads lockfile.

jenkins/check-for-build.jenkinsfile
peternied: > A lot of the hardcoding and even the current 1.1.1 are linked to bundle-build as a name on S3.
Adding this change would cause a lot of build failures.

@peterzhuamazon This does not update any of the previous manifests or artifacts, those will stay named exactly as they are, this will change the names / paths for new builds which shouldn't have the work 'bundle' in them per https://github.com/opensearch-project/opensearch-build/issues/1132

gaiksaya: Are you changing the `TARGET_JOB_NAME` back to `bundle-build` since now renaming will be done entirely in new PR?

peternied: Renaming will happen entirely on Jenkins, no new PRs will be created.  This change requires a deployment step after it has merged to create the new job and rename the old ones

vars/README.md
```diff
@@ -30,6 +31,10 @@ A cross platform build divides the all-in-one build into steps that can be execu
 
 See [buildDockerImage](./buildDockerImage.groovy).
 
+### Avoiding rebuilding
```
dblock, line 34: Avoiding Rebuilds (capital)

vars/README.md
```diff
@@ -30,6 +31,10 @@ A cross platform build divides the all-in-one build into steps that can be execu
...
+A [SHA of the manifest lock](../src/build_workflow/README.md#avoiding-rebuilds) can be generated using [getManifestSHA](./getManifestSHA.groovy), and checking it against previously published artifacts.  If a build is consider released, its manifestSha can be uploaded with [buildAndUploadManifestSHA](./buildAndUploadManifestSHA.groovy).
```
dblock, line 35: manifestSha -> manifest SHA

vars/buildArchive.groovy
```diff
@@ -1,40 +1,10 @@
...
-
-        def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: sha.lock))
-        echo "Setting env.BUILD_SHA_${inputManifest.build.platform}_${inputManifest.build.architecture}_SHA=${sha.sha}"
-        env."BUILD_SHA_${inputManifest.build.platform}_${inputManifest.build.architecture}_SHA" = sha.sha
-        echo "Setting env.BUILD_SHA_${inputManifest.build.platform}_${inputManifest.build.architecture}_LOCK=${sha.lock}"
-        env."BUILD_SHA_${inputManifest.build.platform}_${inputManifest.build.architecture}_LOCK" = sha.lock
-        echo "Setting env.BUILD_SHA_${inputManifest.build.platform}_${inputManifest.build.architecture}_PATH=${sha.path}"
-        env."BUILD_SHA_${inputManifest.build.platform}_${inputManifest.build.architecture}_PATH" = sha.path
-    }
+    def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: args.manifest))
```
dblock, line 9: No longer needed?

dblock (changes requested):

gaiksaya (approved):

vars/buildArchive.groovy
peternied: No, and thanks for calling this out, it looks like the dashboard build might be busted with this change, working on it...

peternied:
I'm still working on this, I'm adding a way to regression test our main build jobs so we can be sure if we are breaking functionality or not, trial and error takes a long time.

jenkins/README.md
```diff
@@ -8,6 +10,18 @@
...
 * [Test](./test)
 
+
+### Check for Build
+
+To ensure that builds are trigger as changes come in the [Check-for-build](check-for-build.jenkinsfile) job is setup with a cron schedule to check manifests and run them at the appropriate build job.  From in a manifest a stable build version can be created and reference against uploaded versions.  If there is an exist match the build will be skipped, if not the job will be triggered in its own build workflow.
+
+For concurrent builds the Check-for-build job has an internal locks to prevent interruption for the same manifest and target job name, any requests that match an existing lock will be skipped to prevent a backup of queued build jobs.
+
+#### Add jobs to run periodically
```
dblock, line 20: CaPs. or just as a regular paragraph.

vars/buildAndUploadManifestSHA.groovy
```diff
@@ -0,0 +1,9 @@
+void call(Map args = [:]) {
+    lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+
+    def sha = getManifestSHA(args)
+
+    withAWS(role: 'opensearch-bundle', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+        s3Upload(bucket: "${ARTIFACT_BUCKET_NAME}", file: sha.lock, path: sha.path)
+    }
+}
```
dblock, line 9: We don't include `And` in the file name elsewhere.

vars/buildArchive.groovy
```diff
@@ -1,40 +1,11 @@
...
-        echo "Setting env.BUILD_SHA_${inputManifest.build.platform}_${inputManifest.build.architecture}_PATH=${sha.path}"
-        env."BUILD_SHA_${inputManifest.build.platform}_${inputManifest.build.architecture}_PATH" = sha.path
-    }
+    buildManifest(args)
+
+    echo "Archiving into zip: builds/**, ${args.manifest}"
+    zip(
+        zipFile: "archived-builds.zip",
+        archive: true,
+        glob: 'builds/**'
```
dblock, line 9: Where do we make `archived-builds.zip`? 

These archives are used to pass output between different steps of Dashboards build to do a cross-platform build.

dblock (changes requested):

vars/buildArchive.groovy
peternied: This is where we make those artifacts, this function call is a little funky

tests/test_run_build.py
```diff
@@ -158,6 +157,6 @@ def test_main_manifest_new_lock(self, mock_logging, mock_to_file, mock_stable, *
     def test_main_manifest_new_lock_with_overrides(self, mock_logging, mock_to_file, mock_stable, *mocks):
         with self.assertRaises(SystemExit):
             main()
-        mock_stable.assert_called_with(architecture="arm64", platform="windows", snapshot=True)
+        mock_stable.assert_called_with()
```
dblock, line 160: not: `assert_called()`

dblock (approved):

Issue #1259, opened 2021-12-06
Run Dashboards integ test from public endpoint


tianleh:
### Is your feature request related to a problem? Please describe

#604

### Describe the solution you'd like

It will need to know where OpenSearch is.

Potential solutions:

1) use a single public endpoint for both OS and OSD

2) provide two parameters to integ test where one for OS and one for OSD

tianleh:
Looks like we may need to extend the test args to support an optional path for dashboards.

https://github.com/opensearch-project/opensearch-build/blob/1ea1896f1979cfa9471ad7cb74eb9008173ab562/src/test_workflow/test_args.py#L27

tianleh:
@dblock @peterzhuamazon @seraphjiang

PR #1263, opened 2021-12-06
Support separate paths for OpenSearch and OpenSearch Dashboards dur…

tianleh:
…ing integ test

### Description
The CI public endpoint is different for OpenSearch and OpenSearch Dashboards. Thus add a new param for specifying such for OpenSearch Dashboards
 
### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/1259
 
### Test
1. unit test in progress.
2. manual test in progress.

### Check List

### DIFF
--- a/src/test_workflow/README.md
+++ b/src/test_workflow/README.md
@@ -27,7 +27,7 @@ The following options are available.
 |----------------------|-------------------------------------------------------------------------|
 | test-type            | Run tests of a test suite. [integ-test, bwc-test, perf-test]            |
 | test-manifest-path   | Specify a test manifest path                                            |
-| path                 | Location of manifest(s).                                                |
+| --paths              | Location of manifest(s)                                                 |
 | --test-run-id        | Unique identifier for a test run                                        |
 | --component          | Test a specific component in a manifest                                 |
 | --keep               | Do not delete the temporary working directory on both success or error. |
@@ -56,7 +56,14 @@ For example, build locally and run integration tests.
 Or run integration tests against an existing build.
 
 ```bash
-./test.sh integ-test manifests/1.3.0/opensearch-1.3.0-test.yml https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/869/linux/x64 # looks for https://.../builds/opensearch/manifest.yml and https://.../dist/opensearch/manifest.yml
+./test.sh integ-test manifests/1.3.0/opensearch-1.3.0-test.yml --paths opensearch=https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/869/linux/x64 # looks for https://.../builds/opensearch/manifest.yml and https://.../dist/opensearch/manifest.yml
+```
+
+To run OpenSearch Dashboards integ tests
+
+```bash
+./test.sh integ-test manifests/1.3.0/opensearch-1.3.0-test.yml --paths opensearch=https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/869/linux/x64
+opensearch-dashboards=https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/1.2.0/869/linux/x64 
 ```
 
 ### Backwards Compatibility Tests
--- a/src/test_workflow/README.md
+++ b/src/test_workflow/README.md
@@ -62,7 +62,7 @@ Or run integration tests against an existing build.
 To run OpenSearch Dashboards integ tests
 
 ```bash
-./test.sh integ-test manifests/1.3.0/opensearch-1.3.0-test.yml --paths opensearch=https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/869/linux/x64
+./test.sh integ-test manifests/1.3.0/opensearch-dashboards-1.3.0-test.yml --paths opensearch=https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/869/linux/x64
 opensearch-dashboards=https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/1.2.0/869/linux/x64 
 ```
 
--- a/src/test_workflow/README.md
+++ b/src/test_workflow/README.md
@@ -26,10 +26,10 @@ The following options are available.
 | name                 | description                                                             |
 |----------------------|-------------------------------------------------------------------------|
 | test-type            | Run tests of a test suite. [integ-test, bwc-test, perf-test]            |
-| test-manifest-path   | Specify a test manifest path                                            |
-| --paths              | Location of manifest(s)                                                 |
-| --test-run-id        | Unique identifier for a test run                                        |
-| --component          | Test a specific component in a manifest                                 |
+| test-manifest-path   | Specify a test manifest path.                                           |
+| --paths              | Location of manifest(s).                                                |
+| --test-run-id        | Unique identifier for a test run.                                       |
+| --component          | Test a specific component in a manifest.                                |
 | --keep               | Do not delete the temporary working directory on both success or error. |
 | -v, --verbose        | Show more verbose output.                                               |
 
@@ -59,7 +59,7 @@ Or run integration tests against an existing build.
 ./test.sh integ-test manifests/1.3.0/opensearch-1.3.0-test.yml --paths opensearch=https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/869/linux/x64 # looks for https://.../builds/opensearch/manifest.yml and https://.../dist/opensearch/manifest.yml
 ```
 
-To run OpenSearch Dashboards integ tests
+To run OpenSearch Dashboards integration tests.
 
 ```bash
 ./test.sh integ-test manifests/1.3.0/opensearch-dashboards-1.3.0-test.yml --paths opensearch=https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/869/linux/x64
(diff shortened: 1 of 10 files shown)


tianleh:
Tests in progress. Send for early review on the functional part.

tianleh:
During local test, 

```
./test.sh integ-test manifests/1.3.0/opensearch-1.3.0-test.yml https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/1235/linux/x64 https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/1.2.0/528/linux/x64
```

find a bug #1264 where the bundle manifest from the CI endpoint contains wrong location for the bundle artifact. A "opensearch" subdir is missing inside the manifest content.

tianleh:
```
2021-12-07 00:09:16 INFO     Downloading https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/1235/linux/x64/dist/opensearch-1.2.0-linux-x64.tar.gz into /tmp/tmp4roe61sb/local-test-cluster/opensearch-1.2.0-linux-x64.tar.gz ...
```

The link above is wrong.

dblock (approved):
Approve as a step forward, but I think in this case we should not make both arguments positional because they have defaults (`.`), so I should be able to run with an OpenSearch from ci.opensearch.org and a local dashboards without having to specify `.` for the former. So I would want them to be `--opensearch-path` and `--opensearch-dashboards-path`?

But maybe we can do better and future proof this? Instead of

```
./test.sh test.yml --opensearch-path <path> --opensearch-dashboards-path <path>
```

would we prefer

```
./test.sh test.yml --paths opensearch=<path> opensearch-dashboards=<path>
```

This makes it possible to extend to all kinds of other paths under test in the future. There's an example of how to do this in https://sumit-ghosh.com/articles/parsing-dictionary-key-value-pairs-kwargs-argparse-python/.

sichend (approved):

tianleh:
> so I should be able to run with an OpenSearch from ci.opensearch.org and a local dashboards without having to specify `.` for the former.

Do you mean the latter instead of former?

dblock:
> > so I should be able to run with an OpenSearch from ci.opensearch.org and a local dashboards without having to specify `.` for the former.
> 
> Do you mean the latter instead of former?

Yes :)

tianleh:
Related bug fix https://github.com/opensearch-project/opensearch-build/pull/1265 has been merged. Will rebase this PR to latest code. Also once next bundle build job is done, we will be able to test the args change end to end.

tianleh:
Just rebased latest changes from remote

tianleh:
> This makes it possible to extend to all kinds of other paths under test or settings in the future. There's an example of how to do this in https://sumit-ghosh.com/articles/parsing-dictionary-key-value-pairs-kwargs-argparse-python/.

Implemented this Kwargs approach.

tianleh:
Performed local test with this command.

```
./test.sh integ-test manifests/1.3.0/opensearch-dashboards-1.3.0-test.yml -p opensearch=https://ci.opensearch.org/ci/dbc/bundle-build/2.0.0/1346/linux/x64 opensearch-dashboards=https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/2.0.0/638/linux/x64
```

I can see that the artifacts are downloaded with correct URL

```
2021-12-09 08:01:50 INFO     Loading https://ci.opensearch.org/ci/dbc/bundle-build/2.0.0/1346/linux/x64/dist/opensearch/manifest.yml
2021-12-09 08:01:50 INFO     Loading https://ci.opensearch.org/ci/dbc/bundle-build/2.0.0/1346/linux/x64/builds/opensearch/manifest.yml
2021-12-09 08:01:51 INFO     Loading https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/2.0.0/638/linux/x64/dist/opensearch-dashboards/manifest.yml
2021-12-09 08:01:51 INFO     Loading https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/2.0.0/638/linux/x64/builds/opensearch-dashboards/manifest.yml
```

Thus the logic related to the changes in the PR is verified. The integ test itself is failed because 2.0.0 only has min/Core artifacts which is not bundled and thus has no security related components. Some more follow up tests may be needed to trigger a bundle build for OS and OSD 1.2.0 which are fully bundled.

tianleh:
Running against local builds/dist (bundled) is simplified to be

`./test.sh integ-test manifests/1.3.0/opensearch-dashboards-1.3.0-test.yml
`
The tests are running successfully.

```
2021-12-09 08:14:58 INFO     | functionalTestDashboards | with-security        | PASS  |    0 |
2021-12-09 08:14:58 INFO     | functionalTestDashboards | without-security     | PASS  |    0 |
```

tianleh:
Tested a hybrid mode
`./test.sh integ-test manifests/1.3.0/opensearch-dashboards-1.3.0-test.yml -p opensearch=https://ci.opensearch.org/ci/dbc/bundle-build/2.0.0/1346/linux/x64`

Verified that the target paths are correct.

```
2021-12-09 08:20:14 INFO     Loading https://ci.opensearch.org/ci/dbc/bundle-build/2.0.0/1346/linux/x64/dist/opensearch/manifest.yml
2021-12-09 08:20:14 INFO     Loading https://ci.opensearch.org/ci/dbc/bundle-build/2.0.0/1346/linux/x64/builds/opensearch/manifest.yml
2021-12-09 08:20:14 INFO     Loading /usr/share/opensearch/workspace/opensearch-build/dist/opensearch-dashboards/manifest.yml
2021-12-09 08:20:14 INFO     Loading /usr/share/opensearch/workspace/opensearch-build/builds/opensearch-dashboards/manifest.yml
```

tianleh:
Fixed tests and updated docs as well. Please help take a look again. @dblock

src/test_workflow/README.md
```diff
@@ -27,7 +27,7 @@ The following options are available.
 |----------------------|-------------------------------------------------------------------------|
 | test-type            | Run tests of a test suite. [integ-test, bwc-test, perf-test]            |
 | test-manifest-path   | Specify a test manifest path                                            |
-| path                 | Location of manifest(s).                                                |
+| --paths              | Location of manifest(s)                                                 |
```
dblock, line 30: We should have periods in every line here.

src/test_workflow/README.md
```diff
@@ -56,7 +56,14 @@ For example, build locally and run integration tests.
 Or run integration tests against an existing build.
 
 ```bash
-./test.sh integ-test manifests/1.3.0/opensearch-1.3.0-test.yml https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/869/linux/x64 # looks for https://.../builds/opensearch/manifest.yml and https://.../dist/opensearch/manifest.yml
+./test.sh integ-test manifests/1.3.0/opensearch-1.3.0-test.yml --paths opensearch=https://ci.opensearch.org/ci/dbc/bundle-build/1.2.0/869/linux/x64 # looks for https://.../builds/opensearch/manifest.yml and https://.../dist/opensearch/manifest.yml
+```
+
+To run OpenSearch Dashboards integ tests
```
dblock, line 62: Add a period.
integ -> integration

src/test_workflow/test_args.py
```diff
@@ -36,9 +38,21 @@ def __init__(self):
         self.component = args.component
         self.keep = args.keep
         self.logging_level = args.logging_level
-        self.path = args.path if validators.url(args.path) else os.path.realpath(args.path)
+        self.test_manifest_path = self._validate_path(args.test_manifest_path)
+
+        self.opensearch_path = self._validate_paths(args.paths, "opensearch")
+        self.opensearch_dashboards_path = self._validate_paths(args.paths, "opensearch-dashboards")
+
+    def _validate_paths(self, args_paths: dict, name: str) -> str:
```
dblock, line 46: `_validate_paths` is really an existing getter with default: `self.paths.get('opensearch-dashboards', '.')`, I would remove that method

there's a way to move the call of `_validate_path` to the `parser` as well, but I forget how to do that, NBD

src/test_workflow/test_kwargs.py
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: Apache-2.0
```
dblock, line 1: Filename should match class. Add a test file, even if you don't implement a test (you should).

dblock (approved):
Have some should have's, but nothing blocking merging. Nice effort on Kwargs!

src/test_workflow/test_args.py
```diff
@@ -36,9 +38,21 @@ def __init__(self):
         self.component = args.component
         self.keep = args.keep
         self.logging_level = args.logging_level
-        self.path = args.path if validators.url(args.path) else os.path.realpath(args.path)
+        self.test_manifest_path = self._validate_path(args.test_manifest_path)
+
+        self.opensearch_path = self._validate_paths(args.paths, "opensearch")
+        self.opensearch_dashboards_path = self._validate_paths(args.paths, "opensearch-dashboards")
```
peternied, line 44: When running OpenSearch distribution tests the *dashboards_path tests is not needed, could you abstract this validation so its based on a test manifest settings?

src/test_workflow/test_kwargs.py
tianleh: Fixed the name and added tests.

src/test_workflow/test_args.py
tianleh: will have follow ups on this.

src/test_workflow/test_args.py
tianleh: This will require loading manifest content. Seems a bit heavy weight for args parsing stage. We can have a follow up item on this if you prefer such checks.

peternied: Please create a follow up issue

tianleh: https://github.com/opensearch-project/opensearch-build/issues/1276

Issue #1264, opened 2021-12-07
[Bug]: The locations inside Bundle Manifest are missing the product in the path


tianleh:
### Describe the bug

Use this bundle manifest as an example
https://ci.opensearch.org/ci/dbc/bundle-build/1.3.0/1318/linux/x64/dist/opensearch/manifest.yml

It shows 

```
  location: https://ci.opensearch.org/ci/dbc/bundle-build/1.3.0/1318/linux/x64/dist/opensearch-1.3.0-linux-x64.tar.gz

```

However, it shall have a `opensearch` before the artifact

### To reproduce

https://ci.opensearch.org/ci/dbc/bundle-build/1.3.0/1318/linux/x64/dist/opensearch/manifest.yml

tianleh:
This happens to OpenSearch Dashboards as well.

tianleh:
But my local build and assemble are working correctly. Need to check further into the bundle logic.

tianleh:
I am able to reproduce this issue by providing a base url to assemble

```
./assemble.sh "builds/opensearch/manifest.yml" --base-url https://ci.opensearch.org/ci/dbc/bundle-build/1.3.0/1318/linux/x64
```

PR #1265, opened 2021-12-07
Update public url to include filename

tianleh:
### Description
The public url has the product as sub dir. Thus we need to include it to access the content.
 
### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/1264
 
### Check List

### DIFF
--- a/src/assemble_workflow/bundle_recorder.py
+++ b/src/assemble_workflow/bundle_recorder.py
@@ -4,7 +4,6 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
-import logging
 import os
 from typing import Any, Dict
 
@@ -58,8 +57,6 @@ class BundleRecorder:
             component.commit_id,
             self.__get_component_location(rel_path),
         )
-
-        logging.info(f"record_component has {self.__get_component_location(rel_path)}")
 
     def get_manifest(self) -> BundleManifest:
         return self.bundle_manifest.to_manifest()
--- a/src/assemble_workflow/bundle_recorder.py
+++ b/src/assemble_workflow/bundle_recorder.py
@@ -4,6 +4,7 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+import logging
 import os
 from typing import Any, Dict
 
@@ -57,6 +58,8 @@ class BundleRecorder:
             component.commit_id,
             self.__get_component_location(rel_path),
         )
+
+        logging.info(f"record_component has {self.__get_component_location(rel_path)}")
 
     def get_manifest(self) -> BundleManifest:
         return self.bundle_manifest.to_manifest()
--- a/src/assemble_workflow/bundle_recorder.py
+++ b/src/assemble_workflow/bundle_recorder.py
@@ -14,6 +14,7 @@ from manifests.bundle_manifest import BundleManifest
 
 class BundleRecorder:
     def __init__(self, build: BuildManifest.Build, output_dir: str, artifacts_dir: str, base_url: str) -> None:
+        self.product = build.name.lower().replace(" ", "-")
         self.output_dir = output_dir
         self.build_id = build.id
         self.base_url = base_url
@@ -40,7 +41,7 @@ class BundleRecorder:
         return "-".join(parts) + (".zip" if build.platform == "windows" else ".tar.gz")
 
     def __get_public_url_path(self, folder: str, rel_path: str) -> str:
-        path = "/".join((folder, rel_path))
+        path = "/".join((folder, self.product, rel_path))
         return urljoin(self.base_url + "/", path)
 
     def __get_location(self, folder_name: str, rel_path: str, abs_path: str) -> str:
--- a/src/assemble_workflow/bundle_recorder.py
+++ b/src/assemble_workflow/bundle_recorder.py
@@ -14,7 +14,6 @@ from manifests.bundle_manifest import BundleManifest
 
 class BundleRecorder:
     def __init__(self, build: BuildManifest.Build, output_dir: str, artifacts_dir: str, base_url: str) -> None:
-        self.product = build.name.lower().replace(" ", "-")
         self.output_dir = output_dir
         self.build_id = build.id
         self.base_url = base_url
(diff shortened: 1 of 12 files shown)


seraphjiang (approved):

peternied (commented):
> Thus we need to include it to access the content.

Please provide background for this change, its a breaking change to the url schema

peternied (changes requested):
Why isn't the `job-name-opensearch` part of the url sufficiently unique?

dblock (changes requested):
This isn't the product or product name (that would be `OpenSearch` or `OpenSearch Dashboards`). I think it's the build folder or file name. Add a property `build.filename` and call the field in `BundleRecorder` `build_filename` to be consistent with `build_id`. See if there are other places we do `name.lower` and replace those too.

dblock:
> Why isn't the `job-name-opensearch` part of the url sufficiently unique?

If I look at something like `bundle-build` in S3 it's hard to know what that really is I guess. And also think it's good not to rely on the job name in case we move/combine things?

tianleh:
> Please provide background for this change, its a breaking change to the url schema

This is the bug issue. https://github.com/opensearch-project/opensearch-build/issues/1264

peternied:
Thanks for the bug link, I would prefer to see is that there is a directory structure on disk and after the build/assembly process that structure is exactly the same as what is uploaded to s3.  With this change it looks like only the s3 pathing is changing but not the local paths, were the local paths already changed - hence this bug?

tianleh:
> Thanks for the bug link, I would prefer to see is that there is a directory structure on disk and after the build/assembly process that structure is exactly the same as what is uploaded to s3. With this change it looks like only the s3 pathing is changing but not the local paths, were the local paths already changed - hence this bug?

This is how the disk structure:
-builds
  -opensearch
  -opensearch-dashboards
-dist
  -opensearch
  -opensearch-dashboards

This is same as the S3 structure. The current issue is that the location data inside bundle manifest yml is wrong (missing the sub dir). 

The reason why disk bundle is working is that the disk output is updated properly here https://github.com/opensearch-project/opensearch-build/blob/main/src/run_assemble.py#L30

However, the output logic for endpoint is handled separately inside https://github.com/opensearch-project/opensearch-build/blob/c7049633b1f6727f2f8fae7292949ca353b6ae37/src/assemble_workflow/bundle_recorder.py#L47

This reminds me that we may further extend @dblock  's suggestion as follows:
1) add `filename` to `build` manifest so that bundle process no longer needs to worry about inferring such from build name

2) unify the output logic to handle both file and url. https://github.com/opensearch-project/opensearch-build/blob/c7049633b1f6727f2f8fae7292949ca353b6ae37/src/run_assemble.py#L30 Thus base url will be passed into this. Bundle Recorder no longer needs the base_url https://github.com/opensearch-project/opensearch-build/blob/c7049633b1f6727f2f8fae7292949ca353b6ae37/src/assemble_workflow/bundle_recorder.py#L19

src/assemble_workflow/bundle_recorder.py
```diff
@@ -32,15 +33,15 @@ def __init__(self, build: BuildManifest.Build, output_dir: str, artifacts_dir: s
...
+            self.product,
             build.version,
             build.platform,
             build.architecture,
         ]
         return "-".join(parts) + (".zip" if build.platform == "windows" else ".tar.gz")
 
     def __get_public_url_path(self, folder: str, rel_path: str) -> str:
-        path = "/".join((folder, rel_path))
+        path = "/".join((folder, self.product, rel_path))
```
peternied, line 44: Not sure this is the right place for this change, when the relative path comes in at line 61, shouldn't it match the path from the local file system?

peternied: I would hope making by making the `record(...)` call simpler, it would eliminate the need for additional logic

tianleh:
Created a separate PR to handle the new attribute in build manifest to reduce potential size of this PR https://github.com/opensearch-project/opensearch-build/pull/1270

src/assemble_workflow/bundle_recorder.py
tianleh: This change is intended to match the local file system. `self.product` is the sub dir `opensearch` or `opensearch-dashboards`.

Once related PR for `build.filename` https://github.com/opensearch-project/opensearch-build/pull/1270 is merged, I will do some refactoring to make the `record_component` logic simple.

tianleh: The existing code is indeed hard to understand. I have to experiment with many logging statements to fully understand the logic flow. To summarize the confusion points, there are two layers:

1. Disk assemble path logic is handled inside run_assemble https://github.com/opensearch-project/opensearch-build/blob/main/src/run_assemble.py#L28 As a contrast, endpoint assemble path logic is handled inside bundle recorder https://github.com/opensearch-project/opensearch-build/blob/f83f53c7ce7b582b2871cceb1c3a52bc08fad9c5/src/assemble_workflow/bundle_recorder.py#L42
2. Regardless of disk or endpoint assemble, there is always a local disk output saved. Such output could be confusing since both disk and endpoint assemble have it.

I already have a POC in local branch which can encapsulate the complexity mentioned above. Once #1270 is merged, I will eliminate the confusion.

dblock: Great stuff @tianleh !

tianleh:
Related PR https://github.com/opensearch-project/opensearch-build/pull/1270 has been merged, rebasing this branch first to get latest change.

src/assemble_workflow/bundle_recorder.py
```diff
@@ -14,6 +14,7 @@
 
 class BundleRecorder:
     def __init__(self, build: BuildManifest.Build, output_dir: str, artifacts_dir: str, base_url: str) -> None:
+        self.product = build.name.lower().replace(" ", "-")
```
dblock, line 17: Rename to build_filename and use `build.filename`.

tianleh:
Introduce a `BundleLocation` to unify the logic for disk and endpoint bundle logic. Verified locally that location information is correctly set in the bundle manifest for both cases.

Send for review on the functional part. Unit test is in progress.

dblock:
👍

peternied:
Magnificent!

dblock (approved):

peternied (approved):

PR #1181, opened 2021-11-29
Add WhiteSource CVEs scan to Github workflow

zelinh:
### Description
Refer to #1157, WhiteSource integration may not work correctly at this time. They have an issue on their side. We are switching to our Plan B using Github Actions for automating WhiteSource scan using old [script](https://github.com/opendistro-for-elasticsearch/opendistro-build/blob/main/.github/workflows/check-vulnerability-whitesource.yml) from ODFE. This workflow will update the scan results in our testing product on WhiteSource dashboard first so that it won't overwrite the previous manual scan results that team tracked. 
 
### Issues Resolved
Backup plan for #1157 
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #1157, opened 2021-11-24
[Bug]: WhiteSource integration is not working correctly


zelinh:
### Describe the bug

Refer to https://github.com/opensearch-project/OpenSearch/issues/1593
Currently there is an epic issue on WhiteSource side regarding of their version manager of their scanner in WhiteSource Github integration app. This issue causes WhiteSource can't scan OpenSearch repository correctly. 
We will continue work with WhiteSource to get this problem resolved. 
At this time, OpenSearch core team can still squash and merge their PRs disregarding the WhiteSource check to unblock the progress. 

### To reproduce

N/A

bbarani:
FYI... Whitesoure Native integration will support JDK 11 starting Feb 06 2022 - More info - https://whitesource.atlassian.net/wiki/spaces/WD/pages/817037652/Developer+Integrations+Release+Notes#Version-22.1.2-(6-February-2022)

bbarani:
@zelinh Is this still a valid issue? 1.x branch is currently downgraded to JDK 11 and I assume Whitesource should be able to open issues automatically?

zelinh:
I think we can close this issue for now as there is nothing need to fix from our side. Thanks.

PR #1268, opened 2021-12-07
Create a Jenkinsfile to run scan on OpenSearch-projects repos

zelinh:

### Description

Since #1181 is not working because of GHA. We will now use Jenkins to automate WhiteSource scan. All repos included in `wss-scan.config` are tested to be running properly in Jenkins. "OpenSearch" is not included since we have encountered some issues scanning it within docker container. 

Refer to #1157, WhiteSource integration may not work correctly at this time. They have an issue on their side. We are switching to our Plan B using Jenkins for automating WhiteSource scan based on old script from ODFE. This workflow will update the scan results in main product "OpenSearch" on WhiteSource dashboard.

### Issues Resolved
part of #1157 
 
### Check List

### DIFF
--- a/tools/vulnerability-scan/wss-scan.sh
+++ b/tools/vulnerability-scan/wss-scan.sh
@@ -32,8 +32,6 @@ if [ ! -f "wss-unified-agent.jar" ]
 then
   # Download the WhiteSource Agent 
   curl https://unified-agent.s3.amazonaws.com/wss-unified-agent-21.11.1.jar --output wss-unified-agent.jar
-  # The version 21.11.1 has been tested and can be used if a specific version is required
-  #wget -q  https://github.com/whitesource/unified-agent-distribution/releases/download/v21.11.1/wss-unified-agent.jar
 fi
 
 # scan the config file for the user configurations
--- a/tools/vulnerability-scan/wss-scan.sh
+++ b/tools/vulnerability-scan/wss-scan.sh
@@ -31,7 +31,7 @@ fi
 if [ ! -f "wss-unified-agent.jar" ]
 then
   # Download the WhiteSource Agent 
-  curl https://unified-agent.s3.amazonaws.com/wss-unified-agent.jar --output wss-unified-agent.jar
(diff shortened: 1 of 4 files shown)


dblock (approved):

dblock:
What do you think about moving the list of repos to scan into Jenkinsfile and use the list to generate stages that are then executed in parallel?

zelinh:
> What do you think about moving the list of repos to scan into Jenkinsfile and use the list to generate stages that are then executed in parallel? (e.g. kicks off a job with the argument of repo to scan, then waits for all the started jobs, or maybe there's a way to create stages dynamically).

The reason I keep the list of repos in `wss-scan.config` is that I want to make minimum change to what we had in our repo. But I do like the idea that we have this list in Jenkinsfile as stages; passing the repo as argument or just use the default list seems to be a good idea for other people to kickoff scan. Maybe we could create an issue for this feature as enhancement?

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -31,14 +31,16 @@ fi
 if [ ! -f "wss-unified-agent.jar" ]
 then
   # Download the WhiteSource Agent 
-  wget -q https://github.com/whitesource/unified-agent-distribution/releases/latest/download/wss-unified-agent.jar
-  # The version 20.9.2.1 has been tested and can be used if a specific version is required
-  #wget -q  https://github.com/whitesource/unified-agent-distribution/releases/download/v20.9.2.1/wss-unified-agent.jar
+  curl https://unified-agent.s3.amazonaws.com/wss-unified-agent.jar --output wss-unified-agent.jar
+  # The version 21.11.1 has been tested and can be used if a specific version is required
```
peternied, line 35: Can we lock to a version?  I would rather we intentional rev the version and see something break then get surprised when we start downloading a new major release

zelinh: Changed. We will now stick with unified agent v21.11.1 until further change.

peternied (approved):

Issue #1269, opened 2021-12-08
Add new attribute filename to build manifest class property


tianleh:
### Is your feature request related to a problem? Please describe

#604 

https://github.com/opensearch-project/opensearch-build/pull/1265/files

### Describe the solution you'd like

add this attribute

PR #1270, opened 2021-12-08
Add new attribute filename to build manifest class property

tianleh:

### Description
During PR review of https://github.com/opensearch-project/opensearch-build/pull/1265, @dblock has the suggestion to introduce a new attribute `filename` into build manifest.
 
### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/1269
 
### Test
Snippet of the new OpenSearch build manifest
```
build:
  architecture: x64
  filename: opensearch
  id: 19bfc8c9ca4a45a6af22fcffc69c4ac5
  name: OpenSearch
  platform: linux
  version: 1.2.0

```
Snippet of the new OpenSearch Dashboards build manifest
```
build:
  architecture: x64
  filename: opensearch-dashboards
  id: 25ea85b70189487abff3eaaf8c463333
  name: OpenSearch Dashboards
  platform: linux
  version: 1.2.0
```

Verify that bundle manifest has the correct location for local assemble.

### Check List

### DIFF
--- a/src/paths/build_output_dir.py
+++ b/src/paths/build_output_dir.py
@@ -8,5 +8,5 @@ from paths.output_dir import OutputDir
 
 
 class BuildOutputDir(OutputDir):
-    def __init__(cls, name, cwd=None, makedirs=True):
-        super().__init__("builds", name.lower().replace(" ", "-"), cwd, makedirs)
+    def __init__(cls, filename, cwd=None, makedirs=True):
+        super().__init__("builds", filename, cwd, makedirs)
--- a/src/paths/build_output_dir.py
+++ b/src/paths/build_output_dir.py
@@ -9,4 +9,4 @@ from paths.output_dir import OutputDir
 
 class BuildOutputDir(OutputDir):
     def __init__(cls, name, cwd=None, makedirs=True):
-        super().__init__("builds", name, cwd, makedirs)
+        super().__init__("builds", name.lower().replace(" ", "-"), cwd, makedirs)
--- a/src/assemble_workflow/bundle_recorder.py
+++ b/src/assemble_workflow/bundle_recorder.py
@@ -32,7 +32,7 @@ class BundleRecorder:
 
     def __get_package_name(self, build: BuildManifest.Build) -> str:
         parts = [
-            build.name.lower().replace(" ", "-"),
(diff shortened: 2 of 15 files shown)


seraphjiang (approved):

peternied (changes requested):
Looks like it is time to create a 'live' schema version 1.3 since we are adding a required parameter.  OpenSearch/OpenSearch Dashboard releases manifests on our artifacts servers do not include this value and cannot be updated.

dblock (changes requested):
If we're adding this to the schema we have to update the schema version, make the previous version loadable, or make the field optional. I suggest reverting the schema changes and just leaving the dynamic `.filename` method.

tianleh:
> If we're adding this to the schema we have to update the schema version, make the previous version loadable, or make the field optional. I suggest reverting the schema changes and just leaving the dynamic `.filename` method.

Sure. Switched to add a `.filename` method without touching schema.

tianleh:
> Looks like it is time to create a 'live' schema version 1.3 since we are adding a required parameter. OpenSearch/OpenSearch Dashboard releases manifests on our artifacts servers do not include this value and cannot be updated.

Will not touch the schema per db's suggestion.

src/paths/build_output_dir.py
```diff
@@ -9,4 +9,4 @@
 
 class BuildOutputDir(OutputDir):
     def __init__(cls, name, cwd=None, makedirs=True):
-        super().__init__("builds", name, cwd, makedirs)
+        super().__init__("builds", name.lower().replace(" ", "-"), cwd, makedirs)
```
dblock, line 12: Did you also mean to change this?

tianleh: No. This is from InputManifest https://github.com/opensearch-project/opensearch-build/blob/main/src/run_build.py#L27 which do not have the filename property.

dblock: Should we add it?

dblock (approved):

src/paths/build_output_dir.py
tianleh: Added. Now the logic for build and assemble look very unified

Issue #606, opened 2021-09-27
[Bug]: OpenSearch Dashboards needs to rebuild core every time 


dblock:
### Describe the bug

OpenSearch dashboards requires to rebuild core every single time to build any of the components, which makes it very slow to debug component build problems.

### To reproduce

```sh
./build-workflow/build.sh manifests/1.1.0/openseach-dashboards-1.1.0.yml
```

Succeeds for all components.

```
./build-workflow/build.sh manifests/1.1.0/openseach-dashboards-1.1.0.yml --component=alertingDashboards
```

Fails with

```
+ PLUGIN_NAME=alertingDashboards
+ cp -r ../alertingDashboards/ ../OpenSearch-Dashboards/plugins
cp: cannot create directory '../OpenSearch-Dashboards/plugins': No such file or directory
2021-09-27 14:17:21 ERROR    Error building alertingDashboards, retry with: ./bundle-workflow/build.sh manifests/1.1.0/opensearch-dashboards-1.1.0.yml --component alertingDashboards
Traceback (most recent call last):
  File "./bundle-workflow/src/run_build.py", line 80, in <module>
    sys.exit(main())
  File "./bundle-workflow/src/run_build.py", line 66, in main
    builder.build(target)
  File "/home/ubuntu/source/opensearch-project/opensearch-build/dblock-opensearch-build/bundle-workflow/src/build_workflow/builder.py", line 41, in build
    self.git_repo.execute(build_command)
  File "/home/ubuntu/source/opensearch-project/opensearch-build/dblock-opensearch-build/bundle-workflow/src/git/git_repository.py", line 70, in execute
    subprocess.check_call(command, cwd=cwd, shell=True)
  File "/home/ubuntu/.pyenv/versions/3.7.11/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '/home/ubuntu/source/opensearch-project/opensearch-build/dblock-opensearch-build/bundle-workflow/scripts/default/build.sh -v 1.1.0 -a x64 -s false -o artifacts -d true' returned non-zero exit status 1.

```

### Expected behavior

We could adopt something similar to how OpenSearch builds into `~/.m2` for dependencies with npm local, so that the output from the dashboards core could be published and preserved.

kavilla:
We were thinking about to solve this issue we will allow for multiple components to be passed so it builds together

peterzhuamazon:
[...]

PR #1278, opened 2021-12-09
Update dashboards-reports build.sh so it builds with dashboards

peterzhuamazon:

### Description
Update dashboards-reports build.sh so it builds with dashboards
 
### Issues Resolved
Relate to #870
 
### Check List

### DIFF
--- a/scripts/components/reportsDashboards/build.sh
+++ b/scripts/components/reportsDashboards/build.sh
@@ -85,6 +85,11 @@ esac
 
 CHROMIUM_URL="https://github.com/opensearch-project/dashboards-reports/releases/download/chromium-1.12.0.0/$CHROMIUM_TARGET"
 
+
+MINOR_VERSION=${VERSION%.*}
+git clone --branch $MINOR_VERSION --single-branch https://github.com/opensearch-project/OpenSearch-Dashboards ../../OpenSearch-Dashboards || echo repo exists
+mkdir -p ../../OpenSearch-Dashboards/plugins
+
 mkdir -p $OUTPUT/plugins
 # For hybrid plugin it actually resides in 'reportsDashboards/dashboards-reports'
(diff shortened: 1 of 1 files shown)


dblock (approved):
I like this! Can we add it to the default build script(s) too and fix https://github.com/opensearch-project/opensearch-build/issues/606?

peternied (approved):

peterzhuamazon:
> I like this! Can we add it to the default build script(s) too and fix #606?

@dblock this seems like a change needed to be added for every scripts in builds.
Is there any better mechanism you can think of that is similar to this?

Thanks.

dblock:
> > I like this! Can we add it to the default build script(s) too and fix #606?
> 
> @dblock this seems like a change needed to be added for every scripts in builds. Is there any better mechanism you can think of that is similar to this?

We could provide a shared function or a shared .sh for this, but maybe we can just modify the scripts. Maybe someone from dashboards can take a look?

scripts/components/reportsDashboards/build.sh
```diff
@@ -85,6 +85,10 @@ esac
 
 CHROMIUM_URL="https://github.com/opensearch-project/dashboards-reports/releases/download/chromium-1.12.0.0/$CHROMIUM_TARGET"
 
+MINOR_VERSION=${VERSION%.*}
+git clone --branch $MINOR_VERSION --single-branch https://github.com/opensearch-project/OpenSearch-Dashboards ../../OpenSearch-Dashboards || echo repo exists
```
kavilla, line 89: so we are downloading Dashboards if it doesn't exist so it can build the dashboards reports component?

peterzhuamazon: yes

dblock: note that to fix #606 property you would want to also check not just whether that folder exists, but that it was built

Issue #1279, opened 2021-12-09
[Bug]Jenkins Opensearch is building snapshot that will fail on a patch release


peterzhuamazon:
### Describe the bug

There seems like an issue with the jenkins due to it is also running `--snapshot`

```
2021-12-09 18:55:45 ERROR    Error building job-scheduler, retry with: ./build.sh manifests/1.1.1/opensearch-1.1.1.yml --component job-scheduler --snapshot
Traceback (most recent call last):
  File "/var/jenkins/workspace/bundle-build@2/src/build_workflow/opensearch/build_artifact_check_maven.py", line 40, in check
    properties.check_value_in("Implementation-Version", versions)
  File "/var/jenkins/workspace/bundle-build@2/src/system/properties_file.py", line 55, in check_value_in
    raise PropertiesFile.UnexpectedKeyValuesError(key, expected, value)
system.properties_file.UnexpectedKeyValuesError: Expected to have Implementation-Version=any of [None, '1.1.1.0-SNAPSHOT', '1.1.0.0-SNAPSHOT', '1.1.1-SNAPSHOT', '1.1.0-SNAPSHOT'], but was '1.1.0.0'.

```

We are trying to release 1.1.1 based on 1.1.0 core, and the `--snapshot` is trying to build on non-exist `1.1.1` core.

Thanks.

### To reproduce

./build.sh manifests/1.1.1/opensearch-1.1.1.yml --component job-scheduler --snapshot

### Expected behavior

snapshot is disabled for patch releases, or it is building based on the patch core version.

zelinh:
[Triage] @peterzhuamazon Could you please update the progress here.

dblock:
I'm fixing this.

peterzhuamazon:
```

total 8
drwxr-xr-x   3 zhujiaxi  staff    96 Dec  9 17:09 dist
drwxr-xr-x   3 zhujiaxi  staff    96 Dec  9 17:10 libs
-rw-r--r--   1 zhujiaxi  staff  3853 Dec  9 17:10 manifest.yml
drwxr-xr-x  14 zhujiaxi  staff   448 Dec  9 17:10 plugins

```

dblock:
Fixed in https://github.com/opensearch-project/opensearch-build/pull/1281

PR #1280, opened 2021-12-09
Fix: Relax compatible versions checks for snapshot dependencies.

dblock:

### Description

When building a snapshot patch build we reference released components that aren't -SNAPSHOT causing the version checks to trip. 

```
expected version to be one of [None, '1.1.1.0-SNAPSHOT', '1.1.0.0-SNAPSHOT', '1.1.1-SNAPSHOT', '1.1.0-SNAPSHOT'], but was '1.1.0.0'.
```

This is saying that the downloaded artifact was `1.1.0.0`, but we're building `1.1.0.1-SNAPSHOT` and we expected `1.1.0.0-SNAPSHOT` at the very least. 

The first option was to disable snapshot build for the patch releases. But patch releases may very well contain snapshots for things like common-utils or job-scheduler-spi that need to be uploaded to AWS OSS repo, thus that is not a good plan.

The next obvious solution would be to download and use artifacts from a -SNAPSHOT build when building a snapshot version, but we never publish complete -SNAPSHOT build artifacts, since the only purpose of snapshot artifacts is to publish snapshot JARs to maven. On the other hand, if we were able to download 1.1.0.0-SNAPSHOT artifacts, we would probably not want to re-publish them either (they have already been published and we may potentially publish an older version on top of a newer one accidentally).

This PR fixes the build by relaxing the version compatibility checks to allow -SNAPSHOT builds to use both snapshot and non-snapshot older dependencies in the distribution. Since an older version has already been released, this is actually semantically correct.

<strike>But once this build succeeds, I believe our maven snapshot publication script will attempt to publish everything in maven/*, including non-snapshot artifacts, to the AWS OSS snapshot repo. Is this a problem?</strike>

Once this build succeeds,  our maven snapshot publication script attempts to publish everything in `maven/*`. This PR excludes downloading those non-snapshot maven artifacts to avoid pushing them to the AWS OSS snapshot repo.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/1279

### Check List

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
-ignore = E722
-max-line-length = 160
+ignore = E722,W503
+max-line-length = 200
--- a/src/build_workflow/build_target.py
+++ b/src/build_workflow/build_target.py
@@ -28,7 +28,7 @@ class BuildTarget:
         name=None,
         snapshot=True,
         build_id=None,
-        output_dir="artifacts",
+        output_dir="artifacts"
     ):
         self.build_id = os.getenv("BUILD_NUMBER") or build_id or uuid.uuid4().hex
         self.name = name
@@ -45,7 +45,11 @@ class BuildTarget:
 
     @property
     def compatible_opensearch_versions(self):
-        return list(map(lambda version: version + "-SNAPSHOT" if self.snapshot else version, self.compatible_versions))
+        return (
+            [self.version + "-SNAPSHOT" if self.snapshot else self.version]
+            + self.patches
+            + list(map(lambda version: version + "-SNAPSHOT", self.patches))
+        )
 
     @property
     def component_version(self):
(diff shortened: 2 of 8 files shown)


src/build_workflow/build_target.py
```diff
@@ -54,8 +58,11 @@ def component_version(self):
 
     @property
     def compatible_component_versions(self):
-        # BUG: the 4th digit is dictated by the component, it's not .0, this will break for 1.1.0.1
```
peterzhuamazon, line 57: Can you restore this comment as this is still the case.
Thanks.

dblock: This is no longer the case.

peterzhuamazon: Got it.

peterzhuamazon (approved):

.flake8
```diff
@@ -1,3 +1,3 @@
 [flake8]
-ignore = E722
-max-line-length = 160
+ignore = E722,W503
```
peternied, line 2: FYI https://www.flake8rules.com/rules/W503.html

I approve of ignoring this rule

peternied (approved):

.flake8
dblock: Do you prefer W504 or W503?

peternied: I tend towards W504, if someone wanted to roll it out that would be interesting to see.

PR #1288, opened 2021-12-10
Manually generate 1.2.1 release from 1.2.0

peternied:

### Description
Creating a new release manifest for OpenSearch 1.2.1 patch release 

### Check List

### DIFF
--- a/src/build_workflow/build_args.py
+++ b/src/build_workflow/build_args.py
@@ -48,12 +48,6 @@ class BuildArgs:
             action="store_true",
             help="Do not delete the working temporary directory.",
         )
-        parser.add_argument(
-            "--patch",
-            dest="patch",
-            action="store_true",
-            help="Create a patch manifest.",
-        )
         parser.add_argument("-p", "--platform", type=str, choices=self.SUPPORTED_PLATFORMS, help="Platform to build.")
         parser.add_argument("-a", "--architecture", type=str, choices=self.SUPPORTED_ARCHITECTURES, help="Architecture to build.")
         parser.add_argument(
@@ -73,7 +67,6 @@ class BuildArgs:
         self.snapshot = args.snapshot
         self.component = args.component
         self.keep = args.keep
-        self.patch = args.patch
         self.platform = args.platform
         self.architecture = args.architecture
         self.script_path = sys.argv[0].replace("/src/run_build.py", "/build.sh")
--- a/src/build_workflow/build_args.py
+++ b/src/build_workflow/build_args.py
@@ -48,6 +48,12 @@ class BuildArgs:
             action="store_true",
             help="Do not delete the working temporary directory.",
         )
+        parser.add_argument(
+            "--patch",
+            dest="patch",
+            action="store_true",
+            help="Create a patch manifest.",
+        )
         parser.add_argument("-p", "--platform", type=str, choices=self.SUPPORTED_PLATFORMS, help="Platform to build.")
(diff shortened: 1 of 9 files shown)


dblock:
I don't think we should be rebuilding everything from source. If the components don't have changes, use `dist:` with a URL.

peternied:
> I don't think we should be rebuilding everything from source. If the components don't have changes, use `dist:` with a URL.

Just chatted offline, I'll switch to already build dist artifacts for all plugins until we have known replacements.  Lets prove out the patching process by using it here.

src/build_workflow/builder_from_dist.py
```diff
@@ -40,14 +41,28 @@ def export_artifacts(self, build_recorder):
...
                 for artifact in component_manifest.artifacts[artifact_type]:
-                    artifact_url = f"{self.component.dist}/{self.target.platform}/{self.target.architecture}/builds/{self.target_name}/{artifact}"
-                    artifact_dest = os.path.realpath(os.path.join(self.output_path, artifact))
-                    os.makedirs(os.path.dirname(artifact_dest), exist_ok=True)
-                    logging.info(f"Downloading {artifact_url} into {artifact_dest}")
-                    urllib.request.urlretrieve(artifact_url, artifact_dest)
-                    build_recorder.record_artifact(self.component.name, artifact_type, artifact, artifact_dest)
+                    for url_pattern in self.__get_url_patterns():
+                        artifact_url = f"{url_pattern}/{artifact}"
+                        try:
```
dblock, line 46: This swallows too many exceptions and fails silently if a distribution cannot be found. 

1. Add a `distribution_urls` method that returns the list of possible URLs.
2. Add a `distribution_url` method that iterates over `distribution_urls` and returns the first one that has a manifest.yml
3. Use the result from `distribution_url` everywhere.

dblock (changes requested):

src/build_workflow/build_args.py
```diff
@@ -48,6 +48,12 @@ def __init__(self):
             action="store_true",
             help="Do not delete the working temporary directory.",
         )
+        parser.add_argument(
```
dblock, line 51: Unused, remove for now?

src/manifests/distribution.py
```diff
@@ -0,0 +1,26 @@
+import logging
```
dblock, line 1: To be consistent I would make this a class called `Distribution` that takes a base url and product name in the constructor, then add a `find` method.

Needs a test.

src/manifests/distribution.py
```diff
@@ -0,0 +1,26 @@
...
+        f"{base_url}/{platform}/{architecture}/builds"
+    ]
+
+    for distribution_url in possible_urls:
+        manifest_url = f"{distribution_url}/manifest.yml"
+        try:
+            with urllib.request.urlopen(manifest_url):
+                # OK we could access the manifest, return the url
+                return distribution_url
+        except:
```
dblock, line 24: Check that the exception is a 403 or 404, fail on everything else.

dblock (approved):
I'm good with this, we can refactor later. Up to you.

src/manifests/distribution.py
peternied: Lets fast follow on these so the build isn't blocked, https://github.com/opensearch-project/opensearch-build/issues/1289

src/manifests/distribution.py
peternied: Lets fast follow on these so the build isn't blocked, https://github.com/opensearch-project/opensearch-build/issues/1289

zelinh (approved):
Thanks!

Issue #1276, opened 2021-12-09
Read manifest settings to decide the test args parsing


tianleh:
### Is your feature request related to a problem? Please describe

Initiated from this PR comment https://github.com/opensearch-project/opensearch-build/pull/1263#discussion_r765899016

### Describe the solution you'd like

do not need to set dashboards related params if the test manifest indicates that this is an OpenSearch integration test.

PR #1305, opened 2021-12-11
Improve test args

tianleh:
### Description
Address db and Peter's remaining comments as follow up.
Related merged PR https://github.com/opensearch-project/opensearch-build/pull/1263 

### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/1276
 
### Check List


tianleh:
For @dblock 's comment, 
https://github.com/opensearch-project/opensearch-build/pull/1263#discussion_r765833715

I followed it by using the `get` method with default value. We need to have `None` check against `args.paths` since it could be `None`.

In terms of moving validation logic to `parser`, based on my research, there are two ways to do it, either through `type` or `action`. (See this reference https://stackoverflow.com/questions/37471636/python-argument-parsing-validation-best-practices.)

I am very curious whether `type` can support a complicated structure like this.
```
type=Dict[str, TestArgsPathValidator.validate]
```

However, it didn't work. Thus I put the `TestArgsPathValidator.validate` inside the action `TestKwargs`

tianleh:
For @peternied 's comment, I added the test manifest loading and read the content for setting Dashboards args. This indeed makes the unit test more clear since we just need to check `None` for Dashboards args in case of OpenSearch test manifest.

tianleh:
During local test, `./test.sh` is not printing the logs in console. Need to check whether it is because of my local console configuration or some coding logic change.

src/test_workflow/test_args.py
```diff
@@ -38,21 +37,13 @@ def __init__(self):
...
-            path = args_paths[name]
+        self.test_manifest = TestManifest.from_path(args.test_manifest_path)
 
-        return self._validate_path(path)
+        cwd = os.getcwd()
+        self.opensearch_path = args.paths.get("opensearch", cwd) if args.paths else cwd
 
-    def _validate_path(self, path: str) -> str:
-        return path if validators.url(path) else os.path.realpath(path)
+        if self.test_manifest.name == "OpenSearch Dashboards":
```
dblock, line 45: Why do you need the `if`? Just assign it if it's passed in. Or better, assign `self.paths` and use `self.paths.get("...")` as needed.

tests/tests_test_workflow/data/test_manifest_opensearch_dashboards.yml
```diff
@@ -0,0 +1,12 @@
+name: OpenSearch Dashboards
```
dblock, line 1: Since it's a manifest I would use `-` instead of `_` in the filename.

dblock (approved):
Could you please fix the `if` before merge? Good work.

seraphjiang (approved):

peternied (changes requested):
Do not merge until we have completed the production release testing, I'll unblock at that time

src/test_workflow/test_args.py
tianleh: E.g 
1) `./build.sh manifests/1.3.0/opensearch-1.3.0.yml`

2) `./build.sh manifests/1.3.0/opensearch-dashboards-1.3.0.yml`

Both will set the default value for OpenSearch. But only 2) will need to set a default value for Dashboards. I think this is also what Peter means in this comment https://github.com/opensearch-project/opensearch-build/pull/1263#discussion_r765899016

peternied (approved):
Production testing has completed unblocking this PR, thanks for your patience

tianleh:
> Production testing has completed unblocking this PR, thanks for your patience

Thanks for your confirmation. I am still checking why the logging is not printed to console during execution. I will take care of this before merging this change.

src/test_workflow/test_args.py
dblock: First, default paths to `{}` and assign paths to `self`. This way it's always a dictionary and you don't have to do `if args.paths` ever.

```
self.paths = args.paths
```

Now, everywhere you need a dashboards path you could do `self.paths.get("opensearch-dashboards", cwd)`, or if you must, declare it as a property.

```
@property
def opensearch_dashboards_path(self)
    self.paths.get("opensearch-dashboards", cwd)
```

Does this remove the `if` or am I missing something?

tianleh:
Somehow the logging config is changed by this PR. Console still cannot see the execution logs. No obvious clue which change is suspicious. I will comment out changes line by line to see which one causes the issue.

tianleh:
> Somehow the logging config is changed by this PR. Console still cannot see the execution logs. No obvious clue which change is suspicious. I will comment out changes line by line to see which one causes the issue.

Finally finds the reason. 

This is where the log level is configured.

`    console.configure(level=args.logging_level)`

https://github.com/opensearch-project/opensearch-build/blob/main/src/run_integ_test.py#L19

However, this will not work if the preceding execution chains have ever called `logging.info` before it. In my previous commit, I added a logic to read manifest from path. https://github.com/opensearch-project/opensearch-build/blob/12093e78268c6052aeb798987ae97f0f770db1a7/src/manifests/manifest.py#L59 

This has a such preceding `logging.info` which thus failed the console output. This is no longer an issue after I applied db's suggestion. However, this is still an interesting knowledge to share.

src/test_workflow/test_args.py
tianleh: Great suggestion! Originally my mind was limited to solve this problem within the scope of `TestArgs` class. But your suggestion opened my mind to let the caller (who intrinsically knows the situation better than `TestArgs` ) decide the parameters to use. 

Just updated the code and it is much cleaner and elegant!

dblock (approved):

PR #1308, opened 2021-12-11
Fix: dependency installer swallows errors.

dblock:
Signed-off-by: dblock <dblock@dblock.org>

### Description

The integration test flow does not get the product name so its missing the `opensearch` part of the path when running the tests. The `ThreadPoolExecutor` is not checking the future returned by submit, so the download error is swallowed.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/1304
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1315, opened 2021-12-12
fix random dependency_installer unit test failure with thread safe counter

tianleh:

### Description
We have noticed flakeness of unit tests https://github.com/opensearch-project/opensearch-build/pull/1308#issuecomment-991780354 

After numerous experiments (See the PR history for details), I have found that it is caused by thread unsafe of mock and thus introduce a thread safe counter to handle this. Good references links:
https://stackoverflow.com/questions/39332139/thread-safe-version-of-mock-call-count
 
### Test
unit test. Have been rerunning the tests multiple times and see consistent pass.
 
### Check List

### DIFF
--- a/src/system/thread_safe_counter.py
+++ b/src/system/thread_safe_counter.py
@@ -1,0 +1,21 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import threading
+
+
+class ThreadSafeCounter:
+    def __init__(self):
+        self.lock = threading.Lock()
+        self.__call_count__ = 0
+
+    def thread_safe_count(self, *args, **kwargs):
+        with self.lock:
+            self.__call_count__ += 1
+
+    @property
+    def call_count(self):
+        return self.__call_count__
--- a/src/test_workflow/dependency_installer.py
+++ b/src/test_workflow/dependency_installer.py
@@ -42,6 +42,7 @@ class DependencyInstaller(abc.ABC):
                         paths)
             ):
                 logging.debug(f"Written {result}")
+                logging.info(f"Written {result}")
 
     def download_or_copy(self, source, dest):
         os.makedirs(os.path.dirname(dest), exist_ok=True)
--- a/src/test_workflow/dependency_installer.py
+++ b/src/test_workflow/dependency_installer.py
@@ -43,6 +43,7 @@ class DependencyInstaller(abc.ABC):
             ):
                 logging.debug(f"Written {result}")
                 logging.info(f"Written {result}")
+                print(f"Written {result}")
 
     def download_or_copy(self, source, dest):
         os.makedirs(os.path.dirname(dest), exist_ok=True)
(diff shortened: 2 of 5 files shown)


tianleh:
Debug logging shall be enabled https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging

Try `print` to see if we can get useful information in Windows

tianleh:
`                print(f"Written {result}")
`

creates 2375 records which match the expectation of the unit test number. 

This means that result count is correct. But `mock_copyfile.call_count` 2349 is smaller.

This means `self.download_or_copy(*args)` is called by correct times, but inside it, `shutil.copyfile` is wrong

tianleh:
rerun multiple times and cannot reproduce it.

tianleh:
Too many logs give the threads more time to run?

Trying reducing logs

tianleh:
Once the logs are reduced, the error shows up again. It gives me a feeling that it might relate to waiting not functioning.

tianleh:
take a note of the failed run 
https://github.com/opensearch-project/opensearch-build/runs/4497107148?check_suite_focus=true

tianleh:
```
---------------------------- Captured stdout call -----------------------------
inside download, paths has len 2330
inside download, paths has len 15
inside download, paths has len 15
inside download, paths has len 15
```

tianleh:
```
        mock_request.assert_not_called()
>       self.assertEqual(mock_copyfile.call_count, 2375)
E       AssertionError: 2264 != 2375

tests\tests_test_workflow\test_dependency_installer_opensearch.py:38: AssertionError
---------------------------- Captured stdout call -----------------------------
install_maven_dependencies OpenSearch
inside download {'OpenSearch'}, paths has len 2330
exiting inside download OpenSearch, paths has len 2330
install_maven_dependencies common-utils
inside download {'common-utils'}, paths has len 15
exiting inside download common-utils, paths has len 15
install_maven_dependencies job-scheduler
inside download {'job-scheduler'}, paths has len 15
exiting inside download job-scheduler, paths has len 15
install_maven_dependencies alerting
inside download {'alerting'}, paths has len 15
exiting inside download alerting, paths has len 15
```

https://github.com/opensearch-project/opensearch-build/runs/4497269655?check_suite_focus=true

tianleh:
Have merged all artifacts to make a single call to download and we still have the same problem. 

```
  
        dependency_installer.install_maven_dependencies()
>       self.assertEqual(mock_request.call_count, 2375)
E       AssertionError: 2181 != 2375

tests\tests_test_workflow\test_dependency_installer_opensearch.py:57: AssertionError

```

https://github.com/opensearch-project/opensearch-build/runs/4497477049?check_suite_focus=true

My interpretation is that the mock's call_count is not thread safe. Thus we have following options.

1)  switch back to the patch ThreadPoolExecutor
2) remove assert against unsafe mock attributes
3) create a thread safe count

tianleh:
Thread safe solution also failed https://github.com/opensearch-project/opensearch-build/runs/4497618130?check_suite_focus=true

tianleh:
Noticed that this is a good learning opportunity about multithreading in Python and thus just take action to research and fix it. @dblock

dblock (approved):
👏

peternied (approved):
Nicely done!

Seems like this is a bug in the mocking framework, do you know if they have one?

tianleh:
> Seems like this is a bug in the mocking framework, do you know if they have one?

Have seen such discussion on stackoverflow but not an official one.

PR #1330, opened 2021-12-14
Add new job for build ruby distributions

peternied:
### Description
Unifying how ruby distributions are created for release.  In draft until has confirmed the changes will work in dev environment.

This PR begins aligning all ruby release mechanics into a single process, the new expectations for a ruby component are:
- It has a file, `certs/opensearch-rubygems.pem`
- it has a build file in the root, `build.sh`
- it can read a file in the root, `gem-private_key.pem` that contains signing secrets
- it creates artifacts in the `builds` folder

While these requirements were established over a chime call, in future PRs we can being abstracting these into a new type of project OpenSearchRuby which ruby projects will align to, very similar to OpenSearch/OpenSearch-Dashboards.  Additional arguments can be baked into an input manifest, and then we can build out a richer workflow and validation process.
 
### Check List


jenkins/ruby/distribution-build.jenkinsfile
```diff
@@ -0,0 +1,99 @@
...
+        )
+        string(
+            name: 'RELEASE_BRANCH',
+            description: 'Branch name to release, e.g. 1.2.0',
+            trim: true
+        )
+        string(
+            name: 'RUBY_INTERPRETER',
+            description: 'ruby, jruby, etc',
+            defaultValue: 'ruby',
```
dblock, line 22: I think we should be defaulting and installing something more specific, e.g. `ruby-2.7.0`.

jenkins/ruby/distribution-build.jenkinsfile
```diff
@@ -0,0 +1,99 @@
...
+            }
+            steps {
+                script {
+                    git url: "$REPOSITORY", branch: "$RELEASE_BRANCH"
+                    withCredentials([string(credentialsId: 'opensearchproject-rubygems-private-key', variable: 'RUBYGEMS_PRIVATE_KEY')]) {
+                        writeFile file: 'gem-private_key.pem', text: RUBYGEMS_PRIVATE_KEY
+                    }
+                    sh'''
+                        rvm install $RUBY_INTERPRETER
+                        rvm use $RUBY_INTERPRETER
```
dblock, line 44: Add `--default`.

dblock (changes requested):
The whole `getRepositoryName` feels like a total overkill. Two ideas:

1. Make `REPOSITORY` just that name and generate `https://github.com/opensearch-project/$REPOSITORY.git`.
2. Use current directory name, `${PWD##*/}`.
3. Hardcode it in the project being built.

jenkins/ruby/distribution-build.jenkinsfile
```diff
@@ -0,0 +1,99 @@
...
+                    image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211130'
+                    args '-u root' // Needed for installing RUBY_INTERPRETER 
+                    alwaysPull true
+                }
+            }
+            steps {
+                script {
+                    git url: "https://github.com/opensearch-project/$REPOSITORY.git", branch: "$RELEASE_BRANCH"
+                    withCredentials([string(credentialsId: 'opensearchproject-rubygems-private-key', variable: 'RUBYGEMS_PRIVATE_KEY')]) {
+                        writeFile file: 'gem-private_key.pem', text: RUBYGEMS_PRIVATE_KEY
```
gaiksaya, line 40: Should the text value (RUBYGEMS_PRIVATE_KEY) have a `$` to take the value of the variable?

jenkins/ruby/distribution-build.jenkinsfile
```diff
@@ -0,0 +1,99 @@
...
+                    }
+                    sh'''
+                        rvm install $RUBY_INTERPRETER
+                        rvm --default use $RUBY_INTERPRETER
+                        gem cert --add certs/opensearch-rubygems.pem
```
gaiksaya, line 45: Note: This means any ruby project would need to have the public key in exact same location (certs/opensearch-rubygems.pem)
opensearch-ruby has it [there](https://github.com/opensearch-project/opensearch-ruby/blob/main/certs/opensearch-rubygems.pem)

peternied: Yes, lets try to start with this as the convention, we can revisit if this becomes problematic, FYI - logstash-output-opensearch was OK with this

gaiksaya: Sounds good. Thanks!

jenkins/ruby/distribution-build.jenkinsfile
peternied: `withCredentials(...)` assigns a variable named by the parameter to the secret value, and then on the next line when it is referenced with `writeFile(...)` it used as a groovy variable.  This is the expected usage, see https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#for-secret-text-usernames-and-passwords-and-secret-files which has more examples and details

gaiksaya (approved):

jenkins/ruby/distribution-build.jenkinsfile
```diff
@@ -0,0 +1,99 @@
...
+                }
+            }
+            steps {
+                script {
+                    git url: "https://github.com/opensearch-project/${REPOSITORY_NAME}.git", branch: "$RELEASE_BRANCH"
+                    withCredentials([string(credentialsId: 'opensearchproject-rubygems-private-key', variable: 'RUBYGEMS_PRIVATE_KEY')]) {
+                        writeFile file: 'gem-private_key.pem', text: RUBYGEMS_PRIVATE_KEY
+                    }
+                    sh'''
+                        rvm install $RUBY_INTERPRETER
```
dblock, line 43: Do these fail when either of them fails?

dblock (approved):

abhinavGupta16 (approved):

jenkins/ruby/distribution-build.jenkinsfile
peternied: Yes they do

peternied:
Note; I am still fighting with the image itself, it is not properly using the rvm install and preventing me from switching environments :/

dblock:
> Note; I am still fighting with the image itself, it is not properly using the rvm install and preventing me from switching environments :/

You might just need to source it, https://github.com/dblock/dotfiles/blob/72db00702e0e98319e63f80fe4a1439da1ae4ae6/bash/.bashrc#L6

peternied:
@zelinh could you look into getting this working while I am out? Recapping the investigation;
'rvm use' needs to switch interpreter versions, it doesn't work because of
RVM is not a function, selecting rubies with 'rvm use ...' will not work.
  FAQ https://rvm.io/rvm/install
 - Code for testing is at https://github.com/peternied/opensearch-build/tree/testing-ruby-dist
 - Job is named opensearch-ruby-client-release
- Seems that we also need to STOP running as root because of side effects
  - Need to add opensearch user into rvm group, usermod -a -G rvm opensearch
- Not sure if changes from https://github.com/opensearch-project/opensearch-build/commit/2262c006f869ddb18a9b35e0db169f7b8880b448#diff-3c6d9a396ab9df4ba9ec18f551c51545675253ca7eaf94ebc1b24afa8ff74a7d are correct, maybe not?
- How to test this on dev machine?
Blocking https://github.com/opensearch-project/opensearch-build/pull/1330

It works when, the greated gem is named:
logstash-output-opensearch-1.2.0-x86_64-java.gem
NOT logstash-output-opensearch-1.2.0-x86_64-linux.gem


Investigation paths:
 - Create a fresh [dockerfile](https://www.jenkins.io/doc/book/pipeline/docker/#dockerfile), not an image, with rvm and aws cli that can install and run jruby and run the job
 - Revert the rvm environment variable changes and see if things work again?

zelinh (changes requested):
Block this PR from merging since we are looking into the `rvm` issue.

dblock:
> Block this PR from merging since we are looking into the `rvm` issue.

Did my [suggestion](https://github.com/opensearch-project/opensearch-build/pull/1330#issuecomment-995249473) not work?

peternied:
@dblock It didn't work :(

peternied:
@zelinh any progress on this?

zelinh:
> @zelinh any progress on this?

I made some changes and had this running. Should I commit on your branch?

peternied:
@zelinh Could you make a pull request onto https://github.com/peternied/opensearch-build/tree/ruby-dist and we can look over how you fixed it?  Afterward we can merge and it will show up here

zelinh:
PR created here. https://github.com/peternied/opensearch-build/pull/12/files

zelinh (approved):

abhinavGupta16:
@peternied - let's add a test case for `jenkins/ruby/distribution-build.jenkinsfile` here as well

jenkins/ruby/distribution-build.jenkinsfile
```diff
@@ -0,0 +1,104 @@
...
+                        "${REPOSITORY_NAME}",
+                        "${RELEASE_BRANCH}",
+                        "${BUILD_NUMBER}",
+                        "builds"
+                    ].join('/'))
+                    withAWS(role: 'opensearch-bundle', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+                        s3Upload(file: 'builds', bucket: "${ARTIFACT_BUCKET_NAME}", path: "${artifacts_path}")
+                    }
+                    lib.jenkins.Messages.new(this).add("${STAGE_NAME}", "Built artifacts into ${PUBLIC_ARTIFACT_URL}/${artifacts_path}.")
+                    sh """
```
dblock, line 61: Isn't `postCleanup()` doing this?

jenkins/ruby/distribution-build.jenkinsfile
```diff
@@ -0,0 +1,104 @@
...
+            description: 'Repository to release, e.g. opensearch-ruby',
+            trim: true
+        )
+        string(
+            name: 'RELEASE_BRANCH',
+            description: 'Branch name to release, e.g. 1.2.0',
+            trim: true
+        )
+        string(
+            name: 'RUBY_INTERPRETER',
```
dblock, line 20: I think RUBY_VERSION would be better.

jenkins/ruby/distribution-build.jenkinsfile
zelinh: We used the root access for `rvm install` so `postCleanup()` is not able to delete the workspace as user role.

peternied:
OK, this isn't getting the attention it needs, closing it until we can get more effort behind it.  I've created #1593 which when resolved will fix the critical missing peace for this PR.

Thanks again everyone that reviewed and commented.

PR #1346, opened 2021-12-15
Add constrait to prevent pulling in older version of log4j

peternied:

### Description
Add constraint to prevent pulling in older version of log4j
 
### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,15 @@ dependencies {
     compile group: 'com.lesfurets', name:'jenkins-pipeline-unit', version: '1.13'
     compile group: 'com.cloudbees', name: 'groovy-cps', version: '1.31'
     compile group: 'org.yaml', name: 'snakeyaml', version: '1.29'
+    constraints {
+        implementation("org.apache.logging.log4j:log4j-core") {
+            version {
+                strictly("[2.16, 3[")
+                prefer("2.16.0")
+            }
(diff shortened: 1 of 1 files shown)


dblock:
Can we try and do something like this for jenkins-core as well? 

https://github.com/opensearch-project/opensearch-build/issues/1097

build.gradle
```diff
@@ -35,6 +35,15 @@ dependencies {
     compile group: 'com.lesfurets', name:'jenkins-pipeline-unit', version: '1.13'
     compile group: 'com.cloudbees', name: 'groovy-cps', version: '1.31'
     compile group: 'org.yaml', name: 'snakeyaml', version: '1.29'
+    constraints {
+        implementation("org.apache.logging.log4j:log4j-core") {
+            version {
+                strictly("[2.16, 3[")
```
gaiksaya, line 41: The bracket needs to be facing inward `[2.16, 3]`

peternied: This is from the gradle folks https://blog.gradle.org/log4j-vulnerability

gaiksaya: Interesting!!

gaiksaya (approved):

peternied:
Not sure that this worked, going to see if I can get better build logging locally to confirm and I'll fix jenkins at the same time.

dblock:
According to Whitesource that didn't seem to work :(

peternied:
Didn't work, we will need to revisit

Issue #368, opened 2021-09-01
Fix file extension of signature files in signer workflow.


mch2:
Right now the signer workflow is invoking the client and specifying the output with -o as a *.asc file.  In reality, the file returned from the client is a binary (.sig) file and not ascii armored.

Until [this](https://github.com/opensearch-project/opensearch-signer-client/issues/10) issue is resolved to specify the desired file type we should change the extension and correctly mark these as .sig files.

peternied:
[Triage] Still waiting on downstream

gaiksaya:
Proposing a work around:
Instead of changing the code base of signer client side (which is likely to be remodel or be replaced by something else cc: @bbarani ), proposing to making this change via python code. 
It is considerably more easy to change the `.sig` (binary output) to `.asc` (armored output) via gpg commands.
The code can be implemented via interim code base (python code in this repo. More precisely https://github.com/opensearch-project/opensearch-build/blob/main/src/sign_workflow/signer_pgp.py) rather than waiting on upstream.

I'll be looking into this. 
Thanks!

gaiksaya:
Works as expected: https://build.ci.opensearch.org/job/distribution-maven-sign-and-release/26/
![image](https://user-images.githubusercontent.com/61760125/200721697-a6d8efa7-0f1d-403a-9f61-42bd3a1590fc.png)

PR #1352, opened 2021-12-16
added support to sign linux artifacts

### LINKED ISSUE #1351: [META] Automate artifact signing with OpensearchSignerClient for linux

Signing currently is a long painstaking process. It needs to be automated to save time when releasing artifacts. We need a tool to be able to sign all artifacts - 
- opensearch
- opensearch-dashboards

Use the existing `Signer.py` class to sign the artifacts and generate ".sig" file along with ".asc" files.

### Tasks
- [x] #1382
- [x] #1383
- [x] #1385

### Acceptance Criteria
- [ ] User is able to provide the artifact directory and tool is able to sign all the artifacts in the directory
- [ ] User is able to pass the artifact path and the tool is able to sign the artifact

### Next Steps
Next steps would include to extend this process over for mac and windows

abhinavGupta16:

### Description
Automate linux signing to sign any artifact or all artifacts in a directory
 
### Issues Resolved
#1351
#1382

### Check List


src/sign_workflow/sign_existing_artifacts.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self, artifact_path, signature_type):
+        self.artifact_path = artifact_path
+        self.signature_type = signature_type
+
+    def __setattr__(self, name, value):
+        if name == 'artifact_path' and not isinstance(value, Path):
+            raise TypeError('SignExistingArtifacts.artifact_path must of type Path')
+        super().__setattr__(name, value)
+
+    def sign_using_manifest(self):
```
gaiksaya, line 26: did you mean `sign_existing_artifacts` here? using manifest already exists in https://github.com/opensearch-project/opensearch-build/pull/1352/files#diff-4bf5b6c3689571f89b11c463c1a6a4ff6bcdceb0a5e0ed1726d23886cb39ea18R24

abhinavGupta16: No, It is `SignExistingArtifacts` as in class name.

src/sign_workflow/sign_existing_artifacts.py
```diff
@@ -0,0 +1,39 @@
...
+class SignExistingArtifacts:
+
+    def __init__(self, artifact_path, signature_type):
+        self.artifact_path = artifact_path
+        self.signature_type = signature_type
+
+    def __setattr__(self, name, value):
+        if name == 'artifact_path' and not isinstance(value, Path):
+            raise TypeError('SignExistingArtifacts.artifact_path must of type Path')
+        super().__setattr__(name, value)
```
gaiksaya, line 24: Why not add `artifact_path : Path` instead of this?

src/sign_workflow/sign_existing_artifacts.py
gaiksaya: why both methods are named same?

abhinavGupta16: fixed

src/sign_workflow/sign_existing_artifacts.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def sign_single_artifacts(self):
+        signer = Signer()
+        artifacts = [self.artifact_path.name]
+        basename = self.artifact_path.parent
+        signer.sign_artifacts(artifacts, basename, self.signature_type)
+
+    def sign_all_artifacts_in_dir(self):
+        signer = Signer()
+        files_map = {}
```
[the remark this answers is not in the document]
abhinavGupta16: No, this cannot be combined since the Signer class takes one basepath and corresponding list of artifacts to sign as inputs.

src/run_sign.py
```diff
@@ -8,19 +8,22 @@
...
+ACCEPTED_SIGNATURE_FILE_TYPES = [".sig", ".asc"]
+
 
 def main():
     parser = argparse.ArgumentParser(description="Sign artifacts")
-    parser.add_argument("manifest", type=argparse.FileType("r"), help="Path to local manifest file.")
+    parser.add_argument("target", type=Path, help="Path to local manifest file or artifact directory")
     parser.add_argument("--component", nargs="?", help="Component name")
     parser.add_argument("--type", nargs="?", help="Artifact type")
+    parser.add_argument("--sigType", nargs="?", help="Type of Signature file", default=".asc")
```
dblock, line 26: Arguments shouldn’t be camelcase, —sigtype or —signature-type.
Lowercase signature in the description.

src/run_sign.py
```diff
@@ -32,22 +35,20 @@ def main():
     )
     args = parser.parse_args()
 
-    console.configure(level=args.logging_level)
-
-    manifest = BuildManifest.from_file(args.manifest)
-    basepath = os.path.dirname(os.path.abspath(args.manifest.name))
-    signer = Signer()
+    if args.sigType not in ACCEPTED_SIGNATURE_FILE_TYPES:
```
dblock, line 38: There’s a way to add a list of allowed values to args and remove this.

src/run_sign.py
```diff
@@ -32,22 +35,20 @@ def main():
...
-        for artifact_type in component.artifacts:
-            if args.type and args.type != artifact_type:
-                continue
+    console.configure(level=args.logging_level)
 
-            signer.sign_artifacts(component.artifacts[artifact_type], basepath)
+    file_extension = args.target.suffix
 
-    logging.info("Done.")
+    if file_extension == ".yml":
```
dblock, line 45: Add a class called Signers that has a method called from_file that returns either SignWithManifest or SignArtifacts. Both should inherit from the same base class and have a .sign method that you call once. The duplicate logic of creating a Signer should also be refactored - I think you want to get a list of files to sign in different ways but then sign each item in the list in the base class.

dblock (changes requested):

src/sign_workflow/sign_existing_artifacts.py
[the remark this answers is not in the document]
peternied: I am pretty you can but this is nit

src/run_sign.py
```diff
@@ -8,19 +8,22 @@
...
 import sys
+from pathlib import Path
 
-from manifests.build_manifest import BuildManifest
-from sign_workflow.signer import Signer
+from sign_workflow.sign_existing_artifacts import SignExistingArtifacts
+from sign_workflow.sign_with_manifest import SignWithManifest
 from system import console
 
+ACCEPTED_SIGNATURE_FILE_TYPES = [".sig", ".asc"]
```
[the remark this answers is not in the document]
abhinavGupta16: The existing workflow used for maven currently is the same as it is here, which only renames the binary file to be `.asc` and does not create new `.asc` files. They are therefore essentially `.sig` binary files.
Here, I have added support so that the existing maven workflow would remain the same and we can also create `.sig` files when required (eg: opensearch)

src/run_sign.py
```diff
@@ -8,19 +8,22 @@
...
+
 
 def main():
     parser = argparse.ArgumentParser(description="Sign artifacts")
-    parser.add_argument("manifest", type=argparse.FileType("r"), help="Path to local manifest file.")
+    parser.add_argument("target", type=Path, help="Path to local manifest file or artifact directory")
```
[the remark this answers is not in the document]
dblock: Just consistency.

src/sign_workflow/sign_existing_artifacts.py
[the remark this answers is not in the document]
abhinavGupta16: nvm... got it!

src/run_sign.py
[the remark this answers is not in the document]
peternied: The issue is this change does not make properly produced `.asc` files and hides the fact they are not generated correctly.  We should either apply the workaround or not support `.asc` until it works properly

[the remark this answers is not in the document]
peternied: Issue in our codebase #368

src/sign_workflow/sign_existing_artifacts.py
```diff
@@ -34,13 +34,8 @@ def sign_single_artifacts(self):
...
+            artifacts = []
             for file in files:
-                if subdir in files_map:
-                    files_map[subdir].append(file)
-                else:
-                    files_map[subdir] = [file]
-
-        for basename, artifact_list in files_map.items():
-            signer.sign_artifacts(artifact_list, basename, self.signature_type)
+                artifacts.append(file)
```
dblock, line 40: Isn't this just `files`, so `signer.sign_artifacts(files, subdir, self.signature_type)`

src/run_sign.py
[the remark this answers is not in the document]
abhinavGupta16: We can work on #368 in a different PR and maybe use `gpg enarmour` as a workaround?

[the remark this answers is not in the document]
peternied: [Blocking] To keep this PR mergable remove the ASC option, we can follow up on that open issue to add it back in the future.

You don't need to remove the plumbing, just the option from the array

tests/test_run_sign.py
```diff
@@ -32,7 +32,7 @@ def test_usage(self, *mocks):
 
     @patch("os.getcwd", return_value="curdir")
     @patch("argparse._sys.argv", ["run_sign.py", BUILD_MANIFEST])
-    @patch("run_sign.Signer", return_value=MagicMock())
+    @patch("sign_workflow.sign_artifacts.SignArtifacts", return_value=MagicMock())
```
peternied, line 35: How did you figure out the trick to get this working?

vars/signArtifacts.groovy
```diff
@@ -0,0 +1,25 @@
+void call(Map args = [:]) {
```
peternied, line 1: Lets moving this change into its own PR so it doesn't impact the python workflow that is ready to merge

vars/signArtifacts.groovy
```diff
@@ -0,0 +1,25 @@
...
+    environment {
+        // These ENV variables are required by https://github.com/opensearch-project/opensearch-signer-client
+        // This client is invoked internally by the sign script.
+        ROLE = "${SIGNER_CLIENT_ROLE}"
+        EXTERNAL_ID = "${SIGNER_CLIENT_EXTERNAL_ID}"
+        UNSIGNED_BUCKET = "${SIGNER_CLIENT_UNSIGNED_BUCKET}"
+        SIGNED_BUCKET = "${SIGNER_CLIENT_SIGNED_BUCKET}"
+    }
+
+    git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
```
peternied, line 12: Will this have an issue if run multiple times without cleaning the workspace?

vars/signArtifacts.groovy
```diff
@@ -0,0 +1,25 @@
...
+
+    sh('curl https://artifacts.opensearch.org/publickeys/opensearch.pgp -o $WORKSPACE/opensearch.pgp')
+    sh('gpg --import $WORKSPACE/opensearch.pgp')
+
+    def signatureType = ".sig"
```
peternied, line 17: The value should come from args.signatureType not hard coded (same for other parameters)

See https://github.com/opensearch-project/opensearch-build/blob/50552294f6830d7412d66831798fe15acf89e43d/vars/buildManifest.groovy for an example.

vars/signArtifacts.groovy
```diff
@@ -0,0 +1,25 @@
+void call(Map args = [:]) {
+
```
peternied, line 2: [For the end to end scenario, not this PR]

In order to call this function from the promotion script it needs to be in the form of a closure - that part can come afterward when the sha512 changes are made.

It would could make sense to keep this file as is built and create another function that returns the closure version.

peternied (changes requested):

src/run_sign.py
```diff
@@ -8,19 +8,21 @@
...
-import os
 import sys
+from pathlib import Path
 
-from manifests.build_manifest import BuildManifest
-from sign_workflow.signer import Signer
+from sign_workflow.sign_artifacts import SignArtifacts
 from system import console
 
+ACCEPTED_SIGNATURE_FILE_TYPES = [".sig", ".asc"]
```
dblock, line 17: Isn't this already `Signer.ACCEPTED_FILE_TYPES`? Use that.

src/run_sign.py
```diff
@@ -34,20 +36,10 @@ def main():
...
-
-    for component in manifest.components.select(focus=args.component):
-        logging.info(f"Signing {component.name}")
-
-        for artifact_type in component.artifacts:
-            if args.type and args.type != artifact_type:
-                continue
-
-            signer.sign_artifacts(component.artifacts[artifact_type], basepath)
+    sign = SignArtifacts.from_path(path=args.target, component=args.component,
```
dblock, line 39: nit: put each arg on its own line, it's easier to diff when they change

src/sign_workflow/sign_artifacts.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @abstractmethod
+    def __sign__(self):
+        pass
+
+    def sign(self):
+        self.__sign__()
+        logging.info("Done.")
+
+    def __sign_artifacts__(self, artifacts, basepath, signature_type):
```
dblock, line 34: The `signature_type` argument here is redundant because it's already passed into the constructor, so you can do `self.signer.sign_artifacts(artifacts, basepath, self.signature_type)`. Remove it.

src/sign_workflow/sign_artifacts.py
```diff
@@ -0,0 +1,79 @@
...
+        else:
+            return SignArtifactsExistingArtifactFile
+
+    @classmethod
+    def from_path(self, path: Path, component, artifact_type, signature_type):
+        klass = self.__signer_class__(path)
+        return klass(path, component, artifact_type, signature_type)
+
+
+class SignWithManifest(SignArtifacts):
```
dblock, line 52: rename to `SignWithBuildManifest` to be clear what kind of manifest we expect

src/sign_workflow/sign_artifacts.py
```diff
@@ -0,0 +1,79 @@
...
+                    continue
+
+                super().__sign_artifacts__(component.artifacts[component_artifact_type], basepath, self.signature_type)
+
+
+class SignArtifactsExistingArtifactFile(SignArtifacts):
+
+    def __sign__(self):
+        artifacts = [self.target.name]
+        basename = self.target.parent
```
dblock, line 71: These variables aren't reused, just pass the arguments as is below.

Add `__sign_artifact__` that takes one artifact.

src/sign_workflow/signer.py
```diff
@@ -27,32 +26,27 @@ def __init__(self):
         self.git_repo.execute("./bootstrap")
         self.git_repo.execute("rm config.cfg")
 
-    def sign_artifacts(self, artifacts, basepath):
+    def sign_artifacts(self, artifacts, basepath, signature_type):
```
dblock, line 29: There's a caller that passes 1 file and another that passes a collection. Split this function in 2, one that receives a single file and another that receives a collection.

src/sign_workflow/sign_artifacts.py
dblock, line 34: Add a single `__sign_artifact__` to avoid having to make an array in the caller.

tests/tests_sign_workflow/test_sign_artifacts.py
```diff
@@ -0,0 +1,40 @@
+import unittest
+from pathlib import Path
+
+from sign_workflow.sign_artifacts import SignArtifacts, SignArtifactsExistingArtifactFile, SignExistingArtifactsDir, SignWithManifest
+
+
+class TestSignArtifacts(unittest.TestCase):
+
+    def test_signer_class_method(self):
```
dblock, line 9: Remove `_class_method` from the name of the function, use the actual function name.

tests/tests_sign_workflow/test_sign_artifacts.py
```diff
@@ -0,0 +1,40 @@
...
+        path = Path(r"/dummy/path/manifest.yml")
+        klass = SignArtifacts.__signer_class__(path)
+
+        self.assertEqual(type(SignWithManifest), type(klass))
+
+        path = Path(r"/dummy/path/")
```
dblock, line 15: Split up in separate tests.

tests/tests_sign_workflow/test_sign_artifacts.py
```diff
@@ -0,0 +1,40 @@
...
+
+        self.assertEqual(type(SignArtifactsExistingArtifactFile), type(klass))
+
+    def test_from_path_method(self):
+        path = Path(r"/dummy/path/manifest.yml")
+        component = 'maven'
+        artifact_type = 'dummy'
+        sigtype = '.asc'
+
+        klass = SignArtifacts.from_path(path, component, artifact_type, sigtype)
```
dblock, line 31: Same as above, split up into individual tests for each type.

dblock (changes requested):
Close!

tests/test_run_sign.py
abhinavGupta16: It's not working yet.

src/run_sign.py
abhinavGupta16: No, this is accepted Signature types, different from accepted file types

src/sign_workflow/sign_artifacts.py
abhinavGupta16: Variables just make it clear as to what is being passed to the function. I think we can leave it like this

tests/tests_sign_workflow/test_sign_artifacts.py
abhinavGupta16: The name of the method is `__signer_class__`

tests/tests_sign_workflow/test_sign_artifacts.py
abhinavGupta16: I think it should be one method, since it just verifies multiple if conditions with no workflow. It would just increase verbosity and complexity. This is clear too. Thoughts?

tests/tests_sign_workflow/test_sign_artifacts.py
abhinavGupta16: Same reason as before.

tests/tests_sign_workflow/test_sign_artifacts.py
dblock: It would then be called `test_signer_class`.

You should use `assertIs` to avoid to do `type` twice.

Using variables that aren't actually reused is an anti-pattern.

Maybe this reads cleaner?

```python
    def test_signer_class(self):
        self.assertIs(SignArtifacts.__signer_class__(
            Path(r"/dummy/path/manifest.yml")), 
            SignWithBuildManifest)
            
        self.assertIs(SignArtifacts.__signer_class__(
            Path(os.path.dirname(__file__))), 
            SignExistingArtifactsDir)

        self.assertIs(SignArtifacts.__signer_class__(
            Path(r"/dummy/path/artifact.tar.gz")), 
            SignArtifactsExistingArtifactFile)
```

tests/tests_sign_workflow/test_sign_artifacts.py
dblock: It's not a must have, but my reasoning is that these tests can fail independently, and methods are cheap.

tests/tests_sign_workflow/test_sign_artifacts.py
```diff
@@ -0,0 +1,40 @@
...
+        klass = SignArtifacts.__signer_class__(path)
+
+        self.assertEqual(type(SignExistingArtifactsDir), type(klass))
+
+        path = Path(r"/dummy/path/artifact.tar.gz")
+        klass = SignArtifacts.__signer_class__(path)
+
+        self.assertEqual(type(SignArtifactsExistingArtifactFile), type(klass))
+
+    def test_from_path_method(self):
```
dblock, line 25: This ends up instantiating an instance of `Signer` and asks for git credentials. You have three options:

- patch `Signer` so it becomes a mock, this is preferred 
- make it a noop with `@patch("sign_workflow.signer.GitRepository")`, this is not recommended because it implies knowledge of what `Signer` is doing, one level too deep
- remove that instantiating from the constructor of `SignArtifacts`, which is not recommended because it will be executing `git checkout` every time we sign anything

peternied (approved):
Thanks for addressing my comments, great work enhancing this flow.

abhinavGupta16:
Thanks @dblock for helping out with the test cases

tests/test_run_sign.py
abhinavGupta16: It's fixed now

src/sign_workflow/signer.py
```diff
@@ -19,40 +18,40 @@
...
 
     def is_valid_file_type(self, file_name):
         return any(
-            x
-            in [
-                pathlib.Path(file_name).suffix,
-                "".join(pathlib.Path(file_name).suffixes),
-            ]
-            for x in Signer.ACCEPTED_FILE_TYPES
+            file_name.endswith(x) for x in Signer.ACCEPTED_FILE_TYPES
```
dblock, line 45: Use `os.path.splitext` to get the extension from the file name, then `in` to check ether it’s one of the accepted types.

dblock (approved):
Works for me.

I would also remove all the passing around of `Signer()` which is always the same class, and create an instance of it at the lowest level.

src/sign_workflow/signer.py
abhinavGupta16: That does not work since we have artifacts with names like `opensearch-1.0.0.tar.gz`. That is why I changed it to `endswith`

dblock: We could just change .tar.gz into .gz. Technically the extension of that file is not .tar.gz, it’s .gz.

Issue #1355, opened 2021-12-16
[Enhancement] Add tests and improvement to the promotion jenkins workflow


peterzhuamazon:
We are adding tar and native plugin zips auto promotion here: https://github.com/opensearch-project/opensearch-build/pull/1354

There are several improvement needed after the PR is merged:
1. Write a full test for the jenkinsfile, including gradle/groovy test case file, sample jenkinsfile, txt results.
2. Put s3upload params to its own variable.
3. Rename all the `AWS_ACCOUNT_PUBLIC` to `AWS_CENTRAL_ARTIFACT_ACCOUNT` as it is mistakenly added 1st in Jenkins with a wrong name, and rename all the entries related to it.
4. As of now it is designed to trigger multiple times by engineer on different arch/platform, we can definitely add a multi-arch trigger so we dont need to manually trigger it multiple times.
5. Reduce the number of params needed for the jenkinfile.
6. Use build parameter more as that feeds in more information for the task.

gaiksaya:
Closing this as the feature has been implemented via https://github.com/opensearch-project/opensearch-build/pull/4280

PR #1367, opened 2021-12-17
Add sha512 generation to promotion job

peternied:
### Description
The promotion workflow is picking up artifacts from a distribution build and placing them in their final resting place on the artifacts server.  This is a perfect opportunity to do any final polishing steps on artifacts that are ready for release.  Adding this hook for file actions and creating a sha512 generation closure.

This is untested on jenkins, but is the core POC needed to bootstrap the signing workflow.
 
### Issues Resolved
- https://github.com/opensearch-project/opensearch-build/issues/1351
- https://github.com/opensearch-project/opensearch-build/issues/1355
 
### Check List


vars/createSha512.groovy
```diff
@@ -0,0 +1,8 @@
+Closure call() {
```
peternied, line 1: This was a quick'n dirty method to pass around these functions calls - there might be a better way

tests/jenkins/TestPromoteArtifacts.groovy
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    public void testDefault() {
+        super.testPipeline("tests/jenkins/jobs/PromoteArtifacts_Jenkinsfile")
+    }
+
+    @Test
+    public void testWithActions() {
+        super.testPipeline("tests/jenkins/jobs/PromoteArtifacts_actions_Jenkinsfile")
+    }
```
peterzhuamazon, line 71: Nice, so we can have with and without actions testing in both.

tests/jenkins/jobs/PromoteArtifacts_actions_Jenkinsfile
```diff
@@ -0,0 +1,15 @@
+pipeline {
+    agent none
+    stages {
+        stage('promote') {
+            steps {
+                script {
+                    promoteArtifacts(
+                        manifest: 'tests/jenkins/data/opensearch-1.3.0.yml',
+                        fileActions: [createSha512()]
```
peterzhuamazon, line 9: Assuming we can expand this into `fileActions: [createSha512(), createPgpSig()]`

peterzhuamazon: #1352

vars/createSha512.groovy
peterzhuamazon: ELI5: what is the reason that we cannot use a normal `call` here?
From what I read it is meant for simplistic and able to passed as argument?
Therefore you can pass `createSha512` as an argument in promoteArtifacts?

vars/promoteArtifacts.groovy
```diff
@@ -32,22 +37,47 @@ void call(Map args = [:]) {
             String pluginSubFolder = pluginSubPath.split('/')[0]
             String pluginNameWithExt = pluginSubPath.split('/')[1]
             String pluginName = pluginNameWithExt.replace('-' + version + '.zip', '')
+            String pluginNameNoExt = pluginNameWithExt.replace('-' + version, '')
             String pluginFullPath = ['plugins', pluginName, version].join('/')
-            s3Upload(bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "releases/$pluginFullPath/", workingDir: "$WORKSPACE/artifacts/$artifactPath/builds/$filename/$pluginSubFolder/"
-                , includePathPattern: "**/${pluginName}*")
+            for (Closure action : fileActions) {
+                for (file in findFiles(glob: "$WORKSPACE/artifacts/$artifactPath/builds/$filename/$pluginSubFolder/**/${pluginName}*")) {
```
peterzhuamazon, line 43: We can merge here but as #1355 points out we need to save these params into their own var.

peterzhuamazon (approved):
Overall great addition to the promotion workflow and allow us to expand with signer and more actions later.

Thanks!

vars/promoteArtifacts.groovy
peternied: This PR doesn't resolve #1355, but it did some of the things called out there so I figured it was better to mention it.

vars/createSha512.groovy
peternied: Not quite the ELI5 level, this is just copying prior art and convention.  `call(...)` is the method that is invoked when you do `createSha512(...)`.  Its signature supports parameters, typically we accept a map of values and stuff it into an object called args.  Typically we don't return anything, making the return type void.  So this function is following the same generalized pattern, but it doesn't accept parameters (we could add that) and it returns a closure that can be used.

dblock (approved):
I like it! You definitely want to test in Jenkins because permissions might not let you do what you're trying to do :(

dblock (approved):

PR #1370, opened 2021-12-17
Generate and install openssl based self-signed certificates in docker

### LINKED ISSUE #1363: Include the certificate management tool in docker distribution
Today, the OpenSearch docker distribution comes pre-installed with demo certificates. The intention is to replace the insecure demo certificates with self-signed certificates as discussed in opensearch-project/OpenSearch#1618 (point 2).

Package the certificate management tool (provided in opensearch-project/OpenSearch#1633) in the docker distribution. While instantiating the container, check if the certificates are present in the respective mounted volume that holds the `config`. If missing, it should run the certificate management tool to create self-signed certificates with default arguments before bootstrapping OpenSearch. If present, skips the certificate generation and rely on user provided certificates and yml configurations. 

Implementation task details
- [x] Provide openSSL inside docker image
- [x] ~~replace install_demo_configurations.sh with cert-management-tool from opensearch-project/OpenSearch#1633.~~
- [x] Add one time certificate installation logic that installs certs when a new container starts. Subsequent starts/stops should not modify installed certs.
- [x] Provide a way to control self-signed certs generation based on user input.
- [x] Make changes to `docker-compose.yml` based on new dockerfile

Related to https://github.com/opensearch-project/OpenSearch/issues/1618

setiah:
Signed-off-by: Himanshu Setia <setiah@amazon.com>

### Description
1. Provides openssl in docker image for self-signed certificate generation.
2. Adds logic to generate self-signed certs on first time container start
3. Accepts user input in docker run for generating self-signed certs

todo - replace demo certificates with self-signed certs tool
 
### Issues Resolved
#1363 
 
### Check List


peternied (changes requested):
Thanks for this contribution, is there an update that we need to the website or documentation with updated instructions that needs to go in along side this change?

Is there an automated test we could make that would be sure the new pattern is working as expected?

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
```diff
@@ -57,11 +94,10 @@ export OPENSEARCH_JAVA_OPTS="-Dopensearch.cgroups.hierarchy.override=/ $OPENSEAR
 ##Security Plugin
 SECURITY_PLUGIN="opensearch-security"
 if [ -d "$OPENSEARCH_HOME/plugins/$SECURITY_PLUGIN" ]; then
-    if [ "$DISABLE_INSTALL_DEMO_CONFIG" = "true" ]; then
```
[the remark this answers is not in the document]
setiah: Good point @peternied. Updated this to use existing env vars

docker/release/dockerfiles/opensearch.al2.dockerfile
```diff
@@ -57,7 +57,7 @@ ARG OPENSEARCH_HOME=/usr/share/opensearch
 # Update packages
 # Install the tools we need: tar and gzip to unpack the OpenSearch tarball, and shadow-utils to give us `groupadd` and `useradd`.
 # Install which to allow running of securityadmin.sh
-RUN yum update -y && yum install -y tar gzip shadow-utils which && yum clean all
+RUN yum update -y && yum clean all && yum install -y tar gzip shadow-utils which openssl-devel openssl && yum clean all
```
peterzhuamazon, line 60: Why do you need to clean all then install again?

docker/release/dockerfiles/opensearch.al2.dockerfile
```diff
@@ -106,4 +105,4 @@ LABEL org.label-schema.schema-version="1.0" \
   org.label-schema.build-date="$BUILD_DATE"
 
 # CMD to run
-CMD ["./opensearch-docker-entrypoint.sh"]
+ENTRYPOINT ["./opensearch-docker-entrypoint.sh"]
```
peterzhuamazon, line 108: Please pull from main as we have already made a lot of changes to the entrypoint file at this point.

peterzhuamazon (changes requested):
Blocking this PR as we are holding on these changes for 1.2.4 OpenSearch and 1.2.0 Dashboards re-release to fix bugs, as well as waiting for @setiah to rebase with main changes that already change the dynamic of this file.

Thanks.

docker/release/dockercomposefiles/docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3'
+version: '4'
```
peterzhuamazon, line 1: This is compose file version not the version of changes:
https://docs.docker.com/compose/compose-file/compose-versioning/

Please keep it at version 3 for now.

Thanks.

docker/release/dockercomposefiles/docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
         soft: 65536 # maximum number of open files for the OpenSearch user, set to at least 65536 on modern systems
         hard: 65536
     volumes:
+      - opensearch-config:/usr/share/opensearch/config
```
peterzhuamazon, line 21: We shouldnt mount the config file version since users can choose to change them or not.
With mounting it is forcing the inode to be the same without the options to create/change to new inode.

setiah: Not sure if I understood this correctly. The reason for creating a volume mount here is to create a persistent storage for generated certs that helps re-using them across all containers, as well as make the configs agnostic to container restarts.
If a user wants to mount a custom config, they still can do so by binding custom yml to docker like 
```./custom-opensearch.yml:/usr/share/opensearch/config/opensearch.yml```

Comment (peterzhuamazon) [reply docker/release/dockercomposefiles/docker-compose.yml:22]: I still think we should leave the config file dir as it is and let the user decide the mount themselves.
If this file bound to change you also need to update the one on the website download page.

Comment (setiah) [reply docker/release/dockercomposefiles/docker-compose.yml:22]: I think I understood what you're trying to say. Let me clarify - mounting is not forcing users to use the mounted config, it is only providing a secure default option. Users can always override the default mount with a their custom config by making changes in `docker-compose.yml` (like they do today).  

Another option was to put certs in a different folder under config and mount only that folder instead of config/, but in that case `opensearch.yml` would still have references to those certs so if a user is using a custom-opensearch.yml, they would still need to override the certs mount. Another reason to not go with this option is changing certs path might break backward compatibility, and since this is intended for 1.x, that would not be desirable.

File: docker/release/dockerfiles/opensearch.al2.dockerfile (hunk 2 above)
Comment (setiah) [reply docker/release/dockerfiles/opensearch.al2.dockerfile]: Actually it doesn't clean and install all again, but just clears caches and metadata to keep the docker image size lean. Nvm, I have removed this now since i did not find any reduction in image size.

File: docker/release/dockercomposefiles/docker-compose.yml (hunk 5 above)
Comment (setiah) [reply docker/release/dockercomposefiles/docker-compose.yml]: I have changed this further. Instead of mounting the entire config, now it is just mounting the certs/. that way users can still override the `opensearch.yml` with a custom one, in the same manner as they do today.

**Comment by bbarani** [comment]
@peternied @peterzhuamazon @gaiksaya Can you please review and add your comments to this PR?

File: docker/release/dockercomposefiles/docker-compose.yml [hunk 6]
```diff
@@ -19,6 +19,7 @@ services:
         soft: 65536  # maximum number of open files for the OpenSearch user, set to at least 65536 on modern systems
         hard: 65536
     volumes:
+      - opensearch-certs:/usr/share/opensearch/config/certs
```
peterzhuamazon, line 22: If merged, this needs to be updated on our website as well.

peterzhuamazon (changes requested):
We would hold this PR until security has the new install demo script ready.
There are some issues with this change as well.
Since this is quite a bit change, will hold until security repo is clear on this.

Thanks.

peterzhuamazon:
I have heard that the demo install script is not changing anymore.
@bbarani we might just close this PR for now.

Thanks.

bbarani:
@setiah Are you good with closing this PR?

gaiksaya:
Closing this PR now. Please reopen if required! 
Thanks!

PR #1378, opened 2021-12-17
Promote min and bundle artifacts to release-candidate after every bundle build

zylzulu:
Signed-off-by: Yilin Zhang <yilizhan@amazon.com>

### Description
Uploads min and bundle artifacts for both opensearch and dashboards to the below location on every build:
```
https://artifacts.opensearch.org/release-candidates/core/opensearch/<version>/opensearch-min-<version>-linux-x64.tar.gz
https://artifacts.opensearch.org/release-candidates/bundle/opensearch/<version>/opensearch-<version>-linux-x64.tar.gz
https://artifacts.opensearch.org/release-candidates/core/opensearch-dashboards/<version>/opensearch-dashboards-min-<version>-linux-x64.tar.gz
https://artifacts.opensearch.org/release-candidates/bundle/opensearch-dashboards/<version>/opensearch-dashboards-<version>-linux-x64.tar.gz
```

 
### Check List


vars/uploadArtifacts.groovy
```diff
@@ -2,6 +2,10 @@ void call(Map args = [:]) {
     def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
 
     def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: args.manifest))
+    def yamlFile = readYaml(file: args.manifest)
+    def fileLocation = yamlFile.components.artifacts.dist[0]
+    def productName = buildManifest.build.getFilename()
+    def fileName = buildManifest.getPackageName("${productName}")
```
dblock, line 8: This seems to load the same YAML file a second time, then combine its properties. Extend `BuildManifest` to know about `components.artifacts.dist`, and implement it as a single method of `buildManifest`.

vars/uploadArtifacts.groovy
```diff
@@ -11,10 +15,15 @@ void call(Map args = [:]) {
         s3Upload(file: 'dist', bucket: "${ARTIFACT_BUCKET_NAME}", path: "${artifactPath}/dist")
     }
 
+    withAWS(role: "${ARTIFACT_PROMOTION_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_PUBLIC_REAL}", duration: 900, roleSessionName: 'jenkins-session') {
+        s3Upload(file: "builds/test-release-candidates/core/${productName}/${buildManifest.build.version}", bucket: "${ARTIFACT_PUBLIC_BUCKET_NAME}", path: "${artifactPath}/builds/${productName}/${fileLocation}")
```
dblock, line 19: Is "core" some other convention? Can we change it to `min` or something like that?

dblock (changes requested):

peternied:
Please make sure that the new output will handle these cases:
- .../release-candidates/core/opensearch/1.2.2/opensearch-min-1.2.2-linux-x64.tar.gz
- .../release-candidates/core/opensearch/1.2.2/opensearch-min-1.2.2-linux-arm64.tar.gz
- .../snapshots/core/opensearch/1.2.2-SNAPSHOT/opensearch-min-1.2.2-SNAPSHOT-linux-x64-latest.tar.gz
- .../snapshots/core/opensearch/1.2.2-SNAPSHOT/opensearch-min-1.2.2-SNAPSHOT-linux-arm64-latest.tar.gz

I think we've got the first case, but we need the snapshot ones too

vars/uploadArtifacts.groovy
zylzulu: That's the original file system directory structure. @peterzhuamazon Is it ok if we change it to `min`?

vars/uploadArtifacts.groovy
```diff
@@ -11,10 +15,15 @@ void call(Map args = [:]) {
...
+        s3Upload(file: "builds/test-release-candidates/bundle/${productName}/${buildManifest.build.version}", bucket: "${ARTIFACT_PUBLIC_BUCKET_NAME}", path: "${artifactPath}/dist/${productName}/${fileName}")
```
zelinh, line 20: It seems like when you upload it, it will be iterating through each folder named `version`, `platform`, `architecture` since they are joint by `/`.

src/jenkins/BuildManifest.groovy
```diff
@@ -49,6 +49,16 @@ class BuildManifest implements Serializable {
...
+
+    public String getPackageName(String productName) {
+        String packagePrefix = [
+            productName,
+            this.build.version,
+            this.build.platform,
+            this.build.architecture,       
+        ].join('/')
+        return packagePrefix + '.tar.gz'
+    }
```
zelinh, line 61: Can we use the name with version, platform, architecture and distribution from [here](https://github.com/opensearch-project/opensearch-build/blob/ce1547934041707af62ff54dd5a21d1c8ddaf2fc/src/jenkins/InputManifest.groovy#L52-L56)?

dblock:
Can we try to move this forward, please?

vars/uploadArtifacts.groovy
zylzulu: thank you for pointing it out! they should be joint by "-"

src/jenkins/BuildManifest.groovy
```diff
@@ -49,6 +63,14 @@ class BuildManifest implements Serializable {
             this.getArtifactRoot(jobName, buildNumber)
         ].join('/')
     }
-}
-
```
dblock, line 66: I believe the manifest contains the product name, doesn’t it?

src/jenkins/BuildManifest.groovy
```diff
@@ -27,10 +27,24 @@ class BuildManifest implements Serializable {
...
 
+    class Components implements Serializable {
+        String dist
+
+        Components(Map data) {
+            this.dist = data.artifacts.dist
+        }
+
+        String getDistPackageLocation() {
+            return this.dist[0]
```
dblock, line 38: Let’s be a little defensive and raise an error if dist has 0 or more than 1 item?

dblock (commented):
Add tests for the new functions at least, please.

src/jenkins/BuildManifest.groovy
zylzulu: Yes. I've updated the `getPackageName()` by using `this.buid.getFilename()`.

src/jenkins/BuildManifest.groovy
```diff
@@ -49,6 +68,14 @@ class BuildManifest implements Serializable {
             this.getArtifactRoot(jobName, buildNumber)
         ].join('/')
     }
-}
-
 
+    public String getPackageName() {
+        String packagePrefix = [
+            this.buid.getFilename(),
```
dblock, line 74: Typo here.

since this gets all properties from build, the method should be implemented on build as well

dblock:
@zylzulu will review when the build is green and the CR is ready

src/jenkins/BuildManifest.groovy
```diff
@@ -27,10 +27,29 @@ class BuildManifest implements Serializable {
...
+        String dist
+
+        Components(Map data) {
+            this.dist = data.artifacts.dist
+        }
+
+        String getDistPackageLocation() {
+            try {
+                return this.dist[0]
+            } catch (Exception e) {
```
dblock, line 40: This just swallows the error. What we want is a better error.

vars/uploadArtifacts.groovy
peterzhuamazon: stay on core

gaiksaya:
> Please make sure that the new output will handle these cases:
> 
>     * .../release-candidates/core/opensearch/1.2.2/opensearch-min-1.2.2-linux-x64.tar.gz
> 
>     * .../release-candidates/core/opensearch/1.2.2/opensearch-min-1.2.2-linux-arm64.tar.gz
> 
>     * .../snapshots/core/opensearch/1.2.2-SNAPSHOT/opensearch-min-1.2.2-SNAPSHOT-linux-x64-latest.tar.gz
> 
>     * .../snapshots/core/opensearch/1.2.2-SNAPSHOT/opensearch-min-1.2.2-SNAPSHOT-linux-arm64-latest.tar.gz
> 
> 
> I think we've got the first case, but we need the snapshot ones too

Hi @peternied , @dblock,
currently the min and bundle are being uploaded to `build/test-release-candidate` folder (I'll change to `release-candidates` before I merge my code).
Regarding snapshots from where do I retrieve them?

dblock:
> > Please make sure that the new output will handle these cases:
> > ```
> > * .../release-candidates/core/opensearch/1.2.2/opensearch-min-1.2.2-linux-x64.tar.gz
> > 
> > * .../release-candidates/core/opensearch/1.2.2/opensearch-min-1.2.2-linux-arm64.tar.gz
> > 
> > * .../snapshots/core/opensearch/1.2.2-SNAPSHOT/opensearch-min-1.2.2-SNAPSHOT-linux-x64-latest.tar.gz
> > 
> > * .../snapshots/core/opensearch/1.2.2-SNAPSHOT/opensearch-min-1.2.2-SNAPSHOT-linux-arm64-latest.tar.gz
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think we've got the first case, but we need the snapshot ones too
> 
> Hi @peternied , @dblock, currently the min and bundle are being uploaded to `build/test-release-candidate` folder (I'll change to `release-candidates` before I merge my code. Not making any code changes since I can use this for testing ). Regarding snapshots from where do I retrieve them? Also I believe snapshots only applies to min here

Check out https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L50, which calls publish-snapshot to Maven. You'll probably want to publish around there.

gaiksaya:
> > > Please make sure that the new output will handle these cases:
> > > ```
> > > * .../release-candidates/core/opensearch/1.2.2/opensearch-min-1.2.2-linux-x64.tar.gz
> > > 
> > > * .../release-candidates/core/opensearch/1.2.2/opensearch-min-1.2.2-linux-arm64.tar.gz
> > > 
> > > * .../snapshots/core/opensearch/1.2.2-SNAPSHOT/opensearch-min-1.2.2-SNAPSHOT-linux-x64-latest.tar.gz
> > > 
> > > * .../snapshots/core/opensearch/1.2.2-SNAPSHOT/opensearch-min-1.2.2-SNAPSHOT-linux-arm64-latest.tar.gz
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I think we've got the first case, but we need the snapshot ones too
> > 
> > 
> > Hi @peternied , @dblock, currently the min and bundle are being uploaded to `build/test-release-candidate` folder (I'll change to `release-candidates` before I merge my code. Not making any code changes since I can use this for testing ). Regarding snapshots from where do I retrieve them? Also I believe snapshots only applies to min here
> 
> Check out https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L50, which calls publish-snapshot to Maven. You'll probably want to publish around there.

Looking at the above code:
The [buildManifest](https://github.com/opensearch-project/opensearch-build/blob/main/vars/buildManifest.groovy) just builds the snapshot but doesn't upload it anywhere. Its the publish-snapshot that uploads the artifacts to maven.

https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L45-L47

Should I directly upload from there to s3? Also what about arm64? While I ran this to see the generated artifacts I did not see anything generated for arm64 here.

dblock:
@gaiksaya I think it's perfectly OK to do the snapshot uploading as part of another PR, and we can discuss options there; (2) seems like the better thing to do

gaiksaya:
> @gaiksaya I think it's perfectly OK to do the snapshot uploading as part of another PR, and we can discuss options there; (2) seems like the better thing to do

Created the issue https://github.com/opensearch-project/opensearch-build/issues/1442 to address the min snapshots uploading part

src/jenkins/BuildManifest.groovy
```diff
@@ -25,12 +25,56 @@ class BuildManifest implements Serializable {
...
         }
+
+        String getPackageName() {
+            String packagePrefix = [
+                this.getFilename(),
+                this.version,
+                this.platform,
+                this.architecture,
+            ].join('-')
+            return packagePrefix + '.tar.gz'
```
dblock, line 36: Nit: don't introduce variables where they aren't reused.

```groovy
String getPackageName() {
            return [
                this.getFilename(),
                this.version,
                this.platform,
                this.architecture,
            ].join('-') + '.tar.gz'
}
```

src/jenkins/BuildManifest.groovy
```diff
@@ -25,12 +25,56 @@ class BuildManifest implements Serializable {
...
+
+
+
+    class Components extends HashMap<String , Component> {
+
+        Components(ArrayList data) {
+            data.each {
+                item ->
+                    Component component = new Component(item)
+                    (this[component.name] = component)
```
dblock, line 48: The parenthesis are unnecessary. You're assigning the component at the key `component.name`. You could also add a method called `add(Component component)` and the code becomes cleaerer:

```groovy
Components(ArrayList data) {
   data.each { item-> this.add(new Component(item)) }
}

public void add(Component component) {
    this[component.name] = component
}
```

src/jenkins/BuildManifest.groovy
```diff
@@ -25,12 +25,56 @@ class BuildManifest implements Serializable {
...
+                this.getFilename(),
+                this.version,
+                this.platform,
+                this.architecture,
+            ].join('-')
+            return packagePrefix + '.tar.gz'
+        }
+    }
+
+
```
dblock, line 40: Fixup too many spaces ...

tests/jenkins/jobs/BuildManifest_Jenkinsfile
```diff
@@ -22,8 +22,10 @@ pipeline {
                     echo buildManifest.build.getFilename()
                     echo buildManifest.getArtifactRoot('bundle-build', '1')
                     echo buildManifest.getArtifactRootUrl('https://ci.opensearch.org/ci/dbc', 'bundle-build', '1')
+                    echo buildManifest.build.getPackageName()
+                    echo buildManifest.components.get("OpenSearch").artifacts.get("dist").first()
```
dblock, line 26: This should be extracted into a method, I want to be able to write:

```
buildManifest.components.getMinArtifact()
```

The implementation of `getMinArtifact` should know which component to get (the first one? the one that matches the name of the build artifact?) and raise errors if such a component cannot be found, or artifacts is missing, or the artifact named "dist" is missing, so it's easier to debug.

vars/uploadArtifacts.groovy
```diff
@@ -2,6 +2,9 @@ void call(Map args = [:]) {
     def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
 
     def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: args.manifest))
+    def fileLocation = buildManifest.components.get("OpenSearch").artifacts.get("dist").first()
```
dblock, line 5: This is the min artifact location, so probably `def minArtifactPath`?

vars/uploadArtifacts.groovy
```diff
@@ -2,6 +2,9 @@ void call(Map args = [:]) {
...
+    def productName = buildManifest.build.getFilename()
```
dblock, line 6: productFilename

vars/uploadArtifacts.groovy
```diff
@@ -2,6 +2,9 @@ void call(Map args = [:]) {
...
+    def fileName = buildManifest.build.getPackageName()
```
dblock, line 7: packageName

dblock (changes requested):

src/jenkins/BuildManifest.groovy
gaiksaya: Updated

src/jenkins/BuildManifest.groovy
gaiksaya: Removed the parenthesis for now

src/jenkins/BuildManifest.groovy
gaiksaya: Fixed

vars/uploadArtifacts.groovy
gaiksaya: Changed

dblock (changes requested):

src/jenkins/BuildManifest.groovy
```diff
@@ -25,30 +25,76 @@ class BuildManifest implements Serializable {
...
+                Component component = new Component(item)
+                this[component.name] = component
+            }
+        }
+
+        public String getMinArtifact(){
+            try {
+                this.get("OpenSearch").artifacts.get("dist").first()
+            } catch (Exception ex){
+                echo("Min artifact cannot be found", ex.getMessage())
```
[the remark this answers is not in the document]
gaiksaya: Name is the only thing that will distinguish them. However, I checked their build manifest and the build.name is a way to differentiate between opensearch and dashboards. How about this:

```
        public String getMinArtifact(){
            if (build.name == "OpenSearch") {
                this.get("OpenSearch")?.artifacts?.get("dist")?.first()
            } else if (build.name == "OpenSearch Dashboards"){
                this.get("OpenSearch-Dashboards")?.artifacts?.get("dist")?.first()
            }
        }
```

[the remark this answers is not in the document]
dblock: Since it's a function it should just either return `null` or raise an error.

I would just do `this.get(build.name.replace(' ', '-'))?....` and call it a day.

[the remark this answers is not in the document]
gaiksaya: done! Thanks!

dblock (approved):

peterzhuamazon (approved):

PR #1387, opened 2021-12-20
Remove plugins that still need a log4j version bump.

dblock:
Signed-off-by: dblock <dblock@dblock.org>

### Description

Coming from https://github.com/opensearch-project/sql/pull/345, remove the plugins that haven't had the version bump for log4j to get a 1.2.3-SNAPSHOT with https://github.com/opensearch-project/OpenSearch/pull/1774.

Running `./build.sh manifests/1.2.3/opensearch-1.2.3.yml  --snapshot` passes with this change.

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1389, opened 2021-12-21
Revert "Remove plugins that still need a log4j version bump."

dblock:
Reverts opensearch-project/opensearch-build#1387, re-adds sql and PA plugins.


dblock:
GRRR DCO

Issue #1383, opened 2021-12-19
Add support to add signer process to artifacts promotions


abhinavGupta16:
### Description
The artifacts promotion to production for opensearch should be signed and their signatures added before the items are uploaded on s3 buckets

### Acceptance Criteria
- [ ] When I run the job to promote artifacts from staging to production, the artifacts should be signed and uploaded with their corresponding `.sig` files created.

PR #1395, opened 2021-12-21
Integrate signing with artifact promotion jenkins job 

### LINKED ISSUE #1351: [META] Automate artifact signing with OpensearchSignerClient for linux

abhinavGupta16:
### Description
Integrate signing with artifact promotion jenkins job. Created a groovy function for jenkins job which can also be used for jenkins job to sign standalone artifacts
 
### Issues Resolved
#1351
#1383
 
### Check List


vars/signArtifacts.groovy
```diff
@@ -0,0 +1,25 @@
...
+    environment {
+        // These ENV variables are required by https://github.com/opensearch-project/opensearch-signer-client
+        // This client is invoked internally by the sign script.
+        ROLE = "${SIGNER_CLIENT_ROLE}"
+        EXTERNAL_ID = "${SIGNER_CLIENT_EXTERNAL_ID}"
+        UNSIGNED_BUCKET = "${SIGNER_CLIENT_UNSIGNED_BUCKET}"
+        SIGNED_BUCKET = "${SIGNER_CLIENT_SIGNED_BUCKET}"
+    }
+
+    if( !fileExists("$WORKSPACE/sign.sh")) {
```
dblock, line 12: Is the if necessary? Doesn’t it do a `git update` if the code is checked out? Without it you’d get stale files.

vars/signArtifacts.groovy
```diff
@@ -0,0 +1,25 @@
...
+        git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
+    }
+
+    if( !fileExists("$WORKSPACE/opensearch.pgp")) {
```
dblock, line 16: Does it hurt to reimport? We probably need to do it anyway if the key were to change.

dblock (commented):
Needs to actually be used by a Jenkins workflow plus a unit test.

vars/signArtifacts.groovy
abhinavGupta16: yes. We would want to ensure to use the same code for the entire signing process.

vars/signArtifacts.groovy
abhinavGupta16: Same reason, we would want to use the same key for one running instance of a job

vars/signArtifacts.groovy
```diff
@@ -0,0 +1,33 @@
...
+        sh('curl https://artifacts.opensearch.org/publickeys/opensearch.pgp -o $WORKSPACE/opensearch.pgp')
```
peterzhuamazon, line 25: Use `-SL` so that if the original link is redirected, it will still find the file.

vars/signArtifacts.groovy
peterzhuamazon: I get what Abhinav is saying as he is trying to prevent updating main branch if any new code is being committed during the run.

vars/signArtifacts.groovy
peterzhuamazon: Same as above.

vars/signArtifacts.groovy
dblock: I am very confused, maybe someone can explain how the `git` call in a Jenkins pipeline work at all? What I really want is that every time we execute the pipeline, it checks out the same opensearch-build code across all pipeline stages. Maybe we need to check it out at a revision that's being built and not just `main` to make it stable?

- If I run the job, then update sign.sh, then run the job again, doesn't that preserve the workspace and isn't the second run using stale code without this update? How do I ever update sign.sh?
- Aren't docker containers always supposed to give me a clean, throw-away environment, in which case this is always `false` when invoked over multiple stages?
- If I have a job that runs in multiple stages in multiple docker containers, am I still getting different code by having `git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'` with this `if`?

peterzhuamazon:
As of now, the code does not run tests correctly due to the changes we have on @peternied sha generating task.
We will add the test case corrections later.

Thanks.

vars/signArtifacts.groovy
abhinavGupta16: - All the jenkins code is run in docker. Anything that is ever created will be deleted after the job completes it's run.
- The condition will not be always in the case of `promoteArtifacts.groovy` as you see, it is repeatedly being called as an action to sign the artifacts before uploading. So, here for example, suppose the job takes an hour to complete, and someone updates the `main`  branch in the meanwhile, it would then start using a different version of code to sign which is not desirable.

dblock: Isn't it simpler to check out the tooling once in the beginning of the job?

Either way this method won't solve runs across stages.

vars/signArtifacts.groovy
dblock: This should be done once in the job, not every time we try to sign something. Same as above, neither will solve runs across stages.

dblock:
> As of now, the code does not run tests correctly due to the changes we have on @peternied sha generating task. We will add the test case corrections later.
> 
> Thanks.

I see half a dozen other issues, happy to CR when this is ready.

vars/createSha512.groovy
```diff
@@ -1,15 +1,55 @@
...
+            if [ -d "$argsMap.artifactPath" ]
+            then
+                for file_sub_path in $(find $argsMap.artifactPath -type f)
+                do
+                    for file_type in \$file_type_list
+                    do
+                        if echo \$file_sub_path | grep -q \$file_type
+                        then
+                            echo Generating sha for \$file_sub_path
+                             (echo -n $(sha512sum \$file_sub_path | awk "{print $1}") && echo  " $(basename \$file_sub_path)") > \$file_sub_path.sha512
```
peterzhuamazon, line 37: The sha is including the whole path not just the basename, which is an issue we need to fix.
The code here is not running as groovy is complaining about the layers.

vars/signArtifacts.groovy
abhinavGupta16: Yes, but checking it out here makes this function self-sufficient to `signArtifacts()` and would not rely on the job calling it to do so. I can look into the commit id var

vars/signArtifacts.groovy
abhinavGupta16: Again, we want to make this function self sufficient and not rely on the calling functions for a setup

vars/createSha512.groovy
abhinavGupta16: I think we need to eventually make this only groovy eventually

peterzhuamazon: fixed.

vars/createSha512.groovy
```diff
@@ -3,62 +3,24 @@ Closure call() {
...
-            fi
-        """
+        for (file in foundFiles) {
+            acceptTypeFound = false
+            for (fileType in allowedFileTypes) {
+                if (file.endsWith(fileType)) {
+                    echo("Creating sha for ${file}")
+                    final sha512 = sh(script: "sha512sum ${file}", returnStdout: true).split()
+                    // Reading the first index of array since sha512sum return [shasum, filename] as output
+                    writeFile file: "${file}_groovy_.sha512", text: "${sha512[0]}"
```
peterzhuamazon, line 15: The output of sha512sum has to be:
```
hashstring filename
```

Seems like you are only saying hashstring to the file?

abhinavGupta16: Yes, saving only the hash on the file. That's what we want right?

vars/createSha512.groovy
```diff
@@ -1,15 +1,26 @@
...
+                    echo("Creating sha for ${file}")
+                    final sha512 = sh(script: "sha512sum ${file}", returnStdout: true).split()
+                    // Reading the first index of array since sha512sum return [shasum, filename] as output
+                    writeFile file: "${file}.sha512", text: "${sha512[0]}"
+                    acceptTypeFound = true
+                    break
+                }
+            }
+            if (!acceptTypeFound) {
+                echo("Not generating sha for ${argsMap.artifactPath}, doesn't match allowed types ${allowedFileTypes}")
```
dblock, line 21: Message should probably say "none of the files matched ..."

vars/promoteArtifacts.groovy
```diff
@@ -29,55 +28,57 @@ void call(Map args = [:]) {
...
+    String coreFullPath = ['core', filename, version].join('/')
+    String bundleFullPath = ['bundle', filename, version].join('/')
+    for (Closure action : fileActions) {
+        for (file in findFiles(glob: "**/${filename}-min-${version}*.tar.gz,**/${filename}-${version}*.tar.gz")) {
+            argsMap['artifactPath'] = "$WORKSPACE" + "/" + file.getPath()
+            action(argsMap)
+        }
+    }
+
+
```
dblock, line 52: A lot of empty lines, remove, and maybe just do a `println` instead of `Uploading Artifacts`

On that note, elsewhere we use `echo` and not println, consolidate.

vars/promoteArtifacts.groovy
```diff
@@ -29,55 +28,57 @@ void call(Map args = [:]) {
...
+
+    //////////// Uploading Artifacts
     withAWS(role: "${ARTIFACT_PROMOTION_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_ARTIFACT}", duration: 900, roleSessionName: 'jenkins-session') {
-        // Core Plugins
-        println("Start Core Plugin Promotion to artifects.opensearch.org Bucket")
         List<String> corePluginList = buildManifest.components.artifacts."core-plugins"[0]
         for (String pluginSubPath : corePluginList) {
             String pluginSubFolder = pluginSubPath.split('/')[0]
```
dblock, line 59: This (and possibly other assignments below) are no longer used, remove.

OpenSearch Dashboards doesn't have `core-plugins`, what's going to happen here?

dblock (changes requested):
I am confused by the naming of `createSha512` vs. `createSignatureFiles` vs. `signArtifacts`. 

1. Why do we need both `createSignatureFiles` and `signArtifacts`? One just calls the other.  
2. If `createSha512` creates .sha512 files, then the second method should be called `createSig` (it creates .sig files, right?) to match. The other option could be `createSha512Checksums` and `createSignatureFiles`.

I was under the impression that signing calls out to an internal signing service. Is that no longer the case? Why are we importing the GPG key? How do those things relate?

I still have an issue with the attempt to only import the PGP key once. I understand that `signArtifacts` tries to be independent, but I'd rather have a separate function called `importPGPKey()` that needs to run once before `signArtifacts`, which would make it so much clearer about what's going on. And since signing is done with the said PGP key, maybe their names could match, e.g. `importPGPKey()` and `signWithPGPKey()`.

Why are the changes to docker images required?

The documentation needs to be revised in this PR to explain what's going on, start with https://github.com/opensearch-project/opensearch-build#signing-artifacts.

abhinavGupta16:
> Overall this is good, but still needs some work.
> 
> I am confused by the naming of `createSha512` vs. `createSignatureFiles` vs. `signArtifacts`.
> 
> 1. Why do we need both `createSignatureFiles` and `signArtifacts`? One just calls the other.
> 2. If `createSha512` creates .sha512 files, then the second method should be called `createSig` (it creates .sig files, right?) to match. The other option could be `createSha512Checksums` and `createSignatureFiles`.
> 
> I was under the impression that signing calls out to an internal signing service. Is that no longer the case? Why are we importing the GPG key? How do those things relate?
> 
> I still have an issue with the attempt to only import the PGP key once. I understand that `signArtifacts` tries to be independent, but I'd rather have a separate function called `importPGPKey()` that needs to run once before `signArtifacts`, which would make it so much clearer about what's going on. And since signing is done with the said PGP key, maybe their names could match, e.g. `importPGPKey()` and `signWithPGPKey()`.
> 
> Why are the changes to docker images required?
> 
> The documentation needs to be revised in this PR to explain what's going on, start with https://github.com/opensearch-project/opensearch-build#signing-artifacts.

1. One is the groovy file which can be used in jobs. The other is the closure function is to use with `promoteArtifacts.groovy`
2. I'll refactor the function names 

We call the opensearch signer client, but it does not get the gpg as per my understanding. Every usage that I have come across thus far has imported the gpg key.

We can call the gpg key download from the same file as a function.

vars/promoteArtifacts.groovy
abhinavGupta16: @peterzhuamazon

vars/promoteArtifacts.groovy
abhinavGupta16: Maybe we should start using println instead of echo.... using more groovy?

vars/createSha512.groovy
abhinavGupta16: The path could be directory or the artifact

vars/promoteArtifacts.groovy
peterzhuamazon: We will need to add a condition to push core-plugins for OpenSearch only, the rest are for both the opensearch/dashboards.

vars/createSha512.groovy
peterzhuamazon: no, we want hash then 2 spaces then only the basename of the file, see #1406.

vars/promoteArtifacts.groovy
dblock: This tells me that `core-plugins` is one of the things to publish, and needs to come in as an argument instead of being hard-coded. Provide the list of the components to publish from the caller?

vars/createSha512Checksums.groovy
```diff
@@ -0,0 +1,26 @@
...
+        for (file in foundFiles) {
+            acceptTypeFound = false
+            for (fileType in allowedFileTypes) {
+                if (file.endsWith(fileType)) {
+                    echo("Creating sha for ${file}")
+                    final sha512 = sh(script: "sha512sum ${file}", returnStdout: true).split()
+                    // Reading the first index of array since sha512sum return [shasum, filename] as output
+                    writeFile file: "${file}.sha512", text: "${sha512[0]}"
+                    acceptTypeFound = true
+                    break
```
dblock, line 17: Passing the allowed file types into `find` would simplify this a lot. See https://stackoverflow.com/questions/2621513/how-to-use-find-command-to-find-all-files-with-extensions-from-list.

peternied (commented):
The promotion workflow was already complication, and I think this PR is suffering with complexity is being added there isn't clear purpose to each component.  My recommendation is that we have a promotion artifact selector, activities, coordinator and a workflow - which aligns closely to what is already on origin/main.

* Artifact selection finds all artifacts that should be promoted, tosses them into a list
* Activities take in a file path and product additional artifacts that will be included in the upload, not folders
  * CreateSha512, path to file -> sha512 next to file in the workspace
  * CreateSignature, path to file -> signature next to file in the workspace
* Coordinator takes an input bucket, output bucket, selector, ordered list of activity to download everything into the workspace, run a selector(s) on it, run activities on the select artifacts, and upload the artifacts to the output bucket.
  * This code handles the management of the workspace environment, so files vs folders can be handled here with loops rather than making activities more complicated

As @dblock mentioned, we could add further parameterize the inputs to make this work for for any kind of promotion workflow.  Consider if we re-wrote the S3 upload aspect as an activity that is run last.  This might be in scope for the current change or another follow up item.

As there might be additional trade-offs around start-up/tear down time associated within signing system, rather than address them in this PR it might make more sense to circle back if performance is a concern, - even if this workflow takes an hour its a vast improvement from the current process.

Thanks to everyone for the dedication and hardwork on this PR - Merry Christmas!

vars/createSha512Checksums.groovy
abhinavGupta16: I agree. However, I think the current implementation also is good, since it mentions the files that are skipped. It helped to keep a check if anything is skipped when we ran the job.

abhinavGupta16:
> The promotion workflow was already complication, and I think this PR is suffering with complexity is being added there isn't clear purpose to each component. My recommendation is that we have a promotion artifact selector, activities, coordinator and a workflow - which aligns closely to what is already on origin/main.
> 
> * Artifact selection finds all artifacts that should be promoted, tosses them into a list
> * Activities take in a file path and product additional artifacts that will be included in the upload, not folders
>   
>   * CreateSha512, path to file -> sha512 next to file in the workspace
>   * CreateSignature, path to file -> signature next to file in the workspace
> * Coordinator takes an input bucket, output bucket, selector, ordered list of activity to download everything into the workspace, run a selector(s) on it, run activities on the select artifacts, and upload the artifacts to the output bucket.
>   
>   * This code handles the management of the workspace environment, so files vs folders can be handled here with loops rather than making activities more complicated
> 
> As @dblock mentioned, we could add further parameterize the inputs to make this work for for any kind of promotion workflow. Consider if we re-wrote the S3 upload aspect as an activity that is run last. This might be in scope for the current change or another follow up item.
> 
> As there might be additional trade-offs around start-up/tear down time associated within signing system, rather than address them in this PR it might make more sense to circle back if performance is a concern, - even if this workflow takes an hour its a vast improvement from the current process.
> 
> Thanks to everyone for the dedication and hardwork on this PR - Merry Christmas!

We have modified the promoteArtifacts to first sign and create sha for everything we have and then upload everything at last in this PR itself. We use the directory structure to save time when creating signatures. This is much more flexible than giving only the path to artifact as input. 

I agree that we should extend this for any kind of promotion, can create an issue and work on it on a different PR.

peternied:
> We use the directory structure to save time when creating signatures.

I have strong preferences towards maintainability and testability, making trade-offs for performance can be useful, what are the performance metrics for this version?  I am comfortable if this workflow takes an hour - what are the criteria you are using, how would my proposal change the performance?

abhinavGupta16:
> > We use the directory structure to save time when creating signatures.
> 
> I have strong preferences towards maintainability and testability, making trade-offs for performance can be useful, what are the performance metrics for this version? I am comfortable if this workflow takes an hour - what are the criteria you are using, how would my proposal change the performance?

If we don't follow the directory structure, it would involve bringing up the pipenv, cloning the signer, running bootstrap, signing the artifact. Therefore, running the signer on a directory would save us a lot of time. 
Although, I am not sure how maintainability or testability would be affected if we sign or create `.sha`  for all files in a directory?  It rather makes our functionality more flexible as iterating through the dir files is also managed as a black box.

vars/createSha512.groovy
abhinavGupta16: okay, I'll make the changes accordingly.

peterzhuamazon:
> > > We use the directory structure to save time when creating signatures.
> > 
> > 
> > I have strong preferences towards maintainability and testability, making trade-offs for performance can be useful, what are the performance metrics for this version? I am comfortable if this workflow takes an hour - what are the criteria you are using, how would my proposal change the performance?
> 
> If we don't follow the directory structure, it would involve bringing up the pipenv, cloning the signer, running bootstrap, signing the artifact. Therefore, running the signer on a directory would save us a lot of time. Although, I am not sure how maintainability or testability would be affected if we sign or create `.sha` for all files in a directory? It rather makes our functionality more flexible as iterating through the dir files is also managed as a black box. The process flow works for single files as well.

Saving all the files in a dedicated folder then sign/sha all of them in one take is definitely better. Not only this is easily extensible later on if we need to sign/sha more files, but also simplify the setups of the s3uploads. We can have one s3uploads to replace the current 3.

vars/createSha512Checksums.groovy
dblock: The current implementation is twice as long and has poor separation of concerns (finding what to sign and signing it). Compare with:

```
final foundFiles = sh(.... allowedFileTypes)

for (file in foundFiles) {
  echo "Creating sha ..."
  final sha512 = 
  writeFile
}
```

vars/signArtifacts.groovy
dblock: This function is very far from self-sufficient. If you wanted to make it self sufficient, you'd have to apply the same argument further, and also install other tools here, like gpg and check whether they are installed. You're making code harder to maintain by spreading setup all over the place.

abhinavGupta16: Well... if we start adding checks for tools installation, no code will ever be self sufficient. 

Trying to understand your point here since setting up the gpg keys is only used by this function here. If we don't do it here, every function would have to call this function anyways before calling `signArtifacts`. I think that make the code and implementation more difficult to maintain. thoughts?

vars/createSha512Checksums.groovy
abhinavGupta16: How will we know the files that were skipped with this implementation?

vars/signArtifacts.groovy
dblock: First, we can agree to disagree, not a must have for me.

> If we don't do it here, every function would have to call this function anyways before calling signArtifacts.

Every _workflow_ that needs to do signing will need to install the GPG keys, not every function. To me, it's easier to maintain. What you call "self sufficient" sounds like "kitchen sink" to me, the code tries to do too much.

tests/jenkins/jobs/PromoteArtifacts_actions_Jenkinsfile.txt
```diff
@@ -4,103 +4,83 @@
...
                   promoteArtifacts.legacySCM(groovy.lang.Closure)
                   promoteArtifacts.library({identifier=jenkins@20211123, retriever=null})
                   promoteArtifacts.readYaml({file=tests/jenkins/data/opensearch-1.3.0.yml})
                   InputManifest.asBoolean()
                   promoteArtifacts.withAWS({role=downloadRoleName, roleAccount=publicAccount, duration=900, roleSessionName=jenkins-session}, groovy.lang.Closure)
                      promoteArtifacts.s3Download({bucket=artifact-bucket, file=workspace/artifacts, path=vars-build/1.3.0/33/x64/linux/, force=true})
                   promoteArtifacts.readYaml({file=artifacts/vars-build/1.3.0/33/x64/linux/builds/opensearch/manifest.yml})
+                  createSha512.sh({script=find workspace/artifacts/vars-build/1.3.0/33/x64/linux/builds/opensearch/core-plugins -type f, returnStdout=true})
+                  createSha512.echo(Not generating sha for workspace/artifacts/vars-build/1.3.0/33/x64/linux/builds/opensearch/core-plugins, doesn't match allowed types [.tar.gz, .zip])
```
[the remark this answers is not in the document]
abhinavGupta16: Good spot. Fixed it now

tests/jenkins/jobs/PromoteArtifacts_actions_Jenkinsfile.txt
```diff
@@ -4,103 +4,83 @@
...
+            export UNSIGNED_BUCKET=signer_client_unsigned_bucket
+            export SIGNED_BUCKET=signer_client_signed_bucket
+
+            workspace/sign.sh workspace/artifacts/vars-build/1.3.0/33/x64/linux/builds/opensearch/core-plugins --sigtype=.sig --component=null --type=null
+        )
+                  promoteArtifacts.findFiles({glob=**/opensearch-min-1.3.0*.tar.gz,**/opensearch-1.3.0*.tar.gz})
+                  promoteArtifacts.getPath()
+                  createSha512.sh({script=find workspace/workspace/file/found.zip -type f, returnStdout=true})
+                  createSha512.echo(Not generating sha for workspace/workspace/file/found.zip, doesn't match allowed types [.tar.gz, .zip])
+                  createSha512.echo(Not generating sha for workspace/workspace/file/found.zip, doesn't match allowed types [.tar.gz, .zip])
```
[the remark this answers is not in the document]
abhinavGupta16: Good spot. Fixed it now

vars/signArtifacts.groovy
abhinavGupta16: Sorry for the terminology there, I meant workflow when I said function there.

Yes, every workflow who would be signing would need to install the gpg keys. Therefore, it would make sense to create the keys inside of signing right? Otherwise, it would be easy to miss creating them for anyone who may use signing in future.

Let's agree to disagree I guess.

dblock: > Sorry for the terminology there, I meant workflow when I said function there.
> 
> Yes, every workflow who would be signing would need to install the gpg keys. Therefore, it would make sense to create the keys inside of signing right? Otherwise, it would be easy to miss creating them for anyone who may use signing in future.

No, for all the other reasons above.

vars/createSha512Checksums.groovy
dblock: You don't need to know what's skipped, you only need to know which files to sign.

vars/createSha512Checksums.groovy
```diff
@@ -0,0 +1,30 @@
...
+                }
+            }
+            if (!acceptTypeFound) {
+                if(foundFiles.length == 1){
+                    echo("Not generating sha for ${file} with artifact Path ${argsMap.artifactPath}, doesn't match allowed types ${allowedFileTypes}")
+                } else {
+                    echo("Not generating sha for ${file} in ${argsMap.artifactPath}, doesn't match allowed types ${allowedFileTypes}")
+                }
+            }
+        }
```
dblock, line 27: I don't understand what's going on here.

dblock (approved):
Will let @peternied take it from here, you have all my feedback.

vars/createSha512Checksums.groovy
abhinavGupta16: The path can be either a file or a directory. So, I created an if condition to print the appropriate message

vars/createSha512Checksums.groovy
```diff
@@ -11,8 +11,11 @@ Closure call() {
                 if (file.endsWith(fileType)) {
                     echo("Creating sha for ${file}")
                     final sha512 = sh(script: "sha512sum ${file}", returnStdout: true).split()
-                    // Reading the first index of array since sha512sum return [shasum, filename] as output
-                    writeFile file: "${file}.sha512", text: "${sha512[0]}"
+                    //sha512 is an array [shasum, filename]
+                    final basename = sh(script: "basename ${sha512[1]}", returnStdout: true).split()
```
peterzhuamazon, line 15: Why do you need to split for the output?
Isnt `basename` will only return one entry which is the filename.zip?
Thanks.

peterzhuamazon:
Also, pytest issues.

vars/createSha512Checksums.groovy
abhinavGupta16: The first `sha512` would contain `sha /path/to/file.zip`
The basname is extracts `file.zip` from `sha /path/to/file.zip`

abhinavGupta16: I 

> Also, pytest issues.

Did you mean groovy tests? I updated those for shafile changes to be as robust as possible

abhinavGupta16:
@peterzhuamazon - Added changes for sha512 file text to be 
``

> Also, pytest issues.

Did you mean groovy tests? I updated those for shafile changes to be as robust as possible

abhinavGupta16:
> Also, pytest issues.

Did you mean groovy tests? I updated those for shafile changes to be as robust as possible

vars/createSha512Checksums.groovy
peterzhuamazon: Then you dont need the second split for basename.

peterzhuamazon:
> > Also, pytest issues.
> 
> Did you mean groovy tests? I updated those for shafile changes to be as robust as possible

Yes, groovy. Thanks.

abhinavGupta16:
> Will let @peternied take it from here, you have all my feedback.

@peternied is on vacation this week. @peterzhuamazon @bbarani can you help review?

peterzhuamazon (approved):
Thanks.

peterzhuamazon:
Even though we have approved this I also want @peternied to take a look before we merge.
Thanks.

PR #1407, opened 2021-12-23
Add new manifests to crontab.

dblock:

### Description

When a new manifest is added, such as in https://github.com/opensearch-project/opensearch-build/pull/1404, also add it to the cron build so it starts building right away.
 
### Check List

### DIFF
--- a/src/manifests_workflow/input_manifests.py
+++ b/src/manifests_workflow/input_manifests.py
@@ -26,6 +26,14 @@ class InputManifests(Manifests):
         return os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "..", "manifests"))
 
     @classmethod
+    def jenkins_path(self):
+        return os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "..", "jenkins"))
+
+    @classmethod
+    def cron_jenkinsfile(self):
+        return os.path.join(self.jenkins_path(), "check-for-build.jenkinsfile")
+
+    @classmethod
     def files(self, name):
         results = []
         for filename in glob.glob(os.path.join(self.manifests_path(), f"**/{name}-*.yml")):
@@ -52,6 +60,7 @@ class InputManifests(Manifests):
                     path=os.path.join(work_dir.name, self.name.replace(" ", ""), branch),
                     branch=branch,
                 )
+
                 version = c.version
                 logging.info(f"{self.name}#{branch} is version {version}")
                 if version not in main_versions.keys():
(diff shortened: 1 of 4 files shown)


peternied (approved):

src/manifests_workflow/input_manifests.py
```diff
@@ -120,3 +130,24 @@ def write_manifest(self, version, components=[]):
...
         manifest.to_file(manifest_path)
         logging.info(f"Wrote {manifest_path}")
+
+    def add_to_cron(self, version):
+        logging.info(f"Adding new version to cron: {version}")
+        jenkinsfile = self.cron_jenkinsfile()
+        with open(jenkinsfile, "r") as f:
+            data = f.read()
+
+        cron_entry = f"H/10 * * * * %INPUT_MANIFEST={version}/{self.prefix}-{version}.yml;TARGET_JOB_NAME=distribution-build-{self.prefix}\n"
```
peterzhuamazon, line 140: I am worried that the hardcoding of H/10 here will be issue later.
I am more in favor of just adding the cron to jenkinsfile, since it is just one line for each build.

dblock: The cron is to check for changes, takes a few seconds to run.

What are you worried about? We can always change this value, too.

src/manifests_workflow/input_manifests.py
```diff
@@ -120,3 +130,24 @@ def write_manifest(self, version, components=[]):
...
+
+        if cron_entry in data:
+            raise ValueError(f"{jenkinsfile} already contains an entry for {self.prefix} {version}")
+
+        data = data.replace(
+            "parameterizedCron '''\n",
+            f"parameterizedCron '''\n{' ' * 12}{cron_entry}"
```
peterzhuamazon, line 147: Can you help me understand why do you want to replace the entire parameterized block with `'''\n{' ' * 12}{cron_entry}`.
I understand triple quotes can be used to represent spanning strings, but what is `{' ' * 12}` being used for here?

Thanks.

src/manifests_workflow/input_manifests.py
peterzhuamazon: I think I get what you trying to do here.
Thanks.

src/manifests_workflow/input_manifests.py
dblock: It's just 12 spaces to align things nicely, see https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/check-for-build.jenkinsfile#L7

Screen Shot 2021-12-27 at 11 35 26 AM

peterzhuamazon: Thanks.

peterzhuamazon (approved):
Thanks.

zelinh (approved):
LGTM!

PR #1410, opened 2021-12-23
Sign standalone artifacts with jenkins job

### LINKED ISSUE #1385: Create Jenkins job to sign and upload standalone artifacts

Currently, we have to manually sign the standalone artifacts. Also, this process cannot be done by individual teams due the access to credentials. We need a jenkins job which can take the artifacts from an s3 bucket, sign and upload the artifacts to the same bucket. 

Use the existing `signer.py` and `signArtifacts.groovy` workflow to create a jenkins job to sign and upload the artifacts on s3

### Acceptance Criteria
- [ ] Given the jenkins job with the bucket and artifactPath on s3 bucket, the job is able to sign and upload the artifacts to the same bucket.

_No response_

### Additional context

_No response_

abhinavGupta16:
### Description
Users will be able to use the jenkins job to sign and the standalone artifacts given the artifact path or the artifact directory
 
**Note:** Currently only supports linux until we extend `signer.py` for windows and mac

### Issues Resolved
#1385
 
### Check List


jenkins/artifact-signer/Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+                choices: ['.sig'],
+                name: 'SIGNATURE_TYPE',
+                description: 'What is signature file type?'
+        )
+    }
+    stages {
+        stage('sign') {
+            steps {
+                script {
+                    if (ARTIFACT_PATH == '') {
```
dblock, line 32: Isn't it `null` by default?

jenkins/artifact-signer/Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+                        currentBuild.result = 'ABORTED'
+                        error('ARTIFACT_PATH is not set')
+                    }
+                    ARTIFACT_PATH = ARTIFACT_PATH.replaceAll('/$', "")
```
dblock, line 36: What does this do?

tests/jenkins/TestArtifactSigner.groovy
```diff
@@ -0,0 +1,55 @@
...
+        helper.registerAllowedMethod("s3Download", [Map])
+        helper.registerAllowedMethod("s3Upload", [Map])
+        helper.registerAllowedMethod("withAWS", [Map, Closure], { args, closure ->
+            closure.delegate = delegate
+            return helper.callClosure(closure)
+        })
+        helper.registerAllowedMethod('getPath', { args ->
+            return "workspace/file/found.zip"
+        })
+        helper.registerAllowedMethod('findFiles', [Map], { args ->
```
dblock, line 46: I don't think this is used, remove. Check if there are others.

dblock (commented):
How is this used? For example, how does an artifact appear in `$WORKSPACE/artifacts`?

jenkins/artifact-signer/Jenkinsfile
abhinavGupta16: I tried submitting an empty field and this works.

jenkins/artifact-signer/Jenkinsfile
abhinavGupta16: Removes the ending forward slash, if added, since we too add a forward slash when pulling and completing operations when creating path etc. Having 2 slashes creates unnecessary problems, folder named `/`, problems in `s3Download` etc.

tests/jenkins/TestArtifactSigner.groovy
abhinavGupta16: Thanks for pointing it out.

abhinavGupta16:
> How is this used? For example, how does an artifact appear in `$WORKSPACE/artifacts`?

The whole folder given as `ARTIFACT_PATH` is downloaded in `$WORKSPACE/artifacts` and all the items that can be signed in the folder structure are signed recursively. They are then uploaded to `ARTIFACT_PATH/signed` on `s3`

jenkins/artifact-signer/Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+                    set_up()
+                    argsMap = [:]
+                    argsMap['artifactPath'] = "$WORKSPACE/artifacts/${ARTIFACT_PATH}"
+                    argsMap['signatureType'] = "${SIGNATURE_TYPE}"
+                    argsMap['distributionPlatform'] = "${DISTRIBUTION_PLATFORM}"
+                    signArtifacts(argsMap)
```
peternied, line 42: Pass as parameters, no need to make a map

```suggestion
                    signArtifacts(
                       artifactPath: "$WORKSPACE/artifacts/${ARTIFACT_PATH}"
                       signatureType: SIGNATURE_TYPE
                       distributionPlatform: DISTRIBUTION_PLATFORM
```

peternied (commented):
Why make a job?  This seems like a groovy function that could be included as part of the promotion workflow

jenkins/artifact-signer/Jenkinsfile
peterzhuamazon: Agree here.

dblock:
> > How is this used? For example, how does an artifact appear in `$WORKSPACE/artifacts`?
> 
> The whole folder given as `ARTIFACT_PATH` is downloaded in `$WORKSPACE/artifacts` and all the items that can be signed in the folder structure are signed recursively. They are then uploaded to `ARTIFACT_PATH/signed` on `s3`

That's what was confusing me. If this task takes the path where artifacts are, it should then take the entire path, and not have to append `/artifacts` to it. Still, the fact that the path is part of the current workspace is odd.

jenkins/artifact-signer/Jenkinsfile
abhinavGupta16: We need a job for standalone artifacts, since we have to do them manually as of now.

abhinavGupta16:
> Why make a job? This seems like a groovy function that could be included as part of the promotion workflow

We need a job for standalone artifacts, since we have to do signing manually as of now for nay standalone artifacts

abhinavGupta16:
> > > How is this used? For example, how does an artifact appear in `$WORKSPACE/artifacts`?
> > 
> > 
> > The whole folder given as `ARTIFACT_PATH` is downloaded in `$WORKSPACE/artifacts` and all the items that can be signed in the folder structure are signed recursively. They are then uploaded to `ARTIFACT_PATH/signed` on `s3`
> 
> That's what was confusing me. If this task takes the path where artifacts are, it should then take the entire path, and not have to append `/artifacts` to it. Still, the fact that the path is part of the current workspace is odd.

It does take the entire Artifact path as an input. I just download it into artifacts folder inside the workspace to segregate it (we can also call it `temp` if it's causing confusion). I then sign whatever is downloaded and upload it to the same path given  inside a folder called `signed` to segregate the pre-exisiting artifacts and the ones the job uploaded.

I download it into the workspace because the workspace is deleted at the end of the job, which is required here.

jenkins/artifact-signer/Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+                    script {
+                        postCleanup()
+                    }
+                }
+            }
+        }
+    }
+}
+
+void upload_artifacts(){
```
dblock, line 57: Groovy convention is `camelCase`, so `uploadArtifacts()`. And to make this self-explanatory I would rename it to `uploadSignedArtifacts`.

Note though that these functions are not namespaced and I think may conflict with globals in `vars`. I don't know how Jenkins loads these things to be sure. Generally, I am not sure it's worth extracting functions that aren't called multiple times. But if we upload with this convention from `path` to `path/signed`, then extract this into a shared function in `vars`.

jenkins/artifact-signer/Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+    // Upload signature files to S3
+    withAWS(role: "${ARTIFACT_UPLOAD_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName:
+            'jenkins-session') {
+        // upload signature files to S3
+        s3Upload(file: "$WORKSPACE/artifacts/${ARTIFACT_PATH}", bucket: "${ARTIFACT_BUCKET_NAME}", path: "${ARTIFACT_PATH}/signed")
+    }
+}
+
+void set_up() {
```
dblock, line 66: This downloads unsigned artifacts, so should be called `downloadArtifacts()` to match `uploadSignedArtifacts()`

jenkins/artifact-signer/Jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+                        postCleanup()
+                    }
+                }
+            }
+        }
+    }
+}
+
+void upload_artifacts(){
+    // Upload signature files to S3
```
dblock, line 58: Signed artifacts are uploaded to S3, not signature files. I think this comment is misleading, it's pretty clear that `s3Upload` uploads to S3, would just remove it here and the duplicate below. If you name the function `uploadSignedArtifacts` then it explains what it does in the name.

tests/jenkins/TestArtifactSigner.groovy
```diff
@@ -0,0 +1,49 @@
...
+        helper.registerAllowedMethod("s3Upload", [Map])
+        helper.registerAllowedMethod("withAWS", [Map, Closure], { args, closure ->
+            closure.delegate = delegate
+            return helper.callClosure(closure)
+        })
+    }
+
+    @Test
+    void testArtifactSigner() {
+        super.testPipeline("tests/jenkins/jobs/ArtifactSigner_Jenkinsfile")
```
dblock, line 47: Rename `ArtifactSigner` to `SignArtifacts` to be more consistent.

dblock (changes requested):
Ok, now I finally understand what this is trying to do. Let's rename a few things and this will be good to go with me.

dblock (approved):

jenkins/artifact-signer/Jenkinsfile
abhinavGupta16: I created a function to upload to s3 and updated the code in other files to use the function instead.

jenkins/artifact-signer/Jenkinsfile
abhinavGupta16: renamed to `setUpSignArtifacts`

jenkins/sign-artifacts/Jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+                    script {
+                        postCleanup()
+                    }
+                }
+            }
+        }
+    }
+}
+
+void setUpSignArtifacts() {
```
dblock, line 63: Nit: This really does 2 things: `setUpSigningTools` and `downloadArtifacts`, I would split.

abhinavGupta16:
@dblock - Thank you for your review earlier, re-requesting review due to these new changes. https://github.com/opensearch-project/opensearch-build/pull/1410#discussion_r779132160

vars/uploadToS3.groovy
```diff
@@ -0,0 +1,7 @@
+void call(Map args = [:]){
+
+    withAWS(role: "${ARTIFACT_UPLOAD_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+        s3Upload(file: args.sourcePath, bucket: args.artifactBucket, path: args.destinationOnS3)
```
dblock, line 4: I would rename these arguments: 

- `sourcePath` looks good
- `artifactBucket` should just be `bucket`
- `destinationOnS3` should just be `path`

dblock (approved):
Some suggestions below, take it or leave it.

vars/uploadToS3.groovy
```diff
@@ -0,0 +1,7 @@
+void call(Map args = [:]){
+
+    withAWS(role: "${ARTIFACT_UPLOAD_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
```
gaiksaya, line 3: Why put this few lines into its own library/method?
Its gonna be confusing if we want to switch roles and accounts and upload to s3. 
Is this really required?

abhinavGupta16: Added this in response to @dblock's review [comment](https://github.com/opensearch-project/opensearch-build/pull/1410#discussion_r778799905) earlier.  It brings upload to s3 component from all places to one place. I think it's good to have as it trivializes the upload to s3 for anyone to use.

We can always extend this to be able to take in bucket as a parameter as well for future use-cases. Or, I can do it now as well if that's already a use-case.

abhinavGupta16:
@peternied - updated the job to take in urls as input following our discussion.

jenkins/sign-artifacts/Jenkinsfile
```diff
@@ -10,9 +10,14 @@ pipeline {
...
-                name: 'ARTIFACT_PATH',
+                name: 'URLs',
                 trim: true,
-                description: 'Path to the artifact or the artifact directory on s3'
+                description: 'comma separated list of urls of artifacts to be signed'
+        )
+        string(
+                name: 'UPLOAD_PATH',
+                trim: true,
+                description: 'Path to upload the artifact to'
```
peternied, line 20: The name on this parameter implies something like `http://foo/bar/baz`, but based on the final upload path it seems more like a 'project name' such as `data-prepper`?

Nit: Please include an example value in the description.

peternied (approved):
I'd recommend changing the parameter name, but looks good

jenkins/sign-artifacts/Jenkinsfile
abhinavGupta16: Update the variable name to `S3_FILE_UPLOAD_PATH` and updated the description as well with an example

peternied (approved):

Issue #578, opened 2021-09-24
Automate the whitesource scans


bbarani:
### Is your feature request related to a problem? Please describe

We need an automated way to scan our code and surface the CVE's. Currently there is no automated mechanism to scan and surface the CVE's corresponding to a particular repository under OpenSearch organization

### Describe the solution you'd like

An automated scan using Whitesource scanning tool is scheduled to track and surface the CVE's on periodic basis.

Basically fix: https://github.com/opensearch-project/opensearch-build/blob/opensearch-1.0.0/.github/workflows/check-vulnerability-whitesource.yml

### Describe alternatives you've considered

Manually verifying the CVE against public vulnerability database

### Tasks

- [x] #613
- [x] #614
- [x] #615
- [ ] Whitesource scanner implementation in opensearch-build
- [ ] Scanning is automatically triggered, with alarm if it does not run or has failures

dblock:
Is it viable to replace Whitesource with Dependabot and if yes/no why/not?

bbarani:
We can enable Dependabot as an additional scanning tool but we cannot replace Whitesource with it due to the below limitations,

**Here are the high level benefits of using Whitesource (Notes from vendor).**
Dependabot only works with GitHub. WhiteSource works with GitHub and other repos, your build tools, IDEs, and other CI\CD tools. Here is a screenshot of examples of some of our integrations.Not listed here are integrations with IDEs (ItelliJ, Eclipse, Visual Studio, and VS Code) and browsers (Chrome and Edge).

> WhiteSource can enforce policies to proactively prevent open source libraries with security or compliance issues from making into your builds. WhiteSource can fail build pipelines when it identifies a library in the repo used with the build that violates a security or license policy defined in WhiteSource.

> Enterprise reports related to inventory\Bill Of Materials, Security Vulnerabilities, License due diligence, and project risk

> Reliability: Our library matching algorithms are primarily based on SHA1 hashes, which are 100% reliable as methods to match a file in your scan to a file in our inventory. In addition, our team of security researchers validates each and every security vulnerability identified by our machine learning algorithms before they are published to our vulnerability database, Together, our matching algorithms and validation of security vulnerabilities reduce the presence of false positive and false negative security alerts.

> Completeness: WhiteSource supports over 200 programming languages, over a dozen package managers, and has the most comprehensive inventory of open source libraries and their security vulnerabilities. Our vulnerability database is comprised of vulnerability reports sourced from more than 20 security advisory services.  

[...]

PR #1411, opened 2021-12-24
Fix issues for WhiteSource scan workflow

zelinh:

### Description
Fix issues for Gradle dependencies scan on several repos and enable scan on OpenSearch repo 1.x branch. 
 
In addition, WhiteSource has some issues handling Gradle Wrapper. Since they suggests installing Gradle on the environment that running the scan ([Ref](https://whitesource.atlassian.net/wiki/spaces/WSK/pages/2273804326/Configuring+the+Unified+Agent+for+Gradle#Best-Practices)), we will install Gradle 6.6.1 manually.

### Issues Resolved
Part of #578
 
### Check List


peterzhuamazon:
These are the things we discovered during the testing:
1. Whitesource does not run correctly even though the the correct wrapper settings are added. Moreover, it only runs when you set the env to `wrapper` while also setting wrapperPath manually. After you set these two up, the wrapper is able to install gradle, but will still fail the run afterwards:
```
gradle.preferredEnvironment=wrapper
gradle.wrapperPath=./gradlew
```
You have to install the actual gradle prior to the run and set env to gradle specifically, as the default is using wrapper. The whitesource support is also recommending this way, despite their default value is wrapper.
```
gradle.preferredEnvironment=gradle
#gradle.wrapperPath=./gradlew
```

2. Whitesource is looking for `settings.gradle` file to build before scan, and it will search in the directory where the `java -jar` command run. Therefore, we need to ensure these two things:
* The scan needs to run in the root folder of each repo
* The `settings.gradle` file must present or the scan will exit
* We use some scripts to create a simple `settings.gradle` file from the `name` variable in build.gradle to pass scan for repos such as performance-analyzer.

3. Whitesource is trying to scan all the possible dependencies, which cause `OpenSearch` core repo to timeout. The reason is there is a `qa` folder (https://github.com/opensearch-project/OpenSearch/tree/main/qa) in that repo, which contains all the relations to all the previous version of ES/ODFE (https://github.com/opensearch-project/OpenSearch/tree/main/qa/os). It will attempt to download all the deb/rpm/tar/jars for all the versions of ES from elastic repos. The default 900seconds timeout is not enough at all. In our calculation, the downloads would complete in 3-6 hrs and we are still not quite sure about the exact number. Testing these dependencies are irrelavant to OpenSearch and is time consuming, so we remove `qa` folder as a whole to speed up the process.

4. Whitesource can use JDK14 despite the official support page only state JDK8 and JDK11.

5. Whitesource will generate a series if `/tmp/ws*` folders in every run, and will not clean them up after the run. This can cause the runner to fill up space and eventually has readonly issues.

6. If you put more than one whitesource unified agent jar within the same folder, it will automatically use the one with the lower version, despite you are calling the latest version jar in the scan command.

7. We have also discovered that many of the previous scans are actually not complete, as the dependencies reported on whitesource panel is actually maven dependencies, not gradle+maven. This scan currently will not run on `SQL` related repo such as dashboards-reports, since they use subfolders to contain the actual project, not in root of the repo.

jenkins/vulnerability-scan/whitesource-scan.jenkinsfile
```diff
@@ -41,8 +41,13 @@ pipeline {
 
 void scan() {
     git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
-    sh '''
+    sh """
+        curl -SLO https://services.gradle.org/distributions/gradle-6.6.1-bin.zip
```
peterzhuamazon, line 45: This is a short term temp solution as whitesource is not running correctly with gradle wrapper.

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -47,33 +47,70 @@ basepath=$baseDirPath"/repos"
...
  
 # clone the desired Repos for scanning 
 for repo in $gitRepos
 do
   echo "Cloning repo "$gitBasePath$repo
-  git clone "$gitBasePath$repo".git $basepath"/"$repo
+  if [ ${repo} == "OpenSearch" ]
+  then
+    echo "Cloning "$repo" at branch 1.x"
+    git clone -b 1.x "$gitBasePath$repo".git ${basepath}/${repo}
```
peterzhuamazon, line 62: Temp measure to only scan 1.x for opensearch.
We need to add a new config option to specify repo version.
Ideally from the input menifest file of the releases.

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -47,33 +47,70 @@ basepath=$baseDirPath"/repos"
...
+    rm -rf ${basepath}/${repo}/qa
```
peterzhuamazon, line 63: Remove QA is due to issues listed in https://github.com/opensearch-project/opensearch-build/pull/1411#issuecomment-1000945335.

@zelinh please add a comment in case we forget about this.

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -47,33 +47,70 @@ basepath=$baseDirPath"/repos"
...
 # scan the Repos using the WhiteSource Unified Agent
 for repo in $gitRepos
 do
   repo_path=$basepath"/"$repo
   if [ -d "$repo_path" ]
   then
+    cd $repo_path
     echo "Scanning repo: "$gitBasePath$repo " Project: " $repo
-    java -jar wss-unified-agent.jar -c wss-unified-agent.config -d $repo_path -apiKey $wss_apikey -product "$target" -project $repo
+# Generate temporary `settings.gradle` file based on the name in `build.gradle`
```
peterzhuamazon, line 82: Spacing.

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -47,33 +47,70 @@ basepath=$baseDirPath"/repos"
...
+            echo "settings.gradle exist $repo"
+        fi
+    else
+# Exceptions for some repos that have `build.gradle` in sub-folder
+        if [ ${repo} == "observability" ]
+        then
+          cd opensearch-observability
+          settings_gradle_content=`cat build.gradle | grep name | grep -i $repo | head -n 1 | sed 's/name/rootProject.name =/g' | awk '{$1=$1};1'`
+          echo $settings_gradle_content > settings.gradle
+          cat settings.gradle
```
peterzhuamazon, line 102: Why is there only observability repo?
SQL, Reports all with this exception.

Plus, consider put this section into a function, as you are rewriting them multiple times.

peterzhuamazon (changes requested):
Good start and there are things to be improved.
Thanks.

jenkins/vulnerability-scan/whitesource-scan.jenkinsfile
zelinh: Yeah after testing, this is the only way currently working for all repos.

tools/vulnerability-scan/wss-scan.sh
zelinh: Temporarily we hardcoded this to clone 1.x branch only for OpenSearch. We plan to add more configuration options so that we could specify branch to scan for all repos.

abhinavGupta16:
Why does it fail the WhiteSource security check?

tools/vulnerability-scan/wss-scan.sh
zelinh: After testing, dashboards reports contains `[settings.gradle](https://github.com/opensearch-project/dashboards-reports/blob/main/reports-scheduler/settings.gradle)` in its sub-folder, so WhiteSource can successfully track the file location with no needs to go in there. Same for SQL. 

The essence of having WhiteSource running for gradle project is to have the `settings.gradle` present. Otherwise it may look into the `settings.gradle` in build repo in our case and cause the incomplete scan.

zelinh:
> Why does it fail the WhiteSource security check?

That check shouldn't be triggered in this PR. I think now it's passed.

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -19,6 +19,13 @@
...
 ###############################################################################################
 
+# Generate temporary `settings.gradle` file based on the name in `build.gradle`
+function generate_settings_gradle() {
+  settings_gradle_content=`cat build.gradle | grep name | grep -i $repo | head -n 1 | sed 's/name/rootProject.name =/g' | awk '{$1=$1};1'`
+  echo $settings_gradle_content > settings.gradle
+  cat settings.gradle
+}
+
 set -e
```
peterzhuamazon, line 29: Move set -e on top of the function.

tools/vulnerability-scan/wss-scan.sh
peterzhuamazon: We dont mention ES in the code, use ODFE.

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -79,27 +89,23 @@ do
   then
     cd $repo_path
     echo "Scanning repo: "$gitBasePath$repo " Project: " $repo
-# Generate temporary `settings.gradle` file based on the name in `build.gradle`    
+
```
peterzhuamazon, line 92: You removed the comment but leave extra space.

zelinh: Good points thanks!

tools/vulnerability-scan/wss-scan.sh
zelinh: Moved

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -21,6 +21,13 @@
...
 set -e
 
+# Generate temporary `settings.gradle` file based on the name in `build.gradle`
+function generate_settings_gradle() {
+  settings_gradle_content=`cat build.gradle | grep name | grep -i $repo | head -n 1 | sed 's/name/rootProject.name =/g' | awk '{$1=$1};1'`
+  echo $settings_gradle_content > settings.gradle
+  cat settings.gradle
+}
+
 java -version
```
peterzhuamazon, line 31: Sorry, one more.
Please move this to the same line with `gradle -v ......`.
Thanks.

zelinh: But I think checking java is different from checking other installation. 
Java is needed to run `wss-unified-agent.jar` so that's why we had a check here before download it. Moving to one line and checking them all together seem to be meaningless from my perspective. 

Or we could actually remove `gradle -v ...` because I had that line just to double check everything is properly installed and present.

peterzhuamazon: I didnt realize you are having a if condition afterwards, you can just remove everything afterwards since you have `set -e` set.

If java not exist it will fail instantly, therefore your if statement is useless anyway.

peterzhuamazon: If you really really really want a statement just write:
```
! (java -version1> /dev/null 2>&1) || (echo Java not found && exit 1)
```

zelinh: Make sense to me. That check has been removed.

peterzhuamazon (approved):

peterzhuamazon:
@bbarani @abhinavGupta16 please take a look on this.
Thanks.

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -47,33 +49,67 @@ basepath=$baseDirPath"/repos"
...
 target='OpenSearch'
-mvn -v; node -v; npm -v; yarn -v
 
 # scan the Repos using the WhiteSource Unified Agent
 for repo in $gitRepos
 do
   repo_path=$basepath"/"$repo
   if [ -d "$repo_path" ]
   then
+    cd $repo_path
```
abhinavGupta16, line 84: Let's use absolute path's everywhere instead of `cd` in a folder. It would make it very easy then see and keep track of the files we are referring to. It would otherwise be difficult to keep track and extend this script later.

jenkins/vulnerability-scan/whitesource-scan.jenkinsfile
abhinavGupta16: Can we install `gradle` in the docker container being used? Installing there would prevent running the install every time the job is built.

abhinavGupta16 (changes requested):

tools/vulnerability-scan/wss-scan.sh
zelinh: After testing, going into the repository folder and running the scan inside it will improve the scan stability with WhiteSource. I changed all with absolute path though to make them more clear.

jenkins/vulnerability-scan/whitesource-scan.jenkinsfile
zelinh: > Can we install `gradle` in the docker container being used? Installing there would prevent running the install every time the job is built.

We are using the docker container from CI-runner, and I don't think we want to install Gradle on that one because we don't need Gradle to be installed as long as we have Gradle wrapper.

peterzhuamazon:
Add 

```
options {
        timeout(time: 3, unit: 'HOURS')
    }
```

on line 3 of your Jenkinsfile, so it will ensure the run will be exit if it stuck for more than 3 hours.

zelinh:
> Add
> 
> ```
> options {
>         timeout(time: 3, unit: 'HOURS')
>     }
> ```
> 
> on line 3 of your Jenkinsfile, so it will ensure the run will be exit if it stuck for more than 3 hours.

Added.

tools/vulnerability-scan/wss-scan.sh
abhinavGupta16: We can remove the cd from here then?

tools/vulnerability-scan/wss-scan.sh
```diff
@@ -47,33 +49,67 @@ basepath=$baseDirPath"/repos"
...
+        if [ ${repo} == "observability" ]
+        then
+          cd $repo_path/opensearch-observability
+          generate_settings_gradle
+        else
+          echo "build.gradle for $repo not exist, either dashboards repo or missing necessary files"
+        fi
+    fi
+    java -jar $baseDirPath/wss-unified-agent.jar -c $baseDirPath/wss-unified-agent.config -d $repo_path -apiKey $wss_apikey -product "$target" -project $repo
+    cd $baseDirPath && pwd
```
abhinavGupta16, line 107: this `cd` too

zelinh: This is the absolute path. https://github.com/opensearch-project/opensearch-build/blob/ac208d0b7a403a3782d3654fecbe7fe91d70e7d9/tools/vulnerability-scan/wss-scan.config#L1

abhinavGupta16: yes, but if we are using absolute path everywhere, then we can remove all the `cd` statements right?

zelinh: Like I said above, we have to go into each folder to run the scan. Also refer to the second point Peter listed above (https://github.com/opensearch-project/opensearch-build/pull/1411#issuecomment-1000945335)
**The scan needs to run in the root folder of each repo**

abhinavGupta16 (approved):

abhinavGupta16:
We can possibly let go of using cd since this is a temporary solution.

PR #1416, opened 2021-12-28
Add whitesource API call to retrieve reports

### LINKED ISSUE #1415: Support extracting WhiteSource data through API after the scanning.
This is related to @ryanbogan where we need to extract the data through whitesource API after the scan and use it to build the dependency tree later.

This issue is used to record all the related activities.

Thanks.

peterzhuamazon:

### Description
Add whitesource API call to retrieve reports
 
### Issues Resolved
#1415
 
### Check List


jenkins/vulnerability-scan/whitesource-report.jenkinsfile
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+void wss_api_call() {
+    def toJson = {
+        input ->
+        groovy.json.JsonOutput.toJson(input)
+    }
+    def payloadBody = [
+        requestType: "getProjectVulnerabilityReport",
```
bbarani, line 41: We might need to use getProductInventoryReport API rather than getProjectVulnerabilityReport to get the entire dependencies.

peterzhuamazon: @ryanbogan please confirm.
Thanks.

zelinh:
Do we want to do this regularly as cron job?

jenkins/vulnerability-scan/whitesource-report.jenkinsfile
zelinh: Ref to these API call [here](https://whitesource.atlassian.net/wiki/spaces/WD/pages/786137094/HTTP+API+v1.2).
I'm not sure if the `getProjectInventoryReport` contains vulnerability information; or we could possibly use both calls: one for dependencies tree, and the other for vulnerability report?

ryanbogan: If that has the both the vulnerabilities and the dependencies, then we should use getProjectInventory Report instead

peterzhuamazon:
I will tweak to have 2 calls as suggested, and will not put it as cron now since the rest of the code for analyzing json is not ready.

Once @ryanbogan has the analyzing part ready we can add to this workflow, and will get triggered after the whitesource scan complete.

Thanks.

jenkins/vulnerability-scan/whitesource-report.jenkinsfile
peterzhuamazon: Move the httprequests to its own lib and add flexibility for more requestType, with tests.

peterzhuamazon:
@dblock @zelinh @bbarani please take a look at this PR.
Thanks.

dblock:
This seems to be an overkill for something that just needs to do a `curl/wget` with a bunch of parameters. It also stores credentials in the file system, which I think is a big problem.

peterzhuamazon:
> This seems to be an overkill for something that just needs to do a `curl/wget` with a bunch of parameters. It also stores credentials in the file system, which I think is a big problem.

1. It is a start up before the full analysis part comes in.
2. It saves the credentials in memory not FS.

dblock:
> > This seems to be an overkill for something that just needs to do a `curl/wget` with a bunch of parameters. It also stores credentials in the file system, which I think is a big problem.
> 
> 1. It is a start up before the full analysis part comes in.

I don't fully understand the scope, but this part alone does not warrant all this code. Let's see it do something useful in this PR.

> 2. It saves the credentials in memory not FS.

That's my bad.

From just this, I don't think you need `jenkins/vulnerability-scan/whitesource-report-opensearch-opensearch.json` at all, loading it, replacing values in it, seems like a lot of work. I would construct JSON inside the job and assign the values to it.

peterzhuamazon:
> > > This seems to be an overkill for something that just needs to do a `curl/wget` with a bunch of parameters. It also stores credentials in the file system, which I think is a big problem.
> > 
> > 
> > 
> > 1. It is a start up before the full analysis part comes in.
> 
> I don't fully understand the scope, but this part alone does not warrant all this code. Let's see it do something useful in this PR.
> 
> > 2. It saves the credentials in memory not FS.
> 
> That's my bad.
> 
> From just this, I don't think you need `jenkins/vulnerability-scan/whitesource-report-opensearch-opensearch.json` at all, loading it, replacing values in it, seems like a lot of work. I would construct JSON inside the job and assign the values to it.

You suggestion of inline changes is the 1st draft of the PR which I then replace to this version due to :
  * We request multiple calls, not just one.
  * I would like to make this a generalized lib use case instead of one off hard coding, that is why I set it flexible so future workflows can use this lib to make calls.
  * If assembleUpload can be its own lib I think my httpRequest can have its seat as well, since the former is even more simplier than my current setup.

peterzhuamazon:
@peternied please also take a look thanks.

peternied (commented):
This code makes some api calls - why do we need those API calls as there response is not used?  The issue associated with this work has no acceptance criteria

I agree with @dblock that the json object/url wrapping is overkill and makes the code harder to follow.

peterzhuamazon:
Like I said before the code will be expanded once @ryanbogan comes back with the rest of the action.
As of now the point of this task is to create a env to make different API calls to whitesource, and save the response before further analysis.

peternied:
> the code will be expanded

Without details its hard to give constructive feedback on pull requests.

jenkins/vulnerability-scan/whitesource-report.jenkinsfile
```diff
@@ -0,0 +1,31 @@
...
+                }
+            }
+            steps {
+                script {
+                    whitesourceRequest()
+                }
+            }
+            post() {
+                always {
+                    cleanWs disableDeferredWipeout: true, deleteDirs: true
```
abhinavGupta16, line 25:
```suggestion
                    postCleanup()
```

vars/whitesourceRequest.groovy
```diff
@@ -0,0 +1,40 @@
...
+            payloadMap['requestType'] = request
+            println("Sending request: " + payloadMap['requestType'])
+
+            def response = sendHttpRequest(
+                 payloadMap,
+                 'https://saas.whitesourcesoftware.com/api/v1.3',
+                 'POST'
+            )
+            
+            //println(response.status)
```
abhinavGupta16, line 34: commented code

abhinavGupta16 (changes requested):
Need to add test cases for the job and `libtesters`

peterzhuamazon:
Will close this PR as we are not expecting to add it now.

Issue #1372, opened 2021-12-17
Trigger OpenSearch Dashboards integ test post bundle build


tianleh:
### Is your feature request related to a problem? Please describe

#604

### Describe the solution you'd like

Trigger the job automatically

tianleh:
Close as duplicate of https://github.com/opensearch-project/opensearch-build/issues/704

PR #1429, opened 2022-01-05
Automate integration tests jenkins job

### LINKED ISSUE #1114: Trigger `integ-test` workflow after build workflow finishes

After every run of the build workflow, `integ-test` should be triggered. The job requires parameters like opensearch_version, platform, architecture and the build id which can be pick up from the `build.yml` generated.

After distribution builds are created trigger the integration test workflow

### Acceptance Criteria:
- [ ] After the build is complete the pipeline triggers a test pipeline
- [ ] The test pipeline gets all information needed to spin up test workers
- [ ] The test pipeline triggers notifications on success/failure that are actionable
- [ ] Links to the test publisher results are consumable by receivers of the notification

ohltyler:
### Description

This PR updates the existing integ test jenkins workflow, which currently doesn't work and is using outdated parameters.

More specifically:
- moves and updates existing test workflow (`jenkins/test/testsuite/Jenkinsfile`)  to `jenkins/opensearch/integ-test.jenkinsfile`. Params were modified to try to prevent link-building logic which has caused broken builds in the past.
- adds `runIntegTestScript` groovy script to execute the actual `test` shell script
- adds `getBuildManifestInfo` groovy script to use the passed-in build manifest URL to pull out build-related information

There will be follow up PRs to finish automating integration tests:
1. Update [distribution-build workflow](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile) to trigger the integ tests after building x64 and arm64 artifacts
2. Remove legacy jenkins test code (in `jenkins/test`) and associated READMEs

NOTE: this PR is a WIP for now. Things to wrap up before it is ready:
- [x] investigate passing full manifest file to `test.sh` instead of root URL
- [x] adding publishNotifications() back

Screenshot - working integ test job
Screen Shot 2022-01-13 at 9 17 44 AM

### Issues Resolved
Closes #1114 
 
### Check List


jenkins/opensearch/distribution-build.jenkinsfile
```diff
@@ -111,38 +111,64 @@ pipeline {
...
             }
         }
     }
+    // post {
+    //     success {
+    //         node('Jenkins-Agent-al2-x64-c54xlarge-Docker-Host') {
+    //             script {
+    //                 def stashed = lib.jenkins.Messages.new(this).get(['build-x64', 'build-arm64'])
+
+    //                 publishNotification(
```
dblock, line 147: You can comment out the body of `publishNotification` which may be easier than commenting out the steps everywhere ;)

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -0,0 +1,133 @@
...
+    environment {
+        AWS_ROLE_ARN = "arn:aws:iam::${AWS_ACCOUNT_PUBLIC}:role/opensearch-test"
+        AWS_ROLE_SESSION_NAME = "jenkins-test-session"
+    }
+    tools {
+        jdk "JDK14"
+        maven "maven-3.8.2"
+    }
+    parameters {
+        string(
```
dblock, line 14: Aren't many of these redundant with `ARTIFACT_BASE_PATH`? I feel like the only thing I want to supply to an integ test build is the location of the build (a manifest).

ohltyler: Agreed, working on simplifying these:
`OPENSEARCH_VERSION`: can be removed
`ARCHITECTURE`: used for selecting the env to run the tests in (could be generated from `ARTIFACT_BASE_PATH`)
`BUILD_ID`: used in `uploadTestResults`
`TEST_MANIFEST`: needed for running the actual `test.sh` script
`BUILD_MANIFEST`: used in `uploadTestResults` (could be generated from `ARTIFACT_BASE_PATH`)

From my understanding, it can be better to pass the full strings than dynamically generate them, to help prevent broken links in the future, should any of the artifact path naming conventions change, etc. But if these conventions are unlikely to change then I'm ok to generate them directly from the base path too, not sure on the status of that.

Lmk what you think. cc @peternied

dblock: BUILD_ID is _what_ is being tested? In that case the build manifest contains it, no? 

I think if you give a build manifest and a test manifest that makes sense.

ohltyler: Right, BUILD_ID can be pulled from the build manifest too. I'm ok to simplify to just build manifest and test manifest. I'll add some helper fns to pull out the rest of the info.

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -0,0 +1,133 @@
...
+            name: 'ARTIFACT_BASE_PATH',
+            description: 'The artifact base path, e.g. https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.2.2/98/linux/x64',
+            trim: true
+        )
+    }
+    stages {
+        stage('Choose Agent Node') {
+            steps {
+                script {
+                    if ("${ARCHITECTURE}" == 'x64') {
```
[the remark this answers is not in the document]
ohltyler: Refactored in latest commit, where the architecture and build ID are getting pulled from BUILD_MANIFEST_URL. Still agreed it's a good section of code that can be thrown in a groovy function, I'll do that.

vars/runIntegTests.groovy
```diff
@@ -0,0 +1,27 @@
...
+    git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
+
+    def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+    String manifest = args.manifest ?: "manifests/${INPUT_MANIFEST}"
+    def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: manifest))
+
+    String tarball = args.architecture == "x64" ? env.ARTIFACT_URL_linux_x64 : env.ARTIFACT_URL_linux_arm64
+    String artifactRootUrl = args.architecture == "x64" ? env.ARTIFACT_ROOT_URL_linux_x64 : env.ARTIFACT_ROOT_URL_linux_arm64
+    String buildManifest = "${artifactRootUrl}/dist/opensearch/manifest.yml"
+    String testManifest = manifest.replaceAll(".yml", "-test.yml")
```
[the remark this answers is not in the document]
ohltyler: The reason I didn't was to prevent adding a param in `distribution-build.jenkinsfile`, and instead using the existing `INPUT_MANIFEST` to derive it. I'm ok either way, lmk if you have a preference.

vars/runIntegTests.groovy
```diff
@@ -0,0 +1,27 @@
+void call(Map args = [:]) {
+    git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
+
+    def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+    String manifest = args.manifest ?: "manifests/${INPUT_MANIFEST}"
+    def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: manifest))
```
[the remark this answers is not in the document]
ohltyler: Good catch, will remove.

vars/runIntegTests.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+    String tarball = args.architecture == "x64" ? env.ARTIFACT_URL_linux_x64 : env.ARTIFACT_URL_linux_arm64
+    String artifactRootUrl = args.architecture == "x64" ? env.ARTIFACT_ROOT_URL_linux_x64 : env.ARTIFACT_ROOT_URL_linux_arm64
+    String buildManifest = "${artifactRootUrl}/dist/opensearch/manifest.yml"
```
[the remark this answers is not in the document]
ohltyler: Again we could, but it would mean adding a param in `distribution-build.jenkinsfile`. What do you think about putting this logic into a method in `BuildManifest.groovy` called something like `getUrl()` which could build and return the full path? Then if there's future changes, it would only be a single change. `getArtifactRootUrl()` already exists so it would logically fit there.

vars/runIntegTests.groovy
```diff
@@ -0,0 +1,27 @@
+void call(Map args = [:]) {
+    git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
+
+    def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+    String manifest = args.manifest ?: "manifests/${INPUT_MANIFEST}"
+    def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: manifest))
+
+    String tarball = args.architecture == "x64" ? env.ARTIFACT_URL_linux_x64 : env.ARTIFACT_URL_linux_arm64
```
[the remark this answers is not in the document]
ohltyler: Right now the `architecture` parameter is used to determine the root url, which is used to determine the build manifest

vars/runIntegTests.groovy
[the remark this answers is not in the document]
peternied: I'd rather be explicit with a parameter.  Note; if we had input manifests reference the test manifest this would also be a way to hold this information without complicating this interface

vars/runIntegTests.groovy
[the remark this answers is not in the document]
peternied: 👍 Great suggestion, I like it

vars/runIntegTests.groovy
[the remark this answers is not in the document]
peternied: Hmm, I don't love it - because when we add more platforms + architectures we have to copypasta a bunch of code in a bunch of different places but we already have to deal with this problem with the 2 build legs.  Lets not try to solve it now.

tianleh:
FYI. I have a task to trigger OpenSearch Dashboards integ test post bundle build https://github.com/opensearch-project/opensearch-build/issues/1372  We can sync up to see if we can leverage each other's work.

ohltyler:
> Let's have test cases for all the `jenkinsfile` that are added.
> 
> * `distribution-build.jenkinsfile`
> * `integ-test.jenkinsfile`

Tests have been added for `integ-test.jenkinsfile` via `runIntegTests.groovy`, and is tested via `TestRunIntegTests.groovy`.
`distribution-build.jenkinsfile` already exists, and is essentially a compilation of a bunch of these groovy scripts that already have associated tests. Since there is no related groovy script for this one, I don't know how else to test.

I'm not aware of any other ways to test these in an automated way, do you have any suggestion?

vars/runIntegTests.groovy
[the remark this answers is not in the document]
ohltyler: Added as `getDistUrl()` in https://github.com/opensearch-project/opensearch-build/pull/1429/commits/15d3640d54eaddc6ec2b130cbf6575b50f2a7f32

vars/runIntegTests.groovy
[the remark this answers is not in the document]
ohltyler: Sure, added `TEST_MANIFEST` param in https://github.com/opensearch-project/opensearch-build/pull/1429/commits/15d3640d54eaddc6ec2b130cbf6575b50f2a7f32

abhinavGupta16:
> > Let's have test cases for all the `jenkinsfile` that are added.
> > 
> > * `distribution-build.jenkinsfile`
> > * `integ-test.jenkinsfile`
> 
> Tests have been added for `integ-test.jenkinsfile` via `runIntegTests.groovy`, and is tested via `TestRunIntegTests.groovy`. `distribution-build.jenkinsfile` already exists, and is essentially a compilation of a bunch of these groovy scripts that already have associated tests. Since there is no related groovy script for this one, I don't know how else to test.
> 
> I'm not aware of any other ways to test these in an automated way, do you have any suggestion?

`TestRunIntegTests.groovy` should only test `runIntegTests.groovy` like a unit test. `integ-test.jenkinsfile` is a jenkins job that may or may not use more functionalities and libraries in the future apart from `runIntegTests.groovy` and therefore should be tested separately.

We can add tests for the intire job ina very similar way to what we do for our libraries. You can refer to [this](opensearch-build/tests/jenkins/TestSignArtifacts.groovy) as reference

ohltyler:
> > > Let's have test cases for all the `jenkinsfile` that are added.
> > > 
> > > * `distribution-build.jenkinsfile`
> > > * `integ-test.jenkinsfile`
> > 
> > 
> > Tests have been added for `integ-test.jenkinsfile` via `runIntegTests.groovy`, and is tested via `TestRunIntegTests.groovy`. `distribution-build.jenkinsfile` already exists, and is essentially a compilation of a bunch of these groovy scripts that already have associated tests. Since there is no related groovy script for this one, I don't know how else to test.
> > I'm not aware of any other ways to test these in an automated way, do you have any suggestion?
> 
> `TestRunIntegTests.groovy` should only test `runIntegTests.groovy` like a unit test. `integ-test.jenkinsfile` is a jenkins job that may or may not use more functionalities and libraries in the future apart from `runIntegTests.groovy` and therefore should be tested separately.
> 
> We can add tests for the intire job ina very similar way to what we do for our libraries. You can refer to [this](opensearch-build/tests/jenkins/TestSignArtifacts.groovy) as reference. I hope this helps

Link seems to be broken. I'm still unsure what you mean

peternied (commented):
Looking really close, could you remove all the commented out code at this point?

abhinavGupta16:
> > > > Let's have test cases for all the `jenkinsfile` that are added.
> > > > 
> > > > * `distribution-build.jenkinsfile`
> > > > * `integ-test.jenkinsfile`
> > > 
> > > 
> > > Tests have been added for `integ-test.jenkinsfile` via `runIntegTests.groovy`, and is tested via `TestRunIntegTests.groovy`. `distribution-build.jenkinsfile` already exists, and is essentially a compilation of a bunch of these groovy scripts that already have associated tests. Since there is no related groovy script for this one, I don't know how else to test.
> > > I'm not aware of any other ways to test these in an automated way, do you have any suggestion?
> > 
> > 
> > `TestRunIntegTests.groovy` should only test `runIntegTests.groovy` like a unit test. `integ-test.jenkinsfile` is a jenkins job that may or may not use more functionalities and libraries in the future apart from `runIntegTests.groovy` and therefore should be tested separately.
> > We can add tests for the intire job ina very similar way to what we do for our libraries. You can refer to [this](opensearch-build/tests/jenkins/TestSignArtifacts.groovy) as reference. I hope this helps
> 
> Link seems to be broken. I'm still unsure what you mean

https://github.com/opensearch-project/opensearch-build/blob/2ca4c8d849563352003c7bf37dacaebec868f1a9/tests/jenkins/TestSignArtifacts.groovy#L43-L60

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -0,0 +1,122 @@
...
+        }
+        stage('Choose Agent Node') {
+            steps {
+                script {
+                    if ("${ARCHITECTURE}" == 'x64') {
+                        agentLabel = "Jenkins-Agent-al2-x64-m5xlarge"
+                    } else if ("${ARCHITECTURE}" == 'arm64') {
+                        agentLabel = "Jenkins-Agent-al2-arm64-m6gxlarge"
+                    } else {
+                        error("Unknown CPU architecture is provided")
```
dblock, line 45: Log $ARCHITECTURE so we know what it was here.

vars/runIntegTests.groovy
```diff
@@ -0,0 +1,24 @@
+void call(Map args = [:]) {
+    git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
+
+    def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+    String testManifest = args.testManifest ?: "manifests/${TEST_MANIFEST}"
+
+    String tarball = args.architecture == "x64" ? env.ARTIFACT_URL_linux_x64 : env.ARTIFACT_URL_linux_arm64
```
dblock, line 7: I think a proper `if` or `case` is better here to handle `x64` vs. `arm64` vs. something else that doesn't work to avoid suprises.

dblock (changes requested):
Are we using/will we ever use `jenkins/test/orchestrator/Jenkinsfile`?

I'm good with this putting back notifications and removing any Tyler-references ;)

dblock:
> > Are we using/will we ever use `jenkins/test/orchestrator/Jenkinsfile`?
> > I'm good with this putting back notifications and removing any Tyler-references ;)
> 
> I think the framework is worth leaving for now, to help with future work at getting the entire test orchestrator running. I imagine we will want the distribution build and/or other test pipelines to be able to kick off all integ/perf/bwc tests
> 
> cc @peternied

my vote is to nuke anything unused, and move it to a link to a specific commit for anyone who wants to resurrect it when it's ready to be used

peternied:
> my vote is to nuke anything unused (and doesn't work)

+1; however, I am comfortable with making this a new PR focused on removing these components and fixing the documentation

jenkins/test/orchestrator/Jenkinsfile
```diff
@@ -1,44 +1,50 @@
 pipeline {
```
peternied, line 1: Note, I would remove any changes to this file and either delete it in this PR or in a follow up PR

src/jenkins/BuildManifest.groovy
```diff
@@ -49,6 +51,23 @@ class BuildManifest implements Serializable {
             this.getArtifactRoot(jobName, buildNumber)
         ].join('/')
     }
+
+    public String getDistUrl(String publicArtifactUrl = 'https://ci.opensearch.org/ci/dbc', String jobName, String buildNumber) {
```
peternied, line 55: This function is used to reverse engineer the url so we can pass in the root path to the test system.  Instead could we update the python code to accept the full path to the build manifest?  It reduces the number of parameters we pass around greatly

vars/getBuildManifestInfo.groovy
```diff
@@ -0,0 +1,16 @@
...
+    env.BUILD_MANIFEST = buildManifestFileName
+    
+    sh "wget ${BUILD_MANIFEST_URL} -O ${buildManifestFileName}"
+    def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: "${buildManifestFileName}"))
+    String architecture = buildManifest.getArtifactArchitecture()
+    String buildId = buildManifest.getArtifactBuildId()
+    echo "Setting ARCHITECTURE=${architecture}"
+    env.ARCHITECTURE = architecture
+    echo "Setting BUILD_ID=${buildId}"
+    env.BUILD_ID = buildId
```
peternied, line 15: I would prefer that we return these as an output of this function call if possible

vars/runIntegTestScript.groovy
```diff
@@ -0,0 +1,17 @@
...
+    echo "args.testManifest: ${args.testManifest}"
+    def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: args.buildManifest))
+    String artifactRootUrl = buildManifest.getArtifactRootUrl("${jobName}", "${BUILD_ID}")
+    echo "Artifact root URL: ${artifactRootUrl}"
+    
+    sh([
+        './test.sh',
+        'integ-test',
+        "${args.testManifest}",
+        "--paths opensearch=${artifactRootUrl}",
```
peternied, line 14: I would prefer we were able to pass the manifest file directly to the python code instead of having to hack up the parameters.

vars/runIntegTestScript.groovy
```diff
@@ -0,0 +1,17 @@
...
+        "--test-run-id ${BUILD_NUMBER}"
```
peternied, line 15: This test run id can be omitted, this is a part of the legacy design that I don't believe we need to support.

vars/runIntegTests.groovy
```diff
@@ -0,0 +1,24 @@
...
+    if (tarball == null) {
+        echo "Skipping integ tests for ${args.architecture} architecture because ${args.architecture} artifact was not built."
+    } else {
+        IntegTests: {
+            // TODO: will need to override integ-test in Jenkins with the new workflow in jenkins/opensearch/integ-test.jenkinsfile
+            build job: 'Playground/ohltyler-integ-test',
+            parameters: [
+                string(name: 'TEST_MANIFEST', value: "${testManifest}"),
+                string(name: 'BUILD_MANIFEST_URL', value: "${buildManifest}"),
+            ]
```
peternied, line 21: We should add the parameter `wait: false` so that the build workflow isn't blocked
We should also add the parameter `propagate: false` so if the tests pass/fail it doesn't impact the distribution-build job.

peternied (commented):
We did a deep dive today, and that resulted in some things coming out.  I'd recommend doing some smaller PRs that would make this more manageable.

ohltyler:
> > my vote is to nuke anything unused (and doesn't work)
> 
> +1; however, I am comfortable with making this a new PR focused on removing these components and fixing the documentation

Got it, makes sense. I'll update #1447 with some details and resolve in a separate PR

ohltyler:
> We did a deep dive today, and that resulted in some things coming out. I'd recommend doing some smaller PRs that would make this more manageable.

To break this up easier, I'm thinking of splitting this up into 3 parts / PRs:
1. Revive old integ test workflow and get a working version via `jenkins/opensearch/integ-test.jenkinsfile` (this PR)
2. Hook new integ test workflow into distribution build workflow (`jenkins/opensearch/distribution-build.jenkinsfile`) (future PR)
3. Remove legacy integ-test jenkins code in `jenkins/test/` and update related READMEs (future PR)

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -0,0 +1,122 @@
...
+            steps {
+                script {
+                    runIntegTestScript(
+                        buildManifest: "${BUILD_MANIFEST}",
+                        testManifest: "${TEST_MANIFEST}"
+                    )
+                }
+            }
+            post {
+                always {
```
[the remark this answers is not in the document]
ohltyler: Adding a try/catch means the build status won't be changed, which means as long as the tests run it will always show up as successful, which means the publish notifications steps below will always run the `success` branch regardless of tests succeeding/failing.

Are you suggesting parsing the results of running the `test.sh` script to determine success/failure? I feel like this can unnecessary work instead of leveraging jenkins failures to tell us whether the integ tests passed or failed. Lmk what you think.

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -0,0 +1,122 @@
...
+            defaultValue: '',
+            name: 'BUILD_MANIFEST_URL',
+            description: 'The build manifest URL, e.g. https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.2.2/98/linux/x64/dist/opensearch/manifest.yml.',
+            trim: true
+        )
+    }
+    stages {
+        stage('Extract build information from build manifest') {
+            steps {
+                node('Jenkins-Agent-al2-x64-m5xlarge') {
```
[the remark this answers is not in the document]
ohltyler: Fixed in https://github.com/opensearch-project/opensearch-build/pull/1429/commits/e13948505ded132571d25c0f805ae499f9dfaf46

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -0,0 +1,122 @@
...
+                    }
+                }
+            }
+        }
+        stage('Choose Agent Node') {
+            steps {
+                script {
+                    if ("${ARCHITECTURE}" == 'x64') {
+                        agentLabel = "Jenkins-Agent-al2-x64-m5xlarge"
+                    } else if ("${ARCHITECTURE}" == 'arm64') {
```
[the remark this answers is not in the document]
ohltyler: Fixed in https://github.com/opensearch-project/opensearch-build/pull/1429/commits/e13948505ded132571d25c0f805ae499f9dfaf46

jenkins/test/orchestrator/Jenkinsfile
ohltyler: Reverted the changes in https://github.com/opensearch-project/opensearch-build/pull/1429/commits/e13948505ded132571d25c0f805ae499f9dfaf46

Will fix in a follow up PR as mentioned here: https://github.com/opensearch-project/opensearch-build/pull/1429#issuecomment-1010444623

vars/getBuildManifestInfo.groovy
ohltyler: Fixed in https://github.com/opensearch-project/opensearch-build/commit/e13948505ded132571d25c0f805ae499f9dfaf46

vars/runIntegTestScript.groovy
ohltyler: Removed in https://github.com/opensearch-project/opensearch-build/commit/e13948505ded132571d25c0f805ae499f9dfaf46

vars/runIntegTests.groovy
ohltyler: Similar to my question [here](https://github.com/opensearch-project/opensearch-build/pull/1429#discussion_r782623755), how do we want to go about test failures/successes impacting the jenkins jobs?

I see 2 main paths forward:
1. Fail the step if the tests failed, which would cause the entire job to fail
2. Jenkins job should always succeed (assuming no syntax / other workflow-related bugs), and test success/failure is handled via notifications and/or other jenkins jobs that _will_ fail if the tests fail

jenkins/opensearch/integ-test.jenkinsfile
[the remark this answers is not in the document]
peternied: If you want the job to show failure, use `catchError` (its also mentioned in that SO link).

Maybe I am being very pessimistic as I expect the test results to fail often while we are first confronting our intermittent failing tests.  By not reporting failure unless there are workflow related failures - it would bring attention to them more quickly IMO.

I'll leave it up to you what route you would prefer.

vars/runIntegTests.groovy
peternied: 1. I put a more detail reply on that first question, for now only job failure == workflow failure is more useful because we are using as we are building so we are still focused on stability of the workflow

2. As you've written it I think so.  Note; I think we can revisit this when we have an automated action like 'deploy to production' that has the required pre-reqs run.   I think we would want to make sure things like 8+ hour long build jobs are still correcting rolling up state across different workflows.

jenkins/opensearch/integ-test.jenkinsfile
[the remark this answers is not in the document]
dblock: I would start by failing the job if integration tests fail. There's no hope to make them reliable if there aren't loud visible failures.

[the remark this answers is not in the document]
ohltyler: @peternied I was hesitant to go with `catchError` since it was mentioned that it's deprecated. I prefer failing the job if the tests fail for now. Agreed that it may bring more failures first, but it will probably help motivate owners of the test cases to fix flaky / broken tests. If it gets too noisy we can easily throw in a try/catch in the future.

vars/runIntegTests.groovy
ohltyler: Right, I think that makes a lot of sense. We can follow this fail-fast strategy for now.

src/jenkins/BuildManifest.groovy
ohltyler: After looking further, it looks like the root path is used, because both the build manifest (ending in `/builds/opensearch/manifest.yml`) and the bundle manifest (ending in `/dist/opensearch/manifest.yml`) are both used for a few things in the dependency installer - the build manifest is used for downloading maven dependencies, and the bundle manifest is used for downloading the distribution (via `location` param stored in the bundle manifest).

I prefer to keep the root url rather than add parameters here to keep the interface simpler. It's already a bit complex when running Dashboards tests since multiple root urls are needed already. Lmk what you think.

ohltyler: Also note that `getDistUrl()` in particular is only used in the distribution-build workflow to generate the build manifest url to pass to the integ-test workflow. The existing `getArtifactRootUrl()` is the one used for reverse engineering

vars/runIntegTestScript.groovy
ohltyler: Keeping a single thread about this via https://github.com/opensearch-project/opensearch-build/pull/1429#discussion_r782485772

src/jenkins/BuildManifest.groovy
dblock: We should consider taking build output such as maven and promoting it to the distribution manifest so we can use a single manifest as input.

ohltyler: Wouldn't this be blending the purposes of these manifests too much? From my understanding, the build manifest is for listing out the artifacts and related dependencies per component that are included in the build. The bundle/distribution manifest is a simplified list of built components, with references to their build locations.

Agreed there is quite a bit of overlap between the two though, where maybe it's worth considering merging these into a single manifest? I may not have the most context for why it was designed this way to begin with, but I feel that only one may be needed.

dblock: Agreed. It's beyond the scope of this issue, but the idea was that the same build is being used to assemble a different distribution, e.g. cross-platform.

vars/getBuildManifestInfo.groovy
```diff
@@ -0,0 +1,14 @@
+Map call(Map args = [:]) {
```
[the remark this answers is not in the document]
ohltyler: Good catch. Fixed.

vars/getBuildManifestInfo.groovy
```diff
@@ -0,0 +1,11 @@
...
+    def lib = library(identifier: "jenkins@20211123", retriever: legacySCM(scm))
+    String buildManifestUrl = args.buildManifestUrl ?: BUILD_MANIFEST_URL
+
+    String buildManifestPath = 'manifest.yml'
+    sh "wget ${buildManifestUrl} -O ${buildManifestPath}"
+    def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: buildManifestPath))
+    String buildId = buildManifest.getArtifactBuildId()
+
+    return [buildManifestPath, buildId]
+}
```
dblock, line 11: Returning a tuple limits the utility of this shared function. As suggested above, maybe something like `downloadBuildManifest` that downloads the remote manifest and returns the `buildManifest` object, then all the fields are available to the caller.

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -0,0 +1,109 @@
...
+        stage('integ-test') {
+            agent {
+                node {
+                    label "${AGENT_LABEL}"
+                }
+            }
+            steps {
+                script {
+                    // Fetching the build-specific values and assigning to their env vars
+                    (buildManifestPath, buildId) = getBuildManifestInfo()
```
dblock, line 43: This API is a bit strange, I'd prefer it'd download a manifest and return that.

```
def buildManifest = downloadBuildManifest(manifest: "${BUILD_MANIFEST_URL}")
...
```

vars/getBuildManifestInfo.groovy
```diff
@@ -0,0 +1,11 @@
+String[] call(Map args = [:]) {
+    def lib = library(identifier: "jenkins@20211123", retriever: legacySCM(scm))
+    String buildManifestUrl = args.buildManifestUrl ?: BUILD_MANIFEST_URL
```
dblock, line 3: I would require the manifest URL parameter, otherwise it's hard to tell that the job calling this needs `BUILD_MANIFEST_URL`.

dblock (changes requested):
This looks great. I have strong feelings about the utility library, WDYT?

jenkins/opensearch/integ-test.jenkinsfile
peternied: I much prefer this approach.  Functions with no side effects are my jam

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -0,0 +1,109 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
+
+pipeline {
+    agent none
+    environment {
+        AWS_ROLE_ARN = "arn:aws:iam::${AWS_ACCOUNT_PUBLIC}:role/opensearch-test"
+        AWS_ROLE_SESSION_NAME = "jenkins-test-session"
+    }
+    tools {
+        jdk "JDK14"
```
peternied, line 10: Lets create an issue; should we only run tests with JDK14? I think this will influence OpenSearch to be running on that jdk version which we added support to run on jdk11

peternied (approved):
Great work!

jenkins/opensearch/integ-test.jenkinsfile
dblock: Do you mean as implemented now? 

Note that shared functions are global. This one does not "get build manifest info", it "get some very specific build manifest info used in integ-test workflow", are we going to be adding another one like these when we need some other value from a remote manifest?

vars/getBuildManifestInfo.groovy
ohltyler: Agreed this is a better approach, changed to `downloadBuildManifest()` in https://github.com/opensearch-project/opensearch-build/pull/1429/commits/c60435a14f764a93afa35f0ab1b002b3e93750a5

jenkins/opensearch/integ-test.jenkinsfile
ohltyler: Good point, I've created #1495

jenkins/opensearch/integ-test.jenkinsfile
ohltyler: I've updated this, see https://github.com/opensearch-project/opensearch-build/pull/1429#discussion_r786970899

dblock (approved):

PR #1437, opened 2022-01-07
Minor bugfix for signArtifacts.groovy

abhinavGupta16:

### Description
Currently, the pgp key for signing is only imported if the file is downloaded. We only need to download the file ones, however, we need to import the key if the file already exists or when we download the file.
 
### Issues Resolved
pgp key failure will not happen for preexisting pgp key
 
### Check List


vars/signArtifacts.groovy
```diff
@@ -30,8 +30,9 @@ void call(Map args = [:]) {
 }
 
 void importPGPKey(){
+    keyPath = "$WORKSPACE/opensearch.pgp"
     if( !fileExists("$WORKSPACE/opensearch.pgp")) {
```
peternied, line 34: this if statement should use `keypath`

peternied (changes requested):
Since this was a bug, lets add unit tests to confirm it works as expected

vars/signArtifacts.groovy
```diff
@@ -30,8 +30,9 @@ void call(Map args = [:]) {
 }
 
 void importPGPKey(){
+    keyPath = "$WORKSPACE/opensearch.pgp"
```
zelinh, line 33: Is this `keyPath` necessary? I think we are justing taking the `gpg import` out of the if bracket.

abhinavGupta16:
> Since this was a bug, lets add unit tests to confirm it works as expected

I have added a test case for it. I have also changed the test case to be a unit test case only `signArtifacts` lib and added another test case for the actual job. I believe we should do this for every jenkins job we have in the project.

vars/signArtifacts.groovy
abhinavGupta16: Yeah, I realized that the same path is being used at multiple places. So it is better to have in a variable. This makes it less error prone (for any changes in the path in future) and easy to follow.

vars/signArtifacts.groovy
```diff
@@ -30,8 +30,9 @@ void call(Map args = [:]) {
 }
 
 void importPGPKey(){
-    if( !fileExists("$WORKSPACE/opensearch.pgp")) {
-        sh("curl -SL https://artifacts.opensearch.org/publickeys/opensearch.pgp -o $WORKSPACE/opensearch.pgp")
-        sh("gpg --import $WORKSPACE/opensearch.pgp")
+    keyPath = "$WORKSPACE/opensearch.pgp"
+    if( !fileExists("${keyPath}")) {
+        sh("curl -SL https://artifacts.opensearch.org/publickeys/opensearch.pgp -o ${keyPath}")
```
dblock, line 35: Since you import every time, how about never storing this key?
https://www.reddit.com/r/GnuPG/comments/4oz7zd/possible_to_import_gpg_key_from_a_url/

```
curl -sSL http://my.key.url | gpg --import -
```

peternied (approved):

jenkins/sign-artifacts/sign-standalone-artifacts.jenkinsfile.txt
```diff
@@ -0,0 +1,33 @@
+   sign-standalone-artifacts.run()
```
peternied, line 1: Please move this file into the `.../tests/data` path with the other files

peternied (changes requested):

jenkins/sign-artifacts/sign-standalone-artifacts.jenkinsfile.txt
abhinavGupta16: data was being used to store dummy files that are required for tests. I have created a separate folder to hold regression txt files for jenkins job, `tests/jenkins/jenkinsjob-regression-files`. Let me know if the name of the dir works

peternied (approved):
Thanks for updating the path on those tests, looks good

dblock (approved):

Issue #1432, opened 2022-01-06
Add version checks for opensearch-dashboards components to CI


dblock:
### Is your feature request related to a problem? Please describe

Security dashboards plugin was added to the 1.3 manifest in https://github.com/opensearch-project/opensearch-build/pull/1431, but its version was 1.2, so it broke the build.

### Describe the solution you'd like

Implement version checks for dashboards, ala https://github.com/opensearch-project/opensearch-build/blob/main/src/ci_workflow/ci_check_gradle_properties_version.py#L20

```
  checks:
    - gradle:properties:version
```

PR #1446, opened 2022-01-11
Added OpenSearch Dashboards package.json version checks.

dblock:

### Description

Loads package.json and compares versions. Add `package:version` check to any OpenSearch Dashboards component to ensure the version matches the manifest version. Updated 1.3.0, 2.0 and 1.2.0 manifests.
 
### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/1432
 
### Check List

### DIFF
--- a/src/ci_workflow/README.md
+++ b/src/ci_workflow/README.md
@@ -28,7 +28,7 @@ The following checks are available.
 | gradle:properties:version                     | Check version of the component.                                                   |
 | gradle:dependencies:opensearch.version        | Check dependency on the correct version of OpenSearch in gradle properties.       |
 | gradle:publish                                | Check that publishing to Maven local works, and publish.                          |
-| package:version                               | Check dependency on the correct version of OpenSearch Dashboards in package.json. |
+| npm:package:version                           | Check dependency on the correct version of OpenSearch Dashboards in package.json. |
 
 The following example sanity-checks components in the the OpenSearch 1.2.0 manifest.
 
--- a/src/ci_workflow/README.md
+++ b/src/ci_workflow/README.md
@@ -4,7 +4,7 @@
 
 ## Sanity Testing the Distribution
 
-This workflow runs sanity checks on every component present in the bundle, executed as part of the [manifests workflow](../../.github/workflows/manifests.yml) in this repository. It ensures that the component GitHub repositories are correct and versions in those components match the OpenSearch version.
+This workflow runs sanity checks on every component present in the bundle, executed as part of the [manifests workflow](../../.github/workflows/manifests.yml) in this repository. It ensures that the component GitHub repositories are correct and versions in those components match the OpenSearch or OpenSearch Dashboards versions.
 
 To use checks, nest them under `checks` in the manifest.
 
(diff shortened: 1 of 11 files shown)


peterzhuamazon:
Is this checking whether the branch we pulled have the correct version in package.json?
E.g. 1.2.0 manifest for dashboards will have plugin package.json having 1.2.0?

dblock:
> Is this checking whether the branch we pulled have the correct version in package.json? E.g. 1.2.0 manifest for dashboards will have plugin package.json having 1.2.0?

Yes.

tests/tests_ci_workflow/test_ci_check_package.py
```diff
@@ -0,0 +1,31 @@
...
+
+    class DummyProperties(CiCheckPackage):
+        def check(self):
+            pass
+
+    def test_loads_package_json(self):
+        props = TestCiCheckPackage.DummyProperties(
+            component=MagicMock(),
+            git_repo=MagicMock(working_directory=self.DATA),
+            target=CiTarget(version="1.3.0", name="opensearch", snapshot=False),
```
peterzhuamazon, line 26: isnt check package is for dashboards only?
But there is `opensearch` in this line.
Is this ok?

tests/tests_ci_workflow/test_ci_check_package.py
```diff
@@ -0,0 +1,31 @@
...
+        )
+
+        self.assertEqual(props.properties["name"].data, "opensearch-security-dashboards")
```
peterzhuamazon, line 29: Relate to above comment the name does not match.

src/ci_workflow/ci_check_package.py
```diff
@@ -0,0 +1,49 @@
...
+    def __get_properties(self):
+        with open(self.package_json_path, "r") as f:
+            return PropertiesFile(CiCheckPackage.__flattenDict(json.load(f)))
+
+    # https://gist.github.com/higarmi/6708779
+    @classmethod
+    def __flattenDict(cls, d, result=None, index=None, parent_key=None):
+        if result is None:
+            result = {}
+        if isinstance(d, (list, tuple)):
```
peterzhuamazon, line 32: Can we change `d` to actually `dict` or `dictionary`?
It will be very confusing and I actually thought it means directory to start with.

Thanks.

peterzhuamazon (approved):
Some questions.

tests/tests_ci_workflow/test_ci_check_package.py
dblock: This is just a string, can be foobar. I replaced it to avoid confusion.

tests/tests_ci_workflow/test_ci_check_package.py
dblock: It's not supposed to, this is not what the test compares.

src/ci_workflow/ci_check_package.py
dblock: No because it's not always a dictionary, it can be `None`, a tuple or a dictionary, so `d` is as good as anything.

tests/tests_ci_workflow/test_ci_check_package.py
peterzhuamazon: Just checked again, got it now.

src/ci_workflow/ci_check_package.py
peterzhuamazon: I think a better naming would be easier to understand and search later as well, but this is just nitpik from me.

dblock:
@peternied could I get a CR here pls?

src/ci_workflow/README.md
```diff
@@ -23,11 +23,12 @@ To use checks, nest them under `checks` in the manifest.
...
-| gradle:properties:version                     | Check version of the component.                               |
-| gradle:dependencies:opensearch.version        | Check dependency on the correct version of OpenSearch.        |
-| gradle:publish                                | Check that publishing to Maven local works, and publish.      |
+| name                                          | description                                                                       |
+|-----------------------------------------------|-----------------------------------------------------------------------------------|
+| gradle:properties:version                     | Check version of the component.                                                   |
+| gradle:dependencies:opensearch.version        | Check dependency on the correct version of OpenSearch in gradle properties.       |
+| gradle:publish                                | Check that publishing to Maven local works, and publish.                          |
+| package:version                               | Check dependency on the correct version of OpenSearch Dashboards in package.json. |
```
peternied, line 31: This new check has a different naming structure than the previous version check, `gradle:properties:version`.  Can we align this new check to something like `npm:properties:version`?

peternied (approved):

kavilla (commented):
lgtm!

Issue #1335, opened 2021-12-15
Skip old(er) manifests in CI


dblock:
### Is your feature request related to a problem? Please describe

We have 30+ manifests running CI checks in https://github.com/opensearch-project/opensearch-build/pull/1331. That takes more and more time and is not scalable long term even with infinite parallelization.

### Describe the solution you'd like

Skip some manifest checks. One idea could be to skip the last patch release, e.g. skip 1.1.0, 1.1.1, 1.1.2 when we have 1.1.3 for example. Another could be to commit release `.lock` files and check for changes.

zelinh:
[Triage] Adding this issue to our product backlog. We have enough information on this.

PR #1455, opened 2022-01-12
Only perform ref checks on components with stable refs.

dblock:

### Description

The manifest checks will continue taking longer and longer because we keep shipping releases. As of now, the 31 checks in https://github.com/opensearch-project/opensearch-build/actions/runs/1685288124 took 14m. It's not bad because the checks are parallelized.

We have suggested in https://github.com/opensearch-project/opensearch-build/issues/1335 to skip checking manifests for already released versions, but this PR proposes a better solution. 

- Only check that a ref exists without checking out the entire source if there are no actual checks to be performed.
- Update any refs that weren't using tags in manifests for shipped releases.
- Remove checks from components in released manifests to avoid re-running the same checks on stable targets.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/1335.
 
### Check List

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_template.md
+++ b/.github/ISSUE_TEMPLATE/release_template.md
@@ -69,7 +69,7 @@ __REPLACE with OpenSearch wide initiatives to improve quality and consistency.__
 ### Post Release
 
 - [ ] Create [release tags](https://github.com/opensearch-project/opensearch-build/issues/378#issuecomment-999700848) for each component.
-- [ ] Replace refs in [manifests/{{ env.VERSION }}](/opensearch-project/opensearch-build/tree/main/manifests/{{ env.VERSION }}) with tags.
+- [ ] Replace refs in [manifests/{{ env.VERSION }}](/opensearch-project/opensearch-build/tree/main/manifests/{{ env.VERSION }}) with tags and remove checks.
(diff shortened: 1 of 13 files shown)


peternied (commented):
What about ignoring checks on component references with `tags/*`?  IMO this is easier than deleting the 'checks' on legacy manifests.

dblock:
> What about ignoring checks on component references with `tags/*`? Less work than deleting the 'checks' on legacy manifests.

I considered it, but it felt like hidden logic/functionality.

peternied:
Other considerations:
- What about a top level, `disable-checks-on-ref-tags: true` that makes this more transparent?
- What about a manifest promotion tool that effective does automatically what you are doing by hand?

dblock:
> Other considerations:
> 
> * What about a top level, `disable-checks-on-ref-tags: true` that makes this more transparent?
> * What about a manifest promotion tool that effective does automatically what you are doing by hand?

These are good ideas, but why are they better than removing useless checks?

peternied:
Hmm, why keep around these manifests at all, why not delete them?

dblock:
> Hmm, why keep around these manifests at all, why not delete them?

So we can rebuild previous versions.

dblock:
Given that JCenter is now gone we also need to do this (or something like this) for manifests such as 1.1.1.

tianleh (approved):

PR #1458, opened 2022-01-12
Resolve #1130 where sed and tee is trying to access files at the same time

### LINKED ISSUE #1454: [Bug]: Entrypoint for Docker image erases config file
### Describe the bug

After #1130 was merged, [this line](https://github.com/opensearch-project/opensearch-build/pull/1130/files#diff-107c2499e08c900c5ab75b5fd413c22bd9cf1ffe3819aeddabe51836ee4c58f5R72) completely wipes out the configuration file of the docker container on startup.

### To reproduce

```
docker run -it -p 9200:9200 -e "discovery.type=single-node" opensearchstaging/opensearch:1.2.4
```

as `1.2.4` is the first image built with this entrypoint.

### Expected behavior

Container should start up successfully.

### Host / Environment

MacOS, docker.

### Additional context

_No response_

### Relevant log output

_No response_

peterzhuamazon:

### Description
Resolve #1130 where sed and tee is trying to access files at the same time.
If we use sed to access file while tee, it would sometimes cause a conflict and result in tee creating empty file due to file is then being held by sed.
Therefore, we cat the file, pipe to sed, then tee, would resolve issue.

Also, fixed an error where sed is not correctly setup for the pattern.
 
### Issues Resolved
#1454
 
### Check List

### DIFF
--- a/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
+++ b/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
@@ -177,8 +177,8 @@ if [ -d "$OPENSEARCH_DASHBOARDS_HOME/plugins/$SECURITY_DASHBOARDS_PLUGIN" ]; the
     if [ "$DISABLE_SECURITY_DASHBOARDS_PLUGIN" = "true" ]; then
         echo "Disabling OpenSearch Security Dashboards Plugin"
         ./bin/opensearch-dashboards-plugin remove securityDashboards
-        sed "/^opensearch_security/d" $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
-        sed "s/https/http/" $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
+        cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "/^opensearch_security/d" | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
+        cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "s/https/http/g" | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
     fi
 fi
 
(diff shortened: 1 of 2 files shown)


dblock (approved):

abhinavGupta16 (approved):

docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
```diff
@@ -177,8 +177,8 @@ if [ -d "$OPENSEARCH_DASHBOARDS_HOME/plugins/$SECURITY_DASHBOARDS_PLUGIN" ]; the
     if [ "$DISABLE_SECURITY_DASHBOARDS_PLUGIN" = "true" ]; then
         echo "Disabling OpenSearch Security Dashboards Plugin"
         ./bin/opensearch-dashboards-plugin remove securityDashboards
-        sed "/^opensearch_security/d" $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
-        sed "s/https/http/" $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
+        cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "/^opensearch_security/d" | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
```
peternied, line 180: If piping together the operations is causing access issues, should we instead try to copy to a tmp location before updating the file?

Something like https://stackoverflow.com/a/5174368

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
```diff
@@ -66,10 +66,11 @@ if [ -d "$OPENSEARCH_HOME/plugins/$SECURITY_PLUGIN" ]; then
 
     if [ "$DISABLE_SECURITY_PLUGIN" = "true" ]; then
         echo "Disabling OpenSearch Security Plugin"
-        sed "s/plugins.security.disabled.*$/plugins.security.disabled: true" $OPENSEARCH_HOME/config/opensearch.yml | tee $OPENSEARCH_HOME/config/opensearch.yml
+        cat $OPENSEARCH_HOME/config/opensearch.yml | sed "/plugins.security.disabled/d" | tee $OPENSEARCH_HOME/config/opensearch.yml
```
peternied, line 69: Nit; Move the deletion of the line to above 67 so you only have to do it once, then in the if statement append the security setting

docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
peterzhuamazon: no the point of using this method is to avoid inode changes.
More see #1130.

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
peterzhuamazon: This is to prevent deleting existing lines if user run the container once, put it down, then restart without changing settings.

peternied: This script is run on start up and sed runs on both 69 & 73?

peterzhuamazon: yes.

PR #1467, opened 2022-01-13
Replace wget with curl

gaiksaya:

### Description
Got error while running this workflow. Instead of adding new package using what we already have on the system
```
line 1: wget: command not found
```
Replacing wget with curl
 
### Issues Resolved
N/A
 
### Check List


peternied (commented):
You will need to update the regression tests as they will fail

gaiksaya:
> You will need to update the regression tests as they will fail

On it!

dlvenable (approved):

zelinh (approved):

peternied (approved):

jcgraybill:
Hey, as a matter of practice, is there a reason this script doesn't specify the full path to binaries it calls? E.g. why `curl` instead of `/usr/bin/curl`? It seems like a best practice to use the latter.

dblock:
> Hey, as a matter of practice, is there a reason this script doesn't specify the full path to binaries it calls? E.g. why `curl` instead of `/usr/bin/curl`? It seems like a best practice to use the latter.

We have lots of places we call shell scripts without hard-coding a path, so maybe better to open a new issue. 
Note that likely the "correct" way to do this is `/usr/bin/env curl` to be properly cross-platform.

PR #1470, opened 2022-01-13
Remove Dashboards 1.3.0 version since 2.0.0 will be the next version

peternied:

### Description
Dashboards is moving node versions https://github.com/opensearch-project/OpenSearch-Dashboards/issues/920, the 1.3.0 release will not be used 
 
### Check List


gaiksaya (approved):

dblock:
> We have an issue: [opensearch-project/OpenSearch-Dashboards#1099](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1099), to track if it is possible that we can release OpenSearch Dashboards `2.0.0` targeting the OpenSearch `1.3.0` release. Even after verifying, I believe we still want to make sure to propose to the community that we plan to split the version of OpenSearch and OpenSearch Dashboards. Historically and programmatically (like there are checks to make sure that OpenSearch Dashboards version `major.minor` matched OpenSearch `major.minor` before starting up), they have gone hand and hand with major and minor versions.
> 
> So I know it has been merged but I reckon if there is push back or if a plugin breaks compatibility on major versions then we should reconsider adding this back.

That's cool. We can always re-create manifests as needed.

Issue #1457, opened 2022-01-12
Auto-format the auto-generated manifest YML


dblock:
### Is your feature request related to a problem? Please describe

When a new manifest is generated by the manifests workflow it's strangely formatted. When we try to make changes to it by hand we end up auto-formatting it very differently, e.g. https://github.com/opensearch-project/opensearch-build/pull/1449

### Describe the solution you'd like

Write new manifests that are formatted as if they were authored by the default auto-formatter in VSCode (or other editor).

peternied:
An alternative approach, we could apply a check / auto-formatter on all manifest.yml files.

PR #1480, opened 2022-01-14
Fix YAML formatting when writing manifests.

dblock:

### Description

Use a YAML formatter to auto-format manifests before they are written. Solves for both reordering of keys and alignment.

Before this change we output something like https://github.com/opensearch-project/opensearch-build/pull/1476.
Will auto-format existing YAMLs and introduce a yaml linter separately because there are dozens of changes.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/1457
 
### Check List

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -21,6 +21,7 @@ cerberus = "~=1.3.4"
 psutil = "~=5.8"
 atomicwrites = "*"
 validators = "*"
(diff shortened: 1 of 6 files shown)


dblock:
Superseded with #1481

peternied (approved):

PR #1481, opened 2022-01-14
Added a YAML linter.

dblock:
### Description

Following https://github.com/opensearch-project/opensearch-build/pull/1480 added a YAML linter and fixer.

### Check List


.pre-commit-config.yaml
```diff
@@ -1,27 +1,36 @@
...
+        types: [python]
+      - id: flake8
+        stages: [commit]
+        name: flake8
+        entry: bash -c 'pipenv run flake8 .'
+        language: system
+      - id: mypy
+        stages: [commit]
+        name: mypy
+        entry: bash -c 'pipenv run mypy src/build_workflow tests/tests_build_workflow
```
peternied, line 22: Looks like `line-length` is disabled in yamllint, which is my preference.  However, it seems like ymlfix was applied with this setting enabled.  Can we synchronize the rules they are applied between them?

peternied (commented):
Great addition to the codebase, thanks!

.pre-commit-config.yaml
dblock: Unfortunately yaml-fix doesn't support line length configuration (https://github.com/lyz-code/yamlfix/issues/117) currently.

I manually undid the 2-3 instances where the fixer CR/LFed those lines.

peternied: Maybe we enable the line-length rule so we are consistent, WDYT?

peternied (approved):

.pre-commit-config.yaml
dblock: Nan, we actually have a similar problem in Python when we recommend using `black` that auto-formats differently, but it hasn't been a concern.

tianleh (approved):

Issue #1492, opened 2022-01-18
[Feature] Add support for urls to substitute `latest` for the most recent build number


peternied:
From https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1038

Potential Solutions:
- **Add functionality to the url re-writer**
  The cdk deployment that uses a lambda function to rewrite the url on incoming requests to the s3 buckets, this could be updated to handle the `/latest/` string in place of a build number by scanning the s3 bucket.  https://github.com/opensearch-project/opensearch-build/blob/f79e7fd2d8d39290e064fa915800d510dfee40f8/deployment/lib/artifacts-public-access.ts#L23-L29

   *Trade-offs*: Con, the bucket might not contain that artifact and result in a 404.  Pro, could be applied on older builds.

- **When uploading artifacts, also add a `latest` bucket**
  The upload code is on a shard and the artifacts could be uploaded into multiple locations.  https://github.com/opensearch-project/opensearch-build/blob/f79e7fd2d8d39290e064fa915800d510dfee40f8/vars/uploadArtifacts.groovy#L26-L29

   *Trade-offs*: Con, Only works on products that build with that shared library function.  Doesn't work with 1.x kind of functionality

Acceptance Criteria:
- Given a url such as `https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/1.2.0/LATEST/linux/x64/builds/plugins/queryWorkbenchDashboards-1.2.0.zip` gets the same artifact as calling `https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/1.2.0/512/linux/x64/builds/plugins/queryWorkbenchDashboards-1.2.0.zip` (Assuming 512 is the most recent version)
- Given a url such as  `https://ci.opensearch.org/ci/dbc/bundle-build-dashboards/1.2.0/LATEST/linux/x64/builds/plugins/queryWorkbenchDashboards-1.2.0.zip` after the artifact has been updated the url returns the artifact in a way that bypasses caching.

dblock:
I think the best case scenario for "latest" would be a 302 redirect, is that an option 3? That is also something that would allow us to move downloadable files around as we tend to do this quite a bit while we learn the "ideal" url structure.

peternied:
> I think the best case scenario for "latest" would be a 302 redirect

Good callout, I've added it to the acceptance criteria that we need a way to get through a cache.

dblock:
Couple more:

[...]

PR #1490, opened 2022-01-16
Update artifacts promotion Jenkins to support OpenSearch Dashboards and support builds/dist structure

tianleh:
### Description
Introduce the builds/dist structure for the release URL to match how it looks like for regular bundle build URL. This will make it unified experience to run integ tests between release and bundle build URLs.

During the work, we notice that the current artifact promotion Jenkins job doesn't support OpenSearch Dashboards. We thus fix it along with this change.
 
### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/1482
Resolves https://github.com/opensearch-project/opensearch-build/issues/1489
 
### Test
To avoid impacting our release URL, for now, I apply my change to the `releases-test` path.

To highlight the difference, 
prod: https://artifacts.opensearch.org/**releases**/bundle/opensearch
my change: https://artifacts.opensearch.org/**releases-test**/bundle/opensearch

I verify that both OpenSearch and OpenSearch Dashboards integ test can run against the release URL.

```
OpenSearch

./test.sh integ-test manifests/1.2.0/opensearch-1.2.0-test.yml -p opensearch=https://artifacts.opensearch.org/releases-test/1.2.2
```

```
OpenSearch Dashboards
./test.sh integ-test manifests/1.2.0/opensearch-dashboards-1.2.0-test.yml -p opensearch=https://artifacts.opensearch.org/releases-test/1.2.2 opensearch-dashboards=https://artifacts.opensearch.org/releases-test/1.2.0
```

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1493, opened 2022-01-18
updated promote artifacts lib s3 upload

abhinavGupta16:

### Description
Remove s3 upload to `builds` and `dist` directly from `promoteArtifacts.groovy`
 
### Issues Resolved
Remove stray artifacts from release folder
 
### Check List


dblock:
@abhinavGupta16 Care to explain why adding this was a mistake, please?

dblock (approved):

tianleh (approved):

tianleh:
> @abhinavGupta16 Care to explain in the PR description why adding this was a bad idea, please?

Thanks @abhinavGupta16 for the quick fix. 

A bit context: I added the logic to upload `builds/dist` under `releases` url during the tar-promotion job as part of the following PR
https://github.com/opensearch-project/opensearch-build/pull/1490 with the intention to let Dashboards integ test run against a static OpenSearch url.

However, the tar-promotion job https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/promotion/tar-promotion.jenkinsfile only runs once manually during release meeting. Thus the url won't be ready (which also means that the Dashboards integ test won't get the OpenSearch url) until the last minute. 

A better option would be uploading such `builds/dist` in a periodic automatic job, e.g the distribution build job. Also since this build will be ongoing and is not actual `release` artifacts, we can consider uploading to `release-candidates` url.

dblock:
> > A better option would be uploading such `builds/dist` in a periodic automatic job, e.g the distribution build job. Also since this build will be ongoing and is not actual `release` artifacts, we can consider uploading to `release-candidates` url.

I think https://github.com/opensearch-project/opensearch-build/issues/1492 can solve this in a way where "latest" = latest released or latest build.

Issue #1185, opened 2021-11-29
[Bug] Docker Image support running without security and PA plugin


peterzhuamazon:
### Describe the bug

#1154 issue follow up after fixing docker build issues.

https://github.com/tianleh/opensearch-build/blob/main/docker/release/config/opensearch/opensearch-docker-entrypoint.sh#L100-L131

As of now we expect docker image to have security and PA plugin install by default.
If PA failed to start then OS will also terminate.

This behavior is ok for previous releases, but having issues in 1.3.0 since we dont have branches for security and PA yet, and the bundle build pipeline is still triggering snapshots builds on a daily basis for docker images.

We can resolve this by decoupling PA and OS if PA not exists.

Thanks.

### To reproduce

See above.

dblock:
There's no indication of what's failing, either.

```
> docker run -p 9200:9200 -p 9600:9600 -e "discovery.type=single-node" opensearchstaging/opensearch:1.3.0

Enabling execution of install_demo_configuration.sh for OpenSearch Security Plugin
OpenSearch Security Demo Installer
 ** Warning: Do not use on production or public reachable systems **
Basedir: /usr/share/opensearch
OpenSearch install type: rpm/deb on NAME="Amazon Linux"
OpenSearch config dir: /usr/share/opensearch/config
OpenSearch config file: /usr/share/opensearch/config/opensearch.yml
OpenSearch bin dir: /usr/share/opensearch/bin
OpenSearch plugins dir: /usr/share/opensearch/plugins
OpenSearch lib dir: /usr/share/opensearch/lib
Detected OpenSearch Version: x-content-1.3.0
Detected OpenSearch Security Version: 1.3.0.0

### Success
### Execute this script now on all your nodes and then start all nodes
### OpenSearch Security will be automatically initialized.
### If you like to change the runtime configuration
### change the files in ../securityconfig and execute:
"/usr/share/opensearch/plugins/opensearch-security/tools/securityadmin.sh" -cd "/usr/share/opensearch/plugins/opensearch-security/securityconfig" -icl -key "/usr/share/opensearch/config/kirk-key.pem" -cert "/usr/share/opensearch/config/kirk.pem" -cacert "/usr/share/opensearch/config/root-ca.pem" -nhnv
### or run ./securityadmin_demo.sh
### To use the Security Plugin ConfigurationGUI
### To access your secured cluster open https://<hostname>:<HTTP port> and log in with admin/admin.
[...]

PR #1507, opened 2022-01-18
Use docker entrypoint instead of command.

### LINKED ISSUE #1486: Make entrypoint scripts actual entrypoints in docker images

Currently, entrypoints scripts are defined as an docker commands and not as entrypoints:

- https://github.com/opensearch-project/opensearch-build/blob/fb20bc0e81bd6e055e40a9f86fcea22c7165c81a/docker/release/dockerfiles/opensearch.al2.dockerfile#L109
- https://github.com/opensearch-project/opensearch-build/blob/fb20bc0e81bd6e055e40a9f86fcea22c7165c81a/docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile#L87

Extra arguments passed to this entrypoint script are ignored (unless #1484 is merged at some point).

This is quite uncommon and prevents the commands to be easily reused or extended for other purposes.

I'd suggest doing something similar to the [postgres docker image](https://github.com/docker-library/postgres/blob/a83005b407ee6d810413500d8a041c957fb10cf0/14/bullseye/docker-entrypoint.sh#L298) (which I guess is pretty standard in Docker images).

- Define entrypoint scripts as `ENTRYPOINT` in the Dockerfile
- Define `openseach` or `opensearch-dashboards` as default `CMD` (this requires adding `$OPENSEARCH_HOME/bin`/`$OPENSEARCH_DASHBOARDS_HOME/bin` to the `PATH`)
- If the command passed to the entrypoint looks like a flag (i.e. starts with a dash), consider that `opensearch`/`opensearch-dashboards` command is implicit
- If the first argument passed to the entrypoint is `opensearch`/`opensearch-dashboards`, run the setup recipe and exec the command, forwarding extra arguments
- Otherwise, just exec the command

This would allow for the following `docker run` commands:
- `docker run opensearchproject/opensearch opensearch -Ediscovery.type=single-node`
- `docker run opensearchproject/opensearch -Ediscovery.type=single-node`
- `docker run opensearchproject/opensearch bash`

This is also slightly more extensible in case you want to use opensearchproject/opensearch as a base image for other images.

I think this suggestion is 100% backwards compatible, even for people who have overriden the `CMD` in a custom image (because the entrypoint would then invoke their custom `CMD` normally).

_No response_

### Additional context

See also #1484.

I'm happy to contribute if this suggestion seems OK to you.

gilbsgilbs:
_To reviewers:_ for an easier review, consider ignoring whitespace changes: https://github.com/opensearch-project/opensearch-build/pull/1507/files?diff=unified&w=1

### Description

This allows to use the entrypoints:

- as a standalone command ("./opensearch-docker-entrypoint.sh") which
  runs the setup procedure and starts opensearch just like before
- as an opensearch wrapper ("./opensearch-docker-entrypoint.sh
  opensearch" which does the exactly the same
- as a wrapper of any other command ("./opensearch-docker-entrypoint.sh
  bash") which just execs the command without the setup procedure

When wrapping the opensearch or opensearch-dashboard command, it'll also
forward the arguments:

- Option 1: "./opensearch-docker-entrypoint.sh -Ediscovery.type=single-node"
- Option 2: "./opensearch-docker-entrypoint.sh opensearch -Ediscovery.type=single-node"

### Issues Resolved

Fixes #1486
 
### Check List

### DIFF
--- a/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
+++ b/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
@@ -3,8 +3,8 @@
 # Copyright OpenSearch Contributors
 # SPDX-License-Identifier: Apache-2.0
 
-# 
-# Run OpenSearch-Dashboards, using environment variables to 
+#
+# Run OpenSearch-Dashboards, using environment variables to
 # set longopts defining OpenSearch-Dashboards's configuration.
 #
 # eg. Setting the environment variable:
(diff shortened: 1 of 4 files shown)


dblock (commented):
This is a lot better than what we have, will leave @peternied and @peterzhuamazon to do a more thorough CR.

Since the entrypoint scripts are already product-specific, should we standardize the default run/start parameter and pass in "run" or "start" instead of "opensearch" vs. "opensearch-dashboards"?

dblock:
Thanks for the detailed explanation, works for me.

dblock (approved):

docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
```diff
@@ -153,50 +153,70 @@ opensearch_dashboards_vars=(
...
-    fi
-fi
+    # Files created at run-time should be group-writable, for Openshift's sake.
+    umask 0002
+
+    ##Security Dashboards Plugin
+    setupSecurityDashboardsPlugin
 
+    # TO DO:
+    # Confirm with Mihir if this is necessary
```
dblock, line 191: Let's figure this one out and remove/update. @mihirsoni ?

dblock:
@gilbsgilbs Again, much thanks for making these changes. If you are interested, there's https://github.com/opensearch-project/opensearch-build/issues/1185 and a slew of other docker-related problems documented that could use help!

docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
mihirsoni: @dblock We need to test with the latest version of linux and see if we really need this or not.

dblock: Care to open an issue describing what this is about ?

peterzhuamazon:
Will take a closer look on this, will come back with some comments.
Thanks.

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
```diff
@@ -90,22 +67,74 @@ function terminateProcesses {
...
-OPENSEARCH_PID=$!
+    # Start opensearch
+    "$@" "${opensearch_opts[@]}" &
+    OPENSEARCH_PID=$!
 
-# Start performance analyzer agent
-$OPENSEARCH_HOME/bin/performance-analyzer-agent-cli > $OPENSEARCH_HOME/logs/performance-analyzer.log 2>&1 &
-PA_PID=$!
+    # Start performance analyzer agent
+    performance-analyzer-agent-cli > $OPENSEARCH_HOME/logs/performance-analyzer.log 2>&1 &
```
peterzhuamazon, line 112: Not sure if this is the best approach to rely on ` ENV PATH=$PATH:$OPENSEARCH_DASHBOARDS_HOME/bin` to find opensearch and pa cli script location.

I personally would like to give a full path to the script, but this is just nitpick tbh.

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
```diff
@@ -90,22 +67,74 @@ function terminateProcesses {
...
+    if [[ "$(id -u)" == "0" ]]; then
+        echo "OpenSearch cannot run as root. Please start your container as another user."
+        exit 1
+    fi
+
+    # Parse Docker env vars to customize OpenSearch
+    #
+    # e.g. Setting the env var cluster.name=testcluster
+    # will cause OpenSearch to be invoked with -Ecluster.name=testcluster
+    local opensearch_opts=()
```
peterzhuamazon, line 85: good one.

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
```diff
@@ -90,22 +67,74 @@ function terminateProcesses {
...
-echo "OpenSearch exited with code $?"
-wait $PA_PID
-echo "Performance analyzer exited with code $?"
+    # Wait for the child processes to terminate
+    wait $OPENSEARCH_PID
+    local opensearch_exit_code=$?
+    echo "OpenSearch exited with code ${opensearch_exit_code}"
+
+    wait $PA_PID
+    echo "Performance analyzer exited with code $?"
```
peterzhuamazon, line 121: Wonder if we want to return PA return code as well?

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
```diff
@@ -90,22 +67,74 @@ function terminateProcesses {
...
+# Prepend "opensearch" command if no argument was provided or if the first
+# argument looks like a flag (i.e. starts with a dash).
+if [ $# -eq 0 ] || [ "${1:0:1}" = '-' ]; then
+    set -- opensearch "$@"
+fi
+
+if [ "$1" = "opensearch" ]; then
+    # If the first argument is opensearch, then run the setup script.
+    runOpensearch "$@"
+    # Opensearch exit code should be forwarded, but it would be a breaking change.
```
peterzhuamazon, line 135: Where do you want to forward the return code?

peterzhuamazon (approved):
Overall great changes, thanks @gilbsgilbs just a few questions here.

Thanks.

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
gilbsgilbs: I just mean that the script should exit with the same exit code as the opensearch command (which implies that the line that just follows should be uncommented). I rephrased that.

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
gilbsgilbs: Easiest way to return multiple values in bash is to rely on globals isn't it? I can do that if we have a compelling reason to, but in the meantime I don't think we absolutely need to return anything at all. To keep things consistent, I removed the return value and if it turns out useful, we can consider adding some globals to track that. Does that sound ok?

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
gilbsgilbs: The rationale was that I wanted the user to be able to specify `openseach` as a docker CMD, so having `opensearch` in PATH at least made sense for that reason. Then, as opensearch (and consequently PA) were looked up from PATH, why not also starting PA from PATH?

It'd still be possible to not add anything to PATH, and make the entrypoint "hijack" the "opensearch" command (which it already does it in a way) to run it using the absolute path, but I reckon it'd be needlessly convoluted.

Anyways, I don't have strong opinions on that either so I'll just let it up to you.

peterzhuamazon: Yeah, I get what you are saying.
I am not against this approach at all, just a bit questioning whether there are better ways.
Anyway, we can try this out for a few snapshots builds and see if it comes up.

https://hub.docker.com/r/opensearchstaging/opensearch
https://hub.docker.com/r/opensearchstaging/opensearch-dashboards

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
peterzhuamazon: Sounds good to me, thank you.

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
peterzhuamazon: Got it.

peterzhuamazon:
Some checking:

The groovy test issues is not related to this PR.

I have opened an issue to track it.

Has seen similar behavior in PR #1520

https://github.com/opensearch-project/opensearch-build/issues/1521

Thanks.

peterzhuamazon:
Hi @gilbsgilbs #1522 has fixed the issue.
Can you pull from main branch and rebase/push again?

Thanks.

gilbsgilbs:
> Hi @gilbsgilbs #1522 has fixed the issue. Can you pull from main branch and rebase/push again?
> 
> Thanks.

Thanks. Just rebased my branch against main.

peternied:
Looks like the PR checks got blocked, I've retriggered them.  I'll merge this change assuming all goes well.

gilbsgilbs:
Thanks!

PR #1517, opened 2022-01-19
Add integ tests to OpenSearch distribution build jenkins job

ohltyler:
### Description
Adds a call to an integ test groovy script which triggers the integ test jenkins job, inside of the distribution build jenkins job.

More details:
- adds a `TEST_MANIFEST` parameter to the distribution-build job, needed for propagating to integ-test job
- tests are triggered within the async build steps, so both x64 and arm64 integ tests can run in parallel
- renames `manifest` arg used in many of the dependent groovy fns to `inputManifest` or `buildManifest` to make code more readable. Makes the fn calls explicitly set the input/build manifest args rather than relying on implicit env vars (e.g., relying on an existing `INPUT_MANIFEST` var to be set)

PR is currently in draft. Things to do before it is ready for merging:

- [x] add back the notification steps in distribution-build job
- [x] change references to test jobs (`Playground/ohltyler-*`) to production jobs
- [x] wrap integ test fn in a try/catch to prevent the distribution-build job to fail if the tests fail
- [x] propagate the test results in one of two ways:
    1. link the integ-test notification text with an associated distribution-build job (right now it only includes integ-test job info)
    2. propagate integ-test results inside of distribution-build job (not sure on the feasibility of that)

### Check List


jenkins/opensearch/distribution-build.jenkinsfile
```diff
@@ -87,18 +98,35 @@ pipeline {
...
-                            buildAssembleUpload()
+                            def buildManifestObj = buildAssembleUpload(
+                                inputManifest: "manifests/${INPUT_MANIFEST}"
+                            )
+                            String buildManifestUrl = buildManifestObj.getUrl(JOB_NAME, BUILD_NUMBER)
+
+                            echo "buildManifestUrl (x64): ${buildManifestUrl}"
+
+                            String testManifest = "manifests/${TEST_MANIFEST}"
+                            if (fileExists(testManifest)) {
```
ohltyler, line 120: @dblock @peternied any other suggestion for how to handle this case? For example, plugins typically aren't on the next major version (e.g. 2.0.0), and there is no test manifest located in `manifests/2.0.0`, and I'm assuming it's for this reason.

This solution just skips running the tests and adds some output explaining that no tests were ran since no test manifest was found.

peternied: What about not bothering to check? If the integration tests workflow can fail because the config doesn't exist seems appropriate to me.

We avoid having problems with build manifest not existing by having them automatically generated on new version detection [GitHub action](https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/versions.yml), [Manifest workflow](https://github.com/opensearch-project/opensearch-build/tree/main/src/manifests_workflow).  we should update that logic to add a test manifest automatically.  Alternatively this might be an appropriate time to merge the input manifest and test manifests into one

vars/assembleUpload.groovy
```diff
@@ -1,13 +1,8 @@
...
 
     lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
 
     assembleManifest(args)
     uploadArtifacts(args)
-
-    def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: args.manifest))
-    def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: args.manifest))
-    String artifactUrl = inputManifest.getPublicDistUrl("${PUBLIC_ARTIFACT_URL}", "${JOB_NAME}", "${BUILD_NUMBER}", buildManifest.build.platform, buildManifest.build.architecture)
-    echo "Setting env.\"ARTIFACT_URL_${buildManifest.build.platform}_${buildManifest.build.architecture}\"=${artifactUrl}"
```
peternied, line 11: This shouldn't be deleted, it creates the part of the build notification with instructions on where the build is located.

peternied (dismissed):

jenkins/opensearch/distribution-build.jenkinsfile
ohltyler: I see, so the reason there's no test manifest for 2.0.0 is because of the manual burden. And there really should be a test manifest for every input manifest. In that case I agree it makes sense to let the integ test workflow fail if no test manifest is passed.

Regarding automated test manifest generation, I'll look into it as a future PR.

vars/assembleUpload.groovy
ohltyler: From my understanding this isn't used by the build notifications - instead, messages are added in [uploadArtifacts](https://github.com/opensearch-project/opensearch-build/blob/main/vars/uploadArtifacts.groovy#L32) which is called in the build steps via [buildAssembleUpload](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L101). Then, the message with the links to the manifests are added in the notification's `extra` field [here](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L151).

But, these env vars are still used in [buildDockerImage](https://github.com/opensearch-project/opensearch-build/blob/main/vars/buildDockerImage.groovy#L9). These can be removed though, and we could set them explicitly in the pipeline itself (we already derive the value [here](https://github.com/ohltyler/opensearch-build/blob/add-to-distribution-build/jenkins/opensearch/distribution-build.jenkinsfile#L117)). Then they can be passed explicitly to `buildDockerImage()` instead of relying on an env var being set. What do you think?

vars/runIntegTests.groovy
```diff
@@ -0,0 +1,14 @@
+void call(Map args = [:]) {
```
peternied, line 1: I'm not sure this file as indirection is very useful the amount of code to call this function is nearly the same as this file.  Note; this would be more useful if it was automagically handling opensearch dashboards

FYI @tianleh

vars/assembleUpload.groovy
peternied: Ah, thanks for checking in on that.

I would prefer moving to parameters to remove these 'hidden' ENV variables

vars/runIntegTests.groovy
ohltyler: Right, this script isn't doing so much now, and is really only abstracting out the job name. I'm ok to remove and add directly in distribution-build workflow.

Regarding dashboards, ideally the integ-test job won't need new parameters - [runIntegTestScript](https://github.com/opensearch-project/opensearch-build/blob/main/vars/runIntegTestScript.groovy) (which is called in the integ-test job) is where the logic can be added to handle dashboards. It may need some additional args to handle multiple build manifests, for example.

vars/assembleUpload.groovy
ohltyler: Totally agreed, I'll go ahead with that approach

ohltyler:
Still investigating the instability in integ tests. Some are related to plugin-level flaky tests, others are related to the Jenkins envs. Tracking in meta issue here: https://github.com/opensearch-project/opensearch-build/issues/1531

I feel we shouldn't merge these changes until the integ tests become stable enough where there's not constant distribution-build failures because of this. The other option is try/catch the integ test failures for now to prevent the builds from being blocked, and remove at a later time when the tests are more stable.

peternied:
> Still investigating the instability in integ tests. Some are related to plugin-level flaky tests, others are related to the Jenkins envs. Tracking in meta issue here: https://github.com/opensearch-project/opensearch-build/issues/1531
> 
> 
> 
> I feel we shouldn't merge these changes until the integ tests become stable enough where there's not constant distribution-build failures because of this. The other option is try/catch the integ test failures for now to prevent the builds from being blocked, and remove at a later time when the tests are more stable.

I would prefer we run these tests more and start to flush out all the different random failures.

Can merge a version of this PR that does not stop the distribution jobs success notification and we get a second notification for the integ test pass/failure?

ohltyler:
> > Still investigating the instability in integ tests. Some are related to plugin-level flaky tests, others are related to the Jenkins envs. Tracking in meta issue here: #1531
> > I feel we shouldn't merge these changes until the integ tests become stable enough where there's not constant distribution-build failures because of this. The other option is try/catch the integ test failures for now to prevent the builds from being blocked, and remove at a later time when the tests are more stable.
> 
> I would prefer we run these tests more and start to flush out all the different random failures.
> 
> Can merge a version of this PR that does not stop the distribution jobs success notification and we get a second notification for the integ test pass/failure?

Agreed, that's why I've opened that other issue to try to throw in all of the different errors I've seen.

I think it makes sense to merge a non-blocking version for now and continue fixing the integ test failures.

Regarding notifications, the integ-test job itself already publishes those results in its own job. I think we could either:
1. Link the integ-test notification text with an associated distribution-build job (right now it only includes integ-test job info)
2. Propagate integ-test results inside of distribution-build job, but not sure on the feasibility of that just yet

vars/runIntegTests.groovy
tianleh: Thanks for the discussion about the support for OpenSearch Dashboards. It will need the artifact url for Dashboards from its own distribution build job (similar to the one in this PR) and one additional url for OpenSearch (which will be the `latest` url). Maybe there will be some common logic for the first part. However, to avoid using if/else as branching logic, I could image that it is very possible to be two different files. I am ok to perform some refactoring once I start working on the Dashboards part (after your work is merged).

peternied:
I'm holding off on reviewing until this switching over to `ready for review` whenever you are ready @ohltyler!

dblock (approved):
LGTM with the remaining renames

peternied (changes requested):
Super close, just a couple changes and lets get this merged

dblock (approved):

peternied (approved):

PR #1520, opened 2022-01-20
Migrating Maven sign and release jenkins jobs 

### LINKED ISSUE #1357: [Bug]: Maven sign and upload should be done inside a docker container
### Describe the bug

We should be using a docker container to run jenkinsjob. 

https://github.com/opensearch-project/opensearch-infra/blob/main/jenkins/jobs/OpenSearch_CI/release_ci/opensearch_maven_release/Jenkinsfile#L3-L5

Also, should this file be a part of `opensearch-build`?

### To reproduce

Docker container is not used - 

https://github.com/opensearch-project/opensearch-infra/blob/main/jenkins/jobs/OpenSearch_CI/release_ci/opensearch_maven_release/Jenkinsfile#L3-L5

### Expected behavior

_No response_

### Host / Environment

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

zelinh:
### Description
- Move this job from Infra repo to build repo since this could use the libraries in build repo and promote our public visibility and add the unit test for this Jenkins job. 
- Add the docker CI runner image to the workflow.
- Create the function in `signArtifacts.groovy` called `generateArguments` to dynamically add arguments for the sign scripts depending on the artifacts to be signed. E.g. If it's a manifest end with `.yml`, `type` arg will be required and for other artifacts, `signatureType` arg will be required.
- Update the corresponding test cases related to the modification on `signArtifacts.groovy`
- Create a new jenkins library `downloadFromS3.groovy` and add test cases for it.
 
 
### Issues Resolved
#1357 
 
### Check List


jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
```diff
@@ -0,0 +1,94 @@
...
+            }
+            environment {
+                REPO_URL = "https://aws.oss.sonatype.org/"
+                STAGING_PROFILE_ID = "${SONATYPE_STAGING_PROFILE_ID}"
+                BUILD_ID = "${params.BUILD_ID}"
+            }
+            steps {
+                script {
+                    git url: 'https://github.com/opensearch-project/opensearch-build.git', branch: 'main'
+                    downloadFromS3(
```
abhinavGupta16, line 73: Do we need to upload the Artifacts on s3? Since we are downloading them again here.

src/run_sign.py
```diff
@@ -15,13 +15,19 @@
...
+    return val
 
 
 def main():
     parser = argparse.ArgumentParser(description="Sign artifacts")
     parser.add_argument("target", type=Path, help="Path to local manifest file or artifact directory.")
-    parser.add_argument("--component", nargs="?", help="Component name")
+    parser.add_argument("--component", nargs="?", type=null_str, help="Component name")
     parser.add_argument("--type", nargs="?", help="Artifact type")
     parser.add_argument("--sigtype", choices=ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of Signature file", default=".asc")
```
abhinavGupta16, line 32: Could you also change the parser to pick default in case a null or a None is passed?

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
```diff
@@ -0,0 +1,94 @@
...
+
+pipeline {
+    agent {
+        docker {
+            label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+            image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211130'
+            alwaysPull true
+        }
+    }
+    parameters {
```
abhinavGupta16, line 11: Needs description

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
```diff
@@ -0,0 +1,94 @@
...
+        string(
```
abhinavGupta16, line 12: Let's make this is a choice rather than a string input, since this would always be a fixed value of the dropdown list

abhinavGupta16 (changes requested):
Needs test cases for job and lib.

Thanks for migrating the job here.

abhinavGupta16:
ci is failing

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
zelinh: I would prefer we leave what it is right now. As @peternied said, we don't know what architectures we could potentially support in the future and it takes efforts to track and update choices of all jobs.

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
zelinh: That makes sense. In the same workspace, both upload and this download are not needed.

src/run_sign.py
zelinh: The default would not be picked at all if calling from https://github.com/opensearch-project/opensearch-build/blob/main/vars/signArtifacts.groovy, because we specified `--component=${args.component}`. In that case, even we leave the `args.component` empty or null, parser will take it as a `string` type with empty content or string `null` . 
After my change, the argument type will be set as `NoneType` so still no default would be needed.

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
abhinavGupta16: Okay. I suggested a menu since it creates confusion as to what values needs to be filled in, which in this case is only `x64`. 

Out of curiosity, is it possible to have a centralized parameterized list for cases like these, params for a job?

src/run_sign.py
abhinavGupta16: Yes, you are correct. Maybe we should change to code such that if the value being passed is 'null' or 'None', the default value gets picked automatically. That way, anyone calling the script with null or None should be able to run with defaults. Thoughts?

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
```diff
@@ -0,0 +1,94 @@
...
+            name: 'BUILD_ID',
+            trim: true
+        )
+        string(
+            name: 'VERSION',
+            trim: true
+        )
+    }
+    environment {
+        ARTIFACT_PATH = "distribution-build-opensearch/${params.VERSION}/${params.BUILD_ID}/${params.PLATFORM}/${params.ARCH}/builds"
```
abhinavGupta16, line 32: don't use `params.ARCH`, only `ARCH` would also work. using `params.ARCH` will create problems when setting variable bindings in tests

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
```diff
@@ -0,0 +1,94 @@
...
+        docker {
+            label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+            image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211130'
+            alwaysPull true
+        }
+    }
+    parameters {
+        string(
+            defaultValue: 'x64',
+            name: 'ARCH',
```
abhinavGupta16, line 14: Unrelated thought @peternied @zelinh - do we really need this argument (`ARCH` and `PLATFORM`), given they are only used as folder paths?

zelinh: Are you saying we could combine `ARCH` and `PLATFORM` to one parameter like called `relative path`?

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
zelinh: I think using `param.ARCH` is more reasonable than just `ARCH`. What issue would it cause for testing?

src/run_sign.py
zelinh: I think my way of resolving it is quite same as the circumstance you are saying. If any `null` or empty string passed in, parser will set its type to be `None`; not really necessary to have a default.

abhinavGupta16: Apologies for the confusion there. I meant for other parameters as well and default value in case of `--sigType`

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
abhinavGupta16: `parms.ARCH` variables are not set and picked up properly when used with bindings. I faced a similar problem when writing tests for ecr.

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
abhinavGupta16: We could remove this parameter altogether since the value is fixed.

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
abhinavGupta16: Upon further looking into this - if don't provide a dropdown list, the user can enter any value for `ARCH` and `platform`. So, one can save and create artifacts of any platform and architecture given there is no check for the values input as a string. This can be both deceiving and dangerous.

For example, I can simple enter `arm64` and the job would run fine and save artifacts as `arm64` which is incorrect.

gaiksaya: In that case why give an option to enter the ARCH? If we know arm64 won't work or something we do not want, `x64` is the hardcode value that can go here.

zelinh: Is the signer currently only working on `x64` platform? I'm not sure why other architecture is incorrect.

gaiksaya: We upload only x64 on maven. arm64 is out of the picture. I do not know the reason.

abhinavGupta16:
Following up on my discussion with @peternied - The `PLATFORM` and `ARCH` parameters are required for getting the download urls for the artifacts which can we changed easily. Also, path is not fixed and any changes to the path will cause this job to fail. We should remove `PLATFORM` and `ARCH` parameters. 

We can alternatively do the following - 
- Take manifest file as input (or get one using the version number) and get the artifact path from the file directly
- Take in the artifact path as input and print the artifact path as a part of the build so it is easily available. 

I personally prefer the first one, since it truly removes manual work and automates the process, using our existing one stop shop of `manifest` files which are assured to have the most up to date values.

jenkins/release_ci/opensearch_maven_release/maven-sign-release.jenkinsfile
abhinavGupta16: Exactly, but this can potentially have `arm64` as input, which is incorrect

vars/signArtifacts.groovy
```diff
@@ -24,11 +26,21 @@ void call(Map args = [:]) {
             export UNSIGNED_BUCKET=${SIGNER_CLIENT_UNSIGNED_BUCKET}
             export SIGNED_BUCKET=${SIGNER_CLIENT_SIGNED_BUCKET}
 
-            $WORKSPACE/sign.sh ${args.artifactPath} --sigtype=${args.signatureType} --component=${args.component} --type=${args.type}
+            $WORKSPACE/sign.sh ${arguments}
```
zelinh, line 29: Removed the previous modification on `run_sign.py` for circumstance of pass `None` but add the check here so that if any arguments are not specified, it wouldn't pass to the sign script.

vars/signArtifacts.groovy
```diff
@@ -24,11 +26,21 @@ void call(Map args = [:]) {
...
         """
     }
 }
 
+String generateArguments(args) {
+    String arguments = args.artifactPath
+    for (entry in args) {
+        if (!entry.key.equals("artifactPath")) {
+            arguments += " --${entry.key}=${entry.value}"
```
abhinavGupta16, line 38: entry.key would not work with `--sigType`

zelinh: We didn't have a unified name for the signature type here. According to (https://github.com/opensearch-project/opensearch-build/blob/f3d8bcda4bc015786a7c167e1563fb22632ff763/src/run_sign.py#L26) we should then change the related Jenkins files arguments with arg like `sigtype: .sig` instead of `signatureType` as before.

abhinavGupta16: I think changing `opensearch-build/src/run_sign.py` would be better since `signArtifacts` is being called from a lot of places, and would require to you to test out a lot of jobs.

zelinh: That makes sense to me. Changed.

abhinavGupta16: We can also remove the function and do the following - 

```groovy

// Since the first argument should be artifactPath
String arguments = args.artifactPath

args.each(entry -> entry.key.equals('artifactPath') ? "" : arguments += " --${entry.key}=${entry.value}")

```

abhinavGupta16: adding this has only one issue, it's hard to recognize what arguments does this function take. I think it would be a good idea to add a javadoc like comment on top describing the params. WDYT?

gaiksaya: Looks like out of scope changes. @zelinh Maybe refactoring come in later PR and this PR should concentrate on maven job only?

abhinavGupta16: The maven job would require these changes to run with opensearch-build setup.

zelinh: We could add a description comment on top of this file for the instruction of this library which will tell people how the acceptable arguments.

zelinh:
Add the unit test based on @abhinavGupta16 PR for new jenkins test framework. This is blocked by https://github.com/opensearch-project/opensearch-build/pull/1471/ until like `LibFunctionTester.groovy` and new `BuildPipelineTest.groovy` are merged.

abhinavGupta16:
> Add the unit test based on @abhinavGupta16 PR for new jenkins test framework. This is blocked by #1471 until like `LibFunctionTester.groovy` and new `BuildPipelineTest.groovy` are merged.

Added dependency here

abhinavGupta16:
> Add the unit test based on @abhinavGupta16 PR for new jenkins test framework. This is blocked by #1471 until like `LibFunctionTester.groovy` and new `BuildPipelineTest.groovy` are merged.

This PR is now merged. You can use the instructions from https://github.com/opensearch-project/opensearch-build/issues/1560 to write the test cases. Thanks!

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
```diff
@@ -0,0 +1,46 @@
+import static org.hamcrest.CoreMatchers.notNullValue
+import static org.hamcrest.MatcherAssert.assertThat
+
+class TestDownloadFromS3LibTester extends LibFunctionTester {
```
abhinavGupta16, line 4: Let's rename the class to `DownloadFromS3LibTester`

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    String libFunctionName(){
+        return 'downloadFromS3'
+    }
+
+    void configure(helper, binding){
+        binding.setVariable('ARTIFACT_DOWNLOAD_ROLE_NAME', 'Dummy_Download_Role')
+        binding.setVariable('AWS_ACCOUNT_PUBLIC', 'dummy_account')
+        binding.setVariable('ARTIFACT_BUCKET_NAME', 'dummy_bucket_name')
```
abhinavGupta16, line 39: This variable is unused

vars/signArtifacts.groovy
```diff
@@ -6,6 +6,9 @@
  * compatible open source license.
  */
 
+/*
+This library will take in arguments Map with only these options as Keys [artifactPath, component, type, signatureType, distributionPlatform] with case-sensitive
+*/
```
abhinavGupta16, line 11: we could add something similar to javadocs format for all params
```groovy
/*
SignArtifacts signs the given artifacts and saves the signature in the same directory

@param Map[artifactPath] - Path to yml or artifact file
@param Map[component] <optional> - Path to yml or artifact file
*/
```

abhinavGupta16 (changes requested):
Almost there! Let's add a test case for `maven-sign-release.jenkinsfile`. Also, we can get rid of `release_ci` in the directory structure to be consistent with other jobs

vars/signArtifacts.groovy
```diff
@@ -14,6 +17,8 @@ void call(Map args = [:]) {
 
     importPGPKey()
 
+    String arguments = generateArguments(args)
```
abhinavGupta16, line 20:
```suggestion
    
// Since the first argument should be artifactPath
String arguments = args.artifactPath

args.each(entry -> entry.key.equals('artifactPath') ? "" : arguments += " --${entry.key}=${entry.value}")

```

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
```diff
@@ -0,0 +1,46 @@
...
+        this.bucket = bucket
+        this.path = path
+        this.force = force
+    }
+
+    void parameterInvariantsAssertions(call){
+        assertThat(call.args.sourcePath.first(), notNullValue())
+        assertThat(call.args.bucket.first(), notNullValue())
+        assertThat(call.args.path.first(), notNullValue())
+        assertThat(call.args.force.first(), notNullValue())
```
abhinavGupta16, line 22: We can check if the value for force is a boolean here

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    boolean expectedParametersMatcher(call) {
+        return call.args.sourcePath.first().toString().equals(this.sourcePath)
+                && call.args.bucket.first().toString().equals(this.bucket)
+                && call.args.path.first().toString().equals(this.path)
+                && call.args.force.first() == (this.force)
```
abhinavGupta16, line 29:
```suggestion
                && call.args.force.first().toString().equals(this.force)
```

vars/signArtifacts.groovy
gaiksaya: This change feels over complicated. Why add a loop to get all the arguments when you can get it directly form `args.variable`or maybe I am misunderstanding something?

zelinh: I was trying to iterating through the `args` without knowing what arguments would be passed in. So use the loop to iterate through the keys of the `args` map.

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
zelinh: This file name is `TestUploadToS3LibTester` so I think these two are similar. https://github.com/opensearch-project/opensearch-build/blob/main/tests/jenkins/lib-testers/TestUploadToS3LibTester.groovy

vars/signArtifacts.groovy
zelinh: Ok, this seems to be more concise.

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
zelinh: This is boolean why we check if the string is equal()? Like I define on line 9.

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
zelinh: Will remove it.

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
abhinavGupta16: ohh. That's my bad then. I'll rename that file as well. The other 2 are named correctly. Thanks for pointing it out!

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
abhinavGupta16: I think everything inside `call.args.force.first()` is a string. `this.force` is boolean.

zelinh: If that's the case, we will need something like this 
`call.args.force.first().toString().equals(this.force.toString())` to compare the actual context of string for argument `force`

tests/jenkins/lib-testers/TestDownloadFromS3LibTester.groovy
abhinavGupta16: https://github.com/opensearch-project/opensearch-build/pull/1562

tests/jenkins/TestMavenSignReleaseJob.groovy
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.Test
+
+class TestMavenSignReleaseJob extends BuildPipelineTest {
+
+    @Before
+    void setUp() {
+
+        def localPath = "${this.workspace}/artifacts"
+
+        def artifactsPath = "${this.workspace}/artifacts/distribution-build-opensearch/1.0.0/123/linux/x64/builds/opensearch/manifest.yml"
```
abhinavGupta16, line 18: Is this the `manifestPath`?

tests/jenkins/TestMavenSignReleaseJob.groovy
```diff
@@ -0,0 +1,46 @@
...
+
+        def bucketPath = 'distribution-build-opensearch/1.0.0/123/linux/x64/builds'
```
abhinavGupta16, line 20: I think this is the `artifactPath` right? We could use this variable to set the values on line 24 and 33

vars/signArtifacts.groovy
```diff
@@ -6,13 +6,23 @@
  * compatible open source license.
  */
 
+/*
+SignArtifacts signs the given artifacts and saves the signature in the same directory
+@param Map[artifactPath] <Required> - Path to yml or artifact file.
+@param Map[component] <Optional> - Path to yml or artifact file.
+@param Map[type] <Optional> - Artifact type in the manifest, [type] is required for signing yml.
+@param Map[signatureType] <Optional> - The signature type of signing artifacts. e.g. '.sig'. Required for non-yml artifacts signing.
+@param Map[distributionPlatform] <Required> - The distribution platform for signing. Choices: ['linux']
```
abhinavGupta16, line 15: Let's remove `Choices: ['linux']` from here since this is not being tested in code.

abhinavGupta16 (commented):
Suggested some minor changes in variable names for better understanding. The rest looks good to me

jenkins/opensearch_maven_release/maven-sign-release.jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+    environment {
+        ARTIFACT_PATH = "distribution-build-opensearch/${VERSION}/${BUILD_ID}/linux/x64/builds"
+    }
+    stages {
+        stage('sign') {
+            steps {
+                script {
+                    sh 'echo "Downloading from S3."'
+                    downloadFromS3(
+                        localPath: "$WORKSPACE/artifacts",
```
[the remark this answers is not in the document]
zelinh: I could rename the `localPath` to `destPath`; for `path`, it's actually originally from Jenkins (https://www.jenkins.io/doc/pipeline/steps/pipeline-aws/#s3download-copy-file-from-s3), so I would like to keep it as what it is. WDYT?

tests/jenkins/TestMavenSignReleaseJob.groovy
zelinh: More precisely yes.

tests/jenkins/TestMavenSignReleaseJob.groovy
zelinh: Good point!

tests/jenkins/TestMavenSignReleaseJob.groovy
```diff
@@ -0,0 +1,46 @@
...
+        this.registerLibTester(new DownloadFromS3LibTester(destPath, bucketName, artifactsPath, true))
+
+        this.registerLibTester(new SignArtifactsLibTester( '.sig', 'linux',  manifestPath, 'maven', null))
+
+        super.setUp()
+
+        // Variables for Maven Sign Release job
+        binding.setVariable('VERSION', '1.0.0')
+        binding.setVariable('BUILD_ID', '123')
+        binding.setVariable('ARTIFACT_PATH', 'distribution-build-opensearch/1.0.0/123/linux/x64/builds')
```
abhinavGupta16, line 33: This is `artifactPath` too

abhinavGupta16, line 33: This is `artifactPath` too.

abhinavGupta16 (approved):

gaiksaya (changes requested):
Please run and test this workflow before merging. This jenkins job is not something that is exercised regularly so want to be sure before merging.

zelinh:
> Please run and test this workflow before merging. This jenkins job is not something that is exercised regularly so want to be sure before merging.

I tested with upload and maven stage commented out. Should I test the uploading part?

gaiksaya:
> > Please run and test this workflow before merging. This jenkins job is not something that is exercised regularly so want to be sure before merging.
> 
> I tested with upload and maven stage commented out. Should I test the uploading part?

Yes. It won't publish anything. It will stage the repo which we can drop later on.

jenkins/opensearch_maven_release/maven-sign-release.jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+        )
+    }
+    environment {
+        ARTIFACT_PATH = "distribution-build-opensearch/${VERSION}/${BUILD_ID}/linux/x64/builds"
+    }
+    stages {
+        stage('sign') {
+            steps {
+                script {
+                    sh 'echo "Downloading from S3."'
```
dblock, line 30: just echo

jenkins/opensearch_maven_release/maven-sign-release.jenkinsfile
```diff
@@ -0,0 +1,72 @@
+lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
```
dblock, line 1: Folders here use `-` and not `_`, so make it `opensearch-maven-release`.

jenkins/opensearch_maven_release/maven-sign-release.jenkinsfile
```diff
@@ -0,0 +1,72 @@
...
+        docker {
+            label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+            image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211130'
+            alwaysPull true
+        }
+    }
+    parameters {
+        string(
+            name: 'BUILD_ID',
+            description: 'Build ID from Jenkins OpenSearch distribution build used for artifact path',
```
dblock, line 14: The description "used for artifact path" doesn't tell me anything. Do you mean "Build ID of the artifacts to stage"?.

vars/signArtifacts.groovy
```diff
@@ -24,11 +34,19 @@ void call(Map args = [:]) {
...
 
-            $WORKSPACE/sign.sh ${args.artifactPath} --sigtype=${args.signatureType} --component=${args.component} --type=${args.type}
+            $WORKSPACE/sign.sh ${arguments}
         """
     }
 }
 
+String generateArguments(args) {
+    // artifactPath is mandatory and the first argument
+    String arguments = args.artifactPath
```
dblock, line 44: Make a copy of `args` without `artifactsPath`, then iterate through the rest.

dblock (changes requested):

vars/signArtifacts.groovy
zelinh: I updated the method that iterates through the `args` with `artifactsPath` removed.

zelinh:
Did a test run on this job with parameters: [`BUILD_ID`: 295, `VERSION`: 1.2.3] and it ran successfully. 
Validated the artifacts are correct and signed on staging maven with @gaiksaya, and dropped after validation.

gaiksaya (approved):

src/run_sign.py
```diff
@@ -23,7 +23,9 @@ def main():
     parser.add_argument("target", type=Path, help="Path to local manifest file or artifact directory.")
     parser.add_argument("--component", nargs="?", help="Component name")
     parser.add_argument("--type", nargs="?", help="Artifact type")
-    parser.add_argument("--sigtype", choices=ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of Signature file", default=".asc")
+    parser.add_argument("--signatureType", choices=ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of Signature file", default=".asc")
+    # This distributionPlatform argument is added for future use to extend the functionality of this library.
+    parser.add_argument("--distributionPlatform", nargs="?", help="The distribution platform")
```
dblock, line 28: To be consistent with other tools i think this should just be `platform`. I would default it to `linux` and remove the comment above.

dblock (approved):
LGTM. Consider renaming `distributionPlatform`. No strong feelings.

src/run_sign.py
zelinh: Removed the comments and add the default value. I will leave the argument name to be `distributionPlatform` since that's what it used in some jenkinsfiles. e.g. [sign-standalone-artifacts.jenkinsfile](https://github.com/opensearch-project/opensearch-build/blob/0a20ec18c3a39770e870b2dade3c23b147962651/jenkins/sign-artifacts/sign-standalone-artifacts.jenkinsfile#L47https://github.com/opensearch-project/opensearch-build/blob/0a20ec18c3a39770e870b2dade3c23b147962651/jenkins/sign-artifacts/sign-standalone-artifacts.jenkinsfile#L47); [distribution-artifacts.jenkinsfile](https://github.com/opensearch-project/opensearch-build/blob/0a20ec18c3a39770e870b2dade3c23b147962651/jenkins/data-prepper/distribution-artifacts.jenkinsfile#L34)

src/run_sign.py
```diff
@@ -23,7 +23,8 @@ def main():
     parser.add_argument("target", type=Path, help="Path to local manifest file or artifact directory.")
     parser.add_argument("--component", nargs="?", help="Component name")
     parser.add_argument("--type", nargs="?", help="Artifact type")
-    parser.add_argument("--sigtype", choices=ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of Signature file", default=".asc")
+    parser.add_argument("--signatureType", choices=ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of Signature file", default=".asc")
+    parser.add_argument("--distributionPlatform", nargs="?", help="The distribution platform", default="linux")
```
dblock, line 27: Actually I think we should change this to match the other tools. Jenkins argument is beyond the point, but we we don't use camelCase elsewhere, e.g. https://github.com/opensearch-project/opensearch-build/blob/e45558ecca8c190f694d1fa52707a184d999c3c3/src/assemble_workflow/assemble_args.py#L19.

So I think this should be back to `--sigtype` and `--platform` or we're introducing a different pattern in Python tools.

dblock (changes requested):
See below.

src/run_sign.py
zelinh: @abhinavGupta16 Do you have any objections we moving this back to `sigtype`? 
We would also need to change the arguments in quite some places where `signArtifacts.groovy` is called.

zelinh:
Name convention for the type of signature (e.g. `.sig`) has been changed from `signatureType` to `sigtype` across all jenkins job calling `signArtifacts` to be consistent with other tools as we didn't use camelCases for our arguments or variables.

dblock (approved):
This works for me.

Generally the boundary should be at interface points, so I don't think you needed to change the convention in groovy code which generally uses camelCase, only in python.

PR #1526, opened 2022-01-20
Signing by ASC should be supported

peternied:
### Description
Removing 'choice' parameters that are hiding potential parameters used
in downstream systems.  Ran into this while running a deployment and had to fall back to manual
steps.

Cleaned up the documentation and provided links to lookup the available
options.

### Check List

### DIFF
--- a/jenkins/sign-artifacts/sign-standalone-artifacts.jenkinsfile
+++ b/jenkins/sign-artifacts/sign-standalone-artifacts.jenkinsfile
@@ -19,15 +19,17 @@ pipeline {
                 trim: true,
                 description: 'Path to upload to artifacts and signatures on s3. Eg: dummy_project/1.0'
         )
-        choice(
-                choices: ['linux'],
+        string(
+                defaultValue: 'linux',
                 name: 'DISTRIBUTION_PLATFORM',
+                trim: true,
                 description: 'What platform is this distribution build for?'
         )
-        choice(
-                choices: ['.sig'],
+        string(
+                defaultValue: '.sig',
                 name: 'SIGNATURE_TYPE',
-                description: 'What is signature file type?'
+                trim: true,
+                description: 'What is signature file type?  See https://github.com/opensearch-project/opensearch-build/blob/main/src/sign_workflow/README.md for all options'
         )
     }
     stages {
--- a/src/run_sign.py
+++ b/src/run_sign.py
@@ -15,7 +15,7 @@ from sign_workflow.sign_artifacts import SignArtifacts
 from sign_workflow.signer import Signer
 from system import console
 
-ACCEPTED_SIGNATURE_FILE_TYPES = [".sig"]
+ACCEPTED_SIGNATURE_FILE_TYPES = [".sig", ".asc"]
 
 
 def main():
--- a/src/sign_workflow/README.md
+++ b/src/sign_workflow/README.md
@@ -13,10 +13,12 @@ The signing step (optional) takes the manifest file created from the build step 
 
 The following options are available. 
 
-| name          | description                                                                           |
-|---------------|---------------------------------------------------------------------------------------|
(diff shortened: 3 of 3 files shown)


src/run_sign.py
```diff
@@ -15,7 +15,7 @@
 from sign_workflow.signer import Signer
 from system import console
 
-ACCEPTED_SIGNATURE_FILE_TYPES = [".sig"]
+ACCEPTED_SIGNATURE_FILE_TYPES = [".sig", ".asc"]
```
gaiksaya, line 18: How will this work? The signer client does not support file types yet. `sig` is what it returns by default. If we add this, either the signer client needs to support this or we need to add additional code on our end to convert `sig` files to `.asc` files

peterzhuamazon: I agree with @gaiksaya that we need to 1st let signer client to have this function, before adding it to the signing workflow.

peternied: AFAIK It does support this functionality, if you look further down the default is `.asc` if no type is supplied

peternied: If this is not correct and we cannot use `.asc` then that is another issue entirely - and yes then I should just close this PR until that is resolved

peterzhuamazon: I thought the default is to binary sig file?
It does not have a enarmor function to change to ascii?
Please correct me if not the case.

gaiksaya: Here is the related issue https://github.com/opensearch-project/opensearch-signer-client/issues/10

gaiksaya: Default is `.asc` here https://github.com/opensearch-project/opensearch-build/blob/c18e1363910031d0368aa4a949f6dc507c442d72/src/run_sign.py#L26 where is technically not an enarmor output

zelinh: Out of curiosity, If signer client doesn't have functionality to sign `.asc`, how did we do signing files for maven release [here](https://github.com/opensearch-project/opensearch-infra/blob/f71d0d72f79c02dcd7cd00c7aabadc0289e12d12/jenkins/jobs/OpenSearch_CI/release_ci/opensearch_maven_release/Jenkinsfile#L58)? Wers this output fake `.asc` files without actually converted?

jenkins/sign-artifacts/sign-standalone-artifacts.jenkinsfile
```diff
@@ -19,15 +19,17 @@ pipeline {
...
-                choices: ['linux'],
+        string(
+                defaultValue: 'linux',
                 name: 'DISTRIBUTION_PLATFORM',
+                trim: true,
                 description: 'What platform is this distribution build for?'
         )
-        choice(
-                choices: ['.sig'],
+        string(
```
abhinavGupta16, line 28: I don't think we should change this to string as it would allow user to enter any value for creating signature. It needs to be from a fixed set of list. The sole reason of dropdown menus is to prevent this.

src/run_sign.py
abhinavGupta16: The reason we left out the `.asc` here was because the `.asc` files created by the signer client are technically incorrect as @peternied pointed it out in the prev PR - 

https://github.com/opensearch-project/opensearch-build/pull/1352#discussion_r771459100

I think a better solution would be to make sure the params pick out default value in case a `null`, `None` or an empty string is passed

peternied: We've been working around opensearch-project/opensearch-signer-client#10 since OpenSearch 1.0.0  Lets just implement the work around in this code base as we aren't getting traction on that issue.

peternied:
Thanks for the great dialog everyone, I didn't get around to adding the 'enarmor' working into the signing code so I'm going to close this PR out.  In the future please feel free to pick this up if I don't get back to it first :)

PR #1350, opened 2021-12-16
[Dashboards] Bumps Node version to 14.18.2 for Dashboards Node upgrade

boktorbb-amzn:
Signed-off-by: Bishoy Boktor <boktorbb@amazon.com>

### Description
Bumps Node version to 14.18.2 for Dashboards Node upgrade 
 
### Issues Resolved
#1320 
 
### Check List
- [X] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1527, opened 2022-01-21
Add CI Image for Dashboards Build

peternied:
### Description
I've copied the existing dashboard build image into a fresh file so these can diverge and included the `DOCKER_BUILD_SCRIPT_WITH_COMMANDS` in the image itself to keep the tagging consistent as changes are made.

See the image at https://hub.docker.com/r/opensearchstaging/ci-runner/tags?page=1&ordering=last_updated&name=ci-runner-centos7-opensearch-dashboards-build

### Issues Resolved
- Follows up #1350 that configured the image
 
### Check List

### DIFF
--- a/docker/ci/dockerfiles/ci-runner.centos7.opensearch-dashboards.build.dockerfile
+++ b/docker/ci/dockerfiles/ci-runner.centos7.opensearch-dashboards.build.dockerfile
@@ -1,0 +1,180 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+# This is a docker image specifically for standardize the ci/cd environment
+# for both developers and ci/cd tools in OpenSearch / OpenSearch-Dashboards
+# Please read the README.md file for all the information before using this dockerfile
+
+# DOCKER_BUILD_SCRIPT_WITH_COMMANDS 
+#  cd docker/ci && bash ./build-image-multi-arch.sh -v ci-runner-centos7-opensearch-dashboards-build -f dockerfiles/ci-runner.centos7.opensearch-dashboards.build.dockerfile
+
+FROM centos:7
+
+ARG MAVEN_DIR=/usr/local/apache-maven
+
+# Ensure localedef running correct with root permission
+USER 0
+
+# Add AdoptOpenJDK Repo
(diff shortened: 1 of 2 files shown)


kavilla (approved):

boktorbb-amzn (approved):
Looks good. One question, why is the node version being removed from the image name?

peternied:
> Looks good. One question, why is the node version being removed from the image name?

When we are attempting to 'manage' all of the docker images, they all begin to blur together, while the name I've created is maybe not the best name, its at least easier to understand its intent and what would potentially break if we updated it

docker/ci/dockerfiles/ci-runner.centos7.opensearch-dashboards.build.dockerfile
```diff
@@ -0,0 +1,180 @@
...
+            MAJOR=${BASH_REMATCH[1]}; \
+            curl -LfsSo /tmp/openjdk-${MAJOR}.tar.gz ${BINARY_URL}; \
+            echo "${ESUM} */tmp/openjdk-${MAJOR}.tar.gz" | sha256sum -c -; \
+            mkdir -p /opt/java/openjdk-${MAJOR}; \
+            cd /opt/java/openjdk-${MAJOR}; \
+            tar -xf /tmp/openjdk-${MAJOR}.tar.gz --strip-components=1; \
+            rm -rf /tmp/openjdk-${MAJOR}.tar.gz; \
+            echo "export JAVA${MAJOR}_HOME=/opt/java/openjdk-${MAJOR}" >> /etc/profile.d/java_home.sh; \
+        fi; \
+    done;
```
peterzhuamazon, line 90: There is no reason to include these JDKs in dashboards building.

peterzhuamazon (changes requested):
I have questions on this approach as it includes everything just like the main image.
Which to me is just duplications.

I would rather this image is simple and clean with only dashboards related libs installed.

Thoughts?

peternied:
> I have questions on this approach as it includes everything just like the main image.

As we have a functional system today and limited automating testing, I'd like pair down what is included iteratively

docker/ci/dockerfiles/ci-runner.centos7.opensearch-dashboards.build.dockerfile
peternied: Note; I'd prefer to do this as a separate PR

boktorbb-amzn: I can confirm that removing them shouldn't affect us @peternied . Will defer to you for whether it should be done in this PR or opening a separate one

peterzhuamazon (approved):

Issue #1510, opened 2022-01-18
[Python Typechecking] Paths `pipenv run mypy src/paths tests/tests_paths`


peternied:
See full instructions at https://github.com/opensearch-project/opensearch-build/issues/1237

Acceptance criteria:
 - Run `pipenv run mypy src/paths tests/tests_paths` passes without error

PR #1546, opened 2022-01-25
Added Python typechecking for mypy for `src/paths` `tests/tests_paths`

abhinavGupta16:
### Description
`pipenv run mypy src/paths tests/tests_paths` passes without errors
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1510
 
### Check List


peternied (changes requested):
Please also update the following to enforce mypy on the new paths you've added:
- .github/workflows/python-tests.yml
- .pre-commit-config.yaml

You can see how this is done in https://github.com/opensearch-project/opensearch-build/pull/1224/files

peterzhuamazon:
The name should be something like `Adding Type Checking ......` not fixing mypy issues. I was quite confused for a while and think there is issue on how to use mypy in our workflows.

src/paths/build_output_dir.py
```diff
@@ -8,5 +8,5 @@
 
 
 class BuildOutputDir(OutputDir):
-    def __init__(cls, filename, cwd=None, makedirs=True):
+    def __init__(cls, filename: str, cwd: str=None, makedirs: bool=True) -> None:
```
gaiksaya, line 11: Shouldn't the cwd be of type `Path`?

abhinavGupta16: Changed it to `Union[str, PathLike[str]]`

peternied: I think this should stay as a string, in the unit tests the only type we use is string.  In the codebase its only reference is a string in `run_build.py`

peternied (approved):

src/paths/build_output_dir.py
peternied: Note; the impact of this type is limited, whereas in https://github.com/opensearch-project/opensearch-build/pull/1564 its much broader in scope.

I would still prefer the return type changed to string

gaiksaya: Makes sense. @abhinavGupta16 are you changing it back to string?

abhinavGupta16: yes, I changed it all to `str` now

dblock (approved):

Issue #1511, opened 2022-01-18
[Python Typechecking] Test system `pipenv run mypy src\system tests\tests_system`


peternied:
See full instructions at https://github.com/opensearch-project/opensearch-build/issues/1237

Acceptance criteria:
 - Run `pipenv run mypy src/system tests/tests_system` passes without error

PR #1564, opened 2022-01-27
Python typechecking, mypy issues for src/systems and tests/tests_system

abhinavGupta16:
### Description
Python typechecking, mypy issues for src/systems and tests/tests_system
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1511
 
### Check List


src/system/execute.py
```diff
@@ -6,11 +6,15 @@
 
 import logging
 import subprocess
+from os import PathLike
+from typing import Any, Tuple, Union
 
 
-def execute(command, dir, capture=True, raise_on_failure=True):
+def execute(command: str, dir: Union[str, PathLike[str]], capture: bool = True, raise_on_failure: bool = True) -> Tuple[int, Any, Any]:
```
peternied, line 13: Union types without different behavior based on the type don't make much sense to me.  Lets convert this to a `str` and patch up any usages

src/system/process.py
```diff
@@ -70,19 +71,19 @@ def terminate(self):
...
         return True if self.process else False
 
     @property
-    def pid(self):
+    def pid(self) -> int:
         return self.process.pid if self.started else None
 
     @property
-    def stdout_data(self):
+    def stdout_data(self) -> Union[Text, bytes]:
```
peternied, line 82: Since `Process` is our construct, we should return a consistent types by converting when needed.  This makes it easier for our usage of this class to be consistent.  I think `stdout_data(self) -> str:` is the easiest usage for our codebase.

Note; I think we should redesign the all the return types in this class to be based on our usage requirements.  Otherwise there are runtime behaviors become unpredictable.  If the intent is to keep the class type scheme as is we need a bunch of additional unit tests to cover these scenarios

peternied (changes requested):

src/system/process.py
abhinavGupta16: I have removed Union and updated the function to return `Any` here. `str` won't work because of how the function behaves. 

Error that pops up when returning `str` 
```bash
src/system/process.py:80: error: Returning Any from function declared to return "str"  [no-any-return]
```

peternied: This error is because on line 20-21 these types are declared as any, they should be a specific type

abhinavGupta16: I cannot have them as `str`. This is because `this.stdout` is of type `_TemporaryFileWrapper`(line 26 and 27). It is later used to call the function `read()` and `close()` (line 80). `_TemporaryFileWrapper` is a private class. So I marked `this.stdout` as Any.

peternied: By changing how the function works to prevent a leaky abstraction would work around this problem.

Note; While I think the design of this function could use some improvement, I also am not aware of problems related to how it works that should prevent this CR from merging, might be worthwhile to create an issue to resolve if/when it becomes more of a problem.

peternied (approved):

dblock (approved):

Issue #1565, opened 2022-01-27
[Bug]: BuildManifest throws nullpointerexception when no `artifacts`


kavilla:
### Describe the bug

When a component doesn't have any `artifacts` it will throw an exception when trying to instantiate a BuildManifest [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/jenkins/BuildManifest.groovy#L65).

This prevents assembling within Jenkins if passing a component without `artifacts` [here](https://github.com/opensearch-project/opensearch-build/blob/main/vars/assembleManifest.groovy#L3).

One example is the [functional test repo](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/1.2.0/opensearch-dashboards-1.2.0.yml#L16). This repo is a repo of Cypress tests and helper functions and not a plugin so it doesn't produce any actual artifacts until a testing stage is added.

### To reproduce

Run the Jenkinsfile for OpenSearch Dashboards for 1.2.0 [here](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/1.2.0/opensearch-dashboards-1.2.0.yml)

### Expected behavior

OpenSearch Dashboards distribution is produced and published.

### Host / Environment

Jenkins

PR #1568, opened 2022-01-27
[Dashboards] check BuildManifest has artifacts

kavilla:
### Description
Verifies if the build manifest has artifacts or not before using the value.
This allows for components that don't produce build artifacts to be passed
to the groovy class without failing.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1565

### Check List


kavilla:
Build 27 was with these changes and build 28 is where I temp removed this fix. So this fixes it the issue I was seeing.

Screen Shot 2022-01-27 at 1 53 18 PM

src/jenkins/BuildManifest.groovy
```diff
@@ -62,7 +62,7 @@ class BuildManifest implements Serializable {
             this.ref = data.ref
             this.commit_id = data.commit_id
             this.repository = data.repository
-            this.artifacts = new HashMap<>(data.artifacts)
+            this.artifacts = data.artifacts ? new HashMap<>(data.artifacts) : new HashMap<>()
```
peternied, line 65: Could we add a test, or alter one of the existing tests to exercise this functionality?

peternied (changes requested):

src/jenkins/BuildManifest.groovy
peternied: Note; I see that you altered the test data, but its not very clear how that alterative prevents the nullpointer, I was hoping you could create a new test that invokes `BuildManifest(...)` with artifacts undefined

peternied (approved):

dblock (approved):

PR #1571, opened 2022-01-28
Fix #1521: use a test build manifest.

### LINKED ISSUE #1521: [BUG] TestAssembleUpload in Jenkins tests is throwing manifest not found error
```
jenkins.tests.TestBuildAssembleUpload > testJenkinsfile FAILED
    java.lang.reflect.UndeclaredThrowableException
        at com.sun.proxy.$Proxy23.accept(Unknown Source)
        at java.base/java.util.LinkedHashMap$LinkedEntrySet.forEach(LinkedHashMap.java:671)
......
 at jenkins.tests.TestBuildAssembleUpload.testJenkinsfile(TestBuildAssembleUpload.groovy:52)

        Caused by:
        java.io.FileNotFoundException: builds/opensearch/manifest.yml (No such file or directory)
            at java.base/java.io.FileInputStream.open0(Native Method)
            at java.base/java.io.FileInputStream.open(FileInputStream.java:219)
            at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
            at groovy.util.CharsetToolkit.<init>(CharsetToolkit.java:78)
```

Seems like it caused multiple PR from merging.
Thanks.

dblock:

### Description

The test calls build and assemble and uses the output of build in assemble.  Thus depending on what you have in `builds/opensearch/manifest.yml` the test will succeed or fail. To reproduce #1521 `rm builds/opensearch/manifest.yml`.

The test used to work because another test output a `builds/opensearch/manifest.yml`, but the execution order of tests is not predictable. The fix is to copy a well known manifest.yml before running the test.

### Issues Resolved

Closes #1521.
 
### Check List

### DIFF
--- a/tests/jenkins/TestBuildAssembleUpload.groovy
+++ b/tests/jenkins/TestBuildAssembleUpload.groovy
@@ -44,11 +44,15 @@ class TestBuildAssembleUpload extends BuildPipelineTest {
     }
 
     @Test
-    @Ignore("https://github.com/opensearch-project/opensearch-build/issues/1521")
     public void testJenkinsfile() {
         helper.registerAllowedMethod("s3DoesObjectExist", [Map], { args ->
             return true
         })
+
+        Path sourceBuildManifest = Path.of("tests/data/opensearch-build-1.1.0.yml")
+        Path targetBuildManifest = Path.of("builds/opensearch/manifest.yml")
+        Files.createDirectories(targetBuildManifest.getParent())
+        Files.copy(sourceBuildManifest, targetBuildManifest, StandardCopyOption.REPLACE_EXISTING)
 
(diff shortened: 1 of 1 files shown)


peternied (approved):
Seems like the root cause of this was that each test isn't responsible for cleaning up after itself.  Could we 'clean' the environment between each test to make sure we don't author tests that could be broken like this?

dblock:
> Seems like the root cause of this was that each test isn't responsible for cleaning up after itself. Could we 'clean' the environment between each test to make sure we don't author tests that could be broken like this?

Probably a good idea to add a base cleanup function that deletes everything in non-git folders.

kavilla (approved):

PR #1574, opened 2022-01-31
Adding checksum for latest snapshots artifacts

### LINKED ISSUE #1497: [Bug]: checksum check failing for OpenSearch
### Describe the bug

OpenSearch Dashboards out-of-the-box experience downloads a snapshot of OpenSearch to start up and run some integration tests and functional tests. However, there is [logic](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/packages/osd-opensearch/src/artifact.js#L427) before starting OpenSearch artifact automatically downloaded to verify the checksum. As of last week (roughly 1/14), we started to notice this logic has been failing with the result being:

```
    artifact downloaded from https://artifacts.opensearch.org/snapshots/core/opensearch/2.0.0-SNAPSHOT/opensearch-min-2.0.0-SNAPSHOT-linux-x64-latest.tar.gz does not match expected checksum
      expected: 02479d4417c501750b0afc39f74bfb5969f086ea6f6a31943298bbbf1dc06be48de3e32051087e58cca8712e3de5066aa48ec1f803d88508ffad96b5246d2f03
      received: 6944c39d9d32d82c2c0604426ba8a98737e725108e9ca7559e90efbf44928e48634983f8696d34a49c53d2973c9e07a91a3ec7f5affe0410c46779b983941873
```
[[SOURCE](https://github.com/opensearch-project/OpenSearch-Dashboards/runs/4836313804?check_suite_focus=true)]

### To reproduce

This is specifically relying out of the out of box experience. If you have an OpenSearch artifact downloaded and pointing to it with `TEST_OPENSEARCH_FROM` it will avoid this logic. 

To reproduce:
* Ensure no running version of OpenSearch
* `unset TEST_OPENSEARCH_FROM`
* `git checkout main`
* `yarn osd bootstrap`
* `yarn test:jest_integration`

Notice failure in integration tests.

### Expected behavior

Integration tests work out of the box and thus the CI will succeed. Developers should be able to run `yarn test:jest_integration` without setting `TEST_OPENSEARCH_FROM`

### Screenshots

n/a

### Host / Environment

Local ubuntu & github ci

### Additional context

Example: https://github.com/opensearch-project/OpenSearch-Dashboards/runs/4836313804?check_suite_focus=true

### Relevant log output

```shell
artifact downloaded from https://artifacts.opensearch.org/snapshots/core/opensearch/2.0.0-SNAPSHOT/opensearch-min-2.0.0-SNAPSHOT-linux-x64-latest.tar.gz does not match expected checksum
      expected: 02479d4417c501750b0afc39f74bfb5969f086ea6f6a31943298bbbf1dc06be48de3e32051087e58cca8712e3de5066aa48ec1f803d88508ffad96b5246d2f03
      received: 6944c39d9d32d82c2c0604426ba8a98737e725108e9ca7559e90efbf44928e48634983f8696d34a49c53d2973c9e07a91a3ec7f5affe0410c46779b983941873
```

peterzhuamazon:

### Description
Adding checksum for latest snapshots artifacts
 
### Issues Resolved
#1497
 
### Check List


tests/jenkins/TestUploadMinSnapshotsToS3.groovy
```diff
@@ -8,31 +8,21 @@
...
+import jenkins.tests.BuildPipelineTest
 import org.junit.*
-import java.util.*
 
 class TestUploadMinSnapshotsToS3 extends BuildPipelineTest {
 
     @Before
     void setUp() {
+        List <Closure> fileActions = ['createSha512Checksums']
+        this.registerLibTester(new UploadMinSnapshotsToS3LibTester( fileActions, 'tests/jenkins/data/opensearch-1.3.0.yml' ))
```
peterzhuamazon, line 20: Since we are not able to import the lib from vars, I treat this one as a string for comparison @dblock

peterzhuamazon: If there is a way to import it into unit test, please let me know about it.

vars/uploadMinSnapshotsToS3.groovy
```diff
@@ -1,15 +1,42 @@
...
-    minArtifactSourceName = "opensearch-min-${version}-${platform}-${architecture}.tar.gz"
-    minArtifactName = "opensearch-min-${version}-${platform}-${architecture}-latest.tar.gz"
+    id = buildManifest.build.id
+
+    // Setup src & dst variables for artifacts
+    srcDir = "${WORKSPACE}/builds/${productName}/dist"
+    dstDir = "snapshots/core/${productName}/${version}"
+    baseName = "${productName}-min-${version}-${platform}-${architecture}"
+    tarSrcName = "${baseName}.tar.gz"
+    tarDstNameId = "${baseName}-${id}.tar.gz"
```
dblock, line 22: Nit: I find additional variables to be overhead that makes code harder to follow. The ones like `tarDstNameId` are only used in 1 place below, I would directly use the value there. No need to change anything unless you agree, just something to think about.

vars/uploadMinSnapshotsToS3.groovy
```diff
@@ -1,15 +1,42 @@
...
+    tarDstNameLatest = "${baseName}-latest.tar.gz"
+    shaSrcName = "${tarSrcName}.sha512"
+    shaDstNameId = "${tarDstNameId}.sha512"
+    shaDstNameLatest = "${tarDstNameLatest}.sha512"
+
+    // Create checksums
+    echo("Create .sha512 for Min Snapshots Artifacts")
+    argsMap = [:]
+    argsMap['artifactPath'] = srcDir
+    for (Closure action : fileActions) { // running createSha512Checksums()
```
dblock, line 32: Nit: Does it run `createSha512Checksums`? If so, then a) hard-code that value here b) remove it from `args` c) default `fileArgs` above to it. 

At least either remove the comment. or change the comment to something like (`e.g. createSha512Checksums()`).

tests/jenkins/TestUploadMinSnapshotsToS3.groovy
dblock: I don't understand the problem, what would you like to write here instead of what's in the code?

vars/uploadMinSnapshotsToS3.groovy
peterzhuamazon: It is written that way in case new actions such as createSignature() or more can be called as well.
The same logic has been used in promotion lib.
If you find it to be too complicated I can change it, but other than that I think it is ok.

tests/jenkins/TestUploadMinSnapshotsToS3.groovy
peterzhuamazon: The original code is:
```
         List <Closure> fileActions = [createSha512Checksums()] 
```
which does not run because it cannot figure out what is `createSha512Checksums()` as I cant seems to find a way to import it from `/vars/createSha512Checksums.groovy`.

vars/uploadMinSnapshotsToS3.groovy
peterzhuamazon: Will try to make these simpler, tbh I also get a bit confused myself on this.

peterzhuamazon: fixed.

kavilla (approved):

abhinavGupta16 (approved):
Thanks for going beyond the requirements for the task and adding the `libtester` for the exiting `UploadMinSnapshotsToS3LibTester.groovy` library.

tests/jenkins/TestUploadMinSnapshotsToS3.groovy
dblock: I see, you could pass another dummy function in here instead of `createSha512Checksums`?

dblock (approved):

tests/jenkins/TestUploadMinSnapshotsToS3.groovy
peterzhuamazon: I can, but that kinda defeat the purpose of testing this.
Will try in a different test tho.
Thanks.

peterzhuamazon: @abhinavGupta16 thoughts?

abhinavGupta16: I don't think a dummy function would do the job either since the passed function in the job is a closure call. The `class.name` fetches `createSha512Checksums$closurecall_1` and not `createSha512Checksums`

dblock: I tried declaring a closure but it doesn't get a name. I would just leave this as is.

Issue #1567, opened 2022-01-27
[BUG]Error when running Docker image


qmonitoring:
**Describe the bug**

When I try to run a docker image (following the [documentation](https://opensearch.org/docs/latest/opensearch/install/docker/#configure-opensearch)) I get this error:

`tee: /usr/share/opensearch/config/opensearch.yml: Permission denied`

The container won't start.

**To Reproduce**

` docker run -p 9200:9200 -p 9600:9600 -e "discovery.type=single-node" -v /usr/share/opensearch/config/custom-opensearch.yml:/usr/share/opensearch/config/opensearch.yml opensearchproject/opensearch:1.2.4`

**Expected behavior**

Opensearch up and running

**Host/Environment (please complete the following information):**

centos-release-8.2-2.2004.0.1.el8.x86_64

**Additional context**

I tried many options: change the configuration file, use a old version of the image, but the result was equally unsuccessful. For _opensearchproject/opensearch:1.2.3_ the error is:

`tee: /usr/share/opensearch/config/opensearch.yml: Permission denied`
`sed: cannot rename /usr/share/opensearch/config/sedZbSzcg: Device or resource busy`

peterzhuamazon:
Hi @qmonitoring please refer to #1529 for more context and discussion as well as temp fixes that I am preparing to resolve the race condition.

As of now if your yml file is root owned it cannot be updated as that requires a complete change in how we handle the startup variable. Or you can make your yml file owned by opensearch and it should run.

Sorry for inconvinience,
Thanks.

peterzhuamazon:
Close this issue please continue the discussion in #1529 as we have a fix now for you to test:
https://github.com/opensearch-project/opensearch-build/issues/1529#issuecomment-1027319799

Thanks.

PR #1579, opened 2022-02-01
Permanent fix for OpenSearch and temp fix for Dashboards startup configurations

### LINKED ISSUE #1529: [BUG] OpenSearch v1.2.4 Docker image is broken while v1.2.3 just works
This command works
Captura de 2022-01-21 17-47-07

But this one doesn't

Captura de 2022-01-21 17-48-23

Errors are shown below

Captura de 2022-01-21 17-49-08
Captura de 2022-01-21 17-49-28
Captura de 2022-01-21 17-49-47
Captura de 2022-01-21 17-50-06

peterzhuamazon:

### Description
Permanent fix for OpenSearch and temp fix for Dashboards startup configurations.

This PR utilize internal parsing of env var to permanently allow security plugin to be disabled, without the need of tweaking the config file on the fly for OpenSearch.

As for Dashboards, we still need to tweak the configuration file if we want to disable security plugin, so a new tweak is added to temporarily resolve the race condition introduced in #1130 #1458.

As for permanent fix for Dashboards I have opened an issue with Security Dashboards Plugin repo:
https://github.com/opensearch-project/security-dashboards-plugin/issues/896

Thanks.
 
### Issues Resolved
#1529 
#1567
https://github.com/opensearch-project/helm-charts/issues/198
 
### Check List

### DIFF
--- a/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
+++ b/docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
@@ -160,8 +160,12 @@ function setupSecurityDashboardsPlugin {
         if [ "$DISABLE_SECURITY_DASHBOARDS_PLUGIN" = "true" ]; then
             echo "Disabling OpenSearch Security Dashboards Plugin"
             ./bin/opensearch-dashboards-plugin remove securityDashboards
-            cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "/^opensearch_security/d" | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
-            cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "s/https/http/g" | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
+
+            # Remove all security related parameters as well as changing HTTPS to HTTP
+            # Temporary fix before security-dashboards plugin implement a parameter to disable the plugin entirely
+            # https://github.com/opensearch-project/security-dashboards-plugin/issues/896
+            NEW_CONFIG=`cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "/^opensearch_security/d" | sed "s/https/http/g"`
+            echo "$NEW_CONFIG" > $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
         fi
     fi
 }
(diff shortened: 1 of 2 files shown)


docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
```diff
@@ -160,8 +160,12 @@ function setupSecurityDashboardsPlugin {
         if [ "$DISABLE_SECURITY_DASHBOARDS_PLUGIN" = "true" ]; then
             echo "Disabling OpenSearch Security Dashboards Plugin"
             ./bin/opensearch-dashboards-plugin remove securityDashboards
-            cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "/^opensearch_security/d" | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
-            cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "s/https/http/g" | tee $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml
+
+            # Remove all security related parameters as well as changing HTTPS to HTTP
+            # Temporary fix before security-dashboards plugin implement a parameter to disable the plugin entirely
+            # https://github.com/opensearch-project/security-dashboards-plugin/issues/896
+            NEW_CONFIG=`cat $OPENSEARCH_DASHBOARDS_HOME/config/opensearch_dashboards.yml | sed "/^opensearch_security/d" | sed "s/https/http/g"`
```
peternied, line 167: Note, I am very paranoid about this change

How do we know that `var=cat | sed | sed` `echo $var > ` is any more / less reliable than what we did in the past?

dblock, line 167: UPDATED_CONFIG maybe?

dblock (approved):

docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
peterzhuamazon: It is more reliable because commands in a shell script are executed synchronously in sequence.

When you run something like this:
```
cmd1
cmd2
```

it is same as to run:
```
cmd1; cmd2
```

In previous iteration we are running cat/sed/tee in one single line/command, which cause the race condition to happen, as `tee` can set the file to empty before `cat` complete the reading of the file, since cat read in block size, thus cause the race condition to happen:
https://unix.stackexchange.com/questions/245499/how-does-cat-know-the-optimum-block-size-to-use

In new iteration, we allow cat and sed to complete processing of the entire file and save to the variable, before writing the content back to the file.

Since they are two different command, the 1st one will always complete, set the return code to 0, before the second command even start.

Thanks.

peterzhuamazon: Thanks, tweaked.

peternied (approved):

docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint.sh
peternied: Thanks for the detailed explaination!

PR #1581, opened 2022-02-01
 Rename internal filename of sha512 so checksum pass 

### LINKED ISSUE #1497: [Bug]: checksum check failing for OpenSearch

peterzhuamazon:
### Description
Rename internal filename of sha512 so checksum pass:
```
$ sha512sum opensearch-min-2.0.0-SNAPSHOT-linux-x64-latest.tar.gz
8dc5b8c1c899d8f3250a79f1fc56fc595fd03fe0b09dd5ca9036f969e26248cf321d151e6253e2378378818bc83dae1fe9adbbb90440c01e1b988e047b215d69  opensearch-min-2.0.0-SNAPSHOT-linux-x64-latest.tar.gz


$ cat opensearch-min-2.0.0-SNAPSHOT-linux-x64-latest.tar.gz.sha512
8dc5b8c1c899d8f3250a79f1fc56fc595fd03fe0b09dd5ca9036f969e26248cf321d151e6253e2378378818bc83dae1fe9adbbb90440c01e1b988e047b215d69  opensearch-min-2.0.0-SNAPSHOT-linux-x64.tar.gz
```

In #1574 we simply rename the checksum file without replacing the internal filename to the new one.
This PR we fix this issue by using sed to replace the names.
 
### Issues Resolved
#1497
 
### Check List


vars/uploadMinSnapshotsToS3.groovy
```diff
@@ -1,15 +1,41 @@
...
+    argsMap['artifactPath'] = srcDir
+    for (Closure action : fileActions) { // running createSha512Checksums()
+        action(argsMap)
+    }
+
     withAWS(role: "${ARTIFACT_PROMOTION_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_ARTIFACT}", duration: 900, roleSessionName: 'jenkins-session') {
-        s3Upload(file: "$WORKSPACE/builds/opensearch/dist/${minArtifactSourceName}", bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "snapshots/core/opensearch/${version}/${minArtifactName}")
+        sh """
+            cp ${srcDir}/${baseName}.tar.gz.sha512 ${srcDir}/${baseName}-${id}.tar.gz.sha512 ${srcDir}/${baseName}-latest.tar.gz.sha512
+            sed -i "s/.tar.gz.sha512/-${id}.tar.gz.sha512/g" ${srcDir}/${baseName}-${id}.tar.gz.sha512
```
gaiksaya, line 33: We do not need id in snapshots name anywhere

gaiksaya: Sorry I was thinking about release-candidates and not snapshots. But even for snapshots do we need build number included in the name? I don't think so.
cc: @dblock

kavilla (approved):

vars/uploadMinSnapshotsToS3.groovy
peterzhuamazon: I think we still need it @peternied can you help confirm?
Adding id copy would help us know whether we have successfully uploaded or not, even retain an older copy of it.
Thanks.

gaiksaya: It defeats the purpose of having snapshots doesn't it? If it is not successful to upload the job will fail and we would know then. Why keep a copy when you are already uploading the `-latest`? We have a copy in staging bucket.

peternied: I don't think its useful to keep around the 'id' artifacts, I would prefer that we consult the log for the build job if we need to confirm behavior

peterzhuamazon: Updated now.

peterzhuamazon:
@gaiksaya @peternied please take a look for the changes with id removed.

Thanks.

vars/uploadMinSnapshotsToS3.groovy
```diff
@@ -28,9 +28,12 @@ void call(Map args = [:]) {
...
 
     withAWS(role: "${ARTIFACT_PROMOTION_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_ARTIFACT}", duration: 900, roleSessionName: 'jenkins-session') {
-        s3Upload(file: "${srcDir}/${baseName}.tar.gz", bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "${dstDir}/${baseName}-${id}.tar.gz")
-        s3Upload(file: "${srcDir}/${baseName}.tar.gz", bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "${dstDir}/${baseName}-latest.tar.gz")
-        s3Upload(file: "${srcDir}/${baseName}.tar.gz.sha512", bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "${dstDir}/${baseName}-${id}.tar.gz.sha512")
-        s3Upload(file: "${srcDir}/${baseName}.tar.gz.sha512", bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "${dstDir}/${baseName}-latest.tar.gz.sha512")
+        sh """
+            cp ${srcDir}/${baseName}.tar.gz ${srcDir}/${baseName}-latest.tar.gz
+            cp ${srcDir}/${baseName}.tar.gz.sha512 ${srcDir}/${baseName}-latest.tar.gz.sha512
+            sed -i "s/.tar.gz/-latest.tar.gz/g" ${srcDir}/${baseName}-latest.tar.gz.sha512
```
gaiksaya, line 34: What is the difference between `cp` and `sed` commands here?

vars/uploadMinSnapshotsToS3.groovy
```diff
@@ -28,9 +28,12 @@ void call(Map args = [:]) {
     }
 
     withAWS(role: "${ARTIFACT_PROMOTION_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_ARTIFACT}", duration: 900, roleSessionName: 'jenkins-session') {
```
abhinavGupta16, line 30: Instead of doing this, we can use the upload library that was recently created

vars/uploadMinSnapshotsToS3.groovy
```diff
@@ -28,9 +28,12 @@ void call(Map args = [:]) {
...
-        s3Upload(file: "${srcDir}/${baseName}.tar.gz", bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "${dstDir}/${baseName}-${id}.tar.gz")
-        s3Upload(file: "${srcDir}/${baseName}.tar.gz", bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "${dstDir}/${baseName}-latest.tar.gz")
-        s3Upload(file: "${srcDir}/${baseName}.tar.gz.sha512", bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "${dstDir}/${baseName}-${id}.tar.gz.sha512")
-        s3Upload(file: "${srcDir}/${baseName}.tar.gz.sha512", bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}", path: "${dstDir}/${baseName}-latest.tar.gz.sha512")
+        sh """
```
abhinavGupta16, line 31: Do we need assume role for these `sh` commands? Maybe we can move them outside of this.

abhinavGupta16 (changes requested):

vars/uploadMinSnapshotsToS3.groovy
peterzhuamazon: cp is to make a new copy of the sha512 with `-latest` in file name.
sed is to replace the internal content and add `-latest`.

vars/uploadMinSnapshotsToS3.groovy
peterzhuamazon: This is a bug fix PR not a feature enhancement PR.
I disagree this is the reason to block this PR.
We can have a separate PR later if we want to improve the process.

vars/uploadMinSnapshotsToS3.groovy
peterzhuamazon: moved. thanks.

vars/uploadMinSnapshotsToS3.groovy
peterzhuamazon: Also that lib is only having one mode which is using file to file.
It does not support wildcard push as of yet.
So not really suitable for us to use it now.

peterzhuamazon: Thanks.

abhinavGupta16 (approved):
LGTM

dblock:
If we did a 302 redirect from `-latest.tar.gz` to `-1234.tar.gz` we wouldn't have this issue, consider this as a followup.

PR #1584, opened 2022-02-02
Adding 1.2.5 manifest

abhinavGupta16:
### Description
Adding manifest files for `1.2.5`
  
### Check List


jenkins/check-for-build.jenkinsfile
```diff
@@ -8,6 +8,7 @@ pipeline {
     triggers {
         parameterizedCron '''
             H/10 * * * * %INPUT_MANIFEST=1.3.0/opensearch-dashboards-1.3.0.yml;TARGET_JOB_NAME=distribution-build-opensearch-dashboards
+            H/10 * * * * %INPUT_MANIFEST=1.2.5/opensearch-1.2.5.yml;TEST_MANIFEST=1.2.5/opensearch-1.2.5-test.yml;TARGET_JOB_NAME=distribution-build-opensearch
             H/10 * * * * %INPUT_MANIFEST=1.2.4/opensearch-1.2.4.yml;TEST_MANIFEST=1.2.4/opensearch-1.2.4-test.yml;TARGET_JOB_NAME=distribution-build-opensearch
```
gaiksaya, line 12: We should remove 1.2.4 from this cron. The version is already out we do not need to keep rebuilding it

manifests/1.2.5/opensearch-1.2.5.yml
```diff
@@ -0,0 +1,99 @@
...
+  name: OpenSearch
+  version: 1.2.5
+components:
+  - name: OpenSearch
+    repository: https://github.com/opensearch-project/OpenSearch.git
+    ref: 1.x
+    checks:
+      - gradle:publish
+      - gradle:properties:version
+  - name: common-utils
```
gaiksaya, line 17: I do not believe we add all components at first. Most of the components build might fail due to version incompatibility if the version is main is not incremented.

peternied (commented):
We shouldn't merge this, after OpenSearch creates a 1.2.5 branch this manifest will be created automatically via this repos github actions

abhinavGupta16:
> We shouldn't merge this, after OpenSearch creates a 1.2.5 branch this manifest will be created automatically via this repos github actions

I'll make this PR a draft

dblock (changes requested):
This won't work. In order to produce a 1.2.5 successful build there needs to be version increments, then a -SNAPSHOT build with common-utils, then another with job-scheduler, etc. This is an incremental process that is semi-automated, but it runs reliable CI and ensures that we have a working 1.2.5 without ever being broken.

I would just close this without merging.

abhinavGupta16:
Closing The PR since this process is automated

PR #1585, opened 2022-02-02
Add TEST_MANIFEST param to check-for-build jenkins job

ohltyler:
Signed-off-by: Tyler Ohlsen <ohltyler@amazon.com>

### Description
Currently distribution builds are broken due to a missing `TEST_MANIFEST` parameter in the check-for-build jenkins job, which was introduced in the `distribution-build-opensearch` job as part of #1517. This PR adds that.

Note that only `distribution-build-opensearch` is using a test manifest, so there's separate logic for the `distribution-build-opensearch-dashboards` job to exclude that parameter for know. In the future these may get merged back, once Dashboards supports integ tests and using a test manifest.
 
### Check List

### DIFF
--- a/jenkins/check-for-build.jenkinsfile
+++ b/jenkins/check-for-build.jenkinsfile
@@ -26,6 +26,11 @@ pipeline {
             trim: true
         )
         string(
+            name: 'TEST_MANIFEST',
+            description: 'Test manifest under the manifests folder, e.g. 2.0.0/opensearch-2.0.0-test.yml. Currently only applicable for distribution-build-opensearch job',
+            trim: true
+        )
+        string(
             name: 'TARGET_JOB_NAME',
             description: 'Job to trigger if build has changed',
             trim: true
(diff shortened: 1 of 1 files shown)


peternied (approved):

gaiksaya (approved):
Wondering if we can pass null or none instead of if else block. won't matter as param will not be used anywhere and will be consistent

ohltyler:
@gaiksaya we can, but then it will show up in Jenkins UI with an extra `TEST_MANIFEST` parameter in the build info, and may add confusion since it's not being used. Prefer to leave as an if/else for now and merge together again once Dashboards supports a test manifest similar to core OpenSearch.

jenkins/check-for-build.jenkinsfile
```diff
@@ -66,9 +71,17 @@ pipeline {
...
+                                if (TARGET_JOB_NAME == 'distribution-build-opensearch') {
+                                    build job: "${TARGET_JOB_NAME}", parameters: [
+                                    string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}"),
+                                    string(name: 'TEST_MANIFEST', value: "${TEST_MANIFEST}")
+                                    ], wait: true
+                                } else if (TARGET_JOB_NAME == 'distribution-build-opensearch-dashboards') {
+                                    build job: "${TARGET_JOB_NAME}", parameters: [
+                                    string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}")
+                                    ], wait: true
+                                }
```
dblock, line 84: This is a hack, but at least it should have `else error`, otherwise if this is ever passed to a job that's not distribution-build-... there's no failure and nothing is kicked off.

Then, does passing TEST_MANIFEST into `distribution-build-opensearch-dashboards` that doesn't expect it actually fails?

dblock: It's NBD, but maybe a comment would suffice.

PR #1590, opened 2022-02-03
Add invalid job case & improve error handling in check-for-build

ohltyler:
Signed-off-by: Tyler Ohlsen <ohltyler@amazon.com>

### Description
Adds a check for handling invalid jobs passed. Properly fails the job if there is any failure seen in the build or uploading steps. Follow-up to #1585 

Sanity test with job name `distribution-build-invalid` (Jenkins job `Playground/ohltyler-check-for-build-2` `#3`):
```
ERROR: Job invalid-job is invalid
```
 
### Check List

### DIFF
--- a/jenkins/check-for-build.jenkinsfile
+++ b/jenkins/check-for-build.jenkinsfile
@@ -70,30 +70,20 @@ pipeline {
                         if (sha.exists) {
                             echo "Skipping, ${sha.path} already exists."
                         } else {
-                            try {
-                                // TEST_MANIFEST param isn't applicable for distribution-build-opensearch-dashboards job yet,
-                                // so we don't pass it to that job to prevent confusion from Jenkins UI
-                                if (TARGET_JOB_NAME == 'distribution-build-opensearch') {
-                                    build job: "${TARGET_JOB_NAME}", parameters: [
-                                    string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}"),
-                                    string(name: 'TEST_MANIFEST', value: "${TEST_MANIFEST}")
-                                    ], wait: true
-                                } else if (TARGET_JOB_NAME == 'distribution-build-opensearch-dashboards') {
-                                    build job: "${TARGET_JOB_NAME}", parameters: [
-                                    string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}")
-                                    ], wait: true
-                                } else {
-                                    throw new Exception("Job ${TARGET_JOB_NAME} is not a valid option")
-                                }
-                                echo "Build succeeded, uploading build SHA for that job"
-                                buildUploadManifestSHA(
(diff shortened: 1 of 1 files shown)


dblock (approved):

jenkins/check-for-build.jenkinsfile
```diff
@@ -81,6 +82,8 @@ pipeline {
                                     build job: "${TARGET_JOB_NAME}", parameters: [
                                     string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}")
                                     ], wait: true
+                                } else {
+                                    throw new Exception("Job ${TARGET_JOB_NAME} is not a valid option")
```
peternied, line 86: Use `error` [doc](https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#error-error-signal) to make sure the build is marked a failure.  I am not a fan of exceptions as standard flow of control mechanism.

jenkins/check-for-build.jenkinsfile
```diff
@@ -89,6 +92,7 @@ pipeline {
                                 )
                             } catch (err) {
```
peternied, line 93: Can we remove this catch block, its just swallowing errors that should be failing the build

jenkins/check-for-build.jenkinsfile
```diff
@@ -71,7 +71,8 @@ pipeline {
                             echo "Skipping, ${sha.path} already exists."
                         } else {
                             try {
-                                // Currently only opensearch build takes an additional test manifest parameter
+                                // TEST_MANIFEST param isn't applicable for distribution-build-opensearch-dashboards job yet,
+                                // so we don't pass it to that job to prevent confusion from Jenkins UI
```
peternied, line 75: Could we just add the param (and not use it) to that job and avoid the error entirely?

peternied (changes requested):

gaiksaya (approved):

jenkins/check-for-build.jenkinsfile
zelinh: +1 here. I think the exceptions threw above will be caught here and won't fail the job.

jenkins/check-for-build.jenkinsfile
ohltyler: Fixed in latest commit by checking the job name matches the 2 currently-supported jobs (`distribution-build-opensearch`, `distribution-build-opensearch-dashboards`

jenkins/check-for-build.jenkinsfile
ohltyler: Good catch - I was adding on to existing code but see this would silently fail. Removed in latest commit.

jenkins/check-for-build.jenkinsfile
ohltyler: Removed if/else logic, and passing the same params regardless of job.

peternied (approved):

PR #1591, opened 2022-02-03
Change the issue title for release to be 'proposed'

peternied:
The current langauge makes the autocut release tickets seem like a
commitment for a release, whereas they are a best guess effort
that a release could happen.  E.g. there might ne a patch release in
flight and then don't release that in favor of a minor release that
also includes the contents as the patch release.

### Check List

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_template.md
+++ b/.github/ISSUE_TEMPLATE/release_template.md
@@ -1,6 +1,6 @@
 ---
 name: Release
-title: "[RELEASE] Release version {{ env.VERSION }}"
+title: "[RELEASE] Release proposal version {{ env.VERSION }}"
 labels: untriaged, release, v{{ env.VERSION }}
 ---
 


dblock (changes requested):
Every open issue is a proposal or a bug or something like that. Not sure who feels that just because we have a ticket open that says "Release v1.2.3" that it's some kind of promise or commitment. Can you please point me to a discussion about that?

I am afraid this will re-auto-cut all old version tickets, see https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/releases.yml#L32, so this implementation doesn't work.

peternied:
> Can you please point me to a comment where someone is confused about that?

@bbarani can you provide background on this?

> I am afraid this will re-auto-cut all old version tickets

Yes, if updating the title is the right path I could spend more time looking into how we can prevent recuts.  Making an update to the issue creation GHA could be a way.

dblock:
> Yes, if updating the title is the right path I could spend more time looking into how we can prevent recuts. Making an update to the issue creation GHA could be a way.

FYI I had forked the original GHA given its lack of maintenance, https://github.com/dblock/create-a-github-issue

peternied:
@bbarani I haven't seen any updates, feel free to reopen this PR or one like it if you feel this change is needed

PR #1597, opened 2022-02-04
Align sign off with DCO expectation

peternied:
```
Error: The SoB (DCO) check failed

  aa4cd71b[5](https://github.com/opensearch-project/opensearch-build/runs/5060869826?check_suite_focus=true#step:3:5)90c[6](https://github.com/opensearch-project/opensearch-build/runs/5060869826?check_suite_focus=true#step:3:6)327bb3a9e5f37ae8ee089cc13da    Expected "dblock <dblock@users.noreply.github.com>", but got "opensearch-ci-bot <opensearch-infra@amazon.com>".
```

Updating the authored commit to include @dblock's username/email as is expected by the tool

### Check List


dblock:
I think different times the bot uses different usernames to make the PR. I am not sure how it chooses. We should be specifying this in the template and including the corresponding DCO check.

dblock:
I force pushed https://github.com/opensearch-project/opensearch-build/pull/1595, I think we should close this.

dblock (changes requested):
I think the fix here is to specify the author of the commit via `author` as well. See https://github.com/peter-evans/create-pull-request for docs.

peternied:
I've added the option to trigger the workflow manually, testing with this enabled,  If this works I'll update the documentation too so we can just trigger on demand if we don't want to wait

dblock (approved):

peternied:
Workflow was successful, pull request was created, https://github.com/peternied/opensearch-build/pull/11.

Waiting on the DCO check to make sure it passes, https://github.com/peternied/opensearch-build/runs/5070310880?check_suite_focus=true.

dblock (approved):

gaiksaya (approved):

Issue #438, opened 2021-09-10
Add release notes collection to build workflow assembly


dblock:
**Is your feature request related to a problem? Please describe.**

We're collecting and generating release notes manually. The bundle workflow can do it for us.

https://github.com/opensearch-project/opensearch-plugins/issues/120 for individual plugin release notes.

Acceptance criteria:
- [ ] The release notes tools should be able to automatically collate the release notes across multiple repositories to create a master release notes corresponding to a specific distribution
- [ ] The release notes tool should also include link checker, spelling checker and any other validation that can be run as part of a pull request workflow.

peternied:
[Triage] this can be automated we have a script already

peterzhuamazon:
https://github.com/opendistro-for-elasticsearch/opendistro-build/blob/main/release-tools/scripts/releasenotes.sh

peterzhuamazon:
[Grooming] Each Plugin Team should have their release notes ready as soon as their code is ready for release.

peterzhuamazon:
Make a proposal on what we can do to force plugin owners push a release note.

bbarani:
https://github.com/opensearch-project/OpenSearch/issues/1868

bbarani:
I am closing this issue as its actively tracked in https://github.com/opensearch-project/OpenSearch/issues/1868

gaiksaya:
Created another one for consolidated release notes https://github.com/opensearch-project/opensearch-build/issues/4009

PR #1598, opened 2022-02-04
1.0.0 changelog: mention no breaking changes

aviau:
I know this is a past release but some of us are still migrating from ES and a big BREAKING CHANGES with title 10 thousand things under it is scary until you realize its an empty section. 🙃

### DIFF
--- a/release-notes/opensearch-release-notes-1.0.0.md
+++ b/release-notes/opensearch-release-notes-1.0.0.md
@@ -12,9 +12,6 @@ OpenSearch [Release Notes](https://github.com/opensearch-project/OpenSearch/blob
 
 OpenSearch Dashboards [Release Notes](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/release-notes/opensearch-dashboards.release-notes-1.0.0.md).
 
-
-## BREAKING CHANGES
-
 ## FEATURES
 
 ### Opensearch Index Management


peterzhuamazon:
Hi @aviau sorry that this is a miss from our side.
Can you consider completely remove the `Breaking Changes` section?
This is also our standard practice since ODFE.

Thanks.

aviau:
@peterzhuamazon done!

zelinh (approved):
Thanks!

peterzhuamazon:
@aviau you need to sign off DCO.
Please do `git commit --amend --no-edit -s -S` and push again.
Thanks.

peternied (changes requested):
I don't think that we should remove this section, its important to call out that we did not introduce any breaking changes.

Could you help me understand how when you saw the release notes it looked like there were many breaking changes, maybe a screenshot?

aviau:
I won't put any more efforts here, I think my issue description is self-explainatory. Empty sections aren't obvious in rendered markdown. If you don't agree then please just close my PR.

dblock:
I think it should either say None or be removed. Don’t care either way. 

@aviau thank you!

@peternied ok merging as is or do you want to PR an alternative?

peternied:
@aviau Thanks for the contribution - lets merge this as is and we can make additional PRs in the future.

There is an issue https://github.com/opensearch-project/opensearch-build/issues/438 tracking the codification of the release notes process.

peternied (approved):

PR #1599, opened 2022-02-04
Add python typechecking for mypy on src/manifests_workflow and tests/tests_manifests_workflow

### LINKED ISSUE #1509: Manifests workflow `pipenv run mypy src\manifests_workflow tests\tests_manifests_workflow`

zelinh:
### Description
` pipenv run mypy src/manifests_workflow tests/tests_manifests_workflow` passes without error

### Issues Resolved
#1509 
 
### Check List

### DIFF
--- a/src/manifests_workflow/component_opensearch.py
+++ b/src/manifests_workflow/component_opensearch.py
@@ -6,6 +6,7 @@
 
 import logging
 import subprocess
+from typing import Any
 
 from git.git_repository import GitRepository
 from manifests_workflow.component import Component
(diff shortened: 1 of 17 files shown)


dblock (approved):

src/manifests_workflow/component_opensearch.py
```diff
@@ -16,13 +17,13 @@ class ComponentOpenSearch(Component):
...
-        snapshot=False,
-        working_directory=None,
-    ):
+        name: str,
+        path: str,
+        opensearch_version: str,
+        branch: str = "main",
+        snapshot: bool = False,
+        working_directory: str = None,
+    ) -> 'ComponentOpenSearch':
```
gaiksaya, line 26: Should this be `ComponentOpenSearch` instead of `'ComponentOpenSearch'`?

zelinh: Good point! Actually no, it will say unresolved reference since I think it must be defined before we could use it in an annotation. The solution could be quote it as a string. 
Ref: https://www.python.org/dev/peps/pep-0484/#forward-references

gaiksaya: I see, Thanks for explaining.

gaiksaya (approved):

Issue #1199, opened 2021-11-30
Create an account at crates.io to publish artifacts of Rust Client


adityaj1107:
We need crates.io account to publish the artifacts of the opensearch rust client for community consumption. Please help create this account.

aditjind:
Repo for which artifacts need to be published: [opensearch-rs](https://github.com/opensearch-project/opensearch-rs)

gaiksaya:
[Offline Discussion] @aditjind is figuring out how to sign and publish the artifacts on crates.io
https://doc.rust-lang.org/cargo/reference/publishing.html
On glancing at the above document looks like the process can be automated. We can use `opensearch-ci-bot` as the GH account required to publish to crates. Need to check with the permissions and stuff.
However, since we are running short on time @aditjind requested to publish the initial version of client using manual documented process. He will work parallely on automating this for future releases.
The automation would include adding a jenkins workflow for opensearch-rs that will have right set of credentials to publish.
cc: @bbarani

VijayanB:
@gaiksaya In the meantime, shall we create an opensearch-project account and use that to publish either manual or automation.

gaiksaya:
> @gaiksaya In the meantime, shall we create an opensearch-project account and use that to publish either manual or automation.

I do not see an option to create a new account. All see is `Log in with GitHub`. Am I missing something here?

bbarani:
@aditjind @VijayanB Can we directly link opensearch-ci-bot GitHub account with crate.io and publish the artifacts?

aditjind:
As per my understanding, we will use the opensearch-ci-bot GitHub account to publish the first release of the client manually. We will then have automation scripts to build, sign and publish the artifacts via Infra Jenkins setup. 
cc: @gaiksaya (Correct me if I missed anything).

aditjind:
No staging area support for Crates:

* [Issue](https://github.com/rust-lang/crates.io/issues/1503) 
* [Discussion](https://github.com/rust-lang/crates.io/discussions/4330)

PR #1605, opened 2022-02-09
added support for .crate files in signer for crate.io

abhinavGupta16:

### Description
Added support to sign `.crate` files with `pgp` as a part of https://github.com/opensearch-project/opensearch-build/issues/1199
 
### Check List

### DIFF
--- a/src/sign_workflow/signer.py
+++ b/src/sign_workflow/signer.py
@@ -18,7 +18,7 @@ The signed artifacts will be found in the same location as the original artifact
 
 
 class Signer:
-    ACCEPTED_FILE_TYPES = [".zip", ".jar", ".war", ".pom", ".module", ".tar.gz", ".whl"]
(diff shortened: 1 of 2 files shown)


aditjind (approved):

peternied (approved):
Approving to unblock, but it asks the question, why do we only sign files on this list?  Seems like a pain every time we encounter a new one

zelinh (commented):
Does our signing client support signing `.crate` file?

abhinavGupta16:
> Does our signing client support signing `.crate` file?

It does now

gaiksaya:
> Approving to unblock, but it asks the question, why do we only sign files on this list? Seems like a pain every time we encounter a new one

I think this was in place when signing maven artifacts was implemented and while looping through the directory, we wanted to avoid signing `.sig` or `.asc` files which were also stored in same directory.

zelinh (approved):
Thanks.

Issue #1543, opened 2022-01-25
[RPM M1] The build workflow should generate the contents of the RPM packages as pre-requisite


peterzhuamazon:
Tasks | Estimate | Status | Notes | Dependencies
-- | -- | -- | -- | --
Re-use the existing build process to generate the OpenSearch/Dashboards min + all of the plugins artifacts for RPM package to use | 2 | Completed |   |  
The artifacts should be built with LINUX platform specified, as we only plan to support RPM package on LINUX | 1 | No Plan |   |  20220211: See next section of adding `--distribution` as that one take care of this one.
Tweak the build workflow to have a “—distribution” parameter so the internal API call can show “rpm” text, instead of the default “tar” text, as well as removing tar related files | 2 | OS ✅  <br><br>OSD ✅     | This probably need changes in the opensearch core so need core team support | opensearch-project/OpenSearch changes: https://github.com/opensearch-project/OpenSearch/issues/2043 <br><br>~~20220211: https://github.com/opensearch-project/opensearch-build/pull/1619~~ <br><br>20220215: https://github.com/opensearch-project/opensearch-build/pull/1629 ✅ <br><br>20220217: ~~Dashboards does not build rpm at all even with correct settings~~: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1259 (20220218: it builds x64 rpm but not arm64 rpm) <br><br>20220301: @kavilla PR to add arm64 deb/rpm support in dashboards https://github.com/opensearch-project/OpenSearch-Dashboards/pull/1285 <br><br>20220217_2: Only send -d to core components: https://github.com/opensearch-project/opensearch-build/issues/1651 https://github.com/opensearch-project/opensearch-build/pull/1652 ✅ <br><br>20220218: dashboards build.sh https://github.com/opensearch-project/opensearch-build/pull/1656 ✅ <br><br>20220302: @kavilla add the functionality for dashboards x64/arm64 rpm will update the build scripts to consume those https://github.com/opensearch-project/opensearch-build/pull/1699 ✅

peterzhuamazon:
Add new issue for core team to add this option for tweak build_type for rpm: https://github.com/opensearch-project/OpenSearch/issues/2043

peterzhuamazon:
20220217: Add a new issue to track problems with dashboards on building rpm:
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1259

peterzhuamazon:
[...]

PR #1619, opened 2022-02-12
Update build workflow to build based on distribution (OpenSearch)

peterzhuamazon:

### Description
Update build workflow to build based on distribution (OpenSearch).
You can now use `--distribution` to build tar/zip/rpm for OpenSearch-min.
Not completely bound to `platform` anymore.

This would also setup the stage for bundled RPM creation in assemble_workflow (#2073)
 
### Issues Resolved
Part of #1543
 
### Check List


dblock:
This is going to require updating all `build.sh`'s in all the active branches across all components to support `-d`. I think it's a lot of work, maybe only pass `-d` if specified, and default it to `None`?

peterzhuamazon:
Hi @dblock I have implemented the changes you suggest, as well as tweaking the scripts to take None as default.
The problem right now is the build_recorder does not record `distribution: something` into the manifest.yml after builds.

Could you take a look at the code again?
Thanks.

scripts/components/OpenSearch/build.sh
```diff
@@ -77,19 +81,54 @@ cp -r ./build/local-test-repo/org/opensearch "${OUTPUT}"/maven/org
 
 [ -z "$PLATFORM" ] && PLATFORM=$(uname -s | awk '{print tolower($0)}')
 [ -z "$ARCHITECTURE" ] && ARCHITECTURE=`uname -m`
+[ -z "$DISTRIBUTION" ] && DISTRIBUTION="None"
```
dblock, line 84: Default it to `tar` here, not another keyword `None`.

scripts/components/OpenSearch/build.sh
```diff
@@ -77,19 +81,54 @@ cp -r ./build/local-test-repo/org/opensearch "${OUTPUT}"/maven/org
...
 
 case $PLATFORM in
     linux*)
-        PACKAGE="tar"
-        EXT="tar.gz"
+        case $DISTRIBUTION in
+            tar|None)
```
dblock, line 89: Remove `None`. It's in the same script.

scripts/components/OpenSearch/build.sh
```diff
@@ -77,19 +81,54 @@ cp -r ./build/local-test-repo/org/opensearch "${OUTPUT}"/maven/org
...
+                exit 1
+                ;;
+        esac
         ;;
+
     darwin*)
-        PACKAGE="tar"
-        EXT="tar.gz"
+        case $DISTRIBUTION in
+            tar|None)
```
dblock, line 108: Remove `None`.

scripts/components/OpenSearch/build.sh
```diff
@@ -77,19 +81,54 @@ cp -r ./build/local-test-repo/org/opensearch "${OUTPUT}"/maven/org
...
+                exit 1
+                ;;
+        esac
         ;;
+
     windows*)
-        PACKAGE="zip"
-        EXT="zip"
+        case $DISTRIBUTION in
+            zip|None)
```
dblock, line 122: Only `zip`.

scripts/components/OpenSearch/build.sh
```diff
@@ -77,19 +81,54 @@ cp -r ./build/local-test-repo/org/opensearch "${OUTPUT}"/maven/org
 
 [ -z "$PLATFORM" ] && PLATFORM=$(uname -s | awk '{print tolower($0)}')
 [ -z "$ARCHITECTURE" ] && ARCHITECTURE=`uname -m`
+[ -z "$DISTRIBUTION" ] && DISTRIBUTION="None"
 
 case $PLATFORM in
```
dblock, line 86: I would combine and do a `case $PLATFORM-$DISTRIBUTION`, it would make the whole thing not nested and much simpler.

scripts/components/OpenSearch/build.sh
```diff
@@ -99,26 +138,43 @@ esac
 
 case $ARCHITECTURE in
```
dblock, line 139: Same as above, `case $ARCHITECTURE-$DISTRIBUTION`

src/build_workflow/build_recorder.py
```diff
@@ -60,6 +60,7 @@ def __init__(self, target: BuildTarget):
             self.data["build"]["version"] = target.opensearch_version
             self.data["build"]["platform"] = target.platform
             self.data["build"]["architecture"] = target.architecture
+            self.data["build"]["distribution"] = target.distribution if target.distribution else "None"
```
dblock, line 63: Leave this as `target.distribution`, it should always have a value here coming from the caller.

src/build_workflow/builder_from_source.py
```diff
@@ -29,22 +29,27 @@ def checkout(self, work_dir: str) -> None:
...
+                    self.target.platform,
+                    "-a",
+                    self.target.architecture,
+                    distribution_parameter,
+                    "-s",
+                    str(self.target.snapshot).lower(),
+                    "-o",
+                    self.output_path,
+                ]
+            )
```
dblock, line 52: You can rewrite this a bit cleaner:

```python
filter(None,
 "bash",
 build_script,
 f"-v {self.target.version}",
 ...
f"-d {self.target.distribution}" if self.target.distribution else None,
...
]

File: tests/data/opensearch-build-1.1.0.yml [hunk 9]
```diff
@@ -2,6 +2,7 @@
 build:
   platform: linux
   architecture: x64
+  distribution: None
```
dblock, line 5: Version 1.1.0 never had a distribution field, remove.

tests/data/opensearch-dashboards-build-1.2.0.yml
```diff
@@ -1,6 +1,7 @@
 ---
 build:
   architecture: x64
+  distribution: None
```
dblock, line 4: Same as for 1.1.0, remove here and everywhere, except for 1.3.0 build output that should now write a value of distribution.

tests/tests_test_workflow/test_integ_workflow/integ_test/data/build_manifest_missing_components.yml
```diff
@@ -2,6 +2,7 @@
 build:
   platform: linux
   architecture: x64
+  distribution: None
```
dblock, line 5: Add a 1.3.0 manifest that has a value for distribution. Possibly a new test.

`distribution: None` is not a thing, I don't expect to ever find this in a build file, when we don't pass distribution it defaults to something (tar?) doesn't it?

dblock (changes requested):
It's on the right track, but should be simpler and have fewer changes if you consider that you can never have `distribution: None`.

scripts/components/OpenSearch/build.sh
peterzhuamazon: I am so surprised that I actually didnt even remotely think about that.
Time to reflect on my approach.
Thanks.

src/build_workflow/build_recorder.py
peterzhuamazon: @dblock but if I leave as is, if no one call `--distribution` then it will default to `None`, which then will error out during recording.

tests/tests_test_workflow/test_integ_workflow/integ_test/data/build_manifest_missing_components.yml
peterzhuamazon: I was following the suggestion to have `None` as default, while previously my commit I use `tar` as default.
I am bit confused here what exactly should I use?
Thanks.

tests/data/opensearch-dashboards-build-1.2.0.yml
peterzhuamazon: So to be clear this is to update a new test for 1.3.0 the release, not the schema?

src/build_workflow/builder_from_source.py
peterzhuamazon: Fixed. Thanks.

scripts/components/OpenSearch/build.sh
peterzhuamazon: I raised it to `$PLATFORM-$DISTRIBUTION-$ARCHITECTURE` and it is now even simpler compares to original.

scripts/components/OpenSearch/build.sh
peterzhuamazon: fixed.

scripts/components/OpenSearch/build.sh
peterzhuamazon: Removed.

scripts/components/OpenSearch/build.sh
peterzhuamazon: Removed.

scripts/components/OpenSearch/build.sh
peterzhuamazon: Fixed. Thanks.

src/build_workflow/build_recorder.py
peterzhuamazon:
```
ValueError: Invalid manifest schema: {'build': [{'distribution': ['null value not allowed']}]}
```

peterzhuamazon: Set to `tar` if `None` for now.

peterzhuamazon:
I somehow messed up this branch and could not recover during a rebase, will open a new PR soon.

peterzhuamazon:
Reopen PR: https://github.com/opensearch-project/opensearch-build/pull/1629

PR #1628, opened 2022-02-15
Disable propagate argument when triggering child job

### LINKED ISSUE #1627: check-for-build job always fails if child job fails
This issue has been solved in previous #1285. However, in recent change, catch block was removed from this job. Therefore this job will always fail and send out notification if its child job (e.g. distribution-build-opensearch-dashboards) fails. 
This notification should only happen if there is an internal error (e.g. `TARGET_JOB_NAME` is invalid) in this job. 
https://github.com/opensearch-project/opensearch-build/blob/bddbc06dd5ee6b67d3be3230cab922a083d4f15a/jenkins/check-for-build.jenkinsfile#L75-L78

### Acceptance Criteria:
We should only see failure notifications if there is an error generated inside of this job, not any child jobs.

zelinh:

### Description
Currently our check-for-build jenkins job will fail if its child job fails, which also sends out failure notification. Disabled propagate argument will not propagate the results of child job and continue the workflow on the current job. This job only fails if there is an issue in check-for-build itself.

### Issues Resolved
#1627 
 
### Check List

### DIFF
--- a/jenkins/check-for-build.jenkinsfile
+++ b/jenkins/check-for-build.jenkinsfile
@@ -72,16 +72,19 @@ pipeline {
                                 TARGET_JOB_NAME != 'distribution-build-opensearch-dashboards') {
                                 error "Job ${TARGET_JOB_NAME} is invalid"
                             }
-                            build job: "${TARGET_JOB_NAME}", parameters: [
-                            string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}"),
-                            string(name: 'TEST_MANIFEST', value: "${TEST_MANIFEST}")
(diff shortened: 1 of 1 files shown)


peternied:
Instead add `propagate: false` on the build job, see https://www.jenkins.io/doc/pipeline/steps/pipeline-build-step/

zelinh:
> Instead add `propagate: false` on the build job, see https://www.jenkins.io/doc/pipeline/steps/pipeline-build-step/

Great suggestion! Working on the change. Thanks!

peternied (approved):

gaiksaya (approved):

Issue #1468, opened 2022-01-13
[META] Add tests for existing jenkins jobs


abhinavGupta16:
### Is your feature request related to a problem? Please describe

Following up on the comment - https://github.com/opensearch-project/opensearch-build/issues/1444#issuecomment-1010324850

- [ ] Create a separate folder for tests for jenkins job
- [ ] #1712

The existing jenkins jobs needs tests - 

- [ ] [docker-build](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/docker/Jenkinsfile) job
- [ ] [opnesearch-dashboards-distribution-build](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/distribution-build.jenkinsfile) job
- [ ] [opensearch-ruby](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-ruby/Jenkinsfile) job
- [ ] [opensearch-distribution-build](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile) job
- [ ] [tar-promotion](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/promotion/tar-promotion.jenkinsfile) job
- [ ] [whitesource-scan](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/vulnerability-scan/whitesource-scan.jenkinsfile) job
- [ ] [test-orchestrator](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/test/orchestrator/Jenkinsfile) job
- [ ] [testsuite](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/test/testsuite/Jenkinsfile) job
- [ ] [check-for-build](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/check-for-build.jenkinsfile) job
- [ ] Add a CI check to ensure all jenkins job have a test associated with it.

### Describe the solution you'd like

Every job should have a test associated with it, with a regression `txt` file inside `jenkinsjob-regression-files`

### Acceptance Criteria
- All the libraries should have a lib tester associated with it
- All the jenkins job should have a test associated with it that uses only lib testers for libraries
- A check is added in CI making sure every job has a test

dblock:
> Let's create an issue for every task here when the work is begun. Thank you!

[...]

PR #1630, opened 2022-02-16
[WIP] Refractor the check-for-build method and tests

zelinh:
### Description
I found out a problem regarding of the PR just merged #1628. When we disable `propagate` when trigger the job, the following 
```
echo "Build succeeded, uploading build SHA for that job"
buildUploadManifestSHA(
    inputManifest: "manifests/${INPUT_MANIFEST}",
    jobName: "${TARGET_JOB_NAME}"
)
``` 
will be executed after triggered job failed which not what we expected. We shouldn't upload it if the distribution-build failed. 

The use of try-catch block will avoid this situation, if the `distribution-build` failed, nothing below the job call in the try block will be executed. 
 
### Issues Resolved
continue on #1627 
 
### Check List

### DIFF
--- a/jenkins/check-for-build.jenkinsfile
+++ b/jenkins/check-for-build.jenkinsfile
@@ -72,16 +72,19 @@ pipeline {
                                 TARGET_JOB_NAME != 'distribution-build-opensearch-dashboards') {
                                 error "Job ${TARGET_JOB_NAME} is invalid"
                             }
-                            build job: "${TARGET_JOB_NAME}", parameters: [
-                            string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}"),
-                            string(name: 'TEST_MANIFEST', value: "${TEST_MANIFEST}")
-                            ], wait: true
-                            
-                            echo "Build succeeded, uploading build SHA for that job"
-                            buildUploadManifestSHA(
-                                inputManifest: "manifests/${INPUT_MANIFEST}",
-                                jobName: "${TARGET_JOB_NAME}"
-                            )
+                            try {
+                                build job: "${TARGET_JOB_NAME}", parameters: [
+                                    string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}"),
+                                    string(name: 'TEST_MANIFEST', value: "${TEST_MANIFEST}")
+                                ], wait: true
+                                echo "Build succeeded, uploading build SHA for that job"
+                                buildUploadManifestSHA(
+                                    inputManifest: "manifests/${INPUT_MANIFEST}",
+                                    jobName: "${TARGET_JOB_NAME}"
+                                )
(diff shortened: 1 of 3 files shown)


dblock:
Should this be closed in favor of https://github.com/opensearch-project/opensearch-build/pull/1628?

peternied (changes requested):
I think that miss on he previous PR could have been caught by unit tests.  Before we do more work here can we add tests for this jenkinsfile that cover the success and failure build cases?

Note; we should revert the previous PR if this is impacting the builds negatively

zelinh:
> > I think that miss on he previous PR could have been caught by unit tests. Before we do more work here can we add tests for this jenkinsfile that cover the success and failure build cases?
> > Note; we should revert the previous PR if this is impacting the builds negatively
> 
> Oh my bad, read too quickly. Sounds like the SHA should be uploaded inside the `build` job instead?

Yes, that's another way I was considering.

zelinh:
> I think that miss on he previous PR could have been caught by unit tests. Before we do more work here can we add tests for this jenkinsfile that cover the success and failure build cases?
> 
> Note; we should revert the previous PR if this is impacting the builds negatively

Yes, adding a test for `check-for-build` is definitely needed as we have a meta issue here. #1468; while that's not in the scope of this PR. This PR provides an approach for resolving #1627 and moving SHA upload to build could also be another approach but it will require to be added in both `distribution-build-opensearch` & `distribution-build-opensearch-dashboards`. WDYT?  @dblock

dblock:
> > I think that miss on he previous PR could have been caught by unit tests. Before we do more work here can we add tests for this jenkinsfile that cover the success and failure build cases?
> > Note; we should revert the previous PR if this is impacting the builds negatively
> 
> Yes, adding a test for `check-for-build` is definitely needed as we have a meta issue here. #1468; while that's not in the scope of this PR. This PR provides an approach for resolving #1627 and moving SHA upload to build could also be another approach but it will require to be added in both `distribution-build-opensearch` & `distribution-build-opensearch-dashboards`. WDYT? @dblock

You can't use try/catch because it will swallow the error from uploading the SHA, and we'll never know.

jenkins/check-for-build.jenkinsfile
```diff
@@ -76,11 +76,6 @@ pipeline {
                                 string(name: 'INPUT_MANIFEST', value: "${INPUT_MANIFEST}"),
                                 string(name: 'TEST_MANIFEST', value: "${TEST_MANIFEST}")
                             ], wait: true, propagate: false
-                            echo "Build succeeded, uploading build SHA for that job"
```
peternied, line 79: We should still add tests to confirm the behavior is correct

peternied (changes requested):

zelinh:
Closing this for now since this is not being worked actively.

PR #1631, opened 2022-02-16
Update security to use gradle instead of maven build

peternied:
### Description
Update security to use gradle instead of maven build
 
### Issues Resolved
Fixed the distribution build break
 
### Check List

### DIFF
--- a/scripts/components/security/build.sh
+++ b/scripts/components/security/build.sh
@@ -17,7 +17,6 @@ function usage() {
     echo -e "-p PLATFORM\t[Optional] Platform, ignored."
     echo -e "-a ARCHITECTURE\t[Optional] Build architecture, ignored."
     echo -e "-o OUTPUT\t[Optional] Output path, default is 'artifacts'."
-    echo -e "-d DASHBOARDS\t[Optional] Build OpenSearch Dashboards, default is 'false'."
     echo -e "-h help"
 }
 
--- a/scripts/components/security/build.sh
+++ b/scripts/components/security/build.sh
@@ -60,10 +60,16 @@ if [ -z "$VERSION" ]; then
     exit 1
 fi
 
+OPENSEARCH_RELEASE=$VERSION
 [[ "$SNAPSHOT" == "true" ]] && VERSION=$VERSION-SNAPSHOT
 [ -z "$OUTPUT" ] && OUTPUT=artifacts
 
-mkdir -p $OUTPUT
+# see https://github.com/opensearch-project/security/pull/1409
+PLUGIN_VERSION=$(cat plugin-descriptor.properties | grep ^version= | cut -d= -f2 | sed "s/-SNAPSHOT//")
+[[ "$SNAPSHOT" == "true" ]] && PLUGIN_VERSION=$PLUGIN_VERSION-SNAPSHOT
+
+sed -i -e "s/\(^opensearch\.version=\).*\$/\1${OPENSEARCH_RELEASE}/" plugin-descriptor.properties
+sed -i -e "s/\(^version=\).*\$/\1${PLUGIN_VERSION}/" plugin-descriptor.properties
 
 ./gradlew assemble --no-daemon --refresh-dependencies -DskipTests=true -Dopensearch.version=$VERSION -Dopensearch_version=$VERSION -Dbuild.snapshot=$SNAPSHOT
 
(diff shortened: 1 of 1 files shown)


scripts/components/security/build.sh
```diff
@@ -59,22 +60,16 @@ if [ -z "$VERSION" ]; then
...
-# see https://github.com/opensearch-project/security/pull/1409
-PLUGIN_VERSION=$(cat plugin-descriptor.properties | grep ^version= | cut -d= -f2 | sed "s/-SNAPSHOT//")
-[[ "$SNAPSHOT" == "true" ]] && PLUGIN_VERSION=$PLUGIN_VERSION-SNAPSHOT
+mkdir -p $OUTPUT
 
-sed -i -e "s/\(^opensearch\.version=\).*\$/\1${OPENSEARCH_RELEASE}/" plugin-descriptor.properties
-sed -i -e "s/\(^version=\).*\$/\1${PLUGIN_VERSION}/" plugin-descriptor.properties
-sed -i -e "s/\(<opensearch.version>\).*\(<\/opensearch.version>\)/\1${VERSION}\2/g" pom.xml
-sed -i -e "1,/<version>/s/\(<version>\).*\(<\/version>\)/\1${PLUGIN_VERSION}\2/g" pom.xml
+./gradlew assemble --no-daemon --refresh-dependencies -DskipTests=true -Dopensearch.version=$VERSION -Dopensearch_version=$VERSION -Dbuild.snapshot=$SNAPSHOT
```
peternied, line 68: Note, there is a typo in the security build.gradle adding, so I am adding the additional version name until the fix can be made / merged into security.  This duplicate will get cleaned up out of band so the build isn't block while we wait.

https://github.com/opensearch-project/security/blob/5b994af53f8f5a9b258b2c66f844ce1ac88d6c1e/build.gradle#L60

scripts/components/security/build.sh
```diff
@@ -59,22 +60,16 @@ if [ -z "$VERSION" ]; then
     exit 1
 fi
 
-OPENSEARCH_RELEASE=$VERSION
 [[ "$SNAPSHOT" == "true" ]] && VERSION=$VERSION-SNAPSHOT
```
peternied, line 63: The properties file is still including the `-SNAPSHOT` on the version, trying to figure out what isn't working there, this shouldn't be merged until that has been resolved

scripts/components/security/build.sh
```diff
@@ -17,6 +17,7 @@ function usage() {
     echo -e "-p PLATFORM\t[Optional] Platform, ignored."
     echo -e "-a ARCHITECTURE\t[Optional] Build architecture, ignored."
     echo -e "-o OUTPUT\t[Optional] Output path, default is 'artifacts'."
+    echo -e "-d DASHBOARDS\t[Optional] Build OpenSearch Dashboards, default is 'false'."
```
dblock, line 20: What’s this?!

dblock:
Since security is gradle based now, could it build with a default build.sh?

scripts/components/security/build.sh
peternied: This was from the default build.sh in opensearch.

peternied:
> Since security is gradle based now, could it build with a default build.sh?

Nearly so, I fixed the build issues so we could unblock the distribution, we can fix up the typo in the opensearch.version, the plugin-descriptor being hardcoded and the extra admin zip file in that system and then cut back over to the default

scripts/components/security/build.sh
dblock: https://github.com/opensearch-project/opensearch-build/issues/1634

dblock (approved):

peterzhuamazon:
See this error during the build with opensearch-build, not sure if it is related here:
```

+ sed -i -e 's/\(^opensearch\.version=\).*$/\11.3.0/' plugin-descriptor.properties
sed: can't read plugin-descriptor.properties: No such file or directory
2022-02-17 20:49:25 ERROR    Error building security, retry with: ./build.sh manifests/1.3.0/opensearch-1.3.0.yml --component security
  File "./src/run_build.py", line 79, in <module>
    sys.exit(main())
  File "./src/run_build.py", line 67, in main
    builder.build(build_recorder)
  File "/local/home/zhujiaxi/opensearch-build-peterzhuamazon/src/build_workflow/builder_from_source.py", line 49, in build
    self.git_repo.execute(build_command)
  File "/local/home/zhujiaxi/opensearch-build-peterzhuamazon/src/git/git_repository.py", line 83, in execute
    subprocess.check_call(command, cwd=cwd, shell=True)
  File "/usr/lib64/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
```

Thanks.

dblock:
Gradle should remove the need for hacking the build.sh for security. Currently it tries to replace the version from -SNAPSHOT to something else.

peternied:
> See this error during the build with opensearch-build, not sure if it is related here:

Updated this PR to remove that logic as its been fixed on the security side of things

peterzhuamazon (approved):

PR #1652, opened 2022-02-18
Add distribution supported components list so only specific scripts can take -d parameter

### LINKED ISSUE #1651: [BUG] Only send `-d` to core components, instead of everything.
As of now, build workflow will not send `-d` if user does not specify `--distribution`.
However, if user specify things such as `--distribution rpm` and try to build core+all plugins,
`-d` will be sent to plugins as well.

The plugin build script does not know how to interpret `-d` thus fail.
```
+ echo 'Invalid option: -?'
Invalid option: -?
+ exit 1
2022-02-17 23:58:36 ERROR    Error building common-utils, retry with: ./build.sh manifests/1.3.0/opensearch-1.3.0.yml --component common-utils
  File "./src/run_build.py", line 79, in <module>
    sys.exit(main())
  File "./src/run_build.py", line 67, in main
    builder.build(build_recorder)
  File "/local/home/zhujiaxi/opensearch-build-peterzhuamazon/src/build_workflow/builder_from_source.py", line 49, in build
    self.git_repo.execute(build_command)
  File "/local/home/zhujiaxi/opensearch-build-peterzhuamazon/src/git/git_repository.py", line 83, in execute
    subprocess.check_call(command, cwd=cwd, shell=True)
  File "/usr/lib64/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command 'bash /local/home/zhujiaxi/opensearch-build-peterzhuamazon/scripts/components/common-utils/build.sh -v 1.3.0 -p linux -a x64 -d rpm -s false -o builds' returned non-zero exit status 1.
```

Need to add a condition where if component != OpenSearch/OpenSearch-Dashboards, then `-d` will not be sent even if not None.

peterzhuamazon:

### Description
Add distribution supported components list so only specific scripts can take -d parameter
 
### Issues Resolved
#1651
 
### Check List

### DIFF
--- a/src/build_workflow/builder_from_source.py
+++ b/src/build_workflow/builder_from_source.py
@@ -19,8 +19,6 @@ Artifacts found in "<build root>/artifacts/<maven|plugins|libs|dist|core-plugins
 
 
 class BuilderFromSource(Builder):
-
-
     def checkout(self, work_dir: str) -> None:
         self.git_repo = GitRepository(
             self.component.repository,
--- a/src/build_workflow/builder_from_source.py
+++ b/src/build_workflow/builder_from_source.py
@@ -19,6 +19,8 @@ Artifacts found in "<build root>/artifacts/<maven|plugins|libs|dist|core-plugins
 
 
 class BuilderFromSource(Builder):
+
+
     def checkout(self, work_dir: str) -> None:
         self.git_repo = GitRepository(
             self.component.repository,
@@ -28,6 +30,11 @@ class BuilderFromSource(Builder):
         )
 
     def build(self, build_recorder: BuildRecorder) -> None:
+
+        # List of components whose build scripts support `-d` parameter
+        # Bundled plugins do not need `-d` as they are java based zips
+        DISTRIBUTION_SUPPORTED_COMPONENTS = ["OpenSearch", "OpenSearch-Dashboards"]
+
         build_script = ScriptFinder.find_build_script(self.target.name, self.component.name, self.git_repo.working_directory)
 
         build_command = " ".join(
@@ -39,7 +46,7 @@ class BuilderFromSource(Builder):
                     f"-v {self.target.version}",
                     f"-p {self.target.platform}",
                     f"-a {self.target.architecture}",
-                    f"-d {self.target.distribution}" if self.target.distribution else None,
+                    f"-d {self.target.distribution}" if self.target.distribution and (self.component.name in DISTRIBUTION_SUPPORTED_COMPONENTS) else None,
                     f"-s {str(self.target.snapshot).lower()}",
                     f"-o {self.output_path}",
                 ]
--- a/tests/tests_build_workflow/test_builder_from_source.py
+++ b/tests/tests_build_workflow/test_builder_from_source.py
@@ -40,6 +40,18 @@ class TestBuilderFromSource(unittest.TestCase):
             ),
         )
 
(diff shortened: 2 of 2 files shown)


peterzhuamazon:
It fixes issues coming from #1629.

dblock:
This is a hack and I dislike it very much. I propose we go through the pain of adding support to `-d` to all these build.sh scripts instead. That should motivate us to get rid of custom steps in scripts so we can have default ones everywhere.

src/build_workflow/builder_from_source.py
```diff
@@ -39,7 +44,7 @@ def build(self, build_recorder: BuildRecorder) -> None:
                     f"-v {self.target.version}",
                     f"-p {self.target.platform}",
                     f"-a {self.target.architecture}",
-                    f"-d {self.target.distribution}" if self.target.distribution else None,
+                    f"-d {self.target.distribution}" if self.target.distribution and (self.component.name in DISTRIBUTION_SUPPORTED_COMPONENTS) else None,
```
dblock, line 47: Checking the name will break silently. 

Is the component type here `ComponentOpenSearchMin` or `ComponentOpenSearchDashboardsMin`? I believe it is, and in that case at least check whether `type(self.component) in [ComponentOpenSearchMin, ComponentOpenSearchDashboardsMin]`.

dblock (changes requested):
I'm ok with the hack but not checking by name.

src/build_workflow/builder_from_source.py
peterzhuamazon: Could you elaborate more on this? Why do we need to check component type?

Thanks.

dblock: The instance of `component` is not `Component`, it's something that's a child of that class, either `ComponentOpenSearchMin` or `ComponentOpenSearchDashboardsMin` therefore you don't need to check whether `.name == "OpenSearch"`, that's already done elsewhere, where we needed to know whether it's one or the other. Checking the name here would duplicate that logic.

dblock: Actually looking at the code it looks like we only do that in the maniests workflow :(

I'll approve this hack for now and we can refactor later.

dblock (approved):

src/build_workflow/builder_from_source.py
peterzhuamazon: After checking, it is actually `manifests.input_manifest.InputComponentFromSource` object and the name is directly taking from github repo name.

```
class InputComponentFromSource(InputComponent):
    def __init__(self, data: Any) -> None:
        super().__init__(data)
        self.repository = data["repository"]
        self.ref = data["ref"]
        self.working_directory = data.get("working_directory", None)

    def __stabilize__(self) -> None:
        ref, name = GitRepository.stable_ref(self.repository, self.ref)
        logging.info(f"Updating ref for {self.repository} from {self.ref} to {ref} ({name})")
        self.ref = ref

    def __to_dict__(self) -> dict:
        return {
            "name": self.name,
            "repository": self.repository,
            "ref": self.ref,
            "working_directory": self.working_directory,
            "checks": list(map(lambda check: check.__to_dict__(), self.checks)),
            "platforms": self.platforms,
        }
```

Will use this method for now.

abhinavGupta16 (approved):

Issue #1381, opened 2021-12-18
Support arm64 when running Dashboards integ test in Jenkins


tianleh:
### Is your feature request related to a problem? Please describe

#604

### Describe the solution you'd like

use proper Docker image and label

tianleh:
ensure proper Cypress version

Issue #704, opened 2021-10-06
Automate integration tests for OpenSearch Dashboards


kavilla:
### Is your feature request related to a problem? Please describe

Automate integration tests for OpenSearch Dashboards and plugins

### Describe the solution you'd like

Automatically runs through defined integration tests.

Issue #1566, opened 2022-01-27
[Bug]: copyArtifactPermission path requires current job without path


kavilla:
### Describe the bug

https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/distribution-build.jenkinsfile#L13

`copyArtifactPermission` currently takes the `JOB_NAME` provided. If you have a job called `jobA` and run the job from a nested location for example `Playground`, `JOB_NAME` will be `Playground\jobA`. So Jenkins will configure the run to allow permissions from `Playground\jobA` but Jenkins requires it to be `jobA`.

### To reproduce

Run an OpenSearch Dashboards job from a nested location. Notice the failure similar to this

```
hudson.AbortException: Unable to find project for artifact copy: Playground/avillk-distribution-build-opensearch-dashboards
This may be due to incorrect project name or permission settings; see help for project name in job configuration.
	at hudson.plugins.copyartifact.CopyArtifact.perform(CopyArtifact.java:452)
	at jenkins.tasks.SimpleBuildStep.perform(SimpleBuildStep.java:123)
	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:99)
	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:69)
	at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution.lambda$start$0(SynchronousNonBlockingStepExecution.java:47)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

### Expected behavior

Nested pipelines should be able to copy the artifact.

### Host / Environment

Jenkins

### Relevant log output

```shell
hudson.AbortException: Unable to find project for artifact copy: Playground/avillk-distribution-build-opensearch-dashboards
This may be due to incorrect project name or permission settings; see help for project name in job configuration.
	at hudson.plugins.copyartifact.CopyArtifact.perform(CopyArtifact.java:452)
[...]

PR #1653, opened 2022-02-18
[Dashboards] integ-test in Jenkins

kavilla:
### Description
Execute the integ-test workflow for OpenSearch Dashboards including
tests.

As opposed to OpenSearch, we need an image with the appropriate libraries
so we pass the image but it does not have the wget so the steps were broken
to access the build manifest.

Hardcoding the OpenSearch build id for now until this get resolves:
https://github.com/opensearch-project/opensearch-build/issues/1492

ARM tests will fail still until this is resolved:
https://github.com/opensearch-project/opensearch-build/issues/1381

### Issues Partially Resolved
https://github.com/opensearch-project/opensearch-build/issues/704
 
### Check List


kavilla:
Screen Shot 2022-02-17 at 9 20 35 PM

Build 52 was for linux x64 while build 53 was for linux arm64.

Screen Shot 2022-02-17 at 9 21 14 PM

tests/jenkins/TestRunIntegTestScript.groovy
```diff
@@ -0,0 +1,25 @@
...
+
+import jenkins.tests.BuildPipelineTest
+import org.junit.Before
+import org.junit.Test
+
+
+class TestRunIntegTestScript extends BuildPipelineTest {
+
+    @Test
+    public void TestRunIntegTestScript() {
```
abhinavGupta16, line 17: Can we add a lib-tester for the library?  Reference - https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#tests-for-jenkins-libraries

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,132 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
```
abhinavGupta16, line 1: We'll need to add this job to https://github.com/opensearch-project/opensearch-build/issues/1468 once PR is merged, so we can add test cases for this when `libTesters` are ready

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,132 @@
...
+            trim: true
+        )
+        string(
+            defaultValue: '',
+            name: 'AGENT_LABEL',
+            description: 'The agent label where the tests should be executed, e.g. Jenkins-Agent-al2-x64-c54xlarge-Docker-Host.',
+            trim: true
+        )
+        string(
+            defaultValue: '',
```
abhinavGupta16, line 31: Do we need an empty default value? Same for the rest

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,132 @@
...
+            post {
+                always {
+                    script {
+                        uploadTestResults(
+                            buildManifestFileName: "${BUILD_MANIFEST}",
+                            jobName: "${INTEG_TEST_JOB_NAME}",
+                            buildNumber: "${BUILD_ID}"
+                        )
+                    }
+                    cleanWs disableDeferredWipeout: true, deleteDirs: true
```
abhinavGupta16, line 88: Let's use `postCleanup()` function instead. Same with the rest

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,132 @@
...
+                }
+            }
+        }
+    }
+
+    post {
+        success {
+            node('Jenkins-Agent-al2-x64-m5xlarge') {
```
abhinavGupta16, line 96: Can we use the same node for `post` as in the stages?

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,132 @@
...
+            description: 'The agent label where the tests should be executed, e.g. Jenkins-Agent-al2-x64-c54xlarge-Docker-Host.',
+            trim: true
+        )
+        string(
+            defaultValue: '',
+            name: 'AGENT_IMAGE',
+            description: 'The agent label where the tests should be executed, e.g. opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028.',
+            trim: true
+        )
+    }
```
abhinavGupta16, line 36: We should add a stage to fail the job in case `AGENT_LABEL` and `AGENT_LABEL` are not specified. Otherwise, the job will be stuck trying to find a node in case it's blank or null.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,132 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
+
```
abhinavGupta16, line 2: We should a timeout for this job

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -2,12 +2,22 @@ lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
 
 pipeline {
     agent none
+    environment {
```
abhinavGupta16, line 5: Can you help add a timeout for this job? Ideally, it should have been added from when it was first created, but better late than never :)

Thank you!

abhinavGupta16 (changes requested):

dblock:
Would it be simpler and future-proof to read the agent from the test manifest just like we do from the build manifest? 

https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.0.0/opensearch-2.0.0.yml#L3

kavilla:
> Would it be simpler and future-proof to read the agent from the test manifest just like we do from the build manifest?
> 
> https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.0.0/opensearch-2.0.0.yml#L3

Will it change often? Regardless, I believe it would look cleaner and enables forks to have their own agents defined.

kavilla:
Thanks for your comments @abhinavGupta16, made the updates as suggested.

tests/jenkins/lib-testers/RunIntegTestScriptLibTest.groovy
```diff
@@ -0,0 +1,40 @@
...
+        this.testManifest = testManifest
+        this.buildId = buildId
+    }
+
+    void configure(helper, binding) {
+        // N/A
+    }
+
+    void parameterInvariantsAssertions(call) {
+        assertThat(call.args.jobName.first(), notNullValue())
```
abhinavGupta16, line 24: I think the job name can be null, since we substitute the default value for the jobname with `distribution-build-opensearch`

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -6,36 +6,49 @@ pipeline {
...
             name: 'AGENT_LABEL',
             description: 'The agent label where the tests should be executed, e.g. Jenkins-Agent-al2-x64-c54xlarge-Docker-Host.',
             trim: true
         )
     }
     stages {
+        stage('verify-parameters') {
+            agent {
+                node {
+                    label "Jenkins-Agent-al2-x64-m5xlarge"
```
abhinavGupta16, line 36: same as before

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,144 @@
...
+pipeline {
+    options {
+        timeout(time: 3, unit: 'HOURS')
+    }
+    agent none
+    environment {
+        AWS_ROLE_ARN = "arn:aws:iam::${AWS_ACCOUNT_PUBLIC}:role/opensearch-test"
```
abhinavGupta16, line 9: Where is `AWS_ROLE_ARN` and `AWS_ROLE_SESSION_NAME` being used?

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,144 @@
...
+            steps {
+                script {
+                    if (AGENT_LABEL == '') {
+                        currentBuild.result = 'ABORTED'
+                        error("Integration Tests failed to start. Missing parameter: AGENT_LABEL.")
+                    }
+                    if (AGENT_IMAGE == '') {
+                        currentBuild.result = 'ABORTED'
+                        error("Integration Tests failed to start. Missing parameter: AGENT_IMAGE.")
+                    }
```
abhinavGupta16, line 52: Do we need to add checks for other parameters?

abhinavGupta16:
@kavilla Thank you for making the changes. I have added a few more suggestions.

abhinavGupta16:
@peterzhuamazon - We have a few stages that run without docker. Should we add docker to all the stages?

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: Looks the original attention when added to OpenSearch integration test was to have it binded and used for the uploadTestResults. But we currently hardcode the role and session so I went ahead and deleted it from both Jenkinsfile.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: Just checked, fails regardless if the other parameters don't get passed.

kavilla:
@opensearch-project/engineering-effectiveness, anyway to get more reviews on this PR? When opened it was on parity with the initial OpenSearch integration tests pipeline but with added missing tests and it would be a big win to have this running as opposed to not have any integration tests running for distributions.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,137 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
+
+pipeline {
+    options {
+        timeout(time: 3, unit: 'HOURS')
```
gaiksaya, line 5: Does it take 3 hours to run this job?

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,137 @@
...
+    }
+    agent none
+    environment {
+        BUILD_MANIFEST = "build-manifest.yml"
+        INTEG_TEST_JOB_NAME = 'integ-test-opensearch-dashboards'
+    }
+    parameters {
+        string(
+            name: 'TEST_MANIFEST',
+            description: 'Test manifest under the manifests folder, e.g. manifests/2.0.0/opensearch-dashboards-2.0.0-test.yml.',
```
gaiksaya, line 15: We are already adding manifest path as `manifests/` here: https://github.com/opensearch-project/opensearch-build/pull/1653/files#diff-3f7479e22eeeedaf481e3de253091fff48673c90624a9e1101a57924864d9b8eR149
I think the example should be corrected to `/2.0.0/opensearch-dashboards-2.0.0-test.yml`

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -3,39 +3,45 @@ lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
 pipeline {
     agent none
     environment {
-        AWS_ROLE_ARN = "arn:aws:iam::${AWS_ACCOUNT_PUBLIC}:role/opensearch-test"
-        AWS_ROLE_SESSION_NAME = "jenkins-test-session"
         BUILD_MANIFEST = "build-manifest.yml"
+        INTEG_TEST_JOB_NAME = 'integ-test'
```
gaiksaya, line 7: Instead of hard-coding you can retrieve the job_name by $JOB_NAME env variable right?

vars/runIntegTestScript.groovy
```diff
@@ -4,11 +4,28 @@ void call(Map args = [:]) {
...
+}
+
+String generatePaths(buildManifest, artifactRootUrl) {
+    String name = buildManifest.build.name
+    String version = buildManifest.build.version
+    String platform = buildManifest.build.platform
+    String architecture = buildManifest.build.architecture
+
+    // TODO: Needs to be latest
+    String tempBuildId = "740"
```
gaiksaya, line 26: When should this change be done? Can you create an issue if not now so that we don't miss it?

gaiksaya (commented):
Not sure why `tests/jenkins/data/opensearch-1.3.0-build.yml` file is being added.

vars/runIntegTestScript.groovy
kavilla: If this PR gets merged in before this https://github.com/opensearch-project/opensearch-build/pull/1569, I can add a comment to that PR to update that piece in the PR.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: As of right now no, but I do not have metrics to determine the time the job will take once a full test suite of plugins onboarded to the new integration test repo with security and without. So opted for the a long time now and then once it gets running I think it can be revisited. Most likely being capped at 2 hours  (from the actual run time of the integration tests when I run them locally)

kavilla:
> Not sure why `tests/jenkins/data/opensearch-1.3.0-build.yml` file is being added.

For testing purposes, the original jenkins/data files were added in here https://github.com/opensearch-project/opensearch-build/commit/6c54533f20233721963dff6d3d353ad1220f3c60

But to add it a test for runIntegTestScript it needed a build manifest  but for the purpose of the test we added a build ID to a regular manifest. So it gave it my room to test on how manifests that go into the groovy script look like. I understand that there is https://github.com/opensearch-project/opensearch-build/tree/main/tests/data but given that more recent hash is in jenkins/data and the tests in the previous commit used jenkins/data it made sense to me to add the new file along with OpenSearch Dashboards.

However, if asking how come OpenSearch and not just OpenSearch Dashboards this PR comes with inclusion of new tests.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: Correct but if triggering the integ-test job manually it will still need that manifests, for example: manifests/2.0.0/opensearch-dashboards-2.0.0-test.yml will work but /2.0.0/opensearch-dashboards-2.0.0-test.yml will show `No such file or directory: '/var/jenkins/workspace/integ-test-opensearch-dashboards/1.2.0/opensearch-dashboards-1.2.0-test.yml`

jenkins/opensearch/integ-test.jenkinsfile
kavilla: This is a really great catch! Thanks!

The only was I've been able to test is by hardcoding the JOB_NAME  used when calling runIntegTestScript. In OpenSearch, it's not passing that value and thus able to us the correct job name but in OpenSearch Dashboards we will need to pass job name. But the assumption how JOB_NAME being used is incorrect. Will fix! Thanks.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,137 @@
...
+                    if (AGENT_IMAGE == '') {
+                        currentBuild.result = 'ABORTED'
+                        error("Integration Tests failed to start. Missing parameter: AGENT_IMAGE.")
+                    }
+                }
+            }
+        }
+        stage('download-manifest') {
+            agent {
+                node {
```
abhinavGupta16, line 51: Should this be in a docker?

jenkins/opensearch/integ-test.jenkinsfile
abhinavGupta16: @kavilla - Do we want `JOB_NAME` or JOB_BASE_NAME`? I think this should be `JOB_BASE_NAME` if we only want the job name

kavilla: With the update, `JOB_NAME` is just used to upload test results to S3. Even though it's not really supported with this PR, I'd imagine if I built from my Playground pipeline (Playground/xyz) and my distribution be located in `Playground/xyz/builds`, I would expect my test results to be in `Playground/xyz/test-results`.

But this is actually super awesome as it resolves another issue I was seeing https://github.com/opensearch-project/opensearch-build/issues/1566, so thanks for that! Will create a PR for this bug so it makes it easier for me to have a playground environment.

TBH, I think most of this confusion can be mitigated with if the build manifest includes like a `JOB_NAME`. So instead of hard coding `BUILD_JOB_NAME` I can just get it it from the manifest. If we think that's a good idea I can create an issue for it. Let me know!

Thanks!

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: Within OpenSearch integ-test, they are able to have a single step that downloads and runs the test. However, the image that is required to run the integration tests for OpenSearch Dashboards does not support `wget` which is what the `downloadBuildManifest` requires. 

Would the suggestion be to hardcode this stage to utilize a docker image that would normally be used by OpenSearch?

Also for my own curiosity, what's the benefit of running within a docker container vs running within a node? Does this prevent potential conflicts?

abhinavGupta16: Running within container ensures a fresh run every-time because the container will be destroyed after every run. Also, as you rightly mentioned, it also helps prevent conflicts. It would also ensure that the node is not affected by the job.

We can specify any image that would help support this operation. 

Also, why are we taking image name as input from the user? Can these possibly change without changes to the job?

kavilla: > We can specify any image that would help support this operation.

Do you know of which I can use? Used a couple docker images and all fail with `line 1: wget: command not found`. It would appear that's why OpenSearch is doing the same for integ tests: https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/integ-test.jenkinsfile#L37

> Also, why are we taking image name as input from the user? 

When OpenSearch Dashboards builds it already detected the docker image and will require to use the appropriate image for example: https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.0.0/opensearch-dashboards-2.0.0.yml#L8. So when the build pipeline triggers the job it will pass the values it already has on hand.

> Can these possibly change without changes to the job?

Apologies I'm not clear what you mean here could you clarify?

abhinavGupta16: > When OpenSearch Dashboards builds it already detected the docker image and will require to use the appropriate image for example: https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.0.0/opensearch-dashboards-2.0.0.yml#L8. So when the build pipeline triggers the job it will pass the values it already has on hand.

Does this job pick the docker image from the manifest file? If yes, we should not have an input param,  but assign the image variable locally using manifest.

kavilla: > You can `curl` instead of `wget` I believe ...

Secret sauce, was able to update.

kavilla: > Does this job pick the docker image from the manifest file? If yes, we should not have an input param, but assign the image variable locally using manifest.

As of right now, the actual build manifest doesn't include the CI image stuff. The `manifest` do. There could be potential to pass `MANIFEST`, `TEST_MANIFEST`, and `BUILD_MANIFEST_URL`. But if we are going that route I would rather we open an issue and address that in both OpenSearch and OpenSearch Dashboards.

vars/runIntegTestScript.groovy
kavilla: Addressed!

kavilla:
@opensearch-project/engineering-effectiveness

bbarani:
> @opensearch-project/engineering-effectiveness

We are looking in to it.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
gaiksaya: So are the args different for manual vs automated run of this file?

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -0,0 +1,123 @@
...
+        string(
+            name: 'AGENT_LABEL',
+            description: 'The agent label where the tests should be executed, e.g. Jenkins-Agent-al2-x64-c54xlarge-Docker-Host.',
+            trim: true
+        )
+        string(
+            name: 'AGENT_IMAGE',
+            description: 'The agent label where the tests should be executed, e.g. opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028.',
+            trim: true
+        )
```
peterzhuamazon, line 32: This is a bit confusing as agent label is to the agent node label, while the image is the container image running on top of the agent.

Using container image probably is more clear as agents are VM not containers.

peterzhuamazon (commented):
I dont have too much issues with this PR except a few naming convention.

Thanks.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: So to clarify you would like this `AGENT_IMAGE` ~> `CONTAINER_IMAGE`

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: Should be the same, 

Manually we can include the `manifests` but at the time of automation we have to append `manifests` to the test run arg: 
https://github.com/opensearch-project/opensearch-build/pull/1653/files#diff-3f7479e22eeeedaf481e3de253091fff48673c90624a9e1101a57924864d9b8eR149.

OpenSearch is doing the same but I understand the confusion. When building, we call groovy scripts adding `manifests`. So instead of passing it to the test job. We can just pass `/2.0.0/opensearch-dashboards-2.0.0-test.yml` and then in the test script we can append `manifests` like how we do while building. For example [this](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/distribution-build.jenkinsfile#L46).

gaiksaya: I see but when we run the above example file we pass manifest as `2.0.0/test.yml`
image
Was wondering if we can keep the manual and automated way as the same?

jenkins/opensearch-dashboards/integ-test.jenkinsfile
peterzhuamazon: sure.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: Correct that is for the build pipeline, for the test pipeline this is how it looks:

Screen Shot 2022-03-02 at 12 15 06 PM

I went ahead and updated the OpenSearch Dashboards and OpenSearch to match what the build pipeline is doing.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: Addressed!

peterzhuamazon (approved):

peterzhuamazon:
Need @abhinavGupta16 to remove his block before we merge.
Thanks.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
abhinavGupta16: @kavilla - In that case, can we code in the docker image rather than taking it as an input parameter? It would reduce one parameter for the user. 
Once the manifest supports holding the image, the job would then parse the build manifest and get the image from there

abhinavGupta16:
> Need @abhinavGupta16 to remove his block before we merge. Thanks.

Added comments. I believe we can get rid of the image name parameter. One less thing for the user to input and worry about

gaiksaya (approved):

jenkins/opensearch-dashboards/integ-test.jenkinsfile
kavilla: We could utilize https://github.com/opensearch-project/opensearch-build/issues/1687, which would have the image available. But that wouldn't resolve the issue about the number of params. I wouldn't see any issue if we did include the CI information into the actual build manifest to reduce the amount of params.

I created an issue for that: https://github.com/opensearch-project/opensearch-build/issues/1694 since that will require more refactoring within the actual build repo.

abhinavGupta16 (approved):

abhinavGupta16:
I have approved it for now, however, we should look at resolving these new issues at the earliest. 

@kavilla @bbarani @gaiksaya @peterzhuamazon

Issue #1544, opened 2022-01-25
[RPM M1] The assemble workflow should have a function to wrap build artifacts into a RPM package


peterzhuamazon:
Tasks | Estimate | Status | Notes | Dependencies
-- | -- | -- | -- | --
~~Just like build workflow, add a “--distribution” parameter, so that the assemble workflow knows which distribution to create based on the input artifacts from build workflow~~ <br><br>20220220: Design change, since build workflow already have `--distribution` ready, assemble workflow can directly take distribution info from build manifest. Then assemble the new rpm after zips installation. | 3 | Complete |   |  
The existing distribution generating code of TAR and ZIP should still function after the above change | 3 | Complete | 20220220: #1659 ✅  <br><br>20220228: Improvements on maps: https://github.com/opensearch-project/opensearch-build/pull/1676 ✅   |

peterzhuamazon:
20220220: 
* https://github.com/opensearch-project/opensearch-build/pull/1659

peterzhuamazon:
20220228: Improvements on maps: https://github.com/opensearch-project/opensearch-build/pull/1676 ✅

PR #1659, opened 2022-02-20
[RPM M1] Assemble artifacts based on distribution not file extension

peterzhuamazon:

### Description
Hi,

This PR is to assemble artifacts based on distribution value in build manifest.
Instead of based on build workflow artifact file extension.
We will use distribution value in build manifest to assemble artifact.

This is the PR to follow up on #1629, as a pre-requisite to add RPM distribution extract/install/assemble.

Thanks.

 
### Issues Resolved
Part of #1544
 
### Check List


peterzhuamazon:
@dblock need some help on the mypy error:
```

tests/tests_assemble_workflow/test_dist.py:49: error: "DistTar" has no attribute "_Dist__find_min_archive_path"  [attr-defined]
tests/tests_assemble_workflow/test_dist.py:56: error: "DistTar" has no attribute "_Dist__rename_archive_path"  [attr-defined]
tests/tests_assemble_workflow/test_dist.py:64: error: "DistTar" has no attribute "_Dist__rename_archive_path"  [attr-defined]
```

My test case is able to call the function and it exists in the code:
```
def test_find_min_archive_path(self) -> None:
        self.assertEqual(
            self.distTar._Dist__find_min_archive_path(self.artifacts_path),
            self.artifacts_path + "opensearch-1.3.0"
        )
```
The reason I call `self.distTar._Dist__find_min_archive_path` is because `self.distTar.__find_min_archive_path` will throw errors during the pytest.

If you have any experience on this or any advice, please let me know about it.

Thanks.

src/manifests/build_manifest.py
```diff
@@ -22,6 +22,8 @@
   version: string
   platform: linux, darwin or windows
   architecture: x64 or arm64
+  distribution: tar, zip, and more
```
gaiksaya, line 25: Be specific here instead of more. `tar, zip and rpm`

src/manifests/bundle_manifest.py
```diff
@@ -23,6 +23,8 @@ class BundleManifest(ComponentManifest['BundleManifest', 'BundleComponents']):
           version: string
           platform: linux, darwin or windows
           architecture: x64 or arm64
+          distribution: tar, zip, or more
```
gaiksaya, line 26: Same as above

tests/tests_assemble_workflow/data/artifacts/dist/opensearch-1.3.0/opensearch.txt
```diff
@@ -0,0 +1 @@
+This page intentionally left blank.
```
gaiksaya, line 1: what is this file for?

tests/tests_assemble_workflow/data/opensearch-dashboards-build-windows-1.3.0.yml
```diff
@@ -0,0 +1,26 @@
+---
```
gaiksaya, line 1: Why are we adding windows changes in this PR? I think its better to keep the scope of this PR limited to rpm changes

peterzhuamazon:
@dblock need some help on the mypy error:
```

tests/tests_assemble_workflow/test_dist.py:49: error: "DistTar" has no attribute "_Dist__find_min_archive_path"  [attr-defined]
tests/tests_assemble_workflow/test_dist.py:56: error: "DistTar" has no attribute "_Dist__rename_archive_path"  [attr-defined]
tests/tests_assemble_workflow/test_dist.py:64: error: "DistTar" has no attribute "_Dist__rename_archive_path"  [attr-defined]
```

My test case is able to call the function and it exists in the code:
```
def test_find_min_archive_path(self) -> None:
        self.assertEqual(
            self.distTar._Dist__find_min_archive_path(self.artifacts_path),
            self.artifacts_path + "opensearch-1.3.0"
        )
```
The reason I call `self.distTar._Dist__find_min_archive_path` is because `self.distTar.__find_min_archive_path` will throw errors during the pytest.

If you have any experience on this or any advice, please let me know about it.

Thanks.



> @dblock need some help on the mypy error:
> 
> ```
> 
> tests/tests_assemble_workflow/test_dist.py:49: error: "DistTar" has no attribute "_Dist__find_min_archive_path"  [attr-defined]
> tests/tests_assemble_workflow/test_dist.py:56: error: "DistTar" has no attribute "_Dist__rename_archive_path"  [attr-defined]
> tests/tests_assemble_workflow/test_dist.py:64: error: "DistTar" has no attribute "_Dist__rename_archive_path"  [attr-defined]
> ```
> 
> My test case is able to call the function and it exists in the code:
> 
> ```
> def test_find_min_archive_path(self) -> None:
>         self.assertEqual(
>             self.distTar._Dist__find_min_archive_path(self.artifacts_path),
>             self.artifacts_path + "opensearch-1.3.0"
>         )
> ```
> 
> The reason I call `self.distTar._Dist__find_min_archive_path` is because `self.distTar.__find_min_archive_path` will throw errors during the pytest.
> 
> If you have any experience on this or any advice, please let me know about it.
> 
> Thanks.

@peternied do you have any experience on this situation?
These functions exist as __<functionname>, if you call with __<functionname>, the test wont run but if you express them into _<ClassName>__<functionname>they will run, but then mypy complain about it.

Thanks.

Thanks.

src/manifests/build_manifest.py
peterzhuamazon: I will update.

src/manifests/bundle_manifest.py
peterzhuamazon: I will update.

tests/tests_assemble_workflow/data/opensearch-dashboards-build-windows-1.3.0.yml
peterzhuamazon: This is not to add windows changes, rather adding distribution related tests.
The PR itself is also not adding rpm, but adding assemble based on distribution not extension.
Therefore, the windows distribution manifest provide a great scenario of `zip` aside from the default `tar` for us to test.

src/manifests/bundle_manifest.py
```diff
@@ -80,6 +83,7 @@ def __init__(self, data: Dict[str, str]):
             self.version = data["version"]
             self.platform = data["platform"]
             self.architecture = data["architecture"]
+            self.distribution: str = data.get('distribution', None)
```
zelinh, line 86: I think `data.get('distribution')` will return `None` in default if the key not exists, no need to specify it.

tests/tests_assemble_workflow/data/artifacts/dist/opensearch-1.3.0/opensearch.txt
zelinh: Then we can say this file is a placeholder so people won't be confused?

src/manifests/bundle_manifest.py
peterzhuamazon: It will error our in my previous setup thus I add it.
This is the same change in #1629.
Thanks.

tests/tests_assemble_workflow/data/artifacts/dist/opensearch-1.3.0/opensearch.txt
peterzhuamazon: It already said ` This page intentionally left blank. `

src/assemble_workflow/dists.py
```diff
@@ -0,0 +1,24 @@
...
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import logging
+
+from assemble_workflow.dist import Dist, DistTar, DistZip
+
+
+class Dists:
+    DISTRIBUTIONS_MAP = {
```
mch2, line 13: We now have two maps to maintain, here and in BundleRecorder - perhaps add a class method on DistTar/Zip for fetching the extension and import this map into BundleRecorder?

peternied:
You can still access private methods, see this stack-overflow on how it can be done

https://stackoverflow.com/a/50164564

src/assemble_workflow/dists.py
peterzhuamazon: Will try that method thanks for advice.

peterzhuamazon:
> What about making `find_min_archive_path` a non-private function, seems like that would resolve this problem?
> 
> It is anti-pythonic to create private methods as there is a notion that call code can be seen / used, see this SO post for [additional opinions](https://stackoverflow.com/a/2064400).

Will take a look on that, never thought about this aspect in Python.
Thanks.

dblock (approved):
LGTM! The comment above doesn't feel like a must have.

src/assemble_workflow/dists.py
peterzhuamazon: I will create a secondary PR after this one to implement this, already have clues.

Thanks.

tests/tests_assemble_workflow/test_dist.py
```diff
@@ -4,21 +4,63 @@
...
+            self.distTar.rename_archive_path("temp_path/opensearch-1.3.0"),
+            "temp_path/opensearch-1.3.0"
+        )
+
+    @patch("os.path.dirname", return_value="temp_path")
+    @patch("os.rename")
+    def test_rename_archive_path_rename(self, os_path_dirname: Mock, os_rename: Mock) -> None:
+        self.assertEqual(
+            self.distTar.rename_archive_path("temp_path/opensearch-x.y.z"),
+            "temp_path/opensearch-1.3.0"
```
zelinh, line 65: It seems like this is failing the python check for windows because of the path format.

peterzhuamazon: fixed.

src/assemble_workflow/dists.py
peterzhuamazon: https://github.com/opensearch-project/opensearch-build/pull/1676

PR #1666, opened 2022-02-23
Automate the creation of release tags

### LINKED ISSUE #378: Automate the creation of release tags after every release
Currently the plugin teams are creating release tags manually after every release. This can lead to discrepancies when they cut the release tag using a different commit id compared to the one shipped with the artifact

The build system knows the the commit id associated with each (generated) component artifacts shipped as part of a distribution. We should come up with an automated process (workflow) to cut release tags for all the components participating in a release based on the commit id (in the build manifest file) after every release.

- [ ] Create proof of concept
- [ ] Create a design document
- [ ] Implementation

Manually creating release tags after release

zelinh:

### Description
Create the release tag for plugins from the distribution-build manifest based on the version and build_id. 
This workflow will fetch the commit_id from the manifest for each plugin and create the GitHub tag based on that, which could avoid discrepancies of using different commit id from release. 
 
### Issues Resolved
#378 
 
### Check List


vars/createReleaseTag.groovy
```diff
@@ -0,0 +1,35 @@
...
+    def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+    def buildManifestObj = lib.jenkins.BuildManifest.new(readYaml(file: args.buildManifest))
+
+    def componentsName = buildManifestObj.getComponets()
+    def componetsNumber = componentsName.size()
+    def version = args.tagVersion
+    echo "Creating $version release tag for $componetsNumber components in ths manifest"
+
+    withCredentials([usernamePassword(credentialsId: "${GITHUB_BOT_TOKEN_NAME}", usernameVariable: 'GITHUB_USER', passwordVariable: 'GITHUB_TOKEN')]) {
+        for (component in componentsName) {
```
gaiksaya, line 12: I think its better to add try catch and check if the tag already exists. Some components cut the tag beforehand.

gaiksaya (changes requested):
Please add tests for both groovy and JenkinsFile

vars/createReleaseTag.groovy
peterzhuamazon: Agreed and probably can also check whether the existing tag has the correct commitid.

peterzhuamazon (changes requested):
Not too many issues with the PR but where are the tests?
Could you put this to draft PR and add test cases?
Thanks.

vars/createReleaseTag.groovy
zelinh: I added a check below: If certain tag is existing in the current repo, we will just simply delete it and replace with the new one with the commit ID from the manifest.

gaiksaya: Not sure if we should delete or skip it. @dblock any thoughts on this?

dblock: I would skip it with a note if it's identical, and fail if the existing tag doesn't match.

jenkins/release-tag/release-tag.jenkinsfile
```diff
@@ -0,0 +1,54 @@
...
+pipeline {
+    agent {
+        docker {
+            label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+            image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211130'
+            alwaysPull true
+        }
+    }
+    parameters {
+        string(
```
dblock, line 12: As an idea, this job could take a version number without a build ID, because the URL for the manifest for a well-known version should be known. Or it could take a URL to the manifest. This avoids the need to download anything from S3. It's no big deal though.

jenkins/release-tag/release-tag.jenkinsfile
```diff
@@ -0,0 +1,54 @@
+lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+
```
abhinavGupta16, line 2: Let's add a timeout to the job

jenkins/release-tag/release-tag.jenkinsfile
```diff
@@ -0,0 +1,54 @@
...
+            trim: true
+        )
+        string(
+            name: 'VERSION',
+            description: 'Version of the distribution build manifest.',
+            trim: true
+        )
+    }
+    environment {
+            BUILD_MANIFEST = "builds/opensearch/manifest.yml"
```
abhinavGupta16, line 24: Can this variable be local to the stage? Let's avoid global variables as much as possible

zelinh: Actually no. I initially had this in local but moved this to env variable for testing purpose.

jenkins/release-tag/release-tag.jenkinsfile
zelinh: Added. Thanks.

jenkins/release-tag/release-tag.jenkinsfile
zelinh: Getting the manifest URL will still need the build ID, otherwise we will need to take the manifest URL as the parameter input. I think essentially both way are the same, good for either way.

vars/createReleaseTag.groovy
zelinh: I added another `if` block to check if the commit ID matches the ref for existing tag.

jenkins/release-tag/release-tag.jenkinsfile
abhinavGupta16: Is it used anywhere outside of the job?

jenkins/release-tag/release-tag.jenkinsfile
```diff
@@ -1,6 +1,9 @@
 lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
 
 pipeline {
+    options {
+            timeout(time: 2, unit: 'HOURS')
```
abhinavGupta16, line 5: How long does this job take, in your tests?

jenkins/release-tag/release-tag.jenkinsfile
zelinh: No. This is only used as the place we store the manifest downloaded and used for fetching the info we needed. While I have this as a fixed location in this job, I used a different manifest path from our testing directory. (`tests/data/opensearch-build-1.1.0.yml`). If I defined this variable in the local, we can't perform testing without actually downloading the manifest to here.

jenkins/release-tag/release-tag.jenkinsfile
zelinh: It depends on multiple things: the number of components the manifest, the size of contents for each repo, github fetching time, etc.. I tested with only three components and it normally took around 2 mins.

abhinavGupta16: Then let's change the timeout to be 30mins?

jenkins/release-tag/release-tag.jenkinsfile
abhinavGupta16: I think overriding the variable using binding should work. Let's try this offline

tests/jenkins/TestReleaseTagJob.groovy
```diff
@@ -0,0 +1,46 @@
...
+        this.registerLibTester(new CreateReleaseTagLibTester('tests/data/opensearch-build-1.1.0.yml', '1.1.0'))
+
+        super.setUp()
+
+        // Variables for Maven Sign Release job
+        binding.setVariable('VERSION', '1.1.0')
+        binding.setVariable('BUILD_ID', '123')
+        binding.setVariable('ARTIFACT_BUCKET_NAME', bucketName)
+        binding.setVariable('BUILD_MANIFEST', 'tests/data/opensearch-build-1.1.0.yml')
+        binding.setVariable('WORKSPACE', 'tmp/workspace')
```
abhinavGupta16, line 35: This is defined in the parent class, you can choose to remove it.

tests/jenkins/TestReleaseTagJob.groovy
```diff
@@ -0,0 +1,46 @@
...
+
+        def destPath = "tmp/workspace/${buildManifest}"
+
+        def artifactsPath = 'distribution-build-opensearch/1.1.0/123/linux/x64/builds/opensearch/manifest.yml'
+
+        def bucketName = 'job-s3-bucket-name'
+
+        this.registerLibTester(new DownloadFromS3LibTester(destPath, bucketName, artifactsPath, true))
+
+        this.registerLibTester(new CreateReleaseTagLibTester('tests/data/opensearch-build-1.1.0.yml', '1.1.0'))
```
abhinavGupta16, line 26: Lets use the variable from line 16 here.

tests/jenkins/TestReleaseTagJob.groovy
```diff
@@ -0,0 +1,46 @@
...
+
+        super.setUp()
+
+        // Variables for Maven Sign Release job
+        binding.setVariable('VERSION', '1.1.0')
+        binding.setVariable('BUILD_ID', '123')
+        binding.setVariable('ARTIFACT_BUCKET_NAME', bucketName)
+        binding.setVariable('BUILD_MANIFEST', 'tests/data/opensearch-build-1.1.0.yml')
```
abhinavGupta16, line 34: Lets use the variable from line 16 here.

jenkins/release-tag/release-tag.jenkinsfile
zelinh: No it won't. That's what I tried initially and why I switched to what it's now.

jenkins/release-tag/release-tag.jenkinsfile
zelinh: I think the timeout here is the max time this job possibly takes. I would like to have this longer just in case. Shouldn't be a big deal IMO.

peterzhuamazon: 2 hr is a ok time range, I normally have 2x 3x time for max range.

jenkins/release-tag/release-tag.jenkinsfile
```diff
@@ -0,0 +1,57 @@
+lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+
+pipeline {
+    options {
+            timeout(time: 2, unit: 'HOURS')
+    }
+    agent {
+        docker {
+            label 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+            image 'opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211130'
```
peterzhuamazon, line 10: Try to see if you can use `alpine:latest` instead of this full ledge image for testing.
Since this is literally a git command pipeline.

peterzhuamazon: Eh, seems like you need aws cli as well.

vars/createReleaseTag.groovy
```diff
@@ -0,0 +1,53 @@
...
+                            exit 1
+                        else
+                            echo "Tag $version has been created with identical commit ID. Skipping creating new tag for $component."
+                        fi
+                    else
+                        git tag $version
+                    fi
+                else
+                    if [[ -n \$(git ls-remote --tags $repo $version.0) ]]; then
+                        if [ \$(git ls-remote --tags $repo $version.0 | awk 'NR==1{print \$1}') != $commitID ]; then
```
peterzhuamazon, line 38: You are mixing the bash buildin brackets and normal brackets everywhere.
I would suggest save the values of command returns to variable then check it to avoid these kind of complications.

vars/createReleaseTag.groovy
gaiksaya: Wouldn't it be easier using a try catch block so that exceptions can be handled per the scenario (Already existing and wrong ones). The code will be more cleaner too.

zelinh: I don't think there is try catch in shell. If we use the try catch outside of the shell block, push existing tag with same commit ID won't throw any exceptions. Error only occurs when we try to push tag with different commit ID.

gaiksaya: Did you try the groovy script or python for the same? Instead of inline code in JenkinsFile we can create a python script or groovy script and just execute it via jenkins as we are doing for other libraries. WDYT?

zelinh: This is the groovy library script that I'm trying to perform all functions and be executed from the Jenkins.

gaiksaya: Let me be more clear. I was recommending to remove the entire shell block and replace it with groovy. 
Just an example: https://stackoverflow.com/questions/45956629/how-to-use-git-commands-diff-log-etc-in-groovy-script-windows/45965983
Python: https://github.com/opensearch-project/opensearch-build/blob/main/src/git/git_repository.py

abhinavGupta16: Thanks @gaiksaya  for bringing this up. I too agree that we should remove the shell block, as much as we can, it is better for error handling and testing etc. 

@dblock - what are your thoughts here?

jenkins/release-tag/release-tag.jenkinsfile
gaiksaya: Also, We are bundling this manifest with the released artifact as well. Does it makes sense to download from publicly downloadable link 
```
Example: https://artifacts.opensearch.org/releases/bundle/opensearch/1.2.4/opensearch-1.2.4-linux-x64.tar.gz
```
We can download from public url , unzip and use the manifest. In that case just the version is required.

zelinh: I liked this idea of using this public link for transparency purpose. Just make sure few things: 
This release artifacts matches the build ID that we will be using for each release right?
Also I actually uses the builds manifest instead of the dist manifest from link like `https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/$version/$build_number/linux/x64/build/opensearch/manifest.yml`; using the dist manifest from the release artifacts will not include some components that won't redistribute anything (e.g. `common-util`). So we will still need to manually tag for them.

gaiksaya: Makes sense. Thanks!

vars/createReleaseTag.groovy
bbarani: @zelinh Can you work with the team to come up with an effort estimate to move the shell code with groovy? I would prefer to go that route if its a quick effort.

dblock: I dislike the implementation here, but with the rule of thumb that this is better than what we had (manual process) I vote to merge this as is.

dblock (approved):

abhinavGupta16:
> I dislike the implementation here, but with the rule of thumb that this is better than what we had (manual process) I vote to merge this as is.

@zelinh - Can you create an issue to look into this in the future?

jenkins/release-tag/release-tag.jenkinsfile
dblock: Implemented in https://github.com/opensearch-project/opensearch-build/pull/1705

vars/createReleaseTag.groovy
zelinh: Changed the implementation here and remove shell block as much as possible.

tests/jenkins/TestReleaseTagJob.groovy
```diff
@@ -0,0 +1,91 @@
...
+
+        super.setUp()
+
+        // Variables for Release Tag Job
+        binding.setVariable('VERSION', '1.1.0')
+        binding.setVariable('BUILD_ID', '123')
+        binding.setVariable('ARTIFACT_BUCKET_NAME', bucketName)
+        binding.setVariable('BUILD_MANIFEST', buildManifest)
+
+        helper.registerAllowedMethod("checkout", [Map], {})
```
abhinavGupta16, line 36: These method registrations should be a part of `libTester` so anyone who uses this library is able to use them.

tests/jenkins/TestReleaseTagJob.groovy
```diff
@@ -0,0 +1,91 @@
...
+        helper.registerAllowedMethod("dir", [Map], {})
+        helper.addShMock("git ls-remote --tags https://github.com/opensearch-project/OpenSearch.git 1.1.0 | awk 'NR==1{print \$1}'") { script ->
+            return [stdout: "", exitValue: 0]
+        }
+        helper.addShMock("git ls-remote --tags https://github.com/opensearch-project/common-utils.git 1.1.0.0 | awk 'NR==1{print \$1}'") { script ->
```
abhinavGupta16, line 41: Why do we want to mock these sh values, since they don't return anything? These mocks would simply be ignored in case this is not found.

Maybe assertions in the lib tester that these sh mocks exists is a better way to go.

vars/createReleaseTag.groovy
```diff
@@ -0,0 +1,54 @@
...
+    echo "Creating $version release tag for $componetsNumber components in the manifest"
+
+    withCredentials([usernamePassword(credentialsId: "${GITHUB_BOT_TOKEN_NAME}", usernameVariable: 'GITHUB_USER', passwordVariable: 'GITHUB_TOKEN')]) {
+        for (component in componentsName) {
+            def commitID = buildManifestObj.getCommitId(component)
+            def repo = buildManifestObj.getRepo(component)
+            def push_url = "https://$GITHUB_TOKEN@" + repo.minus('https://')
+            echo "Tagging $component at $commitID ..."
+
+            dir (component) {
```
abhinavGupta16, line 18: This is much better than before! Thanks!

abhinavGupta16 (changes requested):

vars/createReleaseTag.groovy
```diff
@@ -0,0 +1,54 @@
...
+                    ).trim()
+                    if (tag_id == "") {
+                        echo "Creating $version tag for $component"
+                        sh "git tag $version"
+                        sh "git push $push_url $version"
+                    } else if (tag_id == commitID) {
+                        echo "Tag $version has been created with identical commit ID. Skipping creating new tag for $component."
+                    } else {
+                        error "Tag $version already existed in $component with a different commit ID. Please check this."
+                    }
```
gaiksaya, line 34: Instead of duplicating the logic, just change the `tag_version` if the component if OpenSearch

abhinavGupta16: +1

zelinh: Thanks for the advice! Updated.

peterzhuamazon (approved):

vars/createReleaseTag.groovy
```diff
@@ -0,0 +1,43 @@
...
+                ).trim()
+                if (tag_id == "") {
+                    echo "Creating $tag_version tag for $component"
+                    sh "git tag $tag_version"
+                    sh "git push $push_url $tag_version"
+                } else if (tag_id == commitID) {
+                    echo "Tag $tag_version has been created with identical commit ID. Skipping creating new tag for $component."
+                } else {
+                    error "Tag $tag_version already existed in $component with a different commit ID. Please check this."
+                }
```
zelinh, line 39: The reason I use `if` instead of `try-catch` block here is that when I compare the tag_id, even if the tag already exists or with different commitID, there won't be an exception threw automatically. If we chose to throw the exception and catch it by ourselves, it will be the same logic as what I have. Therefore no need to to use `try-catch` here.

tests/jenkins/TestReleaseTagJob.groovy
zelinh: It returns the commit ID. I mocked different cases for when the tag is not existed so they don't return anything..

tests/jenkins/TestReleaseTagJob.groovy
zelinh: No. Anyone uses the library will have possibility using different manifest. The method registrations here are specifically for the manifest I used.

tests/jenkins/TestReleaseTagJob.groovy
abhinavGupta16: shMock returning empty strings will not serve any purpose here.

tests/jenkins/TestReleaseTagJob.groovy
abhinavGupta16: Since checkout and dir are part of the library, they should be a part of the lib tester

tests/jenkins/TestReleaseTagJob.groovy
abhinavGupta16: These should be a part of libtester since they are specific to the library

zelinh: Had a discussion with @abhinavGupta16 , and new changes are updated.

tests/jenkins/TestReleaseTagJob.groovy
zelinh: Had a discussion with @abhinavGupta16 , and new changes are updated.

jenkins/release-tag/release-tag.jenkinsfile
abhinavGupta16: 3x for 10min is 30mins

zelinh: We won't know if this takes 10 mins or 30 mins or even more. I believe 2 hours should be an acceptable range.

abhinavGupta16 (approved):
Worked with @zelinh to add library related `shMock` commands as a part of lib tester. This removes all the mocks from the library test and job test. It also eliminates the need for the user to write `shMock` if they intend to use this library.

@zelinh - Approving the PR. However, I would revisit the timeout. I feel under no circumstance it would take 2 hrs, or even 1hr. It just blocks the node unnecessarily and therefore can be reduced

jenkins/release-tag/release-tag.jenkinsfile
gaiksaya: We can use the manifest directly from released artifact now. Would recommend to use that instead of build id. In that case just the version is required to retrieve the manifest.
Thanks!

jenkins/release-tag/release-tag.jenkinsfile
abhinavGupta16: I feel under no circumstance it would take 2 hrs, or even 1hr. It just blocks the node unnecessarily and therefore can be reduced. It's not a blocker for me, but timeouts should be visited carefully as they are the failsafes that unblock nodes in case a job is stuck

gaiksaya (approved):
Unblocking this PR merge. @zelinh is working on https://github.com/opensearch-project/opensearch-build/pull/1666#discussion_r820024036
Please feel free to merge if that is coming in another PR.
Thanks!

zelinh:
Issue is created to use public release manifest here #1710. Will be pushing this from another PR.

vars/createReleaseTag.groovy
gaiksaya: > I would skip it with a note if it's identical (someone did the right thing and manually created the tag, it's ok), and fail if the existing tag doesn't match (someone created the wrong tag, let's find out loudly).

Hi @dblock What do you think about hard fail at the end rather than failing the entire workflow before all components are parsed? 
Right now this just fails the workflow and until that component fixes the tag we cannot re-run. How about go and finish the workflow and hard fail at the end if there is a mismatch?

zelinh: More elaboration on the current status: when some component already cut the tag with different commit from the commit in the distribution manifest, the current workflow will fail immediately and ask user to fix it until it matches.

Issue #129, opened 2021-08-02
Automate performance testing for OpenSearch


gaiksaya:
Create automated jobs to run performance testing

camerski:
Implement a `PerformanceTestSuite` class and wire it in to `test.py`. This could be similar to `IntegTestSuite` if we need per-component performance tests, or it could be a bundle-level thing. In any case, the interface should be similar: Accept a test cluster endpoint, return when the tests are finished, report results in some sensible fashion.

See https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/python/test.py and https://github.com/opensearch-project/opensearch-build/blob/main/bundle-workflow/python/test_workflow/integ_test_suite.py

peternied:
See https://github.com/opensearch-project/opensearch-build/issues/583 for the remaining work, http://jenki-jenki-fpgmrv2ryxko-1366042710.us-east-1.elb.amazonaws.com/job/perf-test/ is failing, but exists

owaiskazi19:
> > This work has been completed, there are bugs and follow up tasks associated with the actual work
> 
> We have not integrated the performance testing with jenkins yet! @owaiskazi19 correct me if I am wrong.

That's correct. We have not integrated perf test with jenkins

kotwanikunal:
Taking a look into this issue.

Issue #707, opened 2021-10-06
Publish performance test results for every build 


saratvemulapalli:
Add support to publish performance test results.

kotwanikunal:
I will follow up on this once we have the basic infrastructure for #129 in place.

kotwanikunal:
Currently we have nightly performance tests running as a part of #129. The results are published to the artifact bucket post completion ([Example here](https://ci.opensearch.org/ci/dbc/perf-test/2.0.0/3810/linux/x64/tar/test-results/perf-test/with-security/1af44c88-35f4-439d-a897-27416dd95d13.html)). 
The results are unfortunately not directly discoverable. 
We have plans to publish the results publicly using the [opensearch.org](https://opensearch.org/) website in the future.

bbarani:
We are tracking the dashboard creation for viewing performance test runs [here](https://github.com/opensearch-project/opensearch-build/issues/2297)

bbarani:
You can access the published performance test results using the below html url pattern,

```
https://ci.opensearch.org/ci/dbc/perf-test/<version>/<distribution-build-number>/linux/x64/tar/test-results/<job-build-number>/perf-test/<with/without-security>/perf-test.html
```

More info: https://github.com/opensearch-project/opensearch-build/tree/main/src/test_workflow#perftest-job

PR #1671, opened 2022-02-25
Add performance test scripts

kotwanikunal:
Signed-off-by: Kunal Kotwani <kkotwani@amazon.com>

### Description
- Add Jenkins job and groovy scripts for performance tests
 
### Issues Resolved
- https://github.com/opensearch-project/opensearch-build/issues/129
 
### Check List


jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,100 @@
...
+
+                    runPerfTestScript(bundleManifest: "${BUNDLE_MANIFEST}",
+                        buildId: "${BUILD_ID}",
+                        security: false)
+
+                    if(env.HAS_SECURITY) {
+                     runPerfTestScript(bundleManifest: "${BUNDLE_MANIFEST}",
+                        buildId: "${BUILD_ID}",
+                        security: true)
+                    }
```
owaiskazi19, line 81: Can pass `env. HAS_SECURITY` and the check for security can happen in runPerfTestScript? WDYT?
Will reduce duplication.

src/run_perf_test.py
```diff
@@ -35,19 +36,29 @@ def main():
...
+                        help="Security of the cluster should be True/False",
+                        default=False)
     parser.add_argument("--keep", dest="keep", action="store_true", help="Do not delete the working temporary directory.")
     args = parser.parse_args()
 
     manifest = BundleManifest.from_file(args.bundle_manifest)
     config = yaml.safe_load(args.config)
 
+    tests_dir = os.path.join(os.getcwd(), "test-results", "perf-test")
+    os.makedirs(tests_dir, exist_ok=True)
```
owaiskazi19, line 49: @gaiksaya can we use `test_recorder` here?

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: That's on purpose. 🙂 
A bundle with security will kick off two performance tests - one with security enabled and another without.
I have parameterized it so that I can pull it out and parallelize at some point.

src/test_workflow/perf_test/perf_test_suite.py
```diff
@@ -20,16 +18,24 @@ def __init__(self, bundle_manifest, endpoint, security, current_workspace):
...
+
+        print(self.command)
+
     def execute(self):
         try:
-            with WorkingDirectory(self.work_dir):
-                dir = os.getcwd()
-                subprocess.check_call("python3 -m pipenv install", cwd=dir, shell=True)
-                subprocess.check_call("pipenv install", cwd=dir, shell=True)
+            os.chdir(os.path.join(self.current_workspace, self.work_dir))
```
owaiskazi19, line 31: Is there any reason why can't we use `WorkingDirectory` here?

src/run_perf_test.py
```diff
@@ -35,19 +36,29 @@ def main():
...
         with GitRepository(get_infra_repo_url(), "main", current_workspace):
-            security = "security" in manifest.components
             with WorkingDirectory(current_workspace):
-                with PerfTestCluster.create(manifest, config, args.stack, security, current_workspace) as (test_cluster_endpoint, test_cluster_port):
-                    perf_test_suite = PerfTestSuite(manifest, test_cluster_endpoint, security, current_workspace)
+                with PerfTestCluster.create(manifest, config, args.stack, args.security, current_workspace) \
+                        as (test_cluster_endpoint, test_cluster_port):
+                    # Stack creation returns control before user-data script execution is complete and the server starts
+                    # Sleep helps with consistent service discovery and test initialization success.
+                    time.sleep(120)
```
owaiskazi19, line 59: Still thinking about it. Do we really need this?

kotwanikunal: Hah. Trust me, we do!
It took me quite some time to figure out it was as simple as this 
The tests fail to start otherwise most of the times because the server hasn't initialized.

vars/runPerfTestScript.groovy
```diff
@@ -0,0 +1,47 @@
...
+        'perf-test',
+        args.security ? "--stack test-single-security-${args.buildId}" :
+        "--stack test-single-${args.buildId}",
+        "--bundle-manifest ${args.bundleManifest}",
+        "--config config.yml",
+        args.security ? "--security" : ""
+    ].join(' '))
+}
+
+void install_opensearch_infra_dependencies() {
```
abhinavGupta16, line 25: Let's use a docker image to avoid downloading all the dependencies.

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,100 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
+
+pipeline {
+    agent none
+    environment {
+        AWS_ROLE_ARN = "arn:aws:iam::${AWS_ACCOUNT_PUBLIC}:role/opensearch-test"
+        AWS_ROLE_SESSION_NAME = "jenkins-test-session"
```
abhinavGupta16, line 7: Where are `AWS_ROLE_SESSION_NAME` and `AWS_ROLE_ARN` variables used? 
Similar question for the parameters

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,100 @@
...
+        )
+        string(
+            defaultValue: '',
+            name: 'BUNDLE_MANIFEST_URL',
+            description: 'The bundle manifest URL, e.g. https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.2.2/98/linux/x64/builds/opensearch/manifest.yml.',
+            trim: true
+        )
+        string(
+            defaultValue: '',
+            name: 'AGENT_LABEL',
```
abhinavGupta16, line 29: Can the agent label change for this job?

abhinavGupta16 (changes requested):
Need to add tests for jenkinsfile and `vars/runPerfTestScript.groovy`. Thank you!

zelinh (commented):
You can refer to [this](https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#jenkins-pipelines-and-shared-libraries) for tests on jenkins.

src/run_perf_test.py
gaiksaya: Can be done in another PR. We have an issue opened for that here: https://github.com/opensearch-project/opensearch-build/issues/707

vars/runPerfTestScript.groovy
gaiksaya: These dependencies are not a part of any docker image and I don't think it should be. They are specific to mensor lib. This is similar to pip install for all our libraries in this repo.

abhinavGupta16: I see. What about `npm` `nodejs` `aws-cdk` etc?
https://github.com/opensearch-project/opensearch-build/blob/7affa4d4a410100431344e930d57aec042ba16c6/vars/runPerfTestScript.groovy#L33-L47

Apart from the `pipenv` installs, we should have the rest in a docker image

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,100 @@
...
+            name: 'WARMUP_ITERATIONS',
+            description: 'The agent label where the tests should be executed, e.g. Jenkins-Agent-al2-x64-c54xlarge-Docker-Host.',
+            trim: true
+        )
+    }
+
+    stages {
+        stage('perf-test') {
+            agent {
+                node {
```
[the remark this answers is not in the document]
gaiksaya: Recommending to use: `opensearchstaging/ci-runner:ci-runner-centos7-v1` docker image

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: Removed unused variables.

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: Updated it to a docker configuration.

vars/runPerfTestScript.groovy
kotwanikunal: I have removed the nodejs install. 
`aws-cdk` and `cdk-assume-role-credential-plugin` are specific to performance tests.
I think they are appropriate being in this particular script.

jenkins/opensearch/perf-test.jenkinsfile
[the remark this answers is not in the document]
kotwanikunal: Updated to use docker containers with the image `opensearchstaging/ci-runner:ci-runner-centos7-v1`

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,100 @@
...
+            post {
+                success {
+                    script {
+                        uploadTestResults(
+                                buildManifestFileName: "${BUNDLE_MANIFEST}",
+                                jobName: 'perf-test',
+                                buildNumber: "${BUILD_ID}"
+                        )
+                    }
+                    cleanWs disableDeferredWipeout: true, deleteDirs: true
```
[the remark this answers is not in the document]
kotwanikunal: Updated to use `postCleanup()`

vars/runPerfTestScript.groovy
abhinavGupta16: @kotwanikunal - I believe gcc too right?
@peterzhuamazon - what are your thoughts on this?

src/run_perf_test.py
```diff
@@ -45,13 +49,19 @@ def main():
     manifest = BundleManifest.from_file(args.bundle_manifest)
     config = yaml.safe_load(args.config)
 
+    tests_dir = os.path.join(os.getcwd(), "test-results", "perf-test")
+    os.makedirs(tests_dir, exist_ok=True)
+
     with TemporaryDirectory(keep=args.keep, chdir=True) as work_dir:
         current_workspace = os.path.join(work_dir.name, "infra")
         with GitRepository(get_infra_repo_url(), "main", current_workspace):
-            security = "security" in manifest.components
```
owaiskazi19, line 51: Shouldn't we get security from bundle manifest?

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,100 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
+
```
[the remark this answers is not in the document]
kotwanikunal: Added a timeout.

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,100 @@
...
+        AWS_ROLE_SESSION_NAME = "jenkins-test-session"
+        BUNDLE_MANIFEST = "bundle-manifest.yml"
+    }
+    tools {
+        jdk "JDK14"
+        maven "maven-3.8.2"
+    }
+    parameters {
+        string(
+            defaultValue: '',
```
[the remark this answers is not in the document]
kotwanikunal: Updated to appropriate default values

src/run_perf_test.py
kotwanikunal: That would prevent us from parallelizing the tests in the future. We want to kick off two tests for the security based bundles, which can be done from the Jenkins pipeline

src/run_perf_test.py
kotwanikunal: Updated to use a retry call with exponential backoff.

src/run_perf_test.py
owaiskazi19: I feel like security enabled/disabled should be decided based on `security` param in `manifest.components`. To parallelize the test we can pass 2 different bundle manifest, one with security enabled and other disabled.

src/test_workflow/perf_test/perf_test_suite.py
kotwanikunal: Updated to use `WorkingDirectory`

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,153 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
+
+pipeline {
+    agent none
+     options {
+        timeout(time: 10, unit: 'HOURS')
```
abhinavGupta16, line 6: Will this job take 10 hours?

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,153 @@
...
+        )
+        string(
+            name: 'BUNDLE_MANIFEST_URL',
+            description: 'The bundle manifest URL, e.g. https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.2.2/98/linux/x64/builds/opensearch/manifest.yml.',
+            trim: true
+        )
+        string(
+            defaultValue: 'nyc_taxis',
+            name: 'TEST_WORKLOAD',
+            description: 'The agent label where the tests should be executed, e.g. Jenkins-Agent-al2-x64-c54xlarge-Docker-Host.',
```
abhinavGupta16, line 27: All descriptions are the same.

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: Not for the distribution builds - but the nightly builds, yes. 8-10 hours on average

vars/runPerfTestScript.groovy
```diff
@@ -0,0 +1,45 @@
+void call(Map args = [:]) {
+    String jobName = args.jobName ?: 'distribution-build-opensearch'
+    lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+    def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: args.bundleManifest))
+    String artifactRootUrl = buildManifest.getArtifactRootUrl(jobName, args.buildId)
+
+    install_dependencies()
+    install_opensearch_infra_dependencies()
+    withAWS(role: 'opensearch-test', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
```
abhinavGupta16, line 9: Can we use `downloadFromS3.groovy` instead?

src/run_perf_test.py
kotwanikunal: We will have to publish these manifests to our buckets and distributions, which we do not currently.
Moving it out additionally gives more flexibility for extensibility moving forward for other use cases rather than updating/adding manifests for every build.

tests/jenkins/lib-testers/RunPerfTestScriptLibTest.groovy
```diff
@@ -0,0 +1,62 @@
...
+        assertThat(call.args.bundleManifest.first(), notNullValue())
+        assertThat(call.args.buildId.first(), notNullValue())
+        assertThat(call.args.security.first(), notNullValue())
+        assertThat(call.args.workload.first(), notNullValue())
+        assertThat(call.args.testIterations.first(), notNullValue())
+        assertThat(call.args.warmupIterations.first(), notNullValue())
+    }
+
+    boolean expectedParametersMatcher(call) {
+        return call.args.jobName.first().toString().equals(this.jobName)
```
abhinavGupta16, line 50: `jobname` is a nullable string

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,153 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
+
+pipeline {
+    agent none
+     options {
+        timeout(time: 10, unit: 'HOURS')
+    }
+    environment {
+        AGENT_X64 = 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
+        AGENT_IMAGE = 'opensearchstaging/ci-runner:ci-runner-centos7-v1'
```
abhinavGupta16, line 10: This can be a local variable

tests/jenkins/lib-testers/RunPerfTestScriptLibTest.groovy
```diff
@@ -0,0 +1,62 @@
...
+
+    }
+
+    void parameterInvariantsAssertions(call) {
+        assertThat(call.args.bundleManifest.first(), notNullValue())
+        assertThat(call.args.buildId.first(), notNullValue())
+        assertThat(call.args.security.first(), notNullValue())
+        assertThat(call.args.workload.first(), notNullValue())
+        assertThat(call.args.testIterations.first(), notNullValue())
+        assertThat(call.args.warmupIterations.first(), notNullValue())
```
abhinavGupta16, line 46: Can add checks to make sure the value is a number. Same for all relevant params

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,153 @@
...
+        )
+        string(
+            defaultValue: 'nyc_taxis',
+            name: 'TEST_WORKLOAD',
+            description: 'The agent label where the tests should be executed, e.g. Jenkins-Agent-al2-x64-c54xlarge-Docker-Host.',
+            trim: true
+        )
+        string(
+            defaultValue: '1',
+            name: 'TEST_ITERATIONS',
```
abhinavGupta16, line 32: What will happen if I enter a string instead of a number for this? Will `test.sh` fail and fail the job? If no, we need checks so the node is not stuck, else we are good.

jenkins/opensearch/perf-test.jenkinsfile
abhinavGupta16: We need to then increase the time if the expected time is 10hrs, since this will fail the job after timeout

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: Noticed that. Was already on it 🙂 
Updated. Resolving this comment

vars/runPerfTestScript.groovy
abhinavGupta16: thanks!

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,153 @@
...
+                        buildId: "${BUILD_ID}",
+                        security: false,
+                        workload: TEST_WORKLOAD,
+                        testIterations: TEST_ITERATIONS,
+                        warmupIterations: WARMUP_ITERATIONS)
+
+                    lib.jenkins.Messages.new(this).add('perf-test', "[Test]Performance tests without security for ${BUILD_ID} completed")
+
+                    if(env.HAS_SECURITY) {
+                         runPerfTestScript(bundleManifest: "${BUNDLE_MANIFEST}",
```
abhinavGupta16, line 92: Given the amount of time each run takes, can we run this in parallel?

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: We are still narrowing down on the exact numbers which we will know after the test runs. Currently we are looking at 8 hours. I can increase it later after getting in more datapoints.

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: I will parallelize this going forward. We still need a few test runs to iron out all the time outliers and other execution issues, if any.

vars/runPerfTestScript.groovy
kotwanikunal: The `ARTIFACT_DOWNLOAD_ROLE_NAME` configuration does not have access to the bucket path where we currently have the configuration. This serves as a good starting point for the POC runs that will be executed until all the details and locations are finalized.

jenkins/opensearch/perf-test.jenkinsfile
abhinavGupta16: Do we plan to do it in this PR?

vars/runPerfTestScript.groovy
abhinavGupta16: What is the bucket path?

jenkins/opensearch/perf-test.jenkinsfile
abhinavGupta16: We'll definitely would want to revisit this. It would increase in case the job is run serially, but would decrease if run in parallel.

src/run_perf_test.py
gaiksaya: We do publish manifests to our buckets. `without-security` tests is something that will run irrespective of what's in the manifest but for `with-security` can we add a check in the codebase and then start the test only if the security component is present?
If the performance tests are to be run nightly it is highly possible that security component is not present initially in the manifest.

jenkins/opensearch/perf-test.jenkinsfile
CEHENKLE: At this point I'd recommend getting this PR in, and then cut a separate PR.  I'd much rather get in a solid base and iterate than hold for a big PR with everything in it.  Remember: Progress, not perfection! :)

gaiksaya:
@kotwanikunal Looks like there are failing tests. Can you fix those?
Thanks!

jenkins/opensearch/perf-test.jenkinsfile
abhinavGupta16: Sure, let's parallelize the job in a different PR. @kotwanikunal - Can you open an issue to parallelize this job?

Let's fix the failing tests and address the other review comments fixing the new tests that are added before we get this in.

Thanks!

src/run_perf_test.py
```diff
@@ -35,6 +36,9 @@ def main():
     parser.add_argument("--bundle-manifest", type=argparse.FileType("r"), help="Bundle Manifest file.", required=True)
     parser.add_argument("--stack", dest="stack", help="Stack name for performance test")
     parser.add_argument("--config", type=argparse.FileType("r"), help="Config file.", required=True)
+    parser.add_argument("--security", dest="security", action="store_true",
```
dblock, line 39: We already have a configuration file, and we have a manifest that has or does not have security, why are we promoting something so specific to a top level feature/option and how is it going to add up with that?

src/test_workflow/perf_test/perf_test_suite.py
```diff
@@ -23,16 +22,20 @@ def __init__(self, bundle_manifest, endpoint, security, current_workspace, args)
             f" --warmup-iters {self.args.warmup_iters} --test-iters {self.args.test_iters}"
         )
 
+        if test_results_path is not None:
+            self.command = (
+                f"pipenv run python test_config.py -i {self.endpoint} -b {self.manifest.build.id}"
```
dblock, line 27: This is Python code calling python code? It should import the code and call it directly.

vars/runPerfTestScript.groovy
```diff
@@ -0,0 +1,45 @@
...
+    ].join(' '))
+}
+
+boolean isNullOrEmpty(String str) { return (str == null || str.allWhitespace) }
+
+void install_opensearch_infra_dependencies() {
+    sh'''
+        pipenv install "dataclasses_json~=0.5" "aws_requests_auth~=0.4" "json2html~=1.3.0"
+        pipenv install "aws-cdk.core~=1.143.0" "aws_cdk.aws_ec2~=1.143.0" "aws_cdk.aws_iam~=1.143.0"
+        pipenv install "boto3~=1.18" "setuptools~=57.4" "retry~=0.9"
```
dblock, line 33: This shouldn't be necessary. Dependencies needed by the tools should go into Pipfile.

dblock (changes requested):
Add tests for python code, too.

src/test_workflow/perf_test/perf_test_suite.py
kotwanikunal: The code is in a separate package which is not yet open to the public. We pull down the code from the repo and execute it.
Do you mean using `importlib` or something similar for starting the execution?

vars/runPerfTestScript.groovy
kotwanikunal: The is because of the nature in which the packages are separated and it being pulled at run time. `pipenv` does not support nested module pipfile installation, which is why I had to resort to installing it via this script.

src/run_perf_test.py
kotwanikunal: I am trying to parallelize the two jobs (security/non security) at the top level to get clear status and result visibility. Moving it at the python script level will require additional async logic to be added in to make the script execute in parallel.

src/run_perf_test.py
kotwanikunal: Refer to the comment below - https://github.com/opensearch-project/opensearch-build/pull/1671#discussion_r828477283

What I was trying to say is - we have a single manifest which might or might not have security in the components. By enabling the logic at the Jenkins level, we can execute the two runs in parallel, giving us more control over the test runs and statuses. 

Also, it is based directly on the manifest. Look here for more - https://github.com/opensearch-project/opensearch-build/pull/1671/files#diff-4debf5e3ece07145d8395f15df88f49b8b784a274cead94e2a94c8b7152c11efR75

All I am doing is pulling out that logic for better control at the top level of job execution.

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: Follow the thread below for more details - https://github.com/opensearch-project/opensearch-build/pull/1671#discussion_r828480396
Resolving this comment.

vars/runPerfTestScript.groovy
kotwanikunal: It's stored under the Jenkins configuration. `bundles/tests/perf-test-config`

src/run_perf_test.py
dblock: The tooling is trying hard not to be too specific for OpenSearch/OpenSearch Dashboards so I'd think about this twice. My concern though is that with this change we have 2 ways to say "with security" and "without security", you should collapse it in 1 way as part of this change.

vars/runPerfTestScript.groovy
dblock: You should fix this in the other project. Implement a similar wrapper as `test.sh`, a `Pipfile` and have the .sh script run `pipenv install` to get these dependencies. It's not on the caller's responsibility to ensure that the dependencies are met, or you'll be constantly chasing changes in that project here.

src/run_perf_test.py
kotwanikunal: I have updated the config to be more appropriate with the general use case. I can work on abstracting it out/utilizing a test config for example, when we move to add more components to this script.

vars/runPerfTestScript.groovy
kotwanikunal: I can work on this in a follow up PR. I want to get the changes going as they are blocking other performance test related integrations and tests.

kotwanikunal:
The tests succeed on my machine locally -
```
(opensearch-build) kkotwani@3c22fbb2c92e opensearch-build % git commit -m "Add parallel stages for performance tests, add tests" -s
isort....................................................................Passed
flake8...................................................................Passed
mypy.....................................................................Passed
pytest...................................................................Passed
yamllint.................................................................Passed

```

Looks to be an import issue on the GHA distros - 
```
Run pipenv run flake[8](https://github.com/opensearch-project/opensearch-build/runs/5597195180?check_suite_focus=true#step:6:8) .
  File "/home/runner/.local/share/virtualenvs/opensearch-build-ucxYjMk8/bin/flake8", line 5, in <module>
    from flake8.main.cli import main
  File "/home/runner/.local/share/virtualenvs/opensearch-build-ucxYjMk8/lib/python3.7/site-packages/flake8/main/cli.py", line 5, in <module>
    from flake8.main import application
  File "/home/runner/.local/share/virtualenvs/opensearch-build-ucxYjMk8/lib/python3.7/site-packages/flake8/main/application.py", line [11](https://github.com/opensearch-project/opensearch-build/runs/5597195180?check_suite_focus=true#step:6:11), in <module>
    from flake8 import checker
  File "/home/runner/.local/share/virtualenvs/opensearch-build-ucxYjMk8/lib/python3.7/site-packages/flake8/checker.py", line 18, in <module>
    from flake8 import processor
  File "/home/runner/.local/share/virtualenvs/opensearch-build-ucxYjMk8/lib/python3.7/site-packages/flake8/processor.py", line 13, in <module>
    from flake8 import utils
  File "/home/runner/.local/share/virtualenvs/opensearch-build-ucxYjMk8/lib/python3.7/site-packages/flake8/utils.py", line 16, in <module>
    from flake8._compat import lru_cache
  File "/home/runner/.local/share/virtualenvs/opensearch-build-ucxYjMk8/lib/python3.7/site-packages/flake8/_compat.py", line [12](https://github.com/opensearch-project/opensearch-build/runs/5597195180?check_suite_focus=true#step:6:12), in <module>
    import importlib_metadata
ModuleNotFoundError: No module named 'importlib_metadata'
```

dblock:
Python version mismatch? https://github.com/python-poetry/poetry/issues/1487

kotwanikunal:
> Python version mismatch? Is yours passing on 3.8 vs. 3.7? [python-poetry/poetry#1487](https://github.com/python-poetry/poetry/issues/1487)

Thanks!
That seemed to be it. My venv was on 3.7 but it picked up the OS default which was 3.8.
And since the `Pipfile.lock` was updated, it was breaking with 3.7 on the build envs.

jenkins/opensearch/perf-test.jenkinsfile
abhinavGupta16: @kotwanikunal - Can you please confirm? Thank you

jenkins/opensearch/perf-test.jenkinsfile
abhinavGupta16: @kotwanikunal - Can you please post the issue link here after creating it so it can be connected and will be easy to trace? Thank you!

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: I have already updated the Jenkins check to validate its a number. But I also did verify the test fails if we execute the script with a non number otherwise.

jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -0,0 +1,217 @@
...
+                            env.HAS_SECURITY = bundleManifestObj.components.containsKey("security")
+                            env.ARCHITECTURE = bundleManifestObj.getArtifactArchitecture()
+                            echo "BUNDLE_MANIFEST: ${BUNDLE_MANIFEST}"
+                            echo "BUILD_ID: ${BUILD_ID}"
+                            echo "BUILD_ID: ${BUILD_ID}"
+                            echo "Security: ${HAS_SECURITY}"
+                            echo "Architecture: ${ARCHITECTURE}"
+
+                            lib.jenkins.Messages.new(this).add('perf-test', "Performance tests for #${BUILD_ID}")
+                            if (env.HAS_SECURITY) {
```
abhinavGupta16, line 95: This stage will not do anything if Has_Security is false. Can we instead add it as a condition to the stage so that this stage is skipped instead?  For security false, this would run this stage with success which will be misleading.

abhinavGupta16 (changes requested):

jenkins/opensearch/perf-test.jenkinsfile
abhinavGupta16: great!

jenkins/opensearch/perf-test.jenkinsfile
abhinavGupta16: Thanks!

abhinavGupta16:
Can you also add a test case for the jenkins job? Thank you!

bbarani:
@kotwanikunal Can you please take a look at the comments and provide an update?

kotwanikunal:
> Can you also add a test case for the jenkins job?
> 
> Refer [this](https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#tests-for-jenkins-job)
> 
> Thank you!

There are test for both the scenarios now. The dependency on downloading the manifest file was eliminated and instead I have used hard coded values, which makes it on par with the actual test script.

jenkins/opensearch/perf-test.jenkinsfile
kotwanikunal: Updated the code to reflect this.

kotwanikunal:
> @kotwanikunal Can you please take a look at the comments and provide an update?

All the pending change requests have been resolved.

abhinavGupta16:
> > Can you also add a test case for the jenkins job?
> > Refer [this](https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#tests-for-jenkins-job)
> > Thank you!
> 
> There are test for both the scenarios now. The dependency on downloading the manifest file was eliminated and instead I have used hard coded values, which makes it on par with the actual test script.

We need a test for the jenkinsfile, to make sure that it fails if there are any changes to the job. It would also make sure that the job calls the libraries correctly.

abhinavGupta16:
> > > > Can you also add a test case for the jenkins job?
> > > > Refer [this](https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#tests-for-jenkins-job)
> > > > Thank you!
> > > 
> > > 
> > > There are test for both the scenarios now. The dependency on downloading the manifest file was eliminated and instead I have used hard coded values, which makes it on par with the actual test script.
> > 
> > 
> > We would still need a test for the jenkinsfile, to make sure that it fails if there are any changes to the job. It would also make sure that the job calls the libraries correctly. Example [SignStandaloneArtifactsJob](https://github.com/opensearch-project/opensearch-build/blob/main/tests/jenkins/TestSignStandaloneArtifactsJob.groovy), [DataPrepperJob](https://github.com/opensearch-project/opensearch-build/blob/main/tests/jenkins/TestDataPrepperReleaseArtifacts.groovy)
> 
> I'll add and update the PR shortly.

Let me know whenever you add it. Thanks!

cc: @bbarani

kotwanikunal:
> > > > > Can you also add a test case for the jenkins job?
> > > > > Refer [this](https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#tests-for-jenkins-job)
> > > > > Thank you!
> > > > 
> > > > 
> > > > There are test for both the scenarios now. The dependency on downloading the manifest file was eliminated and instead I have used hard coded values, which makes it on par with the actual test script.
> > > 
> > > 
> > > We would still need a test for the jenkinsfile, to make sure that it fails if there are any changes to the job. It would also make sure that the job calls the libraries correctly. Example [SignStandaloneArtifactsJob](https://github.com/opensearch-project/opensearch-build/blob/main/tests/jenkins/TestSignStandaloneArtifactsJob.groovy), [DataPrepperJob](https://github.com/opensearch-project/opensearch-build/blob/main/tests/jenkins/TestDataPrepperReleaseArtifacts.groovy)
> > 
> > 
> > I'll add and update the PR shortly.
> 
> Let me know whenever you add it. Thanks!
> 
> cc: @bbarani

Took longer than expected to get all the cases in. All the cases and jobs should have relevant tests now.

abhinavGupta16:
> > > > > > Can you also add a test case for the jenkins job?
> > > > > > Refer [this](https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#tests-for-jenkins-job)
> > > > > > Thank you!
> > > > > 
> > > > > 
> > > > > There are test for both the scenarios now. The dependency on downloading the manifest file was eliminated and instead I have used hard coded values, which makes it on par with the actual test script.
> > > > 
> > > > 
> > > > We would still need a test for the jenkinsfile, to make sure that it fails if there are any changes to the job. It would also make sure that the job calls the libraries correctly. Example [SignStandaloneArtifactsJob](https://github.com/opensearch-project/opensearch-build/blob/main/tests/jenkins/TestSignStandaloneArtifactsJob.groovy), [DataPrepperJob](https://github.com/opensearch-project/opensearch-build/blob/main/tests/jenkins/TestDataPrepperReleaseArtifacts.groovy)
> > > 
> > > 
> > > I'll add and update the PR shortly.
> > 
> > 
> > Let me know whenever you add it. Thanks!
> > cc: @bbarani
> 
> Took longer than expected to get all the cases in. All the cases and jobs should have relevant tests now.

Thanks for adding the tests. However, we don't want to create a copy for the job, but rather use the original jenkinsfile to make sure that works. If the test is written on the copy, then any change on the original job will not fail the test and therefore, defeat it's purpose. Let's connect offline so we can close this soon.

Thanks!

kotwanikunal:
> > > > > > > Can you also add a test case for the jenkins job?
> > > > > > > Refer [this](https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#tests-for-jenkins-job)
> > > > > > > Thank you!
> > > > > > 
> > > > > > 
> > > > > > There are test for both the scenarios now. The dependency on downloading the manifest file was eliminated and instead I have used hard coded values, which makes it on par with the actual test script.
> > > > > 
> > > > > 
> > > > > We would still need a test for the jenkinsfile, to make sure that it fails if there are any changes to the job. It would also make sure that the job calls the libraries correctly. Example [SignStandaloneArtifactsJob](https://github.com/opensearch-project/opensearch-build/blob/main/tests/jenkins/TestSignStandaloneArtifactsJob.groovy), [DataPrepperJob](https://github.com/opensearch-project/opensearch-build/blob/main/tests/jenkins/TestDataPrepperReleaseArtifacts.groovy)
> > > > 
> > > > 
> > > > I'll add and update the PR shortly.
> > > 
> > > 
> > > Let me know whenever you add it. Thanks!
> > > cc: @bbarani
> > 
> > 
> > Took longer than expected to get all the cases in. All the cases and jobs should have relevant tests now.
> 
> Thanks for adding the tests. However, we don't want to create a copy for the job, but rather use the original jenkinsfile to make sure that works (as in the examples and readme). If the test is written on the copy, then any change on the original job will not fail the test and therefore, defeat it's purpose. Let's connect offline so we can close this soon.
> 
> Thanks!

Sorry took me a while to realize that and then get the actual script to have all the testable stubs and mocks. Jenkins/Python newbie here. 🙂 
Let me know if this resolves your concerns. 
Thanks!

abhinavGupta16 (approved):
Thank you for addressing the review.

src/run_perf_test.py
```diff
@@ -45,14 +49,22 @@ def main():
...
         current_workspace = os.path.join(work_dir.name, "infra")
         with GitRepository(get_infra_repo_url(), "main", current_workspace):
-            security = "security" in manifest.components
             with WorkingDirectory(current_workspace):
-                with PerfTestCluster.create(manifest, config, args.stack, security, current_workspace) as (test_cluster_endpoint, test_cluster_port):
-                    perf_test_suite = PerfTestSuite(manifest, test_cluster_endpoint, security, current_workspace, args)
-                    perf_test_suite.execute()
+                with PerfTestCluster.create(manifest, config, args.stack, security, current_workspace) \
+                        as (test_cluster_endpoint, test_cluster_port):
+                    perf_test_suite = PerfTestSuite(manifest, test_cluster_endpoint, security,
```
dblock, line 65: Nit:  

```
.create(
    manifest,
    config,
    ...
) as ...
```

A bit cleaner.

src/run_perf_test.py
```diff
@@ -45,14 +49,22 @@ def main():
...
+                                                    current_workspace, tests_dir, args)
+                    retry_call(perf_test_suite.execute, tries=3, delay=60, backoff=2)
```
dblock, line 67: Should we expose retry options to the command line?

src/test_workflow/perf_test/perf_test_suite.py
```diff
@@ -23,16 +22,22 @@ def __init__(self, bundle_manifest, endpoint, security, current_workspace, args)
             f" --warmup-iters {self.args.warmup_iters} --test-iters {self.args.test_iters}"
         )
 
+        if test_results_path is not None:
```
dblock, line 25: When is it `None`?

dblock (approved):

src/run_perf_test.py
kotwanikunal: Updated.

src/run_perf_test.py
```diff
@@ -45,14 +49,22 @@ def main():
     manifest = BundleManifest.from_file(args.bundle_manifest)
     config = yaml.safe_load(args.config)
 
+    security = "security" in manifest.components and not args.insecure
+    if security:
+        tests_dir = os.path.join(os.getcwd(), "test-results", "perf-test", "with-security")
+    else:
+        tests_dir = os.path.join(os.getcwd(), "test-results", "perf-test", "without-security")
```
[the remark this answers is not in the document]
kotwanikunal: Updated.

src/run_perf_test.py
```diff
@@ -35,6 +36,9 @@ def main():
     parser.add_argument("--bundle-manifest", type=argparse.FileType("r"), help="Bundle Manifest file.", required=True)
     parser.add_argument("--stack", dest="stack", help="Stack name for performance test")
     parser.add_argument("--config", type=argparse.FileType("r"), help="Config file.", required=True)
+    parser.add_argument("--force-insecure-mode", dest="insecure", action="store_true",
```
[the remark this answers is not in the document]
kotwanikunal: Updated.

src/test_workflow/perf_test/perf_test_suite.py
kotwanikunal: Updated.

src/run_perf_test.py
kotwanikunal: That seems like a good option to have. I do have a few things to follow up on post based on the test run outcomes once everything is up and running. 
Added an issue here for tracking - https://github.com/opensearch-project/OpenSearch/issues/2718

dblock (approved):

dblock:
Merge on green.

kotwanikunal:
GA Run: 
```
62 tests completed, 2 failed, 1 skipped
```

Tests running successfully locally. 

Can someone from @opensearch-project/engineering-effectiveness help with this? 🙂

dblock:
The success message in `TestRunPerfTestScript > testRunPerfTestScript_Pipeline` looks different between expected and not, probably a leaking ENV variable across tests that causes different output.

Diff the failed test output in the log:

```
<                           publishNotification.sh(curl -XPOST --header "Content-Type: application/json" --data '{"result_text":":white_check_mark: perf-test [1236] Performance Tests Successful
<      Build: test://artifact.url
---
>                           publishNotification.sh(curl -XPOST --header "Content-Type: application/json" --data '{"result_text":":white_check_mark:
>      JOB_NAME=perf-test
>      BUILD_NUMBER=[1236]
>      MESSAGE=Performance Tests Successful
>      BUILD_URL: test://artifact.url
>      MANIFEST: null
146,148d149
<                     perf-test.postCleanup()
<                        postCleanup.cleanWs({disableDeferredWipeout=true, deleteDirs=true})
<      "
```

PR #1676, opened 2022-03-01
[RPM M1] Combine two map into one for distribution class/extension

peterzhuamazon:

### Description
Combine two map into one for distribution class/extension
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/pull/1659#discussion_r813427771
 
### Check List

### DIFF
--- a/src/assemble_workflow/bundle_recorder.py
+++ b/src/assemble_workflow/bundle_recorder.py
@@ -41,7 +41,7 @@ class BundleRecorder:
             build.platform,
             build.architecture,
         ]
-        extension = Dists().DISTRIBUTIONS_MAP[self.distribution][1] if self.distribution else Dists().DISTRIBUTIONS_MAP['tar'][1]
+        extension = Dists.DISTRIBUTIONS_MAP[self.distribution]['extension'] if self.distribution else Dists().DISTRIBUTIONS_MAP['tar']['extension']
         return "-".join(parts) + extension
 
     # Assembled output are expected to be served from a separate "dist" folder
--- a/src/assemble_workflow/bundle_recorder.py
+++ b/src/assemble_workflow/bundle_recorder.py
@@ -41,7 +41,7 @@ class BundleRecorder:
             build.platform,
             build.architecture,
         ]
-        extension = Dists.DISTRIBUTIONS_MAP[self.distribution]['extension'] if self.distribution else Dists().DISTRIBUTIONS_MAP['tar']['extension']
+        extension = Dists.DISTRIBUTIONS_MAP[self.distribution].extension if self.distribution else Dists.DISTRIBUTIONS_MAP['tar'].extension
         return "-".join(parts) + extension
 
     # Assembled output are expected to be served from a separate "dist" folder
--- a/src/assemble_workflow/bundle_recorder.py
+++ b/src/assemble_workflow/bundle_recorder.py
@@ -8,15 +8,12 @@ import os
 from typing import Any, Dict
 
 from assemble_workflow.bundle_location import BundleLocation
+from assemble_workflow.dists import Dists
 from manifests.build_manifest import BuildComponent, BuildManifest
 from manifests.bundle_manifest import BundleManifest
 
 
 class BundleRecorder:
-    EXTENSIONS = {
-        "tar": ".tar.gz",
-        "zip": ".zip",
-    }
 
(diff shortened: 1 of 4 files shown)


peterzhuamazon:
@dblock @peternied is there a way to skip mypy for these `errors`:
```

src/assemble_workflow/dists.py:25: error: Returning Any from function declared to return "Dist"  [no-any-return]
src/assemble_workflow/dists.py:25: error: "object" not callable  [operator]
tests/tests_assemble_workflow/test_dists.py:19: error: "object" has no attribute "__name__"; maybe "__ne__" or "__new__"?  [attr-defined]
tests/tests_assemble_workflow/test_dists.py:21: error: "object" has no attribute "__name__"; maybe "__ne__" or "__new__"?  [attr-defined]
src/assemble_workflow/bundle_recorder.py:45: error: Unsupported operand types for + ("str" and "object")  [operator]
```

They are not errors at all and the code is running fine.
I am not sure why mypy is able to detect these and call it error out.

Thanks.

peterzhuamazon:
I have some research and seems like mypy in strict mode is even complaining different types within the same list.
In python, I think you can pretty must have mixed type of vars inside a list.
And this should be allowed, while mypy is still complaining it.

src/assemble_workflow/bundle_recorder.py
```diff
@@ -44,7 +41,7 @@ def __get_package_name(self, build: BuildManifest.Build) -> str:
             build.platform,
             build.architecture,
         ]
-        extension = self.EXTENSIONS[self.distribution] if self.distribution else self.EXTENSIONS['tar']
+        extension = Dists().DISTRIBUTIONS_MAP[self.distribution][1] if self.distribution else Dists().DISTRIBUTIONS_MAP['tar'][1]
```
dblock, line 44: This creates a new instance of `Dists()`, which shouldn't be necessary since it's a constant. `Dists.DISTRIBUTION_MAPS` should work.

src/assemble_workflow/dists.py
```diff
@@ -10,9 +10,10 @@
 
 
 class Dists:
+
     DISTRIBUTIONS_MAP = {
-        "tar": DistTar,
-        "zip": DistZip,
+        "tar": [DistTar, ".tar.gz"],
```
dblock, line 15: The array causes you to do `DISTRIBUTIONS_MAP['tar'][0]` which doesn't really explain what this is. Turn it into a dictionary of dictionaries:

```python
"tar": {
    'klass': DistTar,
    'extension': ".tar.gz"
}
```

Access via `DISTRIBUTIONS_MAP['tar']['klass']`.

You can go further and make this dictionary a class and do

```python
DISTRIBUTIONS_MAP['tar'].klass`
```

dblock (changes requested):

peternied (commented):
If you follow DB's suggestion it will resolve all of the mypy errors you see.

Further, even when code functionally works, mypy is indicating that it is working outside of type safety.  Sometimes this is making sure types are declared, other times its an indication of a logic gap in the code that should be filled with a class/interface.

src/assemble_workflow/bundle_recorder.py
peterzhuamazon: Will try, thanks.

src/assemble_workflow/dists.py
peterzhuamazon: Thanks for the details I will do some test on this.
Thanks.

peterzhuamazon:
Hi @dblock please review again as I have resolved all the issues.
Thanks.

peterzhuamazon:
> If you follow DB's suggestion it will resolve all of the mypy errors you see.
> 
> Further, even when code functionally works, mypy is indicating that it is working outside of type safety. Sometimes this is making sure types are declared, other times its an indication of a logic gap in the code that should be filled with a class/interface.

Seems like new changes does not stop mypy from sending errors.

dblock (approved):
Nit: looks like Python convention is to use `cls` vs. `klass`. Since we have both here, maybe make them `cls` everywhere?

peternied (approved):

peterzhuamazon:
After reading through some documents, since I am trying to mixed the type within a dictionary to assign klass.
A dynamic class can define __eq__ that returns a string, there is no guarantee it is a Dist in return.
I would keep the old way of 2 dicts for now, after spending some time looking at TypedDict, with still Any as the return.

Thanks.


> Nit: looks like Python convention is to use `cls` vs. `klass`. Since we have both here, maybe make them `cls` everywhere?

Fixed now.

Issue #1682, opened 2022-03-01
[Question] Bundle vs Dependency JDK in OpenSearch RPM


peterzhuamazon:
There are several different benefits of bundling a jdk inside RPM vs treat jdk as dependency during install time.

* Bundle:
  * Pros:
    * No need to configure JDK env, opensearch will use the bundled jdk as default
  * Cons:
    * Problem is, every time when you upgrade opensearch jdk is upgraded alongside the opensearch
    * Many companies likes to use a specific version of JDK themselves

* Dependency:
  * Pros:
    * When upgrade opensearch, JDK will stay unless there are changes in new version dependency
  * Cons:
    * During installation time, if our repo does not contain that jdk (corretto? adoptopenjdk?), installation fail, unless user manually install the specific version we demand, or they add the required repo to their yum.repo.d folder
    * User needs to define OPENSEARCH_JAVA_HOME to change to other jdk and rpm will define it by default to ensure OpenSearch can start
    * This seems not too much different compares to bundled jdk, as it not only complicated things, as well as requires manual setups from users if they dont like the default value

I want to see if there is any more thoughts on this from you guys.

Thanks.

@dblock @spotaws

peterzhuamazon:
Related to https://github.com/opensearch-project/OpenSearch/issues/2302.

dblock:
I prefer a non-bundled JDK in RPM, but I would only take a dependency on the same version as tested in the suite. See https://github.com/opensearch-project/project-website/pull/670 for a breakdown across versions (note the actual list is in flux).

spotrh:
Strong preference towards **not** bundling and locking to a specific jdk version. You should also provide corretto packages in your repo to meet the dependency.

Additionally, OPENSEARCH_JAVA_HOME can be set in the systemd config, which is how the majority of users consuming opensearch via RPM will use it (it is also where they can override it to use a different JDK).

I disagree with your assertion that this is "not too much different compares to bundled jdk". It dramatically decreases the size of the opensearch RPM (which is what will be updating in the majority of transactions). We have Corretto packages at our disposal which are already supported by AWS (and prebuilt) to ensure a running JDK is present at all times and a simple mechanism (1 line change) to have the service use a different JDK.

peterzhuamazon:
> Strong preference towards **not** bundling and locking to a specific jdk version. You should also provide corretto packages in your repo to meet the dependency.
> 
[...]

Issue #27, opened 2021-05-12
[Meta] OpenSearch / OpenSearch Dashboards - Support for RPM distribution


bbarani:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Provide signed RPM artifact for OpenSearch / OpenSearch Dashboards 
- [x] discuss approach to release RPM distribution
- [x] document the approach in a release documentation https://github.com/opensearch-project/opensearch-build/issues/956
- [ ] Publish the proposed based design of RPM with new requirements and execution plan
- [x] #1117
  * https://github.com/opensearch-project/opensearch-build/pull/1443
  * Assemble process will have new param `--distribution` to specify which distribution to build
  * Uncouple distribution creation with min tarball/zip file extension
  * RPM will be built based on min tar + plugin zip installation + fpm bundle pkg
  * User should be able to use `./assemble builds/manifests.yml --distribution rpm` to build rpm and get a rpm file in return for installation.
  * We design it to run on a specific Docker Image with ruby, rpm, fpm installed, not able to run directly on macOS directly. We will make changes to the docker files as well
  * Part of the pre/post scripts and sysvinit/systemd will be taken from ODFE.
  * Also need these scripts specifically for pkg from PA: https://github.com/opensearch-project/performance-analyzer/tree/main/packaging
- [x] We will add the documentation to our design docs and propose to the community
    * https://github.com/opensearch-project/opensearch-build/issues/1452
- [ ] #1244 

------
#### [Milestone 1] OpenSearch build process design for RPM generation
- [x] #1543
- [x] #1544
- [x] #1545
#### [Milestone 2] RPM Package Signing
- [ ] #1547
#### [Milestone 3] RPM Package Design
- [x] #1548
#### [Milestone 4] RPM Package Publishing
- [ ] #1549
  * https://github.com/opensearch-project/opensearch-build/issues/2014
- [ ] #1550
#### [Milestone 5] RPM Package installation
- [ ] #1551
- [ ] #1552
- [ ] #1553
#### [Milestone 6] RPM Validation
- [x] #1554
- [x] #1555
- [ ] #1556
- [x] #1557
- [ ] #1558
- [ ] #1951

**Describe alternatives you've considered**
Generating the artifact using the source code 

**Additional context**
Currently there are no RPM artifacts for OpenSearch / OpenSearch Dashboards

justchris1:
[...]

PR #1726, opened 2022-03-09
[RPM M1] Add rpm assemble/bundle code for opensearch using rpmbuild

peterzhuamazon:
### Description
Add rpm assemble/bundle code for opensearch using rpmbuild.
Adding rpm specific block so that cpio is used during extraction, and rpmbuild is used during build.
A specific opensearch spec file is provided, dashboards one will be committed in another PR.
 
### Issues Resolved
Part of #1545
 
### Check List


peterzhuamazon:
Hi @dblock please review the code and I will add test cases later.

Thanks.

scripts/components/OpenSearch-Dashboards/install.sh
```diff
@@ -66,11 +70,22 @@ fi
...
     DIR="$(dirname "$0")"
     echo $DIR
     cd $DIR
     cp ../../../config/opensearch_dashboards.yml "$OUTPUT/config/"
 )
+elif [ "$DISTRIBUTION" = "rpm" ]; then
+(   
+    DIR="$(dirname "$0")"
+    echo $DIR
+    cd $DIR
```
dblock, line 87: Why is this needed? I don't expect scripts to change current directory. 

If there's a good reason, the `DIR` part is common and should be taken out of the `if`.

dblock (commented):
Needs passing tests and documentation updates.

scripts/components/OpenSearch-Dashboards/install.sh
peterzhuamazon: I forget to make it common as I was copy paste during testing, will change.

scripts/components/OpenSearch-Dashboards/install.sh
```diff
@@ -72,20 +72,16 @@ fi
 [ -z "$ARCHITECTURE" ] && ARCHITECTURE=`uname -m`
 [ -z "$DISTRIBUTION" ] && DISTRIBUTION="tar"
 
+# Make sure the cwd is where the script is located
+DIR="$(dirname "$0")"
+echo $DIR
+cd $DIR
```
dblock, line 78: So, why is this necessary?

peterzhuamazon: If this is not there we cannot find the script folder location.
If you call install script outside of its located folder, it will default cwd to other folder.

scripts/components/OpenSearch-Dashboards/install.sh
peterzhuamazon: fixed.

scripts/components/OpenSearch-Dashboards/install.sh
dblock: Changing directories is generally not a good pattern, because it relies on that location. You could use `$DIR/script.sh` to call it for example. The script should probably be agnostic of where it's called from. 

It's not a huge deal, but generally having to `cd` hides bugs in my experience.

peterzhuamazon: This is due to the fact we are using relative paths right now for the copy.
A lot of the scripts we have right now is relying on cd.
Will definitely worth revisit this later.

Thanks.

dblock: Where? Which scripts? I cannot find any instances of such.

Even if that's the case, can we please call these scripts from the right directory as a workaround instead of "fixing" it this way? I am worried this propagates into dozens of scripts and becomes the default pattern that we can never undo. I am 100% sure no other build.sh/install.sh does this and we're going to have to change all of them.

dblock (changes requested):
The `cd` worries me. Can we merge without that?

scripts/components/OpenSearch-Dashboards/install.sh
peterzhuamazon: @dblock I did not add these lines, these are existing lines and we have been using this method since ODFE.

Examples:
https://github.com/opensearch-project/opensearch-build/blob/main/scripts/opensearch-onetime-setup.sh#L12
https://github.com/opensearch-project/opensearch-build/blob/main/scripts/deploy_single_tar_cluster.sh#L18
https://github.com/opensearch-project/opensearch-build/blob/main/scripts/components/OpenSearch-Dashboards/install.sh#L72-L74

peterzhuamazon:
@spotrh would you mind checking the spec file in this PR?

Thanks.

Note: as of now we are using bundled JDK per https://github.com/opensearch-project/opensearch-build/issues/1682#issuecomment-1059525252

And we are also thinking about using Adopt per community suggestions: https://github.com/opensearch-project/OpenSearch/issues/2312#issuecomment-1057524268

dblock (approved):

zelinh (changes requested):
Block this PR for now until Spot review the spec files.

spotrh (approved):
Looks very good to me. I don't understand the tests well enough here to be sure, but I don't think there is any CI integration to confirm that the RPM was properly generated (or generated at all). Not a blocker for this, but probably something you want to do.

peterzhuamazon:
> Looks very good to me. I don't understand the tests well enough here to be sure, but I don't think there is any CI integration to confirm that the RPM was properly generated (or generated at all). Not a blocker for this, but probably something you want to do.

Yes @spotrh we have plans to add tests and verifications here: #27 
See milestone 5 / 6 about the steps we plan to do.

Thanks.

zelinh (dismissed):
Block this one from merging since we haven't generated the artifacts candidates for 1.3.0 release.

Issue #1632, opened 2022-02-16
Support additional qualifiers in OpenSearch and OpenSearch Dashboards


bbarani:
### Is your feature request related to a problem? Please describe

Currently the distribution build process doesn't support passing any additional qualifier other than the 3 digit versions hence we have to manually make changes to generate builds with additional qualifiers 

### Describe the solution you'd like

Add the ability to pass additional qualifier to automated distribution build  process. Distribution build process should pick it from the manifest file used for creating the builds. Possible qualifiers includes but not limited to,

- Beta
- Alpha
- RC
- Preview
- Preview-1
- Preview.1

### Alpha1

The user should be able to build OpenSearch and Dashboards using the above qualifiers via build process

- [x] https://github.com/opensearch-project/opensearch-build/pull/1844
       https://github.com/opensearch-project/opensearch-build/pull/1867
- [x] https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1340
- [ ] https://github.com/opensearch-project/alerting-dashboards-plugin/issues/194
- [x] https://github.com/opensearch-project/alerting/issues/348
       https://github.com/opensearch-project/alerting/pull/366
       https://github.com/opensearch-project/alerting/pull/373
       https://github.com/opensearch-project/alerting/pull/376
       https://github.com/opensearch-project/alerting/pull/377
- [ ] https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/209
- [x] https://github.com/opensearch-project/anomaly-detection/issues/432
       https://github.com/opensearch-project/anomaly-detection/pull/468
       https://github.com/opensearch-project/anomaly-detection/pull/472
       https://github.com/opensearch-project/opensearch-build/pull/1838
       https://github.com/opensearch-project/anomaly-detection/pull/483
       https://github.com/opensearch-project/anomaly-detection/pull/486
- [x] https://github.com/opensearch-project/asynchronous-search/issues/103
        https://github.com/opensearch-project/asynchronous-search/pull/115
        https://github.com/opensearch-project/asynchronous-search/pull/118
        https://github.com/opensearch-project/asynchronous-search/pull/119
        https://github.com/opensearch-project/asynchronous-search/pull/120
        https://github.com/opensearch-project/asynchronous-search/pull/123
- [x] https://github.com/opensearch-project/common-utils/issues/131
       https://github.com/opensearch-project/common-utils/pull/128
       https://github.com/opensearch-project/common-utils/pull/146
[...]

PR #1731, opened 2022-03-09
Added version qualifier support for OpenSearch.

dblock:

### Description

Adds an optional build version qualifier to the input manifest that is passed onto OpenSearch, which already supports qualified version numbers (currently only `alpha`, `beta` and `rc`). 
 
### Issues Resolved

Part of https://github.com/opensearch-project/opensearch-build/issues/1632.
 
### Check List

### DIFF
--- a/manifests/2.0.0/opensearch-2.0.0.yml
+++ b/manifests/2.0.0/opensearch-2.0.0.yml
@@ -7,6 +7,7 @@ ci:
 build:
   name: OpenSearch
   version: 2.0.0
+  qualifier: alpha1
 components:
   - name: OpenSearch
     ref: main
--- a/scripts/components/OpenSearch/build.sh
+++ b/scripts/components/OpenSearch/build.sh
@@ -13,6 +13,7 @@ function usage() {
     echo ""
     echo "Arguments:"
     echo -e "-v VERSION\t[Required] OpenSearch version."
+    echo -e "-q QUALIFIER\t[Optional] Version qualifier."
     echo -e "-s SNAPSHOT\t[Optional] Build a snapshot, default is 'false'."
     echo -e "-p PLATFORM\t[Optional] Platform, default is 'uname -s'."
     echo -e "-a ARCHITECTURE\t[Optional] Build architecture, default is 'uname -m'."
@@ -21,7 +22,7 @@ function usage() {
     echo -e "-h help"
 }
 
-while getopts ":h:v:s:o:p:a:d:" arg; do
+while getopts ":h:v:q:s:o:p:a:d:" arg; do
     case $arg in
         h)
             usage
@@ -29,6 +30,9 @@ while getopts ":h:v:s:o:p:a:d:" arg; do
             ;;
         v)
             VERSION=$OPTARG
+            ;;
+        q)
+            QUALIFIER=$OPTARG
             ;;
         s)
             SNAPSHOT=$OPTARG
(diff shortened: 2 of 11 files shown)


manifests/2.0.0/opensearch-2.0.0.yml
```diff
@@ -7,6 +7,7 @@ ci:
 build:
   name: OpenSearch
   version: 2.0.0
+  qualifier: alpha1
```
gaiksaya, line 10: What will be the qualifier for GA or prod artifact? Will this field be empty?

scripts/components/OpenSearch/build.sh
```diff
@@ -13,6 +13,7 @@ function usage() {
     echo ""
     echo "Arguments:"
     echo -e "-v VERSION\t[Required] OpenSearch version."
+    echo -e "-q QUALIFIER\t[Optional] Version qualifier."
```
gaiksaya, line 16: Do we have any defaults if nothing is specified as a qualifier?

manifests/2.0.0/opensearch-2.0.0.yml
dblock: Yes, we'll just remove this.

scripts/components/OpenSearch/build.sh
dblock: Default is not set, which is today. Since it's optional doesn't make sense to say it twice.

manifests/2.0.0/opensearch-2.0.0.yml
gaiksaya: In that case this gradle command having a null or empty value for qualifier will not be an issue?
https://github.com/opensearch-project/opensearch-build/pull/1731/files#diff-004d8135f0e1267ff669721dc64b44f84729dc6fe871c6c2ef02eb9d1e04959dR75

dblock: It doesn't seem like it would be a problem, similarly we don't always pass a value for `-Dbuild.snapshot=`.

gaiksaya: Okay. Also do we need to have a check to allow only certain values as a qualifier?

setiah: ^Maybe not. This qualifier check logic is already handled [downstream](https://github.com/opensearch-project/OpenSearch/blob/044f53630a08b0021097e422c9f31f15fb98219d/buildSrc/build.gradle#L301-L302) in the component build, so adding it here seems redundant and creates two sources of truth. So good to skip imo.

setiah (approved):

manifests/2.0.0/opensearch-2.0.0.yml
gaiksaya: Looks like it only supports `(alpha|beta|rc)`

gaiksaya (approved):

manifests/2.0.0/opensearch-2.0.0.yml
setiah: I'm not sure if plugins build script inherits the qualifier check logic as well (?) If not, it might make sense to add here.

dblock: Build will fail downstream, but probably makes sense to have a check here to fail on PRs to the manifest (fail earlier). Will add in a future PR.

Issue #1688, opened 2022-03-02
[Bug]: OpenSearch integ tests should run in docker


kavilla:
### Describe the bug

Current implementation: https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/integ-test.jenkinsfile#L37

### To reproduce

n/a

### Expected behavior

Should use
```
docker {
....
}
```

bbarani:
@kavilla Can you please update this issue with the current status?

kavilla:
Hey @bbarani, looking at the OpenSearch integ test run it looks like each component is ran in it's own docker container. But for verification a maintainer to confirm would be good and then this issue can be closed I believe.

Thanks!

prudhvigodithi:
Hey @kavilla the OpenSearch integ tests for CI, yes they do run in docker agents, but when executed locally its not, we can close this issue, wdyth @kavilla ?

bbarani:
@kavilla @prudhvigodithi @gaiksaya Can we close this issue?

jordarlu:
Hi, team, I am gonna close this issue for now as it has been a while without further dialog. Please feel free to reopen it if needed... thanks!!

Issue #1687, opened 2022-03-02
Integ Test Jenkinsfiles Input Manifest Param


kavilla:
### Is your feature request related to a problem? Please describe

Origin: https://github.com/opensearch-project/opensearch-build/pull/1653#discussion_r817145061

Right now we pass information to the integ test jobs like the information on the runner. We need to do this because certain images have different dependencies. For example, all `1.x` builds of OpenSearch Dashboards requires Node 10.24.1 but all builds for `2.x` require 14.18.2. The same image cannot be used build, install, and execute integration tests.

### Describe the solution you'd like

Instead of passing a string about the AGENT, integ test jobs should just accept the input manifest that triggered the build job and read the information from the input manifest.

gaiksaya:
Closing this issue as we now have the described solution implemented
> Instead of passing a string about the AGENT, integ test jobs should just accept the input manifest that triggered the build job and read the information from the input manifest.

Issue #1686, opened 2022-03-02
Build manifest to include JOB_NAME


kavilla:
### Is your feature request related to a problem? Please describe

Origin: https://github.com/opensearch-project/opensearch-build/pull/1653#discussion_r815728318

The name of the job that triggers `integ-test` pipeline along with the build ID is used to figure out which distribution to download and run tests against. This requires hardcoding that value.

### Describe the solution you'd like

I propose we include the job name with the build manifest. For example:
```
---
schema-version: '1.2'
build:
  name: OpenSearch Dashboards
  version: 2.0.0
  platform: linux
  architecture: x64
  distribution: tar
  id: '1417'
  job_name: 'distribution-build-opensearch-dashboards'
components:
  - name: OpenSearch-Dashboards
    repository: https://github.com/opensearch-project/OpenSearch-Dashboards.git
    ref: main
    commit_id: 7cb8297622788b7c074cea1f92e4c2ca9f8f5f5d
    artifacts:
      dist:
        - dist/opensearch-dashboards-min-2.0.0-linux-x64.tar.gz
    version: 2.0.0.0
  - name: functionalTestDashboards
    repository: https://github.com/opensearch-project/opensearch-dashboards-functional-test.git
    ref: main
    commit_id: 24fba9c0cf413db6c16ae4117d0502aae283a019
    version: 2.0.0.0

```

That way when passing the URL of the build manifest to the test job and downloading the manifest. I can get the value of `build.job_name` and pass that to run `runIntegTestScript`.

### Describe alternatives you've considered

An assumption could be that the URL passed is the same format `https://ci.opensearch.org/ci/dbc/**JOB_NAME**/1.3.0/latest/linux/arm64/builds/opensearch/manifest.yml`, and we can do the logic but we have access to the job name while building the build manifest and I think there are no guarantees that the format of this URL will stay the same.

Can also just pass another input in the jenkins params that takes the build job name.

### Additional context

I believe adding the job name to the build manifest could provide insight in the future if we have ever change the URLs or if a job was triggered and end up dumping the artifacts to another bucket that does not equal the expect name. Not sure if possible in the current system but if someone hard codes in the Playground it could happen.

abhinavGupta16:
[triage] We don't want to add the `job_name` in the build manifest, since the manifest should only be responsible for building the artifact. We can add `job-name` as a part of Jenkins pipeline/pipeline.

@kavilla

bbarani:
[...]

PR #1756, opened 2022-03-15
[Build][Test] OS and OSD improvements plus support for Playground

kavilla:
### Description
While using a child folder 'Playground' in Jenkins I had to do a number of
changes to support a successful test run. I also included the ability to
skip publishing the notification not to spam channels from builds from
for example `Playground`. Also ability to skip building docker because
from what I could tell it was publishing docker images from `Playground`
pipelines as well.

Finally, some cleanup to integ tests to use the input manifest to container image.
Also, make integ tests run in docker container for OpenSearch. Use libtester for 
detectTestDockerAgent

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1688
https://github.com/opensearch-project/opensearch-build/issues/1687
https://github.com/opensearch-project/opensearch-build/issues/1686
https://github.com/opensearch-project/opensearch-build/issues/1758
 
### Check List


peterzhuamazon:
#1394 
Ever consider making this change so playground runs the same?

kavilla:
> #1394 Ever consider making this change so playground runs the same?

Couple things about that. We merged that [change](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/distribution-build.jenkinsfile#L6) for copying artifacts for OSD builds which helps with building in `Playground`.

However, since we build in `Playground` it publishes our artifacts to S3 under `Playground\${JOB_NAME}`. For integration tests and soon to be added BWC tests, it will need that full project name to download the correct build if it is under a nested folder.

If the suggestion is to regardless of the job always build and publish `${$JOB_BASE_NAME}` then nested pipelines like `Playground` will also flood the some base folder as the production ready releases. Which I wouldn't recommend but we can go that route if it's preferred.

tianleh:
Great improvement to avoid spamming the notification! We could consider sending such notification to another channel, e.g 'opensearch-distribution-builds-staging' as a future improvement.

dblock (approved):

kavilla:
Just to showcase how it looks,

This is how it looks when manually triggering a build:

Screen Shot 2022-03-18 at 12 55 32 PM

Build is triggered skipping the build docker step:

Screen Shot 2022-03-18 at 12 56 11 PM

Finally showcasing the integ tests (please note this was me passing an non existing test yaml file thus it was caught in the verify params stage and aborted the workflow and threw an error)

Screen Shot 2022-03-18 at 12 57 05 PM

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -23,8 +23,33 @@ pipeline {
...
+            name: 'PUBLISH_NOTIFICATION',
+            description: 'Publish the status of this build job or not.',
+            defaultValue: true
+        )
     }
     stages {
+        stage('verify-parameters') {
+            steps {
+                script {
+                    env.INTEG_TEST_JOB_NAME = INTEG_TEST_JOB_NAME != '' ?
```
zelinh, line 47: Do we need the check here since `INTEG_TEST_JOB_NAME` always has the default value which is the same as the `DEFAULT_INTEG_TEST_JOB_NAME` if not assigned?

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -65,22 +90,19 @@ pipeline {
                             echo "artifactUrl (x64): ${artifactUrl}"
 
                             def integTestResults = 
-                                build job: INTEG_TEST_JOB_NAME,
+                                build job: env.INTEG_TEST_JOB_NAME,
```
zelinh, line 93: I think according to @abhinavGupta16, use `env` prefix may cause some issues with our tester. Could you help confirm it? @abhinavGupta16

jenkins/opensearch/distribution-build.jenkinsfile
```diff
@@ -18,8 +18,33 @@ pipeline {
...
+            defaultValue: true
+        )
+        booleanParam(
+            name: 'PUBLISH_NOTIFICATION',
+            description: 'Publish the status of this build job or not.',
+            defaultValue: true
+        )
     }
     stages {
+        stage('verify-parameters') {
```
zelinh, line 39: Same as above. I don't think this stage is necessary.

jenkins/opensearch/distribution-build.jenkinsfile
```diff
@@ -123,9 +148,9 @@ pipeline {
 
                             echo "buildManifestUrl (x64): ${buildManifestUrl}"
                             echo "artifactUrl (x64): ${artifactUrl}"
-
+
```
zelinh, line 151: You accidentally added some space here.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
kavilla: This is probably me being too protective but technically someone can kick off a build and empty out the string for `INTEG_TEST_JOB_NAME`. I can remove this if we do not want to handle those situations.

jenkins/opensearch/distribution-build.jenkinsfile
kavilla: Good catch! Addressed!

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
kavilla: Here I wanted to ensure the right variable is used. If we go the route not worrying about the empty string case (which would require someone manually triggering the build and clearing it) then I will remove the`DEFAULT` and only use the param and remove the `env` prefix.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
zelinh: Had some offline discussion with @kavilla, this makes sense to me that we could leave the parameter empty with this stage of check. These situation could happen for example when we pass the params from cron jobs. This shouldn't be a blocker for me.

zelinh (approved):
I really like the features of checking if we want to build docker or publish notifications so that I don't have to comment them out anymore for testing.

PR #1760, opened 2022-03-15
Disable build for 1.3 to freeze the code

zelinh:

### Description
Disable the check for build on 1.3 to freeze the code to block any further changes to the branch. 
 
### Issues Resolved
 
### Check List


gaiksaya (approved):

peterzhuamazon (approved):

dblock:
IMO we don't need to do this since the build will stop unless there are more changes. What we should be doing is incrementing the version to 1.3.1.

zelinh:
> IMO we don't need to do this since the build will stop unless there are more changes. What we should be doing is incrementing the version to 1.3.1.

Yes that would be the optimized scenario. We did this because we want to avoid unexpected code changes in the release state (e.g. last minutes commit from Index-management plugin today). We could add these back after release.

PR #1765, opened 2022-03-16
[CI] fix build trigger

kavilla:
### Description
Notifications showing: `check-for-build [xxxxxx] Failed checking for build to trigger`.

### Issues Resolved
n/a
 
### Check List

### DIFF
--- a/manifests/1.4.0/opensearch-1.4.0-test.yml
+++ b/manifests/1.4.0/opensearch-1.4.0-test.yml
@@ -1,0 +1,3 @@
+---
+name: OpenSearch
+schema-version: '1.0'
--- a/jenkins/check-for-build.jenkinsfile
+++ b/jenkins/check-for-build.jenkinsfile
@@ -7,8 +7,8 @@ pipeline {
     agent none
     triggers {
         parameterizedCron '''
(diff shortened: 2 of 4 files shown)


manifests/1.4.0/opensearch-1.4.0-test.yml
```diff
@@ -0,0 +1,3 @@
+---
```
zelinh, line 1: Do we want the test for `OpenSearch` now since we don't really test anything here?

peterzhuamazon (changes requested):
@kavilla you need to fix this as well since the check-for-build.jenkinsfile is auto updated by workflow whenever there is a new version pop up.

https://github.com/opensearch-project/opensearch-build/blob/60d69588621ab2cdbed520d376d2bd6df5bd5697/src/manifests_workflow/input_manifests.py#L144

kavilla:
@peterzhuamazon @zelinh , I'm closing this for this: https://github.com/opensearch-project/opensearch-build/pull/1756

It happens because when we were trying to read the yaml for the test manifest in OSD build it failed. By moving it to integ test I read if the input exists if not fail the pipeline. pipeline error will be caught and announced in integ test but it wont fail the build pipeline it will just be notified that with the check build the integ tests failed.

PR #1791, opened 2022-03-21
[Build] fix expression for when for docker build

kavilla:
### Description
Originally verified only with it skipping and it was skipping but
it was always skipping. It would appear that I can't just use
BUILD_DOCKER. It request params.

### Issues Resolved
n/a

### Check List

### DIFF
--- a/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
+++ b/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
@@ -190,7 +190,9 @@ pipeline {
         stage('docker build') {
             when {
                 beforeAgent true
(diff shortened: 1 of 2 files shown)


kavilla:
Screen Shot 2022-03-21 at 2 47 23 PM

zelinh (approved):
Lookup how this should be executed and found this on https://www.jenkins.io/doc/book/pipeline/syntax/
Thanks. 
```
expression
Execute the stage when the specified Groovy expression evaluates to true, for example: when { expression { return params.DEBUG_BUILD } } Note that when returning strings from your expressions they must be converted to booleans or return null to evaluate to false. Simply returning "0" or "false" will still evaluate to "true".
```

PR #1815, opened 2022-03-23
Support arm64 rpm generation with dynamic version and any number of plugins

### LINKED ISSUE #1545: [RPM M1] Add a new block to call the generation code for RPM
-- | -- | -- | -- | --

peterzhuamazon:

### Description
Support arm64 rpm generation with dynamic version and any number of plugins
 
### Issues Resolved
#1545
 
### Check List


src/assemble_workflow/bundle_rpm.py
```diff
@@ -89,14 +92,27 @@ def build(self, name: str, dest: str, archive_path: str) -> None:
...
+
         # Run bundle rpmbuild
         bundle_cmd = " ".join(
             [
                 'rpmbuild',
                 '-bb',
                 '--define',
                 f"'_topdir {ext_dest}'",
-                f"{self.filename}.rpm.spec",
+                f"{self.filename}.rpm.final.spec",
```
tianleh, line 115: Just curious what `final` means in this context

scripts/pkg/build_templates/opensearch/opensearch.rpm.spec
```diff
@@ -36,10 +36,17 @@ For more information, see: https://opensearch.org/
...
 cp -a etc usr var %{buildroot}
 chmod 0755 %{buildroot}%{product_dir}/bin/*
-chmod 0755 %{buildroot}%{product_dir}/plugins/opensearch-security/tools/*
+if ls %{buildroot}%{product_dir}/plugins | grep opensearch-security; then
+    chmod 0755 %{buildroot}%{product_dir}/plugins/opensearch-security/tools/*
+fi
+# Pre-populate the folders to ensure rpm build success even without all plugins
+mkdir -p %{buildroot}%{config_dir}/opensearch-observability
+mkdir -p %{buildroot}%{config_dir}/opensearch-reports-scheduler
+mkdir -p %{buildroot}%{product_dir}/performance-analyzer-rca
```
tianleh, line 49: Great work! This will allow developers to speed up testing by commenting out most plugins.

tianleh (approved):

src/assemble_workflow/dist.py
```diff
@@ -77,6 +79,7 @@ def extract(self, dest: str) -> str:
         return self.archive_path
 
     def build(self, name: str, dest: str) -> None:
+        print("123123")
```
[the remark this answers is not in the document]
peterzhuamazon: removed.

src/assemble_workflow/bundle_rpm.py
dblock, line 115: As is, the .spec file is not a real .spec (it includes variables such as `###VERSION###` that's not valid syntax for a spec file. So it should be called `.spec.template`, and its output `.spec`. 

But I'd prefer it if you used the built-in variable substitution mechanism and thus .spec was valid in both the source and final version. Since the .spec file is checked in, the version that is generated should be placed in a destination/target folder under the same name.

src/assemble_workflow/dist.py
```diff
@@ -13,11 +13,13 @@
...
 
 from assemble_workflow.bundle_rpm import BundleRpm
+from manifests.build_manifest import BuildManifest
 from system.zip_file import ZipFile
 
 
 class Dist(ABC):
-    def __init__(self, name: str, path: str, min_path: str) -> None:
+    def __init__(self, name: str, path: str, min_path: str, build_kls: BuildManifest.Build) -> None:
+        self.build_kls = build_kls
```
dblock, line 22: What is "kls"? You mean class? That's `cls` in Python.

src/assemble_workflow/dists.py
```diff
@@ -26,11 +27,12 @@ class Dists:
     }
 
     @classmethod
-    def create_dist(cls, name: str, path: str, min_path: str, distribution: str) -> Dist:
+    def create_dist(cls, name: str, path: str, min_path: str, build: BuildManifest.Build) -> Dist:
+        distribution = build.distribution
```
dblock, line 31: distribution = build.distribution or 'tar'

I don't think you need that info log.

src/assemble_workflow/bundle_rpm.py
```diff
@@ -89,14 +92,27 @@ def build(self, name: str, dest: str, archive_path: str) -> None:
...
+            "x64": "x86_64",
+            "arm64": "aarch64",
+        }
+        replaces = ('##VERSION##', build_kls.version), ('##ARCHITECTURE_ALT##', architecture_alt_rpm[build_kls.architecture])
+
+        with open(os.path.join(ext_dest, f"{self.filename}.rpm.spec"), 'r') as fp:
+            spec_lines = fp.read()
+
+        with open(os.path.join(ext_dest, f"{self.filename}.rpm.final.spec"), 'w') as fp:
+            fp.write(reduce(lambda a, key_value: a.replace(*key_value), replaces, spec_lines))
```
dblock, line 106: If you're going to be doing template replacement extract that into at least a method, or ideally into a separate class, or ideally use a built-in way to edit templates in Python, https://stackabuse.com/formatting-strings-with-the-python-template-class/

dblock (changes requested):
You also need tests that verify that the variable substitution was done correctly, especially if you're rolling out your own. But I think you can delete all that code if you use built-in rpm variable substitution (I could be wrong).

scripts/pkg/build_templates/opensearch/opensearch.rpm.spec
```diff
@@ -36,10 +36,17 @@ For more information, see: https://opensearch.org/
 cd %{_topdir} && pwd
 # Create necessary directories
 mkdir -p %{buildroot}%{pid_dir}
+mkdir -p %{buildroot}%{product_dir}/plugins
 # Install directories/files
 cp -a etc usr var %{buildroot}
 chmod 0755 %{buildroot}%{product_dir}/bin/*
-chmod 0755 %{buildroot}%{product_dir}/plugins/opensearch-security/tools/*
+if ls %{buildroot}%{product_dir}/plugins | grep opensearch-security; then
```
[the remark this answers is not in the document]
peterzhuamazon: It does not have any issues to run on bourne shell.
Thanks.

src/assemble_workflow/bundle_rpm.py
```diff
@@ -89,14 +92,27 @@ def build(self, name: str, dest: str, archive_path: str) -> None:
         shutil.move(f"{min_bin_env_path}.backup", min_bin_env_path)
         shutil.move(min_dest_path, min_source_path)
 
+        # Append correct value substitute to the spec file
+        architecture_alt_rpm = {
```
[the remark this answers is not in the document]
peterzhuamazon: Hi @dblock 

There is no need to do vise versa, and it is a completely different conversion on deb.
Having a matrix with clear definition is ok to me.
This is similar to assign extension based on distribution.

If you strongly against it we can open an issue and fix later.


Thanks.

src/assemble_workflow/bundle_rpm.py
```diff
@@ -89,14 +92,27 @@ def build(self, name: str, dest: str, archive_path: str) -> None:
...
+            "x64": "x86_64",
+            "arm64": "aarch64",
+        }
+        replaces = ('##VERSION##', build_kls.version), ('##ARCHITECTURE_ALT##', architecture_alt_rpm[build_kls.architecture])
```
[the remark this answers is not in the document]
peterzhuamazon: Can you elaborate more on this ?
I think that would make the process complicated for now.

Thanks.

src/assemble_workflow/bundle_rpm.py
peterzhuamazon: This is a valid call out, I will create an issue to improve this later.

src/assemble_workflow/dist.py
peterzhuamazon: Valid call outs.

src/assemble_workflow/dists.py
peterzhuamazon: sure.

src/assemble_workflow/bundle_rpm.py
dblock: I'd rather fix this here. It's a trivial change and the issue will just sit there forever.

scripts/pkg/build_templates/opensearch/opensearch.rpm.spec
```diff
@@ -12,13 +12,13 @@
 %define pid_dir %{_localstatedir}/run/%{name}
 
 Name: opensearch
-Version: 1.3.0
+Version: ##VERSION##
```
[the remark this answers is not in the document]
dblock: I am not an expert in RPM, but it feels wrong to me not to use the standard way that RPM says you should be doing this and we're inventing another template replacement process. What do you think @spotaws?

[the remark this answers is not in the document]
peterzhuamazon: It is fixed right now using `--define`.

src/assemble_workflow/bundle_rpm.py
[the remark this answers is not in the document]
peterzhuamazon: fixed now.

src/assemble_workflow/bundle_rpm.py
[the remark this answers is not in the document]
peterzhuamazon: Understand now, thanks.

src/assemble_workflow/bundle_rpm.py
peterzhuamazon: Fixed now.

src/assemble_workflow/dist.py
peterzhuamazon: Removed kls, now it is using the dict as `build_dict`.

src/assemble_workflow/dists.py
peterzhuamazon: Also fixed now.

src/assemble_workflow/bundle_rpm.py
peterzhuamazon: Completely removed this section right now.

peterzhuamazon:
Hi @dblock I have resolved all the comments.
Please review again.

Thanks.

scripts/pkg/build_templates/opensearch/opensearch.rpm.spec
```diff
@@ -10,15 +10,17 @@
 %define data_dir %{_sharedstatedir}/%{name}
 %define log_dir %{_localstatedir}/log/%{name}
 %define pid_dir %{_localstatedir}/run/%{name}
+%{!?_version: %define _version 0.0.0 }
+%{!?_architecture_alt: %define _architecture_alt x86_64 }
```
dblock, line 14: Should this be just `_architecture`?

src/assemble_workflow/bundle.py
```diff
@@ -135,7 +135,7 @@ def __get_min_dist(self, build_components: BuildComponents) -> Dist:
         logging.info(f"Copied min bundle to {min_dist_path}.")
         min_path = f"{self.build.filename}-{self.build.version}".replace("-SNAPSHOT", "")
         logging.info(f"Start creating distribution {self.build.distribution} for {self.min_bundle.name}.")
-        min_dist = Dists.create_dist(self.min_bundle.name, min_dist_path, min_path, self.build.distribution)
+        min_dist = Dists.create_dist(self.min_bundle.name, min_dist_path, min_path, self.build.__to_dict__())
```
dblock, line 138: I think you misunderstood me. Here pass in `self.build`.

src/assemble_workflow/bundle_rpm.py
```diff
@@ -94,8 +96,9 @@ def build(self, name: str, dest: str, archive_path: str) -> None:
             [
                 'rpmbuild',
                 '-bb',
-                '--define',
-                f"'_topdir {ext_dest}'",
+                f"--define '_topdir {ext_dest}'",
+                f"--define '_version {build_dict['version']}'",
```
dblock, line 100: Here you can just do `build.version`.

src/assemble_workflow/bundle_rpm.py
```diff
@@ -94,8 +96,9 @@ def build(self, name: str, dest: str, archive_path: str) -> None:
...
+                f"--define '_architecture_alt {rpm_architecture_alt(build_dict['architecture'])}'",
```
dblock, line 101: Here you can do `build.architecture`.

scripts/pkg/build_templates/opensearch/opensearch.rpm.spec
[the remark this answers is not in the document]
dblock: I don't think I understand, `ls` + `grep` feels like the "wrong" way to check whether a folder exists.

src/system/os.py
```diff
@@ -23,3 +23,14 @@ def current_platform() -> str:
         return "windows"
     else:
         return subprocess.check_output(["uname", "-s"]).decode().strip().lower()
+
+
+def rpm_architecture_alt(architecture: str) -> str:
```
dblock, line 28: I would just call this `rpm_architecture`, lose the `alt`.

tests/tests_system/test_os.py
```diff
@@ -45,3 +45,8 @@ def test_current_platform(self) -> None:
     @patch("subprocess.check_output", return_value="Windows".encode())
     def test_current_platform_lowercase(self, mock_subprocess: MagicMock) -> None:
         self.assertTrue(current_platform() == "windows")
+
+    # rpm_architecture_alt
+    def test_rpm_architecture_alt(self) -> None:
+        self.assertTrue(rpm_architecture_alt('x64') == 'x86_64')
```
dblock, line 51: use `self.assertEquals`

dblock (dismissed):

scripts/pkg/build_templates/opensearch/opensearch.rpm.spec
peterzhuamazon: I want to be consistent between our python code and here.
It is not the version we are using, but alternative ones.

peterzhuamazon: Probably will make it simpler by using architecture.

scripts/pkg/build_templates/opensearch/opensearch.rpm.spec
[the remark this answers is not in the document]
peterzhuamazon: fixed now.

tests/tests_system/test_os.py
peterzhuamazon: fixed.

src/system/os.py
peterzhuamazon: fixed.

src/assemble_workflow/bundle.py
peterzhuamazon: fixed.

src/assemble_workflow/bundle_rpm.py
peterzhuamazon: fixed.

src/assemble_workflow/bundle_rpm.py
peterzhuamazon: fixed.

tianleh (approved):

src/assemble_workflow/bundle_rpm.py
```diff
@@ -94,8 +97,9 @@ def build(self, name: str, dest: str, archive_path: str) -> None:
             [
                 'rpmbuild',
                 '-bb',
-                '--define',
-                f"'_topdir {ext_dest}'",
+                f"--define '_topdir {ext_dest}'",
+                f"--define '_version {build_cls.version}'",
+                f"--define '_architecture_alt {rpm_architecture(build_cls.architecture)}'",
```
dblock, line 102: This looks like a bug, you should be passing `_architecture` into it.

tests/tests_system/test_os.py
```diff
@@ -45,3 +45,8 @@ def test_current_platform(self) -> None:
     @patch("subprocess.check_output", return_value="Windows".encode())
     def test_current_platform_lowercase(self, mock_subprocess: MagicMock) -> None:
         self.assertTrue(current_platform() == "windows")
+
+    # rpm_architecture_alt
```
dblock, line 49: Comment is unnecessary and has the wrong name anyway, remove.

src/assemble_workflow/bundle_rpm.py
tianleh: Fixed in https://github.com/opensearch-project/opensearch-build/pull/1829

Issue #1814, opened 2022-03-23
[Bug]: Integ tests fail if no security plugin is installed


kavilla:
### Describe the bug

When the integ test workflow runs it needs to start a cluster for OpenSearch tests, it needs to start a cluster and an OpenSearch Dashboards for OpenSearch Dashboards tests. 

During this workflow, even if we pass `without-security` to the test manifest for `integ-test` config the workflow makes the assumption that the security plugin is actually there and tries to uninstall it or add configs to a plugin that isn't there. Which fails to start a cluster or a cluster and OpenSearch Dashboards.

The source files are here:
* https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/service_opensearch_dashboards.py#L54
* https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/service_opensearch.py#L38

### To reproduce

Attempt to run integration test workflow and watch it fail to run.

### Expected behavior

Integration test workflow if `without-security` is specified even if the security plugin is not already installed to the distribution.

### Screenshots

n/a

PR #1816, opened 2022-03-23
[Tests][Integ] Fix without-security logic if no security plugin

kavilla:
### Description
When the integ test workflow runs it needs to start a cluster for
OpenSearch tests, it needs to start a cluster and an
OpenSearch Dashboards for OpenSearch Dashboards tests.

During this workflow, even if we pass `without-security` to the
test manifest for `integ-test` config the workflow makes the
assumption that the security plugin is actually there and tries
to uninstall it or add configs to a plugin that isn't there. Which
fails to start a cluster or a cluster and OpenSearch Dashboards

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1814
 
### Check List

### DIFF
--- a/src/test_workflow/integ_test/service_opensearch.py
+++ b/src/test_workflow/integ_test/service_opensearch.py
@@ -33,8 +33,9 @@ class ServiceOpenSearch(Service):
         self.__download()
 
         self.opensearch_yml_dir = os.path.join(self.install_dir, "config", "opensearch.yml")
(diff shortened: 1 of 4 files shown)


tianleh:
Thanks for the fix! Could be even better to have a local manual test by commenting out security plugin from input manifest, run build/assemble and then run test without security.

tianleh (approved):

gaiksaya (approved):

Issue #1672, opened 2022-02-25
[Enhancement] Make build notification in chat show failed build command


peterzhuamazon:
It would be easier for us to debug the failure of build pipelines without the need to log dive.

If the build bot is able to print the failed build command in notification chat, it would be easier to reproduce the issue and save time.

Ex:
```
BUILD FAILED in 26s
1 actionable task: 1 executed
2022-02-25 17:17:28 ERROR    Error building security, retry with: ./build.sh manifests/1.3.0/opensearch-1.3.0.yml --component security --snapshot
```

Thanks.

prudhvigodithi:
We can achieve this in the following ways:

1)Consume the logs of the current job and parse for the error, this is little bit complex as each build fail for different error. 
Can come up with a shared library as
```
def BuildFailureCause() {
    
                     
                          def cause = "Caused:"
                          def error = "ERROR: "

                                build.getLog().eachLine { line ->
                                        if (line =~ /$cause/ || line =~ /$error/ ) {
                                            println "error: $line"  }
                                                          }

                         }

```
2) The another way is to consume the current running build logs and dump them as readable message 
Example as `def log = currentBuild.rawBuild.getLog(100).join('\n')`

3) We can use try catch block for each stage and get catch the error to convert to toString() format to read the error, but with this we need to add try/catch block to every stage out there in jenkinsfile
```
        stage('build') {
            steps {
                node('Agent') {
                    script {
                        FAILED_STAGE=env.STAGE_NAME
                        try {
                            echo "test"
                        }
                        catch (Exception e) {
                            ERROR=e.toString()
                            echo 'Exception occurred: ' + e.toString()
                            sh "exit 1"
                        }
                        
                    
                    }
                }
            }

```
4) We can combine try/catch block with post script to catch the failed stage and dump the error.

peterzhuamazon:
Thanks @prudhvigodithi 

If you can show some examples of each it would be helpful for us to make a decision on which is suits our needs well.

Thanks.

prudhvigodithi:
Fixed, added detailed error message to publish to slack channel
[...]

PR #1819, opened 2022-03-24
Added enhancement to slack publish with build failure cause

prudhvigodithi:
Signed-off-by: pgodithi <pgodithi@amazon.com>

### Description
With this PR, the slack build failure message will be enhanced with actually error.
Example while distribution-build-opensearch with manifest 2.0.0/opensearch-2.0.0.yml
```
MESSAGE=2022-03-24 01:47:24 ERROR    Error building job-scheduler
 retry with: ./build.sh manifests/2.0.0/opensearch-2.0.0.yml --component job-scheduler --snapshot
2022-03-24 01:48:51 ERROR    Error building job-scheduler
 retry with: ./build.sh manifests/2.0.0/opensearch-2.0.0.yml --component job-scheduler --snapshot
```
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1672
 
### Check List

### DIFF
--- a/vars/buildFailureMessage.groovy
+++ b/vars/buildFailureMessage.groovy
@@ -9,7 +9,7 @@ def call(){
     performance_log.close()
     performance_log = null
     logContent.eachLine() { line ->
-        if (line.matches(".*$ERROR_STRING.*")) {
+        if (line.matches(".*?$ERROR_STRING(.*?)")) {
             line=line.replace("\"", "")
             message.add(line)
         }
--- a/vars/buildFailureMessage.groovy
+++ b/vars/buildFailureMessage.groovy
@@ -10,6 +10,7 @@ def call(){
     performance_log = null
     logContent.eachLine() { line ->
         if (line.matches(".*$ERROR_STRING.*")) {
+            line=line.replace("\"", "")
             message.add(line)
         }
     }
--- a/vars/buildFailureMessage.groovy
+++ b/vars/buildFailureMessage.groovy
@@ -1,17 +1,21 @@
+import com.cloudbees.groovy.cps.NonCPS
+import org.apache.commons.io.IOUtils
 @NonCPS
 def call(){
-    def String ERROR_STRING = "ERROR"
-    def String message = ""
-    def String delimiter = ","
-    def performance_log = currentBuild.rawBuild.getLog(Integer.MAX_VALUE).join('\n')
-    performance_log.eachLine() { line ->
+    String ERROR_STRING = "ERROR"
+    List<String> message = []
+    String delimiter = ","
+    Reader performance_log = currentBuild.getRawBuild().getLogReader()
+    String logContent = IOUtils.toString(performance_log)
+    performance_log.close()
+    performance_log = null
+    logContent.eachLine() { line ->
         if (line.matches(".*$ERROR_STRING.*")) {
-            message+=(line+delimiter)
+            message.add(line)
         }
     }
-    message=message.replaceAll(",", "\n")
     if(message.isEmpty()){
-        message="Build failed"
+        message=["Build failed"]
     }
     return message
 }
--- a/vars/buildFailureMessage.groovy
+++ b/vars/buildFailureMessage.groovy
@@ -4,7 +4,6 @@ import org.apache.commons.io.IOUtils
 def call(){
     String ERROR_STRING = "ERROR"
     List<String> message = []
-    String delimiter = ","
     Reader performance_log = currentBuild.getRawBuild().getLogReader()
     String logContent = IOUtils.toString(performance_log)
     performance_log.close()
--- a/vars/buildFailureMessage.groovy
+++ b/vars/buildFailureMessage.groovy
@@ -1,0 +1,17 @@
+@NonCPS
+def call(){
+    def ERROR_STRING = "ERROR"
+    def message = ""
+    String delimiter = ","
+    def performance_log = currentBuild.rawBuild.getLog(Integer.MAX_VALUE).join('\n')
+    performance_log.eachLine() { line ->
+        if (line.matches(".*$ERROR_STRING.*")) {
+            message+=(line+delimiter)
+        }
+    }
+    message=message.replaceAll(",", "\n")
+    if(message.isEmpty()){
+        message="Build failed"
+    }
+    return message
+}
--- a/vars/buildFailureMessage.groovy
+++ b/vars/buildFailureMessage.groovy
@@ -1,10 +1,14 @@
 @NonCPS
+import java.io.Reader;
 def call(){
-    def ERROR_STRING = "ERROR"
-    def message = ""
-    String delimiter = ","
-    def performance_log = currentBuild.rawBuild.getLog(Integer.MAX_VALUE).join('\n')
-    performance_log.eachLine() { line ->
+    def String ERROR_STRING = "ERROR"
+    def String message = ""
+    def String delimiter = ","
+    Reader performance_log = currentBuild.getRawBuild().getLogReader().join('\n')
+    String logContent = IOUtils.toString(performance_log)
+    performance_log.close();
+    performance_log = null
+    logContent.eachLine() { line ->
         if (line.matches(".*$ERROR_STRING.*")) {
             message+=(line+delimiter)
         }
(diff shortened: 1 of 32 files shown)


peterzhuamazon (commented):
@abhinavGupta16 do we want @prudhvigodithi to use this new way of testing scripts or using the old ways of lib-testers?

vars/buildFailureMessage.groovy
```diff
@@ -0,0 +1,17 @@
...
+    def performance_log = currentBuild.rawBuild.getLog(Integer.MAX_VALUE).join('\n')
+    performance_log.eachLine() { line ->
+        if (line.matches(".*$ERROR_STRING.*")) {
+            message+=(line+delimiter)
+        }
+    }
+    message=message.replaceAll(",", "\n")
+    if(message.isEmpty()){
+        message="Build failed"
+    }
```
peterzhuamazon, line 15: Is this assuming the build exit with no `ERROR` keyword?

prudhvigodithi: Hey @peterzhuamazon yes you are right, this is for other outliers, like failed for any reason other than marked as ERROR, then in case instead we capture all of them, just sending a message as `Build failed`

peterzhuamazon: Got it.

vars/buildFailureMessage.groovy
```diff
@@ -0,0 +1,17 @@
+@NonCPS
+def call(){
+    def ERROR_STRING = "ERROR"
+    def message = ""
```
abhinavGupta16, line 4: Let's use definite datatypes. We have spent a lot of time fixing that for Python

vars/buildFailureMessage.groovy
abhinavGupta16, line 15:
```suggestion
    if(message.isEmpty()){
        message="Build failed"
    } else {
        message=message.replaceAll(",", "\n")
    }
```

vars/buildFailureMessage.groovy
```diff
@@ -0,0 +1,17 @@
...
+    String delimiter = ","
+    def performance_log = currentBuild.rawBuild.getLog(Integer.MAX_VALUE).join('\n')
+    performance_log.eachLine() { line ->
+        if (line.matches(".*$ERROR_STRING.*")) {
+            message+=(line+delimiter)
```
abhinavGupta16, line 9: This message can be long if there are a log of errors. Let's make it a `StringBuffer` or a `StringBuilder` instead

vars/buildFailureMessage.groovy
```diff
@@ -0,0 +1,17 @@
+@NonCPS
```
abhinavGupta16, line 1: What's this?

vars/publishNotification.groovy
```diff
@@ -1,11 +1,13 @@
 void call(Map args = [:]) {
     text = ([
-        "${args.icon} ${JOB_NAME} [${BUILD_NUMBER}] ${args.message}",
-        "Build: ${BUILD_URL}",
-        args.manifest,
+        "${args.icon}", 
+        "JOB_NAME=${JOB_NAME}",
+        "BUILD_NUMBER=[${BUILD_NUMBER}]",
```
abhinavGupta16, line 5: Why use an array here? Can there be multiple build numbers in one notification?

vars/buildFailureMessage.groovy
```diff
@@ -0,0 +1,17 @@
...
+def call(){
+    def ERROR_STRING = "ERROR"
+    def message = ""
+    String delimiter = ","
+    def performance_log = currentBuild.rawBuild.getLog(Integer.MAX_VALUE).join('\n')
+    performance_log.eachLine() { line ->
```
zelinh, line 7: I'm wonder how long this normally takes since it's iterating through the log? This could be inefficient if the log is long. Could we possibly read through the bottom of the log where the error is generally close to?

prudhvigodithi: Hey @zelinh, `currentBuild.rawBuild.getLog(Integer.MAX_VALUE)` we can add a value in place of Integer.MAX_VALUE, but this will get only last few lines of that value, and there are high chances we miss the ERROR message in those lines, so which is why I have added all lines and go over each line in running log, also this is part of post() script, which runs as part of same pipeline, so it should go over all the currentBuild log to get that ERROR string.

vars/buildFailureMessage.groovy
prudhvigodithi: so .each methods interpreter adds some overhead, so we should add @NonCPS to bypass the CPS transformation inside that method also when we add @NonCPS on a method, Jenkins will execute the entire method in one go without the ability to pause.

vars/buildFailureMessage.groovy
zelinh: Got it. Thanks!

vars/buildFailureMessage.groovy
prudhvigodithi: We want this to message.replaceAll(",", "\n") even-though message.isEmpty() is true as want to separate the Error messages one per line
Example slack output
```
MESSAGE=2022-03-24 15:30:20 ERROR    Error building job-scheduler
 retry with: ./build.sh manifests/2.0.0/opensearch-2.0.0.yml --component job-scheduler --snapshot
2022-03-24 15:31:41 ERROR    Error building job-scheduler
 retry with: ./build.sh manifests/2.0.0/opensearch-2.0.0.yml --component job-scheduler --snapshot
```

vars/buildFailureMessage.groovy
prudhvigodithi: Pushed in my latest commit

abhinavGupta16:
> @abhinavGupta16 do we want @prudhvigodithi to use this new way of testing scripts or using the old ways of lib-testers?

yes

vars/buildFailureMessage.groovy
abhinavGupta16: if message is an empty string, what will `.replaceAll` replace?

vars/buildFailureMessage.groovy
abhinavGupta16: Are we sure we want to do that, given this method can take very long to execute?

vars/buildFailureMessage.groovy
prudhvigodithi: yes right now this code reads line by line, and filters for the log error message, I'm looking for a buffered stream here, as explained above this is part of Same pipeline as post() call, so we need to read all lines each line as part of same pipeline.

vars/publishNotification.groovy
prudhvigodithi: This is part of already existing code, I have just made more readable.

vars/buildFailureMessage.groovy
prudhvigodithi: Thats a good point, Added Reader reading character streams and pushed in my latest commit.

vars/publishNotification.groovy
abhinavGupta16: Ohh.... I see. I wonder why is it an array though

vars/buildFailureMessage.groovy
abhinavGupta16: Yes, but it currently goes through all the lines twice and makes 2 strings. Can we change it to only once?

vars/buildFailureMessage.groovy
```diff
@@ -0,0 +1,20 @@
+import com.cloudbees.groovy.cps.NonCPS
+import org.apache.commons.io.IOUtils
+@NonCPS
+def call(){
+    String ERROR_STRING = "ERROR"
+    List<String> message = []
+    Reader performance_log = currentBuild.getRawBuild().getLogReader()
+    String logContent = IOUtils.toString(performance_log)
+    performance_log.close()
+    performance_log = null
```
abhinavGupta16, line 10: Nit: Why do we reassign it to null?

prudhvigodithi: Stream object will be eligible for garbage collection

gaiksaya (approved):
Looks good! Thanks!

PR #1836, opened 2022-03-27
[BUILD][OSD] Support version qualifier

kavilla:
### Description
Support version qualifier builds for OpenSearch Dashboards and
OpenSearch Dashboards Plugins.

This will create a zip for plugins with the OpenSearch Dashboards
version + qualifier. But the version of the plugin will remain the
same whilst opensearchDashboardsVersion in the opensearch_dashboards.json
will be the OpenSearch Dashboards version, eg, 1.3.0-alpha1. This is why
the version check needed to be modified a little to ensure that it can
get the version + qualifier and the default version of the plugin.

Updated for plugins as well.

Was able to successfully build and connect OSD 1.3.0-alpha1 to
the release version of OS 1.3.0.

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1340

### Issue partially resolved:
https://github.com/opensearch-project/opensearch-build/issues/1632
 
### Check List

### DIFF
--- a/src/build_workflow/build_target.py
+++ b/src/build_workflow/build_target.py
@@ -55,7 +55,7 @@ class BuildTarget:
         )
 
     @property
-    def compatible_core_versions(self) -> List[str]:
+    def compatible_min_versions(self) -> List[str]:
         return (
             [BuildTarget.__qualify_version(self.version, self.qualifier, self.snapshot)]
             + self.patches
--- a/src/build_workflow/build_target.py
+++ b/src/build_workflow/build_target.py
@@ -81,7 +81,7 @@ class BuildTarget:
 
     @property
     def compatible_opensearch_dashboards_component_versions(self) -> List[str]:
-        versions = self.compatible_component_versions 
+        versions = self.compatible_component_versions
         versions.extend([self.version + ".0"] if self.version + ".0" not in self.compatible_component_versions else [])
         return versions
 
--- a/src/build_workflow/build_target.py
+++ b/src/build_workflow/build_target.py
@@ -55,7 +55,7 @@ class BuildTarget:
         )
 
     @property
-    def compatible_opensearch_versions(self) -> List[str]:
+    def compatible_core_versions(self) -> List[str]:
         return (
             [BuildTarget.__qualify_version(self.version, self.qualifier, self.snapshot)]
             + self.patches
@@ -78,6 +78,10 @@ class BuildTarget:
             + list(map(lambda version: BuildTarget.__qualify_version(version + ".0", self.qualifier, False), self.patches))
             + list(map(lambda version: BuildTarget.__qualify_version(version + ".0", self.qualifier, True), self.patches))
         )
+
+    @property
+    def compatible_opensearch_dashboards_component_versions(self) -> List[str]:
+        return list(set([self.version + ".0"] + self.compatible_component_versions))
 
     @property
     def compatible_versions(self) -> List[str]:
--- a/src/build_workflow/build_target.py
+++ b/src/build_workflow/build_target.py
@@ -80,12 +80,6 @@ class BuildTarget:
         )
 
     @property
-    def compatible_opensearch_dashboards_component_versions(self) -> List[str]:
-        versions = self.compatible_component_versions
-        versions.extend([self.version + ".0"] if self.version + ".0" not in self.compatible_component_versions else [])
-        return versions
-
-    @property
     def compatible_versions(self) -> List[str]:
         versions = [self.version]
         versions.extend(self.patches)
--- a/src/build_workflow/build_target.py
+++ b/src/build_workflow/build_target.py
@@ -81,7 +81,9 @@ class BuildTarget:
 
     @property
     def compatible_opensearch_dashboards_component_versions(self) -> List[str]:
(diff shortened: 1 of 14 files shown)


kavilla:
* Created an "alpha" version of 1.3 OSD (with all the plugins). It was able to connect successfully to release OS 1.3
* Created an "alpha" of 2.0 OSD was able to successfully connect to alpha 2.0 OS.
* Created a 1.3 OSD with no qualifier (with all the plugins). It was able to connect successfully to release OS 1.3

src/build_workflow/build_target.py
```diff
@@ -55,7 +55,7 @@ def opensearch_version(self) -> str:
         )
 
     @property
-    def compatible_opensearch_versions(self) -> List[str]:
+    def compatible_core_versions(self) -> List[str]:
```
dblock, line 58: Let's rename this back to `compatible_opensearch_versions`, we don't use "core" anywhere.

dblock (changes requested):
Something feels fishy. Are you saying that an OpenSearch Dashboards version will be different from OpenSearch? Same question for plugins. Do you have an example of what that looks like?

kavilla:
> Something feels fishy. Are you saying that an OpenSearch Dashboards version will be different from OpenSearch? Same question for plugins. Do you have an example of what that looks like?

I'm sorry I not sure what you are referring to. Can you expand on this?

src/build_workflow/build_target.py
kavilla: Will it be confusing if I utilize this within the OpenSearch Dashboards plugin check?

dblock: Oh I see! I think the right name for this would be `min` then?

dblock:
> > Something feels fishy. Are you saying that an OpenSearch Dashboards version will be different from OpenSearch? Same question for plugins. Do you have an example of what that looks like?
> 
> I'm sorry I'm not sure what you are referring to. Can you expand on this?

Why is `compatible_opensearch_dashboards_component_versions` needed, meaning why is that set different from OpenSearch component versions?

kavilla:
> > > Something feels fishy. Are you saying that an OpenSearch Dashboards version will be different from OpenSearch? Same question for plugins. Do you have an example of what that looks like?
> > 
> > 
> > I'm sorry I'm not sure what you are referring to. Can you expand on this?
> 
> Why is `compatible_opensearch_dashboards_component_versions` needed, meaning why is that set different from OpenSearch component versions?

Got it I was going to bring this up so thanks for mentioning. With this change this makes it so plugins can install to specific version of OpenSearch Dashboards without doing the update manually. For example, https://github.com/opensearch-project/alerting-dashboards-plugin/blob/main/opensearch_dashboards.json#L3 can remain `"version": "1.3.0.0",` but this command will update `"opensearchDashboardsVersion": "1.3.0",` to be the version that is passed in the parameter so for example if building `2.0.0-alpha1` this plugin would look like:

```
{
  "id": "alertingDashboards",
  "version": "2.0.0.0",
  "opensearchDashboardsVersion": "2.0.0-alpha1",
  "configPath": ["opensearch_alerting"],
  "requiredPlugins": [],
  "server": true,
  "ui": true
}
```

But with the current condition check it will expect that `version` to be `2.0.0.0-alpha1` and the only way to change this is if plugins manually update the `version` for every qualifier. But since that is just used internally with the plugin (as far as I know not sure if plugins do some special check but they don't need to) it shouldn't require that to be change for every qualifier release. So as like as we tell that this plugin is for a specific version it build successfully and run successfully.

kavilla:
Not sure why the manifest for OpenSearch 2.0.0 keeps failing but should be good to go.

src/build_workflow/build_target.py
```diff
@@ -79,6 +79,12 @@ def compatible_component_versions(self) -> List[str]:
             + list(map(lambda version: BuildTarget.__qualify_version(version + ".0", self.qualifier, True), self.patches))
         )
 
+    @property
```
dblock, line 82: Last one, I promise, should `compatible_component_versions` now we renamed to `compatible_opensearch_component_versions`?

dblock:
@kavilla Why should it be possible to install a `2.0.0` plugin on top of `2.0.0-alpha` dashboards? That seems broken by design.

gaiksaya:
> > Not sure why the manifest for OpenSearch 2.0.0 keeps failing but should be good to go.
> 
> @gaiksaya is going to add qualifier support for ci workflow soon.

Not sure if I have the bandwidth but if anyone has please feel free to pick up this issue #1839

src/build_workflow/build_target.py
kavilla: After discussing, we decided that OpenSearch Dashboards plugins will need to manually bump their version to add the qualifier and we won't allow for `x.y.z.n` plugins to build with OpenSearch Dashboards `x.y.z-qualifier`.

That said I will remove the `opensearch_dashboards_component_versions` and then we can probably just keep `compatible_component_versions` the same name.

kavilla: Updated

dblock (approved):

PR #1852, opened 2022-03-30
Add dashboards rpm assemble support in opensearch-build repo

### LINKED ISSUE #1545: [RPM M1] Add a new block to call the generation code for RPM

peterzhuamazon:
### Description
Add dashboards rpm assemble support in opensearch-build repo
 
### Issues Resolved
#1545
 
### Check List


peterzhuamazon:
@spotrh could you help check the dashboards spec file again.
Thanks.

src/assemble_workflow/bundle_rpm.py
```diff
@@ -89,7 +89,10 @@ def build(self, name: str, dest: str, archive_path: str, build_cls: BuildManifes
         os.environ.pop('OPENSEARCH_PATH_CONF', None)
 
         # Restore config file and core folder to original location
-        shutil.move(f"{min_bin_env_path}.backup", min_bin_env_path)
+        if os.path.exists(f"{min_bin_env_path}.backup"):
+            logging.info(f"Restore {min_bin_env_path}.backup to {min_bin_env_path}")
+            shutil.move(f"{min_bin_env_path}.backup", min_bin_env_path)
```
dblock, line 94: This looks hacky, why is any of this necessary? Why is it undoing something done in extract?

peterzhuamazon:
> Do you plan to fix the "missing starting space in comment"?

That is straight up copy directly from OpenSearch core and it is expected.

dblock:
> > Do you plan to fix the "missing starting space in comment"?
> 
> That is straight up copy directly from OpenSearch core and it is expected.

Why not fix it everywhere so that the YML look cleaner?

src/assemble_workflow/bundle_rpm.py
peterzhuamazon: This was in for the opensearch only change.
The reason being there is a hardcoded `source /etc/sysconfig/opensearch` line specifically in rpm version of core.
This caused the failure during installing of plugin zips.

In extract block we comment out the line and back up the original file.
Here we just move the backup back.

The change for dashboards is just to check if the file exist, if not then ignore this step.

Thanks.

spotrh:
> 

Looks good to me without the /tmp permission item in the scriptlet.

peterzhuamazon:
> > > Do you plan to fix the "missing starting space in comment"?
> > 
> > 
> > That is straight up copy directly from OpenSearch core and it is expected.
> 
> Why not fix it everywhere so that the YML look cleaner? Can be another PR.
> 
> Why are we copy-pasting an entire file? Doesn't this duplicate the original source of the documentation for it?

OpenSearch core has an empty file of opensearch_dashboards.yml.
As for bundle with plugins we are using the one in config folder of this repo.
I just copy paste the comments here so if this change is not seems to be necessary, I can revert.

This just seems like a good way for user to have a template instead of finding the settings themselves.

peterzhuamazon:
> > 
> 
> Looks good to me without the /tmp permission item in the scriptlet.

Thanks @spotrh I have removed the /tmp line now :)

src/assemble_workflow/bundle_rpm.py
dblock: Why isn't that line removed in the OpenSearch repo in the first place, why is this happening here?

I'll let it slide, but this is a hack on top of a hack.

dblock (approved):

tianleh (approved):

src/assemble_workflow/bundle_rpm.py
peterzhuamazon: Improvement issue: https://github.com/opensearch-project/opensearch-build/issues/1864

Issue #1851, opened 2022-03-30
[Bug]: Integration test for arm64 are failing in jenkins


gaiksaya:
### Describe the bug

The paths for opensearch in integration test workflow is being set to `https://ci.opensearch.org/ci/dbc/integ-test/1.3.1/1584/linux/arm64`

```
BUILD_ID: 1584
[Pipeline] library
Only using first definition of library jenkins
[Pipeline] readYaml
[Pipeline] echo
Artifact root URL: https://ci.opensearch.org/ci/dbc/integ-test/1.3.1/1584/linux/arm64
[Pipeline] echo
Paths: opensearch=https://ci.opensearch.org/ci/dbc/integ-test/1.3.1/1584/linux/arm64
[Pipeline] sh

+ ./test.sh integ-test manifests/1.3.1/opensearch-1.3.1-test.yml --paths opensearch=https://ci.opensearch.org/ci/dbc/integ-test/1.3.1/1584/linux/arm64
Installing dependencies in . ...

Warning: the environment variable LANG is not set!
We recommend setting this in ~/.profile (or equivalent) for proper expected behavior.
Creating a virtualenv for this project...
Pipfile: /var/jenkins/workspace/integ-test/Pipfile
Using /usr/local/bin/python3.7m (3.7.7) to create virtualenv...
created virtual environment CPython3.7.7.final.0-64 in 378ms
  creator CPython3Posix(dest=/usr/share/opensearch/.local/share/virtualenvs/integ-test-B5RGX0jA, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/usr/share/opensearch/.local/share/virtualenv)
    added seed packages: pip==21.3.1, setuptools==58.3.0, wheel==0.37.0
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator


Successfully created virtual environment!
Virtualenv location: /usr/share/opensearch/.local/share/virtualenvs/integ-test-B5RGX0jA
Installing dependencies from Pipfile.lock (16e10a)...

To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
[...]

PR #1854, opened 2022-03-30
Add disclaimer in jenkinsFiles using upstreamBuilds

gaiksaya:

### Description
Add disclaimer in jenkinsFiles using upstreamBuilds.
 
### Issues Resolved
 https://github.com/opensearch-project/opensearch-build/issues/1851
 
### Check List

### DIFF
--- a/jenkins/opensearch-dashboards/bwc-test.jenkinsfile
+++ b/jenkins/opensearch-dashboards/bwc-test.jenkinsfile
@@ -43,7 +43,7 @@ pipeline {
                         currentBuild.result = 'ABORTED'
                         error("BWC Tests failed to start. Test manifest not found in manifests/${TEST_MANIFEST}.")
                     }
-                    /**
+                    /*
                     Rebuilding of this job will result in considering the upstream build as self($JOB_NAME) See https://issues.jenkins.io/browse/JENKINS-61590 for bug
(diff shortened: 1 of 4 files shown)


ohltyler (approved):
Thanks for adding!

gaiksaya:
> What about something like this? https://github.com/kavilla/opensearch-build/pull/6/files (ignoring the other content in there). This checks to verify it's not the job name which will be `integ-test`

Looks good but what was the use case at first to use upstream job? Does it upload the results under that job or something? Want to make sure the playground jobs do not mess with actual distribution jobs.

kavilla:
> > What about something like this? https://github.com/kavilla/opensearch-build/pull/6/files (ignoring the other content in there). This checks to verify it's not the job name which will be `integ-test`
> 
> Looks good but what was the use case at first to use upstream job? Does it upload the results under that job or something? Want to make sure the playground jobs do not mess with actual distribution jobs.

The original case being playground build jobs were not able to actual have playground integ test jobs.  It uses the build job name to access the distribution but the build job name since the distribution gets nested under `Playground/buildjobname`. The error would be similar to what you are seeing when trying to re run it. It will try to access a distribution in a location that doesn't exist and then throw an error. Without it then playground build job names will need to hardcode their build job name for the tests to pass.

PR #1866, opened 2022-03-30
Change consumers of knn zip to expect libs in lib directory

jmazanec15:

### Description
Updates all consumers of k-NN zip to expect the libs in the lib directory as opposed to knnlib. 

Question:  Is [this file](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/legacy/tar/linux/opensearch-tar-build.sh) still in use? I did not update it because for k-NN we have a separate build script. 

Also, this PR is tied to this PR: https://github.com/opensearch-project/k-NN/pull/345. They need to be merged around the same time or else build will break.
 
### Issues Resolved
https://github.com/opensearch-project/k-NN/issues/314
 
### Check List

### DIFF
--- a/docker/release/dockerfiles/opensearch.al2.dockerfile
+++ b/docker/release/dockerfiles/opensearch.al2.dockerfile
@@ -75,7 +75,7 @@ ENV JAVA_HOME=$OPENSEARCH_HOME/jdk
 ENV PATH=$PATH:$JAVA_HOME/bin:$OPENSEARCH_HOME/bin
 
 # Add k-NN lib directory to library loading path variable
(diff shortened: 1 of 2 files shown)


peterzhuamazon:
> To answer your question @jmazanec15 , that file is not used today for building tarball. @opensearch-project/engineering-effectiveness maybe we should deprecate it for coming major as a clean up to the codebase.

it is in use @gaiksaya right here: https://github.com/opensearch-project/opensearch-build/blob/main/scripts/components/OpenSearch/install.sh#L82

jmazanec15:
@peterzhuamazon 

> Also, this PR is tied to this PR: https://github.com/opensearch-project/k-NN/pull/345. They need to be merged around the same time or else build will break.

So this PR should address that, but I wanted to merge them around the same time to prevent build breaking. Not sure if there is a better way to go about this

PR #1905, opened 2022-04-05
[OSD][Build] nvm use

kavilla:
### Description
Within OSD we have a .nvmrc file that is used when `nvm use`
is executing. So we can use the right node version for the
build.

Issue:
n/a

### Issues Resolved
n/a
 
### Check List

### DIFF
--- a/scripts/components/OpenSearch-Dashboards/build.sh
+++ b/scripts/components/OpenSearch-Dashboards/build.sh
@@ -5,6 +5,8 @@
 
 set -ex
 
+source $NVM_DIR/nvm.sh
+
 function usage() {
     echo "Usage: $0 [args]"
     echo ""
--- a/scripts/components/OpenSearch-Dashboards/build.sh
+++ b/scripts/components/OpenSearch-Dashboards/build.sh
@@ -109,6 +109,9 @@ case $PLATFORM-$DISTRIBUTION-$ARCHITECTURE in
         ;;
 esac
 
+echo "Setting node version"
+nvm use
+
 echo "Building node modules for core with $PLATFORM-$DISTRIBUTION-$ARCHITECTURE"
 yarn osd bootstrap
 
--- a/scripts/components/OpenSearch-Dashboards/build.sh
+++ b/scripts/components/OpenSearch-Dashboards/build.sh
@@ -5,8 +5,6 @@
 
 set -ex
 
-source $NVM_DIR/nvm.sh
-
 function usage() {
     echo "Usage: $0 [args]"
     echo ""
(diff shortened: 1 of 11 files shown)


kavilla:
Jenkins is doing stuff that is preventing us from passing the environment variable and running the command that was added https://issues.jenkins.io/browse/JENKINS-51307. Therefore the current node version is being used and having some failures while building `2.0.0`. 

This also doesn't work because `nvm` isn't available but I think we need to add nvm to the PATH and have the docker image published so that it can be used here: https://github.com/opensearch-project/opensearch-build/blob/main/docker/ci/dockerfiles/build.centos7.opensearch-dashboards.x64.arm64.dockerfile#L151

kavilla:
> > Jenkins is doing stuff that is preventing us from passing the environment variable and running the command that was added https://issues.jenkins.io/browse/JENKINS-51307. Therefore the current node version is being used and having some failures while building `2.0.0`.
> > This also doesn't work because `nvm` isn't available but I think we need to add nvm to the PATH and have the docker image published so that it can be used here: https://github.com/opensearch-project/opensearch-build/blob/main/docker/ci/dockerfiles/build.centos7.opensearch-dashboards.x64.arm64.dockerfile#L151
> 
> This does work if you can source the nvm.sh before running nvm.
> 
> ```
> source $NVM_DIR/nvm.sh
> ```

I understand the NVM docs call out to define $NVM_DIR when setting it up but I'm not seeing the preference to source a script from an environment variable that may not exist rather than adding it to the PATH. Would seem like an anti pattern for the repo as well.

kavilla:
If I source $NVM_DIR will make sure to update https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#nvm-and-node to ensure users define it in the environment.

kavilla:
The builds for 2.0.0 have been broken for a couple days now so to unblock this I have created a follow up issue: https://github.com/opensearch-project/opensearch-build/issues/1922 since I still believe adding to `PATH` fits into this repo more.

scripts/components/OpenSearch-Dashboards/build.sh
```diff
@@ -109,6 +109,10 @@ case $PLATFORM-$DISTRIBUTION-$ARCHITECTURE in
         ;;
 esac
 
+echo "Setting node version"
+source $NVM_DIR/nvm.sh
+nvm use
```
peternied, line 114: Add a check if $NVM_DIR isn't defined then send them to the setup guide - this would be a good fast follow

peternied (approved):

scripts/components/OpenSearch-Dashboards/build.sh
kavilla: Oh yes that would actually be a really clean solution that make me feel better about this code being in here. I've created an issue https://github.com/opensearch-project/opensearch-build/issues/1922 but trying my best not to be online so I might not be able to get this as a fast follow.

peterzhuamazon (approved):

Issue #1498, opened 2022-01-18
Add support for creating multiple cluster and setup CCR


ankitkala:
### Changes required
- Create multiple single node cluster.
- Pass multiple endpoint to PerfSuite.
- Connect two cluster via seed nodes setting and configure replication.

PR #1907, opened 2022-04-05
Support for Plugin Performance tests

ankitkala:
### Description
This PR contains changes to allow triggering a performance suite for Opensearch plugin . 
The plugin owner is required to have a script `` in their repo which can be invoked by this package. Using this, we can invoke Jenkin CI for periodic perf runs for different plugins.This approach also allows plugin owners to reuse the existing infra to initialise the test clusters and test suite while ensuring that the plugin specific logic resides with the plugin owners.  

### Changes which are not included in this PR: 
- Jenkin file to trigger the test run.
- Support for publishing test results.
- Support for multi cluster in `PerfTestCluster`
 
### Related change
[155](https://github.com/opensearch-project/opensearch-infra/pull/155)

### Issues Resolved
[1498](https://github.com/opensearch-project/opensearch-build/issues/1498)
[146](https://github.com/opensearch-project/opensearch-infra/issues/146)

 
### Check List

### DIFF
--- a/src/test_workflow/perf_test/perf_test_cluster.py
+++ b/src/test_workflow/perf_test/perf_test_cluster.py
@@ -2,6 +2,11 @@ import json
 import logging
 import os
 import subprocess
+import time
+from contextlib import contextmanager
+
+import requests
+from requests.auth import HTTPBasicAuth
 
 from test_workflow.test_cluster import TestCluster
 
@@ -13,14 +18,14 @@ class PerfTestCluster(TestCluster):
 
     def __init__(self, bundle_manifest, config, stack_name, security, current_workspace):
         self.manifest = bundle_manifest
-        self.work_dir = os.path.join("opensearch-cluster", "cdk", "single-node")
+        self.work_dir = os.path.join(current_workspace, "opensearch-cluster", "cdk", "single-node")
         self.current_workspace = current_workspace
         self.stack_name = stack_name
         self.cluster_endpoint = None
         self.cluster_port = None
         self.output_file = "output.json"
-        self.ip_address = None
-        self.security = "enable" if security else "disable"
+        self.private_ip = None
+        self.security = security
         role = config["Constants"]["Role"]
         params_dict = {
             "url": self.manifest.build.location,
@@ -29,9 +34,10 @@ class PerfTestCluster(TestCluster):
             "account_id": config["Constants"]["AccountId"],
             "region": config["Constants"]["Region"],
             "stack_name": self.stack_name,
-            "security": self.security,
+            "security": "enable" if self.security else "disable",
             "platform": self.manifest.build.platform,
             "architecture": self.manifest.build.architecture,
+            "public_ip": config["Constants"].get("PublicIp", "disable")
         }
         params_list = []
         for key, value in params_dict.items():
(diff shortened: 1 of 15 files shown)


src/test_workflow/perf_test/perf_test_cluster.py
```diff
@@ -71,3 +85,34 @@ def service(self):
 
     def dependencies(self):
         return []
+
+    def wait_for_processing(self, max_wait_sec = 60):
+        if self.public_ip is None:
+            return
+        url = "".join([self.endpoint(), "/_cluster/health"])
+        end_time = time.time() + max_wait_sec
+        while(True):
```
kotwanikunal, line 94: It might be cleaner to use https://pypi.org/project/retry/. We already have this in our current dependencies and is being utilized here - https://github.com/opensearch-project/opensearch-build/blob/main/src/run_perf_test.py#L68

src/test_workflow/perf_test/perf_test_runner.py
```diff
@@ -0,0 +1,27 @@
...
+import os
+
+from manifests.bundle_manifest import BundleManifest
+from test_workflow.perf_test.perf_args import PerfArgs
+from test_workflow.test_recorder.test_recorder import TestRecorder
+from test_workflow.test_result.test_suite_results import TestSuiteResults
+
+
+class PerfTestRunner(abc.ABC):
+    def __init__(self, args: PerfArgs, test_manifest: BundleManifest):
```
kotwanikunal, line 18: Is the test manifest here referring to the distribution/bundle manifest?

ankitkala: yes.

src/test_workflow/perf_test/perf_test_cluster.py
ankitkala: Sure.

kotwanikunal (approved):

tests/tests_test_workflow/test_perf_workflow/perf_test/test_perf_test_suite.py
```diff
@@ -22,14 +22,48 @@ def setUp(self):
...
+                " -b 41d5ae25183d4e699e92debfbe3f83bd -a x64 -p test/results/ --workload nyc_taxis" \
+                " --workload-options '{}' --warmup-iters 0 --test-iters 1 --scenario-type CROSS_CLUSTER_REPLICATION" \
+                " --owner test-owner")
+
+    def test_execute_with_dict_endpoint(self):
+        endpoint = {
+            "default": ["cluster-1"],
+            "second_cluster": ["cluster-2"]
+        }
+        perf_test_suite = PerfTestSuite(bundle_manifest=self.manifest, endpoint=endpoint, security=False,
```
gaiksaya, line 60: Should we also add test with security=True, just to see if the workload options are loaded as expected?

gaiksaya (approved):

tests/tests_test_workflow/test_perf_workflow/perf_test/test_perf_test_suite.py
ankitkala: Workload options in perf suite would be same irrespective of the security True/False
Let me know if you still think we should add one more test with `security=True`

Issue #1877, opened 2022-03-31
[Enhancement] Create a groovy src library for distribution manifest


zelinh:
### Is your feature request related to a problem? Please describe

Currently we only have `BuildManifest` and `InputManifest` Jenkins method to process and fetch information from either build manifest or the manifest for ci pipeline. Although we can use `BuildManifest` to process the data for the distribution bundle manifest but it still lacks of some info like `location`. We should have a Jenkins library specifically for distribution manifest for those who want to use. 
Example distributon manifest from 1.3.1 release: https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.3.1/1584/linux/x64/dist/opensearch/manifest.yml

### Acceptance criteria

- [ ] We have a Jenkins library `DistributionManifest.groovy` used to process info from distribution manifest.

### Describe alternatives you've considered

Keep using the library based on `BuildManifest.groovy` but makes it more general for both types of manifest.

abhinavGupta16:
[triage] It's a good idea to separate the manifests if it has a use case. We'll need to update the tag creation job once this is complete to use the `DistributionManifest.groovy`

zelinh:
Closing this issue as we now have this library available.

PR #1807, opened 2022-03-23
Support subfolders for distribution types for build and assemble locally and in Jenkins

tianleh:
Signed-off-by: Tianle Huang <tianleh@amazon.com>

### Description
Support subfolders for distribution types. E.g

current `builds/opensearch/dist/` -> `builds/opensearch/tar/dist/`
current `dist/opensearch/` -> `dist/opensearch/rpm/`
 
### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/1800
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #1555, opened 2022-01-25
[RPM M6] Validate package installing


peterzhuamazon:
Tasks | Estimate | Status | Notes | Dependencies
-- | -- | -- | -- | --
Install rpm through local yum repository | 1 | No Plan |   | All of these requires us to create pipelines to automatically run these tests for validations
Install rpm through remote yum repository | 1 | Complete |   | All of these requires us to create pipelines to automatically run these tests for validations
Install rpm through installing the package directly without yum | 1 | Complete |   | All of these requires us to create pipelines to automatically run these tests for validations

* https://github.com/opensearch-project/opensearch-build/pull/1912

zelinh:
Our automatic RPM distribution have mostly completed for OpenSearch distribution and we are working on the validation for the OpenSearch Dashboards distribution. Since the dashboards needs OpenSearch installed and running for validation so we plan to use the latest opensearch rpm distribution. We would need a running distribution build workflow for latest OpenSearch distribution to run the validation for dashboards. We are actively working on this.

peterzhuamazon:
Add systemd for rpm to run on it: https://github.com/opensearch-project/opensearch-build/pull/1958

peterzhuamazon:
@zelinh will add jenkinsfile to use the libs committed in #1912.

peterzhuamazon:
Add to distribution:
* https://github.com/opensearch-project/opensearch-build/pull/2021

PR #1912, opened 2022-04-06
Add the distribution validation for RPM

zelinh:

### Description
Add a validation for rpm distribution. This PR will need to follow and be adjusted with https://github.com/opensearch-project/opensearch-build/pull/1807
Given parameters: 
- Workspace location for the distribution manifest. 
- Workspace location of the RPM distribution for either OpenSearch or OpenSearch-Dashboards.

Validation details: 

For both OpenSearch & OpenSearch Dashboards:
- Validate the name convention for the distribution files.
- Validate the meta data of the distribution by running `rpm -qip $distFile` and compare the info with that from the distribution manifest.
- Validate the installation of the distribution with `sudo yum install`. For OpenSearch Dashboards, we will download the distribution of OpenSearch from latest URL with same version.
- Validate the status of service is actively running by `sudo systemctl status $opensearch/$opensearch-dashboards`

For OpenSearch:
- Validate the certs ("esnode-key.pem", "kirk.pem", "esnode.pem", "kirk-key.pem", "root-ca.pem") are existing in `/etc/opensearch/` folder after installation.
- Validate the `install_demo_configuration.log` exists in `/var/log/opensearch/` and it's showing `Success`.
-  After OpenSearch service starts, check the status of it using `curl -s "https://localhost:9200" -u admin:admin --insecure` and see if the name, version and type match the manifest.
- Check the status of cluster using `curl -s "https://localhost:9200/_cluster/health?pretty" -u admin:admin --insecure` and validate if the cluster name matches and the status is green.
- Validate all the components from the manifest are started from `curl -s "https://localhost:9200/_cat/plugins?v" -u admin:admin --insecure`.

For OpenSearch-Dashboards:
- After service starts, validate the name from status matches the name from the manifest and the overall state is green. `curl -s "http://localhost:5601/api/status"`
- Validate all components from the manifest are included in the plugin. `/usr/share/opensearch-dashboards/bin/opensearch-dashboards-plugin list`

This PR also include an initial version of `DistributionManifest.groovy` Jenkins library which we can elaborate on.

 
### Issues Resolved
Part of #1555 
First version of #1877
 
### Check List


src/jenkins/DistributionManifest.groovy
```diff
@@ -0,0 +1,101 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+
+package jenkins
+
+
+class DistributionManifest implements Serializable {
```
dblock, line 13: In Python we call this a bundle manifest, https://github.com/opensearch-project/opensearch-build/blob/main/src/manifests/bundle_manifest.py. I like "distribution manifest" a lot better, but let's be consistent? If you want to take on renaming everything "bundle" to "distribution" elsewhere that's A-OK too.

dblock (commented):
The amount of "if"'s in `rpmDistValidation" tells me that we need separate implementations for rpm distribution validation for different products with common parts extracted in classes. This looks like it will be difficult to maintain.

src/jenkins/DistributionManifest.groovy
zelinh: Rename the Jenkins library to `BundleManifest` to be consistent with what we had for python.

zelinh:
Separated the original `rpmDistValidation` to `rpmOpenSearchDistValidation` and `rpmDashboardsDistValidation` for different product type.

vars/rpmDashboardsDistValidation.groovy
```diff
@@ -0,0 +1,127 @@
...
+
+    def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+    def bundleManifestObj = args.bundleManifestObj
+    def distFile = args.rpmDistribution         //Distribution file location
+    def name = bundleManifestObj.build.getFilename()   //opensearch-dashboards
+    def version = bundleManifestObj.build.version        //1.3.0
+    def architecture = bundleManifestObj.build.architecture
+    def plugin_names = bundleManifestObj.getNames();
+
+    def latestOpenSearchURL = "https://ci.opensearch.org/ci/dbc/Playground/tianleh-test/tianle-opensearch-build-3-22/$version/latest/linux/$architecture/rpm/dist/opensearch/opensearch-$version-linux-${architecture}.rpm"
```
dblock, line 11: Will need changing, externalize the root part of this? (https://ci.opensearch.org/ci/dbc/Playground/tianleh-test/tianle-opensearch-build-3-22)

vars/rpmDashboardsDistValidation.groovy
```diff
@@ -0,0 +1,127 @@
+def call(Map args = [:]) {
+
+    def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+    def bundleManifestObj = args.bundleManifestObj
+    def distFile = args.rpmDistribution         //Distribution file location
```
dblock, line 5: Document args above the function instead.

vars/rpmDashboardsDistValidation.groovy
```diff
@@ -0,0 +1,127 @@
...
+    refMap['License'] = "Apache-2.0"
+    refMap['Relocations'] = "(not relocatable)"
+    refMap['URL'] = "https://opensearch.org/"
+    // The context of meta data should be for OSD
+    refMap['Summary'] = "Open source visualization dashboards for OpenSearch"
+    refMap['Description'] = "OpenSearch Dashboards is the visualization tool for data in OpenSearch\n" +
+            "For more information, see: https://opensearch.org/"
+
+    //Validation for the Meta Data of distribution
+    println("Meta data validations start:")
```
dblock, line 31: I would extract metadata validation into a class and write tests.

vars/rpmDashboardsDistValidation.groovy
```diff
@@ -0,0 +1,127 @@
...
+    //Install the rpm distribution via yum
+    println("Start installation with yum.")
+    sh "sudo yum install -y $distFile"
+    println("RPM distribution for $name is installed with yum.")
+    sh "sudo yum install -y $latestOpensearchDist"
+    println("Latest RPM distribution for OpenSearch is also installed with yum.")
+
+    //Start the installed OpenSearch-Dashboards distribution
+    sh ("sudo systemctl restart $name")
+    sleep 30
```
dblock, line 73: Is there a better way to do this than just wait long enough? try to hit the endpoint in a loop in a `waitForService [url]` function? You could simplify both waiting for OS and OSD by just giving a URL and looking for certain text with a timeout.

vars/rpmDashboardsDistValidation.groovy
```diff
@@ -0,0 +1,127 @@
...
+    for (component in components_list) {
+        def component_with_version = component.replace("-","@") + ".0"  //e.g. anomalyDetectionDashboards@1.3.0.0
+        assert osd_plugins.contains(component_with_version)
+        println("Component $component is present with correct version $version." )
+    }
+
+    sh ("sudo systemctl stop opensearch-dashboards")
+    sh ("sudo yum remove -y opensearch-dashboards")
+    sh ("sudo systemctl stop opensearch")
+    sh ("sudo yum remove -y opensearch")
```
dblock, line 126: Why is this cleanup needed? Don't we throw away the docker container anyway?

dblock (commented):
Still tons of duplication, I would work on extracting some of these further, but lmk when you feel like it's too much and want to merge this.

vars/rpmDashboardsDistValidation.groovy
zelinh: Currently this is not supported running in a docker container since our docker CI container doesn't support using `systemctl` cc: @peterzhuamazon

vars/rpmDashboardsDistValidation.groovy
zelinh: I will investigate on this suggestion

vars/rpmDashboardsDistValidation.groovy
zelinh: Yes. This is a temporary distribution URL and need to be updated once `distribution-build` running for rpm.

vars/rpmDashboardsDistValidation.groovy
zelinh: Makes sense to me. Will be working on this.

gaiksaya:
I am confused about the language and implementation in groovy. 
All our building/testing code is and will be written in python. Groovy is just used as a bridge between python code and our CI workflows.
Are we going on this path for future validation tests as well?
cc: @bbarani @dblock @abhinavGupta16

vars/rpmDashboardsDistValidation.groovy
tianleh: You can refer to this example https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/service.py#L83

peterzhuamazon:
After sync up with @zelinh we will use yum repo to test this once #1960 merged.

peterzhuamazon:
@zelinh due to #1966 your test should also replace `-` with `.` if there is qualifier.

peterzhuamazon (approved):

PR #1921, opened 2022-04-07
Support release iteration value for assemble workflow rpm

### LINKED ISSUE #1545: [RPM M1] Add a new block to call the generation code for RPM

peterzhuamazon:

### Description
Support release iteration value for assemble workflow rpm.

This PR add support for release/iteration value in rpm/deb packages.

It allows us to build packages with the same version without overriding existing packages.
It also allow users to upgrade the package to the same version but latest release/iteration.

```
-rw-r--r-- 1 <> 390033760 Apr  7 02:09 opensearch-1.3.0-1-linux-x64.rpm
-rw-r--r-- 1 <> 390033852 Apr  7 01:59 opensearch-1.3.0-2-linux-x64.rpm
-rw-r--r-- 1 <> 386532844 Apr  7 02:04 opensearch-1.3.1-1-linux-x64.rpm
```
```
opensearch-1.3.0-1.x86_64.rpm
Name        : opensearch
Version     : 1.3.0
Release     : 1
Architecture: x86_64

opensearch-1.3.0-2.x86_64.rpm
Name        : opensearch
Version     : 1.3.0
Release     : 2
Architecture: x86_64
```

It also allows us to assign the $BUILD_NUMBER to release/iteration in Jenkins so we can separate then when pushing to staging yum repo for testing.

Thanks.

 
### Issues Resolved
#1545
 
### Check List

### DIFF
--- a/src/assemble_workflow/assemble_args.py
+++ b/src/assemble_workflow/assemble_args.py
@@ -8,15 +8,29 @@ import argparse
 import logging
 from typing import IO
 
-
 class AssembleArgs:
     manifest: IO
     keep: bool
 
     def __init__(self) -> None:
+
+        def check_positive(value) -> int:
+            int_value = int(value)
+            if int_value <= 0:
+                raise argparse.ArgumentTypeError(f"{value} is invalid, you must enter a positive integer")
+            return str(int_value)
+
         parser = argparse.ArgumentParser(description="Assemble an OpenSearch Distribution")
         parser.add_argument("manifest", type=argparse.FileType("r"), help="Manifest file.")
         parser.add_argument("-b", "--base-url", dest="base_url", help="The base url to download the artifacts.")
+        parser.add_argument(
+            "-r",
+            "--release-iteration",
+            dest="release_iteration",
+            type=check_positive,
+            default="1",
+            help="The release/iteration number of deb/rpm packages, allow multiple revision of same package version (e.g. 2.0.0-1, 2.0.0-2, 2.0.0-3)"
+        )
         parser.add_argument(
             "--keep",
             dest="keep",
@@ -38,3 +52,4 @@ class AssembleArgs:
         self.manifest = args.manifest
         self.keep = args.keep
         self.base_url = args.base_url
+        self.release_iteration = args.release_iteration
--- a/scripts/pkg/build_templates/opensearch-dashboards/opensearch-dashboards.rpm.spec
+++ b/scripts/pkg/build_templates/opensearch-dashboards/opensearch-dashboards.rpm.spec
@@ -11,6 +11,7 @@
 %define log_dir %{_localstatedir}/log/%{name}
 %define pid_dir %{_localstatedir}/run/%{name}
 %{!?_version: %define _version 0.0.0 }
+%{!?_release: %define _release 1 }
 %{!?_architecture: %define _architecture x86_64 }
 
 Name: opensearch-dashboards
--- a/scripts/pkg/build_templates/opensearch/opensearch.rpm.spec
+++ b/scripts/pkg/build_templates/opensearch/opensearch.rpm.spec
@@ -11,11 +11,12 @@
 %define log_dir %{_localstatedir}/log/%{name}
 %define pid_dir %{_localstatedir}/run/%{name}
 %{!?_version: %define _version 0.0.0 }
+%{!?_release: %define _release 1 }
 %{!?_architecture: %define _architecture x86_64 }
 
 Name: opensearch
 Version: %{_version}
-Release: 1
+Release: %{_release}
 License: Apache-2.0
 Summary: An open source distributed and RESTful search engine
 URL: https://opensearch.org/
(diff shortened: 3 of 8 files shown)


peterzhuamazon:
@spotrh I have a question for you on the release value.

Lets say we build rpm for $BUILD_NUMBER 123 in Jenkins.
If we assign `123` to release then it will show:
```
opensearch-1.3.0-123.x86_64.rpm
Name        : opensearch
Version     : 1.3.0
Release     : 123
Architecture: x86_64

```

If we release this directly as the 1st 1.3.0 release, would this be a problem after all?
To my understanding there is no technical issues.
Just the 1st release of 1.3.0 being `123` release is a bit weird.

Thanks.

peterzhuamazon:
Hi @dblock @tianleh,

Please review this I will add test cases later.

Thanks.

src/assemble_workflow/assemble_args.py
```diff
@@ -8,15 +8,29 @@
...
         parser = argparse.ArgumentParser(description="Assemble an OpenSearch Distribution")
         parser.add_argument("manifest", type=argparse.FileType("r"), help="Manifest file.")
         parser.add_argument("-b", "--base-url", dest="base_url", help="The base url to download the artifacts.")
+        parser.add_argument(
+            "-r",
+            "--release-iteration",
+            dest="release_iteration",
+            type=check_positive,
+            default="1",
+            help="The release/iteration number of deb/rpm packages, allow multiple revision of same package version (e.g. 2.0.0-1, 2.0.0-2, 2.0.0-3)"
```
tianleh, line 32: revision -> revisions

src/assemble_workflow/assemble_args.py
```diff
@@ -8,15 +8,29 @@
...
 from typing import IO
 
-
 class AssembleArgs:
     manifest: IO
     keep: bool
 
     def __init__(self) -> None:
+
+        def check_positive(value) -> int:
```
tianleh, line 17: You can add type for the input. `value: str`

peterzhuamazon: This does not allow me to block non-positive inputs.

spotrh:
> Hi @spotrh hope you can help us clear some confusions.
> 
> Let' s say we build rpm for `$BUILD_NUMBER=123` in Jenkins. If we assign `123` to release then it will show:
> 
> ```
> opensearch-1.3.0-123.x86_64.rpm
> Name        : opensearch
> Version     : 1.3.0
> Release     : 123
> Architecture: x86_64
> ```
> 
> If we release this directly as the 1st 1.3.0 release, would this be a problem after all? To my understanding there is no technical issues. Just the 1st release of 1.3.0 being `123` release is a bit weird.

The "Release" field is intended to be incremented when changes are made to the package that do not involve updating to a newer version of the software. When Version increments, you should reset Release (to prevent it from growing absurdly large)

I'm not sure why you'd want to set Release to "123" as a starting value, but it's not harmful.

peterzhuamazon:
> > Hi @spotrh hope you can help us clear some confusions.
> > Let' s say we build rpm for `$BUILD_NUMBER=123` in Jenkins. If we assign `123` to release then it will show:
> > ```
> > opensearch-1.3.0-123.x86_64.rpm
> > Name        : opensearch
> > Version     : 1.3.0
> > Release     : 123
> > Architecture: x86_64
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If we release this directly as the 1st 1.3.0 release, would this be a problem after all? To my understanding there is no technical issues. Just the 1st release of 1.3.0 being `123` release is a bit weird.
> 
> The "Release" field is intended to be incremented when changes are made to the package that do not involve updating to a newer version of the software. When Version increments, you should reset Release (to prevent it from growing absurdly large)
> 
> I'm not sure why you'd want to set Release to "123" as a starting value, but it's not harmful.

I think I misunderstand some concert of my upgrade design.
I will still allow this change to go in but will use release=1 for each build.

Thanks.

src/assemble_workflow/assemble_args.py
dblock, line 17: Make this method private.

src/assemble_workflow/assemble_args.py
dblock, line 32: Help missing period, is too long and has grammar issues. Just say "Release number."

dblock (changes requested):
The fact that this is a new parameter that is only specific to deb/rpm, but is passed into a generic workflow is a red flag to me. Why do we need this at all? We don't vary build numbers for .tar.gz, so why are we trying to vary build numbers for deb? 

- Upgrading between builds is not a supported scenario.
- We distinguish between builds by placing output in different build folders (with job number) via Jenkins, so why do we want to do this for rpm?

If you want to vary the build number, we should do it for all artifacts. Every build would generate a patch number based on, for example, date and time, e.g. 2.0.0.4071558 (April 7, 3:58).

peterzhuamazon:
After some review I think this PR is not needed anymore.
I have figured out a better way to test upgrade across different builds using latest.

Thanks.

Issue #1926, opened 2022-04-07
[BUG][DISTRIBUTION][2.0.0] Build.sh does not support to use qualifier yet


peterzhuamazon:
https://github.com/opensearch-project/opensearch-build/blob/main/scripts/components/security/build.sh
This file can be changed base on this:
https://github.com/opensearch-project/ml-commons/blob/c8f8396e02168b0bb34bffcd48ddc4b80069a99e/scripts/build.sh

```
2022-04-07 16:21:39 INFO     Executing "git rev-parse HEAD" in /tmp/tmpc9l3i2v7/security
2022-04-07 16:21:39 INFO     Checked out https://github.com/opensearch-project/security.git@main into /tmp/tmpc9l3i2v7/security at 274408130b315de313ea04afae4ddff70ba34352
2022-04-07 16:21:39 INFO     Executing "bash /var/jenkins/workspace/distribution-build-opensearch/scripts/components/security/build.sh -v 2.0.0 -q alpha1 -p linux -a arm64 -s true -o builds" in /tmp/tmpc9l3i2v7/security
+ getopts :h:v:s:o:p:a: arg
+ case $arg in
+ VERSION=2.0.0
+ getopts :h:v:s:o:p:a: arg
+ case $arg in
+ echo 'Invalid option: -?'
Invalid option: -?
+ exit 1
2022-04-07 16:21:39 ERROR    Error building security, retry with: ./build.sh manifests/2.0.0/opensearch-2.0.0.yml --component security --snapshot
Traceback (most recent call last):
  File "./src/run_build.py", line 80, in <module>
    sys.exit(main())
  File "./src/run_build.py", line 68, in main
    builder.build(build_recorder)
  File "/var/jenkins/workspace/distribution-build-opensearch/src/build_workflow/builder_from_source.py", line 55, in build
    self.git_repo.execute(build_command)
  File "/var/jenkins/workspace/distribution-build-opensearch/src/git/git_repository.py", line 83, in execute
    subprocess.check_call(command, cwd=cwd, shell=True)
  File "/usr/local/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command 'bash /var/jenkins/workspace/distribution-build-opensearch/scripts/components/security/build.sh -v 2.0.0 -q alpha1 -p linux -a arm64 -s true -o builds' returned non-zero exit status 1.
```

PR #1927, opened 2022-04-07
Remove the custom security build script

peternied:
### Description
Remove the custom security build script

The build was failing because this script was out of date, by removing this script we can use the default script which I have tested locally

### Issue
* Resolves https://github.com/opensearch-project/opensearch-build/issues/1926

### Check List


peterzhuamazon (changes requested):
Notifications is going to remove their custom script soon and move to their own repo.
Please avoid using notifications here.

Thanks.

peterzhuamazon:
> > Notifications is going to remove their custom script soon and move to their own repo. Please avoid using notifications here.
> > Thanks.
> 
> Please give me guidance on how to fix these tests correctly. Switching to the default script doesn't seem to be correct. I am not familiar with the code base to know if this is correct or not. I am just trying to fix the security plugin build break.

You can create a dummy component in `tests/data` to test this feature.
We are going to move away from saving the custom scripts in component folder.
Switching from `security` to `notifications` is just a temp patch that wont run later.

peternied:
Looks like https://github.com/opensearch-project/opensearch-build/pull/1929 will resolve the unit test issue, thanks @peterzhuamazon

peterzhuamazon (approved):

Issue #1950, opened 2022-04-11
Supporting installation of multiple artifacts for plugin components (related to Notifications)


qreshi:
### Is your feature request related to a problem? Please describe

In preparation for 2.0 Preview, we'd like to add the Notifications plugin to the manifest so it begins being bundled in the distribution. The Notifications backend is a bit of an uncommon case in that there are two plugins (`opensearch-notifications` and `opensearch-notifications-core`) that are being produced from the same repo. The latter needs to be installed first since the former extends it and both need to be installed to use Notifications. The current build process does not support this in a way that's as simple as just adding the repo to the manifest and ensuring the plugin build scripts produce the artifacts to the output directory.

[More information is in the 'additional context' section below]

### Describe the solution you'd like

We'd like for the infra build process to ensure that these two plugins are installed and in the correct order to ensure correct assembly of the distribution.

### Additional context

I checked out `opensearch-build` locally and tried testing the building and assembly process with Notifications. After adding it to the manifest file, I ran:

```
./build.sh manifests/2.0.0/opensearch-2.0.0.yml 
```

The resulting output file showed that the Notification component was added with multiple plugin artifacts:
```
  - name: notifications
    repository: https://github.com/opensearch-project/notifications
    ref: main
    commit_id: 731ee820a36edc25bac07a5b6a624c35d1615ce5
    artifacts:
      plugins:
        - plugins/opensearch-notifications-core-2.0.0.0-alpha1.zip
        - plugins/opensearch-notifications-2.0.0.0-alpha1.zip
    version: 2.0.0.0-alpha1
```

**Question**: Is the order of this list of plugins above random? (We'd like it to always be the order that it's in above if the plugins are iterated through in this order)

After building, I attempted to assemble:

```
./assemble.sh tar/builds/opensearch/manifest.yml
```

At this point, it only installed the first plugin (`opensearch-notifications-core`) and completed:
```
2022-04-11 02:57:45 INFO     Recording common-utils
2022-04-11 02:57:45 INFO     Installing notifications
2022-04-11 02:57:45 INFO     Executing "/tmp/tmpqueccp_h/opensearch-2.0.0-alpha1/bin/opensearch-plugin install --batch file:/tmp/tmpqueccp_h/opensearch-notifications-core-2.0.0.0-alpha1.zip" in /tmp/tmpqueccp_h/opensearch-2.0.0-alpha1
[...]

PR #1956, opened 2022-04-12
Support multiple artifacts per bundled component.

dblock:

### Description

Added support for multiple artifacts per distribution component. 

With this change, the notifications plugin collects 2 ZIPs instead of 1, publishes and installs both.

#### Build Manifest

```
  - name: notifications
    repository: https://github.com/opensearch-project/notifications.git
    ref: main
    commit_id: 731ee820a36edc25bac07a5b6a624c35d1615ce5
    artifacts:
      plugins:
        - plugins/opensearch-notifications-core-2.0.0.0-alpha1-SNAPSHOT.zip
        - plugins/opensearch-notifications-2.0.0.0-alpha1-SNAPSHOT.zip
```

#### Bundle Manifest

```
  - name: notifications
    repository: https://github.com/opensearch-project/notifications.git
    ref: main
    commit_id: 731ee820a36edc25bac07a5b6a624c35d1615ce5
    locations:
      - /home/ubuntu/source/opensearch-project/opensearch-build/dblock-opensearch-build/tar/builds/opensearch/plugins/opensearch-notifications-core-2.0.0.0-alpha1-SNAPSHOT.zip
      - /home/ubuntu/source/opensearch-project/opensearch-build/dblock-opensearch-build/tar/builds/opensearch/plugins/opensearch-notifications-2.0.0.0-alpha1-SNAPSHOT.zip
```

It doesn't seem like the order of installation between `notifications` and `notifications-core` matters, but it could be that they are installed in the right order "accidentally". If there's an issue with this I can fix it in a future PR.
 
### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/1950.

Will need to move the custom `build.sh` into notifications repo as `script/build.sh` after merging.
 
### Check List


dblock:
@qreshi

qreshi:
The scripts have actually already been moved to Notifications, there are just two minor PRs open to make some changes to them (https://github.com/opensearch-project/notifications/pull/400 and https://github.com/opensearch-project/notifications/pull/401). So we can merge those in and remove the scripts in this PR if you like.

Also, the installation order does matter. I also brought this up in the tracking issue and you mentioned it here as well, it could be that the build manifest being outputted here happened to list the plugins in order:
```
    artifacts:
      plugins:
        - plugins/opensearch-notifications-core-2.0.0.0-alpha1-SNAPSHOT.zip
        - plugins/opensearch-notifications-2.0.0.0-alpha1-SNAPSHOT.zip
```

But is this order deterministic?

qreshi:
Also @dblock I think @peterzhuamazon was looking into the option of listing separate entries in the manifest and supplying a different `working_directory` for each (since both of those plugins have their own subfolders). If that works, it would guarantee build and installation order so it's worth checking with him if we want to go with that solution.

dblock:
> The scripts have actually already been moved to Notifications, there are just two minor PRs open to make some changes to them ([opensearch-project/notifications#400](https://github.com/opensearch-project/notifications/pull/400) and [opensearch-project/notifications#401](https://github.com/opensearch-project/notifications/pull/401)). So we can merge those in and remove the scripts in this PR if you like.

Ok, I see, deleting this from this PR.

> Also, the installation order does matter. I also brought this up in the tracking issue and you mentioned it here as well, it could be that the build manifest being outputted here happened to list the plugins in order:
> 
> ```
>     artifacts:
>       plugins:
>         - plugins/opensearch-notifications-core-2.0.0.0-alpha1-SNAPSHOT.zip
>         - plugins/opensearch-notifications-2.0.0.0-alpha1-SNAPSHOT.zip
> ```
> 
> But is this order deterministic?

I think it's repeatable in the order files are written, but I don't think we can rely on it.

dblock:
> Also @dblock I think @peterzhuamazon was looking into the option of listing separate entries in the manifest and supplying a different `working_directory` for each (since both of those plugins have their own subfolders). If that works, it would guarantee build and installation order so it's worth checking with him if we want to go with that solution.

I went down that rabbit hole first and it seemed like it's a smaller amount of changes. Then I thought that it would be great to support multiple artifacts produced by a single build anyway, so I did this change. 

@peterzhuamazon Did you make it work with `working_directory`? Can we compare the two PRs?

peterzhuamazon:
> > Also @dblock I think @peterzhuamazon was looking into the option of listing separate entries in the manifest and supplying a different `working_directory` for each (since both of those plugins have their own subfolders). If that works, it would guarantee build and installation order so it's worth checking with him if we want to go with that solution.
> 
> I went down that rabbit hole first and it seemed like it's a smaller amount of changes. Then I thought that it would be great to support multiple artifacts produced by a single build anyway, so I did this change.
> 
> @peterzhuamazon Did you make it work with `working_directory`? Can we compare the two PRs?

I was not sure if you are working on this issue.
If you are not I can take a look.
In theory two entries should run well with two different work directory.
Just need some tweak in the build.sh to take it.

dblock:
@peterzhuamazon https://github.com/opensearch-project/opensearch-build/pull/1957

What do we like better @peterzhuamazon @qreshi?

The pros of this PR is that we can publish multiple ZIPs per component with it. The cons is that it's a bigger change, the order is still an issue, and the bundle manifest has changed.

peterzhuamazon:
As this PR requires a lot more changes and probably will break a thing or two after the push, combined with the fact that notifications situation is a special case, I vote for #1970 as that is the minimum change for maximum outcome.

peterzhuamazon:
The #1970 not only allow use to define the order, but also resolved the issue without adding existing code to further complicated the code base for a one time use case.

dblock:
Closing in favor of https://github.com/opensearch-project/opensearch-build/pull/1957

PR #1957, opened 2022-04-12
Added notifications-core and notifications to 2.0 and add integTest manifest

dblock:

### Description

Simpler alternative to https://github.com/opensearch-project/opensearch-build/pull/1956. Requires https://github.com/opensearch-project/notifications/pull/403

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/1950.
 
### Check List


peterzhuamazon:
As discussed in #1956 we will go with #1957 for now.
This allows us to use existing code base and functionalities to resolve the problem without changing too much for this specific use case.

Thanks.

dblock:
CI was failing on checks because it tries to execute `./gradlew` in `notiications/core` which doesn't exist. I just removed the checks for now. Really notifications/core feels like a very weird hack all around.

peterzhuamazon:
I have tweaked the build.sh so it will build correctly now.

peterzhuamazon:
This will go in with another PR: https://github.com/opensearch-project/notifications/pull/404

peterzhuamazon:
Seems like notifications due to `name` field is not matching the repo name of `notifications`.
It is better to save the build scripts in build repo not notifications repo.

dblock:
> Seems like notifications due to `name` field is not matching the repo name of `notifications`. 

I don't think so. The check for `notifications-core` just failed with `gradlew` being not found, because there's none in `notifications/core`. Build works because we have `notifications/core/scripts`.

> It is better to save the build scripts in build repo not notifications repo.

Build scripts change version-to-version, so they need to be different and follow notifications branching. If you leave them in opensearch-build with these changes, 1.3.x will start failing.

peterzhuamazon (dismissed):
Block for a bit as we are still fixing the build scripts.
Due to the double subfolder structure of notifications the builds folder and plugins are not correctly copied to tar/builds/opensearch/plugins.

tianleh (approved):

gaiksaya:
@peterzhuamazon Can we add the checks back or are they still failing?

peterzhuamazon:
> @peterzhuamazon Can we add the checks back or are they still failing?

They will not pass as notifications structure is different from any other repos.

They need to build within `notifications` folder, but working directory is pointing to `notifications/core` and `notifications/notifications` and those two folders does not have gradle wrapper to run the tests.

peterzhuamazon:
@gaiksaya fixed in https://github.com/opensearch-project/opensearch-build/pull/1959

PR #1971, opened 2022-04-14
Disable security to unblock the RC1 build

peternied:
### Description
Disable security to unblock the RC1 build

> This is a chicken v. egg build problem. The security plugin uses the OpenSearch min tar to make sure it can be installed. Until we've got a rc1 version of OpenSearch min our CI check won't pass.
> 
> Option 1) remove security plugin from the manifest. Cleanest approach for build safety with the tradeoff of consuming a lot of time during this transition phase.
> 
> Option 2) bypass the CI check and cross fingers that the new build doesn't have any unexpected issues.
> 
> My preference is towards option 1 with the understanding it will slow down the whole rollout. This will ensure only correctly building plugins are included in the build, avoiding an utterly broken build.
> 
> Note; These problems are magnified in security-dashboards-plugin as it depends on a full distribution build for its CI checks.
_From https://github.com/opensearch-project/security/pull/1764_
 
### Check List

### DIFF
--- a/manifests/2.0.0/opensearch-2.0.0.yml
+++ b/manifests/2.0.0/opensearch-2.0.0.yml
@@ -100,9 +100,9 @@ components:
     checks:
       - gradle:properties:version
       - gradle:dependencies:opensearch.version
-  - name: security
-    repository: https://github.com/opensearch-project/security.git
-    ref: main
+  # - name: security
+  #   repository: https://github.com/opensearch-project/security.git
+  #   ref: main
   - name: cross-cluster-replication
     repository: https://github.com/opensearch-project/cross-cluster-replication.git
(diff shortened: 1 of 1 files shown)


dblock:
This shouldn’t be necessary, is there no successful rc1 snapshot build @peterzhuamazon ?

peternied:
I am unsure if this change is needed, waiting on a passing build.  Keeping this change around until it is clear it isn't needed.

peterzhuamazon:
Close this as we have a option selected for RPM fix.
This is not needed once RPM PR merged.
https://github.com/opensearch-project/opensearch-build/issues/1545#issuecomment-1099632988
Thanks.

PR #1981, opened 2022-04-15
[OSD] remove check for build for 1.3.2 and 1.2.1

kavilla:
### Description
We have bumped the 1.x branches patch versions but it doesn't have the rpm commands so it's failing. We also don't have plan to release this so for now we can just not build these and add them later if there is a plan.

### Issues Resolved
n/a
 
### Check List


peterzhuamazon:
Fixed in #1979

PR #1990, opened 2022-04-19
Remove BWC test for now as BWC framework does not run correctly as well as rpm integTest as it requires root

peterzhuamazon:

### Description
Remove BWC test for now as BWC framework does not run correctly as well as rpm integTest as it requires root

 
### Check List

### DIFF
--- a/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
+++ b/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
@@ -173,50 +173,27 @@ pipeline {
                             echo "buildManifestUrl (x64, rpm): ${buildManifestUrl}"
                             echo "artifactUrl (x64, rpm): ${artifactUrl}"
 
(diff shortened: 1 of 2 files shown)


bbarani (approved):

kavilla:
BWC tests should work it's just failing due to security plugin.

bbarani:
> BWC tests should work it's just failing due to security plugin.

@kavilla Are you taking about OpenSearch dashboards?

peterzhuamazon:
Merge it now, we can add back later on.

kavilla:
> Merge it now, we can add back later on.

BWC test were running for OpenSearch Dashboards tar distributions they were actually catching some issues that should be followed up. I guess my question instead of full out removing it from everything we remove the BWC tests from the test manifest being called if we don't want to see the error but otherwise it was just show a "failed" notification versus removing this completely.

Issue #1916, opened 2022-04-06
[Meta] Mechanism to publish the plugin zips to maven repo.


prudhvigodithi:
### Is your feature request related to a problem? Please describe

With current setup the plugins zips that are used as dependencies cannot be downloaded dynamically during runtime, as they are not part of version controlled maven system. Only the respective Java jars are available to download using maven coordinates.
More details on https://github.com/opensearch-project/opensearch-build/issues/716

### Describe the solution you'd like

Publish the release candidate zips, distribution build zips (snapshots) to maven repo, so that a user who is dependent on the plugin zip, can download using maven coordinates. There needs to be a process implemented for uploading zips to snapshot and release repos

**Proposed Solution [1]:**
With the current build system the plugin zips are stored under `builds/plugins/`, these are not stored or ordered in maven folders, so that maven can publish them with right artifactID, groupID and version.

for each plugin use `install:install-file` to create the maven coordinates and publish them to local file system maven repo, passing the right artifactID, groupID and version along with proper qualifier if required.
Then from local file system publish them to respective snapshot or release repos, adhering to existing workflows.

To achieve this integrate this solution with existing CI.

**Proposed Solution [2]:** ---> (The chosen one)
Add a gradle sub-project in each component, so that root gradle project can include them, that converts the zips generated inside distribution folder to maven format, then publish them to local maven repo and staging repo, this way all the maven coordinate's (groupID, artifactID, version) generation and other maven heavy lifting is done at component (plugin) level and using CI with already existing code, publish them to maven nexus repo.

### Additional context
**For Proposed Solution [1]:**
Need to come up with an idea to add either a python script,  a groovy shared library or a gradle sub tasks for publishing the zips.

**Snapshot zips:**

Upload the snapshot plugin zips to https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/plugin/
    <plugin_name>/<version>-SNAPSHOT/<.zip_file>
Edit the plan https://github.com/prudhvigodithi/opensearch-build/blob/slack-publish-v1/jenkins/opensearch/distribution-build.jenkinsfile#L82 to upload zips to nexus maven repo

**Release zips:**
**The Release candidates will be signed before staging to make sure files are unaltered.**

Also there needs to be an additional tags inside the POM as example for the zips pom file
https://aws.oss.sonatype.org/service/local/repositories/releases/content/org/opensearch/common-utils/1.0.0.0/common-utils-1.0.0.0.pom
[...]

PR #1999, opened 2022-04-19
default build.sh logic add for Gradle custom java zippublish plugin

prudhvigodithi:
Signed-off-by: pgodithi <pgodithi@amazon.com>

### Description
This change will add custom zippublish plugin publishMavenzipPublicationToZipstagingRepository logic to default build.sh, this can publish plugin zips to maven repo.
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1916
 
### Check List

### DIFF
--- a/scripts/default/opensearch/build.sh
+++ b/scripts/default/opensearch/build.sh
@@ -77,3 +77,14 @@ distributions="$(dirname "${zipPath}")"
 echo "COPY ${distributions}/*.zip"
 mkdir -p $OUTPUT/plugins
 cp ${distributions}/*.zip ./$OUTPUT/plugins
+
+TASK="publishMavenzipPublicationToZipstagingRepository"
+if ./gradlew tasks --all | grep -qw "^$TASK"
+then
+   ./gradlew publishMavenzipPublicationToZipstagingRepository -PzipVersion=$VERSION 
+    echo "Copying the Maven plugin Zips"
+    mkdir -p $OUTPUT/maven/org/opensearch/plugin
+    cp -r ./build/local-staging-repo/org/opensearch/plugin/. $OUTPUT/maven/org/opensearch/plugin
+else
+    echo "Not running publishMavenzipPublicationToZipstagingRepository"
+fi


scripts/default/opensearch/build.sh
```diff
@@ -77,3 +77,15 @@ distributions="$(dirname "${zipPath}")"
 echo "COPY ${distributions}/*.zip"
 mkdir -p $OUTPUT/plugins
 cp ${distributions}/*.zip ./$OUTPUT/plugins
+
+#This is to check if a component has a plugin opensearch.zippublish applied and run publishMavenzipPublicationToZipstagingRepository.
+MAVEN_ZIP_PUBLISH_TASK="publishMavenzipPublicationToZipstagingRepository"
```
gaiksaya, line 82: Can this be renamed to `publishZipsToMavenRepository` or something similar? Much simpler and precise.

prudhvigodithi: @gaiksaya 
so the reason I have added this way is this var MAVEN_ZIP_PUBLISH_TASK is actually referenced same in the java code, https://github.com/opensearch-project/OpenSearch/pull/2988/files#diff-d56f11b358bf18f6e6c28d2a528aa08d4436d01ce9b3c3c45bab1fa970d9cb43R49 and this is the format `publishMavenzipPublicationToZipstagingRepository` gradle expects for publishing a custom repo (mavenzip) to custom staging repo (zipstaging).

prudhvigodithi: just to add this task is coming from the custom gradle plugin (identifies once the plugin is applied).

gaiksaya: I see, Yeah I thought this can be changed in upstream PR as well but looks like we can't since its a gradle requirement.

dblock (commented):
Once the task is available in OpenSearch, doesn't it "just work" with all plugins? Do we need the `if`?

prudhvigodithi:
Hey @dblock this task wont show up until a user explicitly add's as `apply plugin: 'opensearch.zippublish'`, so if there are some components that does not want to publish a zip and has not added the plugin, this task wont show up and will throw an error as task not available.

dblock:
> Hey @dblock this task wont show up until a user explicitly add's as `apply plugin: 'opensearch.zippublish'`, so if there are some components that does not want to publish a zip and has not added the plugin and again relies on this default build scripts this task wont show up and will throw an error as task not available.

Ok, then this is backwards. It should only be added to the plugins that want to publish ZIPs in their own build.sh in their repos.

prudhvigodithi:
@dblock one way its true, but repos like security, anomaly-detection, cross-cluster-replication (could be more), does use the default build.sh script and still generates the zip, so if we have publish them, do you suggest to maintain their own buildscripts ? But the added if condition should be exiting gracefully if the task does not exists. Just we might need inform user in the case we remove it from default script.

prudhvigodithi:
My opinion is to add this task to the default script as this will minimize the code changes of existing repos. Since they are already pointing to the default script, they should continue to do, even if they want to publish the zip.
PS: I'm open for suggestions

dblock:
> @dblock one way its true, but repos like security, anomaly-detection, cross-cluster-replication (could be more), does use the default build.sh script and still generates the zip, so if we have publish them, do you suggest to maintain their own buildscripts ? But the added if condition should be exiting gracefully if the task does not exists. Just we might need inform user in the case we remove it from default script.

Yes, that's the design of this system. Unless the task applies to _all_ plugins, don't include it in the default script.

dblock:
> My opinion is to add this task to the default script as this will minimize the code changes of existing repos. Since they are already pointing to the default script, they should continue to do, even if they want to publish the zip. open for discussion.

Many plugins don't use the default script, so it won't really minimize changes that much. Scripts also vary release-by-release/branch-by-branch. And again, this is not the pattern that has been adopted before, otherwise we'd have a default script with tons of "if task XYZ exists run it" for all kinds of other things.

dblock:
Now, why don't all plugins implement zip publishing?

prudhvigodithi:
> Many plugins don't use the default script, so it won't really minimize changes that much. Scripts also vary release-by-release/branch-by-branch. And again, this is not the pattern that has been adopted before, otherwise we'd have a default script with tons of "if task XYZ exists run it" for all kinds of other things.

So here I was referencing to the plugins that use the default buildscript, who has to re add the script in their repo moving forward.

> Now, why don't all plugins implement zip publishing?

Yes they can implement, but need to execute the task `publishMavenzipPublicationToZipstagingRepository` for this to happen, to call this gradle command `./gradlew publishMavenzipPublicationToZipstagingRepository -Dopensearch.version=$VERSION`, they need to add it to the build flow.

dblock:
> > Now, why don't all plugins implement zip publishing?
> 
> Yes they can implement, but need to execute the task `publishMavenzipPublicationToZipstagingRepository` for this to happen, to call this gradle command `./gradlew publishMavenzipPublicationToZipstagingRepository -Dopensearch.version=$VERSION` they need to add it to the build flow and this build flow (using default build.sh) resides in this repo.

If this were the case (all plugins implement this action), then you can add it to the default script. It would be similar to how all plugins implement `build`.

prudhvigodithi:
> If this were the case (all plugins implement this action), then you can add it to the default script. It would be similar to how all plugins implement `build`.

Make sense @dblock , also I think it would be not a good experience for repo users, to ask them to create and maintain a custom script, just because they want the zips to be in maven.

dblock (changes requested):
I can't agree to having `f ./gradlew tasks -Dopensearch.version=$VERSION --all | grep -qw "^$MAVEN_ZIP_PUBLISH_TASK"`, this is backwards, and not how we've done customization across repos. Add the task to all repos and add it to the default script without the conditional and customize everything else.

prudhvigodithi:
> I can't agree to having `f ./gradlew tasks -Dopensearch.version=$VERSION --all | grep -qw "^$MAVEN_ZIP_PUBLISH_TASK"`, this is backwards, and not how we've done customization across repos. Add the task to all repos and add it to the default script without the conditional and customize everything else.

Hey @dblock as you know right now, not all plugins, use `bundlePlugin` task to generate the zip as well, so these repos cant even use the new plugin until they migrate the zip generation using  `bundlePlugin` task, so even before adding/executing  the` MAVEN_ZIP_PUBLISH_TASK` task, these repos has to migrate to this `bundlePlugin` task, then next they can start using the plugin and then off course the publish task can then be added, while this all happens, if we cant add this task to default build script, then the plugins that depend on the default build.sh has to wait until the other repos migrate and start using `bundlePlugin` task or create their own build.sh and start using it.

dblock:
> > I can't agree to having `f ./gradlew tasks -Dopensearch.version=$VERSION --all | grep -qw "^$MAVEN_ZIP_PUBLISH_TASK"`, this is backwards, and not how we've done customization across repos. Add the task to all repos and add it to the default script without the conditional and customize everything else.
> 
> Hey @dblock as you know right now, not all plugins, use `bundlePlugin` task to generate the zip as well, so these repos cant even use the new plugin until they migrate the zip generation using `bundlePlugin` task, so even before adding/executing the` MAVEN_ZIP_PUBLISH_TASK` task, these repos has to migrate to this `bundlePlugin` task, then next they can start using the plugin and then off course the publish task can then be added, while this all happens, if we cant add this task to default build script, then the plugins that depend on the default build.sh has to wait until the other repos migrate and start using `bundlePlugin` task or create their own build.sh and start using it.

This is a hack that we will never undo, so I am opposed to it. Implement the new task for 1-2 repos where it's easy (e.g. job-scheduler), then run a campaign where each repo does the work of migrating to `bundlePlugin` and using this new task, _then_ add it to the default script.

dblock:
@prudhvigodithi I think you should do this work for 1 plugin (e.g. job-scheduler) before creating a campaign. And document exactly what needs to be done: so things like use `bundlePlugin` should be in there, e.g. "1. Rewrite build to use budlePlugin like so [link] if you aren't doing it already."

dblock:
> > @prudhvigodithi I think you should do this work for 1 plugin (e.g. job-scheduler) before creating a campaign. And document exactly what needs to be done: so things like use `bundlePlugin` should be in there, e.g. "1. Rewrite build to use budlePlugin like so [link] if you aren't doing it already."
> 
> Sure @dblock, However I have already created an enhancement issue across all plugin repos, regrading this :) I will start with job-scheduler ([issue](https://github.com/opensearch-project/job-scheduler/issues/175))

Consider https://github.com/opensearch-project/opensearch-plugin-template-java first!

prudhvigodithi:
Hey @dblock I have removed the if condition as now all the plugins should have the `publishPluginZipPublicationToZipStagingRepository` gradle task. Should we still consider to add `publishPluginZipPublicationToZipStagingRepository` to default `build.sh`?
Thank you

scripts/default/opensearch/build.sh
```diff
@@ -77,3 +77,8 @@ distributions="$(dirname "${zipPath}")"
 echo "COPY ${distributions}/*.zip"
 mkdir -p $OUTPUT/plugins
 cp ${distributions}/*.zip ./$OUTPUT/plugins
+
+# Execute publishMavenzipPublicationToZipstagingRepository. task to publish plugin zips to local maven repo
```
dblock, line 81: I would just say "Publish plugin zips to maven"

dblock (approved):

Issue #1500, opened 2022-01-18
Run CCR Perf tests as part of Jenkins CI


ankitkala:
### Is your feature request related to a problem? Please describe

https://github.com/opensearch-project/opensearch-build/issues/1428

### Describe the solution you'd like

After all tasks mentioned in https://github.com/opensearch-project/opensearch-build/issues/1428 are complete, we need to work towards enabling these tests as part of Jenkins CI.

bbarani:
@ankitkala Can you update the progress of this issue?

bbarani:
@ankitkala Can you update the progress of this issue? Is this still actively worked upon?

bbarani:
@ankitkala Can you update the progress of this issue? Can we close this issue?

PR #2005, opened 2022-04-20
Add Jenkins CI for CCR Performance tests

ankitkala:
Signed-off-by: Ankit Kala <ankikala@amazon.com>

### Description
- Adds the JenkinsFile for triggering CCR Performance tests.
- Add support for creating multi node clusters via `PerfTestCluster`

### Issues Resolved
[1500](https://github.com/opensearch-project/opensearch-build/issues/1500)
 
### Check List


src/test_workflow/perf_test/perf_test_cluster.py
```diff
@@ -57,22 +46,31 @@ def start(self):
...
+            host = public_ip if public_ip is not None else private_ip
+        else:
+            host = cdk_output[self.stack_name]["LoadBalancerEndpoint"]
+            self.is_endpoint_public = True
+
+        if host is not None:
+            self.endpoint = host
+            self.cluster_endpoint_with_port = "".join([scheme, host, ":", str(self.port())])
+
+    def endpoint_with_port(self):
```
dblock, line 66: Make this a `@property` so you can use it without `()`.

src/test_workflow/perf_test/perf_test_cluster.py
```diff
@@ -13,32 +13,20 @@
...
-    Represents a performance test cluster. This class deploys the opensearch bundle with CDK and returns the private IP.
+    Represents a performance test cluster. This class deploys the opensearch bundle with CDK. Supports both single
+    and multi-node clusters
     """
 
-    def __init__(self, bundle_manifest, config, stack_name, security, current_workspace):
+    def __init__(self, bundle_manifest, config, stack_name, cluster_config, current_workspace):
         self.manifest = bundle_manifest
-        self.work_dir = os.path.join(current_workspace, "opensearch-cluster", "cdk", "single-node")
+        self.work_dir = os.path.join(current_workspace, "opensearch-cluster", "cdk",
```
dblock, line 22: You should subclass `PerfTestCluster` into single vs. multi-node. All the methods such as `create_endpoint` and `setup_cdk_params` have different implementations in those cases.

src/test_workflow/perf_test/perf_test_cluster.py
ankitkala: done.

src/test_workflow/perf_test/perf_test_cluster.py
ankitkala: done.

src/test_workflow/perf_test/perf_single_node_cluster.py
```diff
@@ -0,0 +1,55 @@
+
+import os
+from contextlib import contextmanager
+
+from test_workflow.perf_test.perf_test_cluster import PerfTestCluster
+
+
+class PerfSingleNodeCluster(PerfTestCluster):
```
owaiskazi19, line 8: How about if we can create one base class for the PerfCluster and pass the argument as `SingleNode` or `MultiNode` to avoid the duplication?

dblock: The logic is totally different between single-node and multi-node cluster, that's why there are 2 child classes here.

src/test_workflow/perf_test/perf_single_node_cluster.py
```diff
@@ -0,0 +1,55 @@
...
+        work_dir = os.path.join(current_workspace, "opensearch-cluster", "cdk", "single-node")
+        super().__init__(bundle_manifest, config, stack_name, cluster_config, current_workspace, work_dir)
+
+    def create_endpoint(self, cdk_output):
+        scheme = "https://" if self.cluster_config.security else "http://"
+        private_ip = cdk_output[self.stack_name]["PrivateIp"]
+        public_ip = cdk_output[self.stack_name].get("PublicIp", None)
+        self.is_endpoint_public = public_ip is not None
+        host = public_ip if public_ip is not None else private_ip
+        if host is not None:
```
dblock, line 24: Isn't this an error in the case where `host` is `None`? What should happen here? Default to localhost?

src/test_workflow/perf_test/perf_test_cluster.py
```diff
@@ -87,23 +72,13 @@ def dependencies(self):
         return []
 
     def wait_for_processing(self, tries=3, delay=15, backoff=2):
-        if self.public_ip is None:
-            return
-        url = "".join([self.endpoint(), "/_cluster/health"])
+        # Should be invoked only if the endpoint is public.
+        assert self.is_endpoint_public, "wait_for_processing should be invoked only when cluster is public"
+        print("Waiting for domain to be up")
```
dblock, line 77: Use `logging.info`

src/test_workflow/perf_test/perf_test_cluster_config.py
```diff
@@ -0,0 +1,20 @@
...
+    Maintains the cluster level configuration.
+    """
+    def __init__(self, security=False, data_nodes=1, master_nodes=0, ingest_nodes=0, client_nodes=0):
+        self.security = security
+        self.data_nodes = data_nodes
+        self.master_nodes = master_nodes
+        self.ingest_nodes = ingest_nodes
+        self.client_nodes = client_nodes
+
+    def is_single_node_cluster(self):
```
dblock, line 18: Make it a `@property` and maybe just call it `single_node_cluster`

src/test_workflow/perf_test/perf_test_cluster_config.py
```diff
@@ -0,0 +1,20 @@
...
+        return True if (self.data_nodes == 1 and self.master_nodes == 0
```
dblock, line 19: return `self.data_nodes == ...`, no need to do `if/else`, it's already a bool.

dblock (approved):
This is a lot better. 

I have some smaller asks that you might want to consider before merging.

jenkins/cross-cluster-replication/perf-test.jenkinsfile
```diff
@@ -0,0 +1,138 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
+
+pipeline {
+    agent none
+    options {
+        timeout(time: 10, unit: 'HOURS')
+    }
+    environment {
+        AGENT_LABEL = 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host'
```
kotwanikunal, line 9: Please use the agent label - `Jenkins-Agent-al2-x64-m5xlarge-Perf-Test`

src/test_workflow/perf_test/perf_single_node_cluster.py
ankitkala: Yep. Ideally this shouldn't happen. I'll remove the 'None' check.

jenkins/cross-cluster-replication/perf-test.jenkinsfile
ankitkala: ack.

src/test_workflow/perf_test/perf_test_cluster_config.py
ankitkala: done

ankitkala:
> This is a lot better.
> 
> I have some smaller asks that you might want to consider before merging.
> 
> Also fix CI.

Addressed all the comments and fixed the broken tests.

jenkins/cross-cluster-replication/perf-test.jenkinsfile
```diff
@@ -0,0 +1,139 @@
+lib = library(identifier: "jenkins@20211118", retriever: legacySCM(scm))
```
gaiksaya, line 1: Do we  plan to add regression tests for this jenkins file? Similar to https://github.com/opensearch-project/opensearch-build/tree/main/tests/jenkins/jenkinsjob-regression-files
More details on how to add: https://github.com/opensearch-project/opensearch-build/blob/main/DEVELOPER_GUIDE.md#regression-tests

src/test_workflow/perf_test/perf_test_cluster.py
```diff
@@ -57,25 +45,22 @@ def start(self):
...
+
+    @property
     def port(self):
-        self.cluster_port = 443 if self.security else 9200
-        return self.cluster_port
+        return 443 if self.cluster_config.security else 80
 
     def terminate(self):
-        os.chdir(os.path.join(self.current_workspace, self.work_dir))
+        os.chdir(self.work_dir)
```
gaiksaya, line 63: terminate can also be converted to `@property` right?

gaiksaya (commented):
Looks like there are some failing tests for windows.

src/test_workflow/perf_test/perf_test_cluster.py
dblock: It shouldn't. A property is a field/noun (e.g. `object.color` or `object.shape`), invoked without `()` vs. `terminate` is an action/verb, invoked with `()` (e.g. `object.run()`, `object.stop()`).

src/test_workflow/perf_test/perf_multi_node_cluster.py
```diff
@@ -0,0 +1,54 @@
...
+        Set up the cluster. When this method returns, the cluster must be available to take requests.
+        Throws ClusterCreationException if the cluster could not start for some reason. If this exception is thrown, the caller does not need to call "destroy".
+        """
+        cluster = cls(*args)
+
+        try:
+            cluster.start()
+            yield cluster
+        finally:
+            cluster.terminate()
```
dblock, line 54: This method should be in the base `PerfTestCluster` class.

src/test_workflow/perf_test/perf_single_node_cluster.py
```diff
@@ -0,0 +1,53 @@
...
+        Set up the cluster. When this method returns, the cluster must be available to take requests.
+        Throws ClusterCreationException if the cluster could not start for some reason. If this exception is thrown, the caller does not need to call "destroy".
+        """
+        cluster = cls(*args)
+
+        try:
+            cluster.start()
+            yield cluster
+        finally:
+            cluster.terminate()
```
dblock, line 53: Duplicate, move to base class.

src/test_workflow/perf_test/perf_multi_node_cluster.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, bundle_manifest, config, stack_name, cluster_config, current_workspace):
+        assert not cluster_config.is_single_node_cluster, "Cluster is a single node configuration"
+        work_dir = os.path.join(current_workspace, "opensearch-cluster", "cdk", "multi-node")
+        super().__init__(bundle_manifest, config, stack_name, cluster_config, current_workspace, work_dir)
+
+    def create_endpoint(self, cdk_output):
+        scheme = "https://" if self.cluster_config.security else "http://"
+        host = cdk_output[self.stack_name]["LoadBalancerEndpoint"]
+        self.is_endpoint_public = True
+        if host is not None:
```
dblock, line 21: I still question what happens here if `host == None`. Is that valid? We end up without values in `endpoint`, is that expected?

dblock (changes requested):
CI is still failing + some more small stuff

src/test_workflow/perf_test/perf_multi_node_cluster.py
ankitkala: Yes, we'd be without value in endpoint. Not sure if there is any graceful way to handle that. 

I'm thinking we can assert on the host not being `None` instead. Thoughts?

jenkins/cross-cluster-replication/perf-test.jenkinsfile
ankitkala: Added the test for the jenkins file

src/test_workflow/perf_test/perf_multi_node_cluster.py
ankitkala: My bad. Had to move this to subclass during dev testing and forgot to move it back.

src/test_workflow/perf_test/perf_single_node_cluster.py
ankitkala: Done

src/test_workflow/perf_test/perf_multi_node_cluster.py
dblock: Raise a `ValueError`

src/test_workflow/perf_test/perf_single_node_cluster.py
```diff
@@ -0,0 +1,44 @@
...
+        work_dir = os.path.join(current_workspace, "opensearch-cluster", "cdk", "single-node")
+        super().__init__(bundle_manifest, config, stack_name, cluster_config, current_workspace, work_dir)
+
+    def create_endpoint(self, cdk_output):
+        scheme = "https://" if self.cluster_config.security else "http://"
+        private_ip = cdk_output[self.stack_name]["PrivateIp"]
+        public_ip = cdk_output[self.stack_name].get("PublicIp", None)
+        self.is_endpoint_public = public_ip is not None
+        host = public_ip if public_ip is not None else private_ip
+        assert host is not None, "Unable to fetch the cluster endpoint from cdk output"
```
dblock, line 28: I think this error should happen at runtime, there's a legit cdk output that might be missing this data, so I think it's not an assert, but a normal exception. And would need a test that it's raised.

dblock (approved):

src/test_workflow/perf_test/perf_single_node_cluster.py
ankitkala: Sure. Changed to RuntimeError and added the tests for it.

PR #2028, opened 2022-04-22
upload test result to upstream job if exists

tianleh:

### Description
Follow up on https://github.com/opensearch-project/opensearch-build/pull/2024#issuecomment-1106743524
When non-prod build job triggers the integ test, it will override the result from prod build job if the build number happens to be same. Thus add the logic to upload the test result to its upstream job if it exists. This can avoid such conflict and also allow us to use `latest` url to access the test result.
 
### Check List

### DIFF
--- a/jenkins/opensearch-dashboards/bwc-test.jenkinsfile
+++ b/jenkins/opensearch-dashboards/bwc-test.jenkinsfile
@@ -99,9 +99,11 @@ pipeline {
             post {
                 always {
                     script {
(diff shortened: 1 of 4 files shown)


peterzhuamazon (changes requested):
Will revisit this after 2.0.0 release.

gaiksaya:
Hi @tianleh , can you resolve the merge conflicts and confirm that this is working as expected?

tianleh:
Will revisit when there is bandwidth

PR #2061, opened 2022-04-29
Re-add the custom security build

peternied:
### Description
For 2.0.0 builds the default build script works as expected.  There is a standalone library in the
1.X line of that was removed in 2.0.0, which is why this issue has appeared now

### Issues Resolved
* Closes https://github.com/opensearch-project/security/issues/1809

### DIFF
--- a/scripts/components/security/build.sh
+++ b/scripts/components/security/build.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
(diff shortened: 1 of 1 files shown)


dblock:
Shouldn't you be committing this into the security plugin repo instead in the right branch?

zelinh:
I think having this here will affect the build for 2.0.0 right?

peternied:
> Shouldn't you be committing this into the security plugin repo instead in the right branch?

Oh! Yes, we can do that

peternied:
Making this fix in the specific branch rather than the shard codebase https://github.com/opensearch-project/security/pull/1810

PR #2092, opened 2022-05-05
Fix incorrect path in documentation

peternied:
### Description
When I was authoring https://github.com/opensearch-project/security/pull/1811 I had to iterate a couple of extra times and debug the python code because the custom script was not being picked up due to the incorrect path.  

### DIFF
--- a/src/build_workflow/README.md
+++ b/src/build_workflow/README.md
@@ -76,7 +76,7 @@ The following options are available in `build.sh`.
 
 ### Custom Build Scripts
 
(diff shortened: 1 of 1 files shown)


dblock (changes requested):
I believe we support both `scripts/build.sh` and `build.sh`, but should really standardize on one. I'd prefer we correct this to `scripts/build.sh`, and have a separate ticket to get rid of anything at root.

peternied:
> I believe we support both scripts/build.sh and build.sh, but should really standardize on one.

I'm in favor of `scripts/build.sh`, updated the documentation.

> have a separate ticket to get rid of anything at root.

Would you want to go into previously shipped release branches and fix this?  IMO, I'd rather keep the code supporting both methods, but shift the documentation to get folks pointed towards `scripts/build.sh` which seems more correct.

dblock:
> Would you want to go into previously shipped release branches and fix this? IMO, I'd rather keep the code supporting both methods, but shift the documentation to get folks pointed towards `scripts/build.sh` which seems more correct.

I wouldn't want to change code in previous releases. I think the build.sh location at root could be deprecated with a warning.

dblock (approved):

Issue #2076, opened 2022-05-03
[BUG] Add support for promotion workflow for qualifier and rpm


peterzhuamazon:
The TAR promotion workflow to migrate the staging artifacts to production bucket doesn't support qualifiers hence need to enhance the workflow to support the qualifiers.

PR #2094, opened 2022-05-05
Tweak promotion pipeline to support rpm and qualifier

peterzhuamazon:

### Description
Tweak promotion pipeline to support rpm and qualifier

### Issues Resolved
Part of #2076


peterzhuamazon:
Note:
@saratvemulapalli has issued a fix to the core-plugin wrong folder: https://github.com/opensearch-project/OpenSearch/pull/3193.

If that PR get merged this PR needs some changes.

Thanks.

vars/promoteTarArtifacts.groovy
```diff
@@ -16,78 +16,94 @@ void call(Map args = [:]) {
     def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: manifest))
```
dblock, line 16: The name of this file should change to `promoteArtifacts` or maybe `promoteDistributions`?

dblock (commented):
Just one ask for the function name change, otherwise looks good!

vars/promoteTarArtifacts.groovy
peterzhuamazon: I think using distribution is inaccurate compares to artifacts.
We are moving only parts of all the distribution artifacts to production bucket.
If this one is named promoteDistribution then docker/yum should also be promoted in this same lib, which is never the case anyway.

I still think promoteAritfacts is better at this point.
Open to suggestion.

Thanks.

dblock: `promoteAritfacts` works for me, currently it's called ` promoteTarAritfacts`

vars/promoteTarArtifacts.groovy
```diff
@@ -16,78 +16,94 @@ void call(Map args = [:]) {
...
+                        bucket: "${ARTIFACT_PRODUCTION_BUCKET_NAME}",
+                        path: "releases/$pluginFullPath/",
+                        workingDir: "$prefixPath/$artifactPath/builds/$filename/core-plugins/",
+                        includePathPattern: "**/${pluginName}*"
+                    )
+                }
+            }
+            
+            // We will only publish min artifacts for Tar, ignore other distributions
+            if (distribution_local.equals('tar')) {
```
dblock, line 92: This is a bad smell. Is there a way to extract the "special case of handling tar" out of the generic "promoteArtifacts"?

peterzhuamazon: They all have the same file name conversion except the distribution, hence this is the best usage.
We will only release tar min and that will not change for now.

vars/promoteTarArtifacts.groovy
peterzhuamazon: fixed now.

dblock (approved):
Approved to unblock, but the fact that we have to do `distribution_local.equals('tar'))` in `promoteArtifacts` should be refactored into something better.

Issue #2112, opened 2022-05-11
[Question] How to approach `api/status` `api/reporting/stats` route authentication addition in future releases


peterzhuamazon:
With security team merge the change of `adding auth to /api/status api/reporting/stats call` in 2.0 branch, we are facing multiple issues with both build repo scripts/workflows, as well as the approach on future releases.

* Note: This change is not in 1.3 but only 2.0+, means we also need to think about how to maintain either 2 copies of change or choose a one-fit-all direction in build repo scripts.

------
### Behavior changes

* Previously: 
```
curl localhost:5601/api/status
curl localhost:5601/api/reporting/stats
```

* New:
```
curl localhost:5601/api/status -u admin:admin
curl localhost:5601/api/reporting/stats -u admin:admin
```

------
### Two ways to approach this issue:

1. Restore previous behavior in 2.0 by adding this config into the `opensearch_dashboards.yml`:
```
 opensearch_security.auth.unauthenticated_routes: ['/api/reporting/stats', '/api/status']
```
2. Use new behavior by default in 2.0.

------
### Impacts on build:

**1. If we use previous behavior by edit `opensearch_dashboards.yml`:**
* Nothing else changes
* Customer dont see the changes as well.

**2. If we use the new behavior:**
* All calls in build scripts that uses the old `curl localhost:5601/api/status` call needs to make changes
* This include `Test framework`, `RPM validation workflow`.
  * Additional impact on all the workflow runs on 1.3 branch, as 1.3 does not have this change so `/api/status` would run, but in 2.0 we need to run `api/status -u admin:admin`.
* Detailed documentation to inform user on this change as it is pretty big.

------
### Additional items:

1. @cliu123 create documentation website related changes: https://github.com/opensearch-project/documentation-website/issues/563

Thanks.

peterzhuamazon:
cc: @CEHENKLE @bbarani @dblock @cliu123

dblock:
Is this another piece of code that needs to branch version-to-version, aka https://github.com/opensearch-project/opensearch-build/issues/1975?

For this particular problem, will specifying `admin:admin` be just ignored in the 1.3.x version?

peterzhuamazon:
> Is this another piece of code that needs to branch version-to-version, aka #1975?
> 
> For this particular problem, will specifying `admin:admin` be just ignored in the 1.3.x version? Will it hurt to add it without branching?

1. Yes
2. I test on artifacts that do not have these change with these results:

```
[...]

PR #2117, opened 2022-05-12
Add admin authentication to dashboards api/status

zelinh:

### Description
Fixed the authentication for api/status 

### Issues Resolved
Part of #2112 


src/test_workflow/integ_test/service_opensearch_dashboards.py
```diff
@@ -68,7 +68,7 @@ def url(self, path=""):
     def get_service_response(self):
         url = self.url("/api/status")
         logging.info(f"Pinging {url}")
-        return requests.get(url, verify=False, auth=("kibanaserver", "kibanaserver") if self.security_enabled else None)
+        return requests.get(url, verify=False, auth=("admin", "admin"))
```
peterzhuamazon, line 71: This is not correct you should not change the behavior of how to handle enabled security plugin, rather than only focus on the `api/status` I think.

tests/tests_test_workflow/test_integ_workflow/integ_test/test_service_opensearch_dashboards.py
```diff
@@ -208,7 +208,7 @@ def test_get_service_response_with_security(self, mock_url, mock_requests_get):
         service.get_service_response()
 
         mock_url.assert_called_once_with("/api/status")
-        mock_requests_get.assert_called_once_with(mock_url_result, verify=False, auth=("kibanaserver", "kibanaserver"))
+        mock_requests_get.assert_called_once_with(mock_url_result, verify=False, auth=("admin", "admin"))
```
peterzhuamazon, line 211: Same as above.

src/test_workflow/integ_test/service_opensearch_dashboards.py
peterzhuamazon: Also need @cliu123 to chime in on whether `api/status` requires `admin:admin`, or `kibanaserver:kibanaserver` is enough.

peterzhuamazon (changes requested):
Some questions, block for now until it is cleared.

src/test_workflow/integ_test/service_opensearch_dashboards.py
cliu123: Yes, `kibanaserver:kibanaserver` is good enough. Basically, any valid credential will go through authentication here. There is no authorization enforced here, so it doesn't matter what roles a user has. As long as the credential provided is valid, request goes through.

prudhvigodithi: Hey @cliu123 the `kibanaserver` user is commented out on default setup 
https://github.com/opensearch-project/security/blob/main/securityconfig/config.yml#L71
So correct me if i'm wrong `kibanaserver:kibanaserver` wont work I guess, unless we enable the right role for the index?

peterzhuamazon: @prudhvigodithi https://github.com/opensearch-project/opensearch-build/blob/main/config/opensearch_dashboards.yml#L178-L179

peterzhuamazon: Thanks @cliu123 for the clarification here.
@zelinh please restore.

Thanks.

zelinh: Thanks all for the inputs here. Reverted the changes here on integ test. 
@peterzhuamazon Please help review again.

peterzhuamazon (approved):

Issue #1513, opened 2022-01-18
[Python Typechecking] Final push `pipenv run mypy .`


peternied:
See full instructions at https://github.com/opensearch-project/opensearch-build/issues/1237

Note; this should be the last type checking task completed otherwise there will be some duplication with those tasks

Acceptance criteria:
 - Run `pipenv run mypy .` passes without error

Issue #1512, opened 2022-01-18
[Python Typechecking] Test workflow `pipenv run mypy tests\tests_test_workflow`


peternied:
See full instructions at https://github.com/opensearch-project/opensearch-build/issues/1237

Acceptance criteria:
 - Run `pipenv run mypy tests\tests_test_workflow` passes without error

Issue #1237, opened 2021-12-02
[Meta] Python - Typechecking to 100%


peternied:
To get the whole python codebase typechecking I'd suggest we divide it up into the following sections:

- [x] Build workflow `pipenv run mypy src\build_workflow tests\tests_build_workflow`
- [x] Checkout workflow `pipenv run mypy src\checkout_workflow tests\tests_checkout_workflow`
- [x] #1508
- [x] #1509
- [x] #1510
- [x] #1511
- [ ] #1512
- [ ] #1513

### How to add type checking?
In the root of the project run one of the above commands like `pipenv run mypy src\build_workflow tests\tests_build_workflow` then you'll see an output
```
...
tests/tests_build_workflow/test_build_recorder.py:173: error: Function is missing a type annotation  [no-untyped-def]
src/build_workflow/builders.py:16: error: Function is missing a type annotation  [no-untyped-def]
Found 139 errors in 23 files (checked 28 source files)
```
Just go to each line and figure out how to add the types and then make a PR when there are 0 errors.

Checkout this cheat sheet for many many common scenarios https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html

### What if something cannot be typed?
Add a comment `# type: ignore[return-value]`, you can find the code in the output from `pipenv run mypy`

### Any examples of how this was done?
Yes, here are some of the first changes where the defaults were largely enabled, check out these PRs
https://github.com/opensearch-project/opensearch-build/pull/1224
https://github.com/opensearch-project/opensearch-build/pull/1230

peternied:
@opensearch-project/engineering-effectiveness this is a great issue to work on this sprint if you are blocked with your core tasks

PR #2123, opened 2022-05-14
Added missing typings.

dblock:

### Description

- Added missing types.
- Enabled mypy in the complete python codebase.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-build/issues/1513
Closes https://github.com/opensearch-project/opensearch-build/issues/1512
Closes https://github.com/opensearch-project/opensearch-build/issues/1237

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -32,7 +32,7 @@ jobs:
           pipenv run flake8 .
       - name: Run Type Checker
         run: |
-          pipenv run mypy src/build_workflow tests/tests_build_workflow src/checkout_workflow tests/tests_checkout_workflow src/run_assemble.py tests/test_run_assemble.py src/assemble_workflow tests/tests_assemble_workflow src/manifests tests/tests_manifests src/paths tests/tests_paths src/system tests/tests_system src/ci_workflow tests/tests_ci_workflow src/manifests_workflow tests/tests_manifests_workflow
+          pipenv run mypy .
       - name: Run Tests with Coverage
         run: |
           pipenv run coverage run -m pytest --cov=./src --cov-report=xml
--- a/src/test_workflow/integ_test/integ_test_start_properties_opensearch.py
+++ b/src/test_workflow/integ_test/integ_test_start_properties_opensearch.py
@@ -9,6 +9,8 @@ from test_workflow.integ_test.integ_test_start_properties import IntegTestStartP
 
 
 class IntegTestStartPropertiesOpenSearch(IntegTestStartProperties):
-    def __init__(self, path):
+    dependency_installer: DependencyInstallerOpenSearch
+
+    def __init__(self, path: str):
         super().__init__(path, "builds/opensearch/manifest.yml", "dist/opensearch/manifest.yml")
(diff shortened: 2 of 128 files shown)


.github/workflows/python-tests.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           pipenv run flake8 .
       - name: Run Type Checker
         run: |
-          pipenv run mypy src/build_workflow tests/tests_build_workflow src/checkout_workflow tests/tests_checkout_workflow src/run_assemble.py tests/test_run_assemble.py src/assemble_workflow tests/tests_assemble_workflow src/manifests tests/tests_manifests src/paths tests/tests_paths src/system tests/tests_system src/ci_workflow tests/tests_ci_workflow src/manifests_workflow tests/tests_manifests_workflow
+          pipenv run mypy .
```
peterzhuamazon, line 35: In my exp, using `.` will not properly check all the relationships between the types.

dblock: Can you please elaborate? Regardless of how I run it I get the same results, 16.07% imprecise, 18241 LOC.

peterzhuamazon: I think previously I have to specify all the folders for the mypy to check errors, or it will skip or pass even though I have wrong type.

dblock: I am not seeing this, but I could be wrong. How do I know? Running `.` vs. `src tests` makes no difference. AFAIK the reason why we have this list here is that we didn't have typings in all files.

peterzhuamazon: I can probably be wrong here so I was wondering if you can confirm the usages, but if not I am ok we are using `.` for now.

src/test_workflow/integ_test/integ_test_start_properties_opensearch.py
```diff
@@ -9,6 +9,8 @@
 
 
 class IntegTestStartPropertiesOpenSearch(IntegTestStartProperties):
-    def __init__(self, path):
+    dependency_installer: DependencyInstallerOpenSearch
+
+    def __init__(self, path: str):
```
zelinh, line 14: This method should also return `None`?

dblock: Yeah, so why isn’t mypi failing on this? Any ideas?

dblock: Found the answer in https://mypy.readthedocs.io/en/stable/common_issues.html

[__init__](https://docs.python.org/3/reference/datamodel.html#object.__init__) method has no annotated arguments or return type annotation. [__init__](https://docs.python.org/3/reference/datamodel.html#object.__init__) is considered fully-annotated if at least one argument is annotated, while mypy will infer the return type as None.

To be consistent, added missing annotations in [Added missing __init__ annotations.](https://github.com/opensearch-project/opensearch-build/pull/2123/commits/a8832cb9201f2088f9dab08fbce6f0a2cffa212b)

zelinh: I was wondering the same thing. Thanks for the explanation!

zelinh (approved):
LGTM. Thanks.

Issue #1762, opened 2022-03-15
[Enhancement] Add extra docker tags specifically for different builds


peterzhuamazon:
We have seen an issue today were IM committed a change during code freeze time:
https://github.com/opensearch-project/index-management/pull/311

This triggers a new build in jenkins, although we have separate links for tarball, docker only has one tag `1.3.0` and get overwritten.

I propose we have separate tags for docker based on builds.

Ex:

`latest`: always pointing latest
`1.3.0`: pointing to the latest version in 1.3.0
`1.3.0.1234`: pointing to the version with build number 1234 in 1.3.0

This way we can for sure lock all the artifacts to a specific version and will not overwrite without backups.

Thanks.

dblock:
Would be great to get this done in the 2.0 timeframe.

zelinh:
[Triage] For the tag with build number in it, we could activate it when there will be a release candidate coming to avoid too many images.

peterzhuamazon:
Update 20220607:
* https://github.com/opensearch-project/opensearch-build/pull/2163

peterzhuamazon:
Update 20220609:
* #2176

peterzhuamazon:
This is complete and can be closed for now.
Thanks.

PR #2163, opened 2022-06-07
Add new options for distribution build to tag docker with build number

peterzhuamazon:

### Description
Add new options for distribution build to tag docker with build number

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1762


zelinh (approved):

jenkins/docker/docker-build.jenkinsfile
```diff
@@ -40,18 +35,15 @@ pipeline {
...
-                    }
+                    def CREDENTIAL_ID = "jenkins-staging-dockerhub-credential"
                     sh "echo Account: $CREDENTIAL_ID"
                     withCredentials([usernamePassword(credentialsId: CREDENTIAL_ID, usernameVariable: 'DOCKER_USERNAME', passwordVariable: 'DOCKER_PASSWORD')]) {
                         sh '''
-                            set -ex
+                            set -e
                             echo Login to $CREDENTIAL_ID
-                            docker logout && docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD && eval $DOCKER_BUILD_SCRIPT_WITH_COMMANDS
+                            docker logout && echo $DOCKER_PASSWORD | docker login -u $DOCKER_USERNAME --password-stdin && eval $DOCKER_BUILD_SCRIPT_WITH_COMMANDS
```
gaiksaya, line 46: Echoing password is not a good practice. Is this change intended?

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -38,10 +38,10 @@ pipeline {
             defaultValue: "bwc-test-opensearch-dashboards",
             trim: true
         )
-        booleanParam(
+        choice(
             name: 'BUILD_DOCKER',
-            description: 'Build docker image or not.',
-            defaultValue: true
+            description: 'Build docker image or not with options.',
+            choices: ['build_docker', 'build_docker_with_build_number_tag', 'do_not_build_docker'],
```
gaiksaya, line 44: Wanted to confirm if programmatically triggering this workflow might be affected by this change

gaiksaya (changes requested):

jenkins/docker/docker-build.jenkinsfile
peterzhuamazon: @prudhvigodithi 
This is intended and it is actually much safer than the original setup.
Note that the password is pipe directly into --password-stdin instead of stdout.
In that case `docker login` command will not print `login with -u **** -p ****` which is the best practice from docker.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
peterzhuamazon: I have already taken care of all the instances that get touched by this change.
Thanks.

jenkins/docker/docker-build.jenkinsfile
gaiksaya: But when you do echo $DOCKER_PASSWORD it will emit an password right?

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
peterzhuamazon: Jenkins default the 1st entry in the choices list is the default, thus the behavior is exactly the same as before and I have changed all the instance to reflect this change.

gaiksaya: I guess you didn't get my question. I was concerned if choices and parameters have different affect on the workflow. If build-workflow triggers docker-workflow, can we select an option like default choice or something?

jenkins/docker/docker-build.jenkinsfile
peterzhuamazon: It send the password to the pipe it did not emit to anywhere else.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
gaiksaya: Ok got it. This was new to me Thanks!

jenkins/docker/docker-build.jenkinsfile
gaiksaya: Okay still seems risky. Better to monitor and maybe enhance later.

gaiksaya: Why can't we keep what was before? I am not comfortable with echoing password.

gaiksaya: [Offline discussion] Looks like previously docker did emit the password. This way it will not. Hence, this is better. Thanks!

zelinh: I see we also use it here. https://github.com/opensearch-project/opensearch-build/blob/92efa806f8bb3321aa0af40711ba6de3b879b93f/vars/copyContainer.groovy#L16

peterzhuamazon: Thank you!

gaiksaya (approved):
Recommend to add a note that choices first option is the default. Easy to miss thing.
Thanks!

PR #2209, opened 2022-06-17
Hardcode artifact upload role in vars/uploadToS3.groovy

gaiksaya:

### Description
The artifact upload role name is hardcoded in other places already. Example: https://github.com/opensearch-project/opensearch-build/blob/main/vars/buildUploadManifestSHA.groovy#L6, https://github.com/opensearch-project/opensearch-build/blob/main/vars/getManifestSHA.groovy#L24
Hence, hardcoding it here as well instead of getting it's value from environment variable.

### Issues Resolved


vars/uploadToS3.groovy
```diff
@@ -1,6 +1,6 @@
 void call(Map args = [:]){
 
-    withAWS(role: "${ARTIFACT_UPLOAD_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
+    withAWS(role: 'opensearch-bundle', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
         s3Upload(file: args.sourcePath, bucket: args.bucket, path: args.path)
```
prudhvigodithi, line 4: @gaiksaya, It would be better we pull the withAWS() role info from credentials
https://github.com/opensearch-project/opensearch-build/blob/main/vars/copyContainer.groovy#L24-L35

gaiksaya: Its just a name, not the actual value. I checked other places (see PR description) keeping it like that. We have policies in place so random role assumption is restricted.

prudhvigodithi (approved):

Issue #2220, opened 2022-06-21
Move all AWS account numbers, bucket names and other sensitive data to secrets


gaiksaya:
### Is your feature request related to a problem? Please describe

With jenkins going public, we observed that few variables which should be secrets are rather being stored and used as environment variables.
From security perspective, it is better if we move all those to credential store (AWS secrets manager) and use it as secrets rather than exposed as env variables.

Few of them are as follow:

- ARTIFACT_BUCKET_NAME
- ARTIFACT_PRODUCTION_BUCKET_NAME
- SIGN_ASM_ACCOUNT, SIGN_ASM_KEYID, SIGN_ASM_REGION, SIGN_ASM_ROLE (can go as one secret) see [example](https://github.com/opensearch-project/opensearch-build/pull/2156/files#diff-63b9bb4025ab21358c7d296b6204b847ecfffac44b6b249a2363824b30c61b1fR31-R38)
- SIGNER_CLIENT_EXTERNAL_ID, SIGNER_CLIENT_ROLE, SIGNER_CLIENT_SIGNED_BUCKET, SIGNER_CLIENT_UNSIGNED_BUCKET (can go as one secret) See [example](https://github.com/opensearch-project/opensearch-build/pull/2156/files#diff-63b9bb4025ab21358c7d296b6204b847ecfffac44b6b249a2363824b30c61b1fR31-R38)
- SONATYPE_STAGING_PROFILE_ID
- AWS_ACCOUNT_ARTIFACT
- AWS_ACCOUNT_PUBLIC
- AWS_ACCOUNT_TESTING

### Describe the solution you'd like

Move them to AWS secrets manager

bbarani:
@gaiksaya Can we close this issue?

gaiksaya:
Not yet! Still yet to go through jenkinsFiles to see any ad-hoc environment variables usage.

PR #2226, opened 2022-06-22
Update uploadToS3 lib env vars with credentials

gaiksaya:

### Description
Starting with `uploadToS3` library to move environment variables to secrets.
See #2220 for more details. 
Updated the current jenkins with this credential

### Issues Resolved
Part of #2220 


vars/uploadToS3.groovy
```diff
@@ -1,7 +1,8 @@
 void call(Map args = [:]){
 
-    withAWS(role: 'opensearch-bundle', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session') {
-        s3Upload(file: args.sourcePath, bucket: args.bucket, path: args.path)
+    withCredentials([string(credentialsId: 'AWS_ACCOUNT_PUBLIC', variable: 'aws_account_public')]) {
```
prudhvigodithi, line 3: `credentialsId` values should be start with `jenkins-*` prefix, this way we can configure the jenkins master role to only pull secrets that has `jenkins-*` prefix Example as https://github.com/opensearch-project/opensearch-build/blob/main/vars/copyContainer.groovy#L25

prudhvigodithi (approved):

vars/uploadToS3.groovy
gaiksaya: I believe whatever secrets we store in under the account will all belong to jenkins. So won't be it an overhead?

prudhvigodithi: It will also contain secrets that are part of `OpenSearch-CI-Config-Dev `stack
https://github.com/opensearch-project/opensearch-ci, also its possible to to filter with `jenkins-*` prefix
Screen Shot 2022-06-22 at 5 43 13 PM

gaiksaya: True but credentials are only fetched using a particular tag as per the plugin documentation. Since we are not adding these tags in config or CI stacks, they will be on jenkin's credential DB. 
https://plugins.jenkins.io/aws-secrets-manager-credentials-provider/#plugin-content-usage

prudhvigodithi: I just want to eliminate jenkins master able to query all available secret manager credentials in that account, would be good to have only credentials with prefix `jenkins-*` be queried, then it would be jenkins app specific.

gaiksaya: I see. what about existing secrets. Looks like we cannot change name of already stored secrets.

prudhvigodithi: Ya we have to recreate a new one and change the value in jenkinsfile, @peterzhuamazon thoughts please?

prudhvigodithi: Ya we have to recreate a new one and change the value in jenkinsfile, @peterzhuamazon thoughts please?

PR #2231, opened 2022-06-22
Reduce input manifest build frequency for new upcoming versions

peterzhuamazon:

### Description
Reduce input manifest build frequency for new upcoming versions.
Every time when a new input manifest getting created, it often doesnt have all the necessary pieces.
Core not ready, plugin not bumping or supporting it, etc.
It is not very useful to build every 10min for these broken builds.

### Issues Resolved

### DIFF
--- a/src/manifests_workflow/input_manifests.py
+++ b/src/manifests_workflow/input_manifests.py
@@ -146,7 +146,7 @@ class InputManifests(Manifests):
         with open(jenkinsfile, "r") as f:
             data = f.read()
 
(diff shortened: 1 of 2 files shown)


gaiksaya (approved):
Maybe also need to add step in release template to increase the frequency when release is close by

peterzhuamazon:
> Maybe also need to add step in release template to increase the frequency when release is close by

We still need to think about how to approach this cause machines does not know when we want to release 😄

PR #2156, opened 2022-06-06
Add signer to support signing windows artifacts

zelinh:
Signed-off-by: Zelin Hao <zelinhao@amazon.com>

### Description
This PR will enable our sign script to sign for windows distribution and also the jenkins library `signArtifacts` to be able to sign for windows platform.

For windows signing with the scripts, the signed artifacts will be stored in a subfolder called `signed` and keep its origin name. E.g.: the original unsigned file path is `/tmp/opensearch.exe`, the signed artifacts will be stored with file path `/tmp/signed/opensearch.exe`. 

### Issues Resolved
#1430 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2275, opened 2022-06-28
Update signArtifacts lib env vars with credentials

gaiksaya:

### Description
Update signArtifacts lib env vars with credentials.
Confirming that the credentials used here have been updated in current dev jenkins.

### Issues Resolved
Part of #2220 

### DIFF
--- a/vars/signArtifacts.groovy
+++ b/vars/signArtifacts.groovy
@@ -15,15 +15,17 @@ SignArtifacts signs the given artifacts and saves the signature in the same dire
 @param Map[platform] <Required> - The distribution platform for signing.
 */
 void call(Map args = [:]) {
-
     if (args.sigtype.equals('.rpm')) {
-        echo "RPM Add Sign"
+        withCredentials([string(credentialsId: 'jenkins-rpm-signing-props', variable: 'configs')]) {
+                def props = readJSON(text: configs)
+                def signingAccount = props['account']
+                def signingPassphraseSecretsArn = props['passphrase_secrets_arn']
+                def signingSecretKeyIdSecretsArn = props['secret_key_id_secrets_arn']
+                def signingKeyId = props['key_id']
 
-        withAWS(role: "${SIGN_ASM_ROLE}", roleAccount: "${SIGN_ASM_ACCOUNT}", duration: 900, roleSessionName: 'jenkins-signing-session') {
-            withCredentials([
-                string(credentialsId: 'jenkins-rpm-signing-asm-pass-id', variable: 'SIGNING_PASS_ID'),
-                string(credentialsId: 'jenkins-rpm-signing-asm-secret-id', variable: 'SIGNING_SECRET_ID')])
-                {
+            echo 'RPM Add Sign'
+
+            withAWS(role: 'jenki-jenki-asm-assume-role', roleAccount: "${signingAccount}", duration: 900, roleSessionName: 'jenkins-signing-session') {
                     sh """
                         set -e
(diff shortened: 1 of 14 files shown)


gaiksaya:
> This should be conflict with my PR for window signing. #2156. Maybe we update the environment variables after that?

Would prefer to go this in first, then you can need to make minimal changes to `vars` 
Thanks!

vars/signArtifacts.groovy
```diff
@@ -15,15 +15,16 @@ SignArtifacts signs the given artifacts and saves the signature in the same dire
 @param Map[platform] <Required> - The distribution platform for signing.
 */
 void call(Map args = [:]) {
-
     if (args.sigtype.equals('.rpm')) {
-        echo "RPM Add Sign"
+        withCredentials([string(credentialsId: 'jenkins-rpm-signing-props', variable: 'configs')]) {
```
zelinh, line 19: Have you run test signing for rpm package with these changes? We should be careful with this change as this is built in our distribution workflow.

gaiksaya: I have updated the credentials and the regression files seems fine to me.

vars/signArtifacts.groovy
```diff
@@ -89,56 +90,58 @@ void call(Map args = [:]) {
 
                         echo "------------------------------------------------------------------------"
                         echo "Clean up gpg"
-                        gpg --batch --yes --delete-secret-keys $SIGN_ASM_KEYID
-                        gpg --batch --yes --delete-keys $SIGN_ASM_KEYID
+                        gpg --batch --yes --delete-secret-keys C5B7498965EFD1C2924BA9D539D319879310D3FC
```
zelinh, line 93: I don't like we explicit show it here, even though it's public. It may cause other confused.

vars/signArtifacts.groovy
```diff
@@ -15,15 +15,16 @@ SignArtifacts signs the given artifacts and saves the signature in the same dire
...
+                def signingSecretKeyIdSecretsArn = props['secret_key_id_secrets_arn']
+
+            echo 'RPM Add Sign'
 
-        withAWS(role: "${SIGN_ASM_ROLE}", roleAccount: "${SIGN_ASM_ACCOUNT}", duration: 900, roleSessionName: 'jenkins-signing-session') {
-            withCredentials([
-                string(credentialsId: 'jenkins-rpm-signing-asm-pass-id', variable: 'SIGNING_PASS_ID'),
-                string(credentialsId: 'jenkins-rpm-signing-asm-secret-id', variable: 'SIGNING_SECRET_ID')])
-                {
+            withAWS(role: 'jenki-jenki-asm-assume-role', roleAccount: "${signingAccount}", duration: 900, roleSessionName: 'jenkins-signing-session') {
```
zelinh, line 27: This role is still env var?

vars/signArtifacts.groovy
gaiksaya: Consulted with @peterzhuamazon. This is a public key as on our website https://opensearch.org/verify-signatures.html. I dont feel like its an env variable or secret as such.

vars/signArtifacts.groovy
gaiksaya: Nope its an actual role name. No need to keep role names as secrets or env variables.

vars/signArtifacts.groovy
gaiksaya: Tested! Ran successfully

PR #2301, opened 2022-07-06
Exclude manifest checks for 3.0.0 for now

zelinh:

### Description
Exclude manifest checks for 3.0.0 manifest for now since it requires JDK 17.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

### Issues Resolved

### DIFF
--- a/.github/workflows/manifests.yml
+++ b/.github/workflows/manifests.yml
@@ -5,7 +5,8 @@ on:
   push:
   pull_request:
     paths:
-      - manifests/**/*.yml
+      - 'manifests/**/*.yml'
+      - '!manifests/3.**/**'
   schedule:
     - cron: 0 0 * * *
 


.github/workflows/manifests.yml
```diff
@@ -5,7 +5,8 @@ on:
   push:
   pull_request:
     paths:
-      - manifests/**/*.yml
+      - 'manifests/**/*.yml'
+      - '!manifests/3.**/**'
```
gaiksaya, line 9: Instead if excluding can you add java 17 for 3.x? I dont think it is beneficial to exclude the manifests

zelinh: Are you suggesting we adding another identical workflow with JDK 17 that includes check for 3.x manifest?

prudhvigodithi:
Just a thought, we can run 2 jobs one use 14 for versions less than 3.x and other use jdk17 for versions greater than 3.x

peterzhuamazon (changes requested):
Hi.

We should change the matrix to have:

<=1 use jdk11
>=2 use jdk17.

JDK14 has been deprecated so we should not use it anymore.

Thanks.

zelinh:
> Maybe this is a good candidate for a matrix with multiple values like [here](https://github.com/opensearch-project/opensearch-java/blob/main/.github/workflows/test-integration.yml#L17) where one of the values is the version comparison in the manifest files and the other the version of the JDK? Would be a lot shorter?
> 
> Another option to try is to extract the version of the JDK from the manifest dynamically as a step.

Thanks for the input thoughts on here!
I think we would need to add each entry every time we have a new manifest if we have them hardcoded there and this might be getting longer as we have more releases. 

I could dive deep into the possible way to see if we could extract the jdk version from the manifest as an alternative way.

gaiksaya (approved):
CI is blocked right now because of 3.0.0 manifest checks. Approving this to unblock. @zelinh  please create an issue if we need to revisit this approach.
Thanks!

PR #2354, opened 2022-07-20
Add Java home argument to the manifest workflow

### LINKED ISSUE #2044: Generated manifests for OpenSearch 2.x need custom JAVA_HOME

Automation opened a PR to generate the OpenSearch 3.0 manifest. That manifest was missing `JAVA_HOME` that was added manually in https://github.com/opensearch-project/opensearch-build/pull/2042.

Add a way to customize generated manifests for various products/versions, add `JAVA_HOME` to the OpenSearch ones.

Keep doing it manually.

### Additional context

_No response_

zelinh:
### Description
Add Java home argument to the manifest workflow based on the manifest version. If it's 1.x (in our case 1.3.x primarily), it will add jdk 11 argument. If it's 2.x or 3.x, it will add jdk 17 for now. Otherwise, it will set to jdk17 by default. 

### Issues Resolved
#2044 

### DIFF
--- a/src/manifests_workflow/input_manifests.py
+++ b/src/manifests_workflow/input_manifests.py
@@ -126,7 +126,7 @@ class InputManifests(Manifests):
             "ci": {
                 "image": {
                     "name": image_map[self.prefix],
-                    "args": jdk_map[version.split(".")[0]]
+                    "args": jdk_map.get(version.split(".")[0], jdk_map["3"])
                 }
             },
             "components": [],
--- a/src/manifests_workflow/input_manifests.py
+++ b/src/manifests_workflow/input_manifests.py
@@ -111,6 +111,12 @@ class InputManifests(Manifests):
             "opensearch-dashboards": "opensearchstaging/ci-runner:ci-runner-centos7-opensearch-dashboards-build-v2"
         }
 
+        jdk_map = {
+            "1": "-e JAVA_HOME=/opt/java/openjdk-11",
+            "2": "-e JAVA_HOME=/opt/java/openjdk-17",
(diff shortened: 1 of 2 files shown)


dblock (changes requested):
This is going to start adding JDK args into OpenSearch Dashboards' manifests as well, which is certainly not desired.

I think you need to do something more elaborate. Instead of hard-coding both `image_map` and `data: Dict = ...` maybe create a folder structure with templates and load those depending on the product and version being built, falling back to a default that doesn't have anything in it?

```
manifests/templates/opensearch/main/manifest.yml
manifests/templates/opensearch/2.x/manifest.yml
manifests/templates/opensearch-dashboards/main/manifest.yml
```

Another alternative would be to look for the previous version of the manifest, and copy `image` and `args` from there. Maybe simpler?

zelinh:
> This is going to start adding JDK args into OpenSearch Dashboards' manifests as well, which is certainly not desired.
> 
> I think you need to do something more elaborate. Instead of hard-coding both `image_map` and `data: Dict = ...` maybe create a folder structure with templates and load those depending on the product and version being built, falling back to a default that doesn't have anything in it?
> 
> ```
> manifests/templates/opensearch/main/manifest.yml
> manifests/templates/opensearch/2.x/manifest.yml
> manifests/templates/opensearch-dashboards/main/manifest.yml
> ```
> 
> Another alternative would be to look for the previous version of the manifest, and copy `image` and `args` from there. Maybe simpler?
> 
> Finally, if you do go the hard-coded route, create a map of maps to split up OS and OSD.

oops you are right. I didn't think about about the dashboard situation. I will close this PR for now and dive into the solution you propose. Thanks.

PR #2333, opened 2022-07-13
Add legacy-manifests folder

prudhvigodithi:
### Description
Migrate the old manifests to legacy-manifests folder 

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2332

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2281, opened 2022-06-29
Add java args to 2.2.0 manifest

gaiksaya:
Signed-off-by: Sayali Gaikawad <gaiksaya@amazon.com>

### Description
2.2.0 build were failing for opensearch component due to missing java args. Adding it in this PR

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2358, opened 2022-07-21
Add JAVA_HOME arguments in manifest workflow from template manifests.

### LINKED ISSUE #2044: Generated manifests for OpenSearch 2.x need custom JAVA_HOME

zelinh:
### Description
Add Java home argument to the manifest workflow based on the manifest version based on its major version templates stored in the `manifests` folder. If it's 1.x (in our case 1.3.x primarily), it will add jdk 11 argument. If it's 2.x or 3.x, it will add jdk 17 for now. Otherwise, it will set to be corresponded with 2.x templates which has jdk17 by default.

Followup to the previous closed PR #2354 

### Issues Resolved
#2044 

### DIFF
--- a/.github/workflows/manifests.yml
+++ b/.github/workflows/manifests.yml
@@ -5,7 +5,8 @@ on:
   push:
   pull_request:
     paths:
-      - manifests/**/*.yml
+      - 'manifests/**/*.yml'
+      - '!manifests/templates/**/'
   schedule:
     - cron: 0 0 * * *
 
--- a/src/manifests/input_manifest.py
+++ b/src/manifests/input_manifest.py
@@ -136,8 +136,6 @@ class InputManifest(ComponentManifest['InputManifest', 'InputComponents']):
 
             def __to_dict__(self) -> dict:
                 return {
-                    "name": self.name
-                } if self.args is None else {
                     "name": self.name,
                     "args": self.args
                 }
--- a/src/manifests/input_manifest.py
+++ b/src/manifests/input_manifest.py
@@ -136,6 +136,8 @@ class InputManifest(ComponentManifest['InputManifest', 'InputComponents']):
 
             def __to_dict__(self) -> dict:
                 return {
+                    "name": self.name
+                } if self.args is None else {
                     "name": self.name,
                     "args": self.args
                 }
--- a/src/manifests_workflow/component.py
+++ b/src/manifests_workflow/component.py
@@ -28,4 +28,9 @@ class Component:
         return branches
 
     def to_dict(self) -> Any:
-        return Manifest.compact({"name": self.name, "repository": self.git_repo.url, "ref": self.git_repo.ref, "checks": self.checks})
+        return Manifest.compact({
+            "name": self.name,
+            "repository": self.git_repo.url,
+            "ref": self.git_repo.ref,
+            "checks": self.checks
+        })
--- a/src/manifests_workflow/component.py
+++ b/src/manifests_workflow/component.py
@@ -28,9 +28,4 @@ class Component:
         return branches
 
     def to_dict(self) -> Any:
-        return Manifest.compact({
-            "name": self.name,
-            "repository": self.git_repo.url,
-            "ref": self.git_repo.ref,
-            "checks": self.checks
-        })
+        return Manifest.compact({"name": self.name, "repository": self.git_repo.url, "ref": self.git_repo.ref, "checks": self.checks})
--- a/src/manifests_workflow/input_manifests.py
+++ b/src/manifests_workflow/input_manifests.py
@@ -11,7 +11,7 @@ import re
 from abc import abstractmethod
 from typing import Dict, List, Type, Union
 
-from manifests.input_manifest import InputManifest
+from manifests.input_manifest import InputComponents, InputManifest
 from manifests.manifests import Manifests
 from manifests_workflow.component_opensearch import ComponentOpenSearch
 from manifests_workflow.component_opensearch_dashboards_min import ComponentOpenSearchDashboardsMin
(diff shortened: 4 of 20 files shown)


zelinh:
This PR is intended to refactor some of our hardcoding in our `manifest_workflow` and resolved the issue to add JAVA_HOME argument manually as what we did in [this PR](https://github.com/opensearch-project/opensearch-build/pull/2281).

gaiksaya (approved):
LGTM. Thanks!

src/manifests/input_manifest.py
```diff
@@ -136,6 +136,8 @@ def __init__(self, data: dict) -> None:
 
             def __to_dict__(self) -> dict:
                 return {
+                    "name": self.name
+                } if self.args is None else {
```
dblock, line 140: You don't need to do this because `to_dict` should compact the outcome. Just return what you have in `else`.

src/manifests_workflow/input_manifests.py
```diff
@@ -106,22 +106,23 @@ def update(self, min_klass: Union[Type[ComponentOpenSearchMin], Type[ComponentOp
...
-        }
+        templates_base_path = os.path.join(self.manifests_path(), "templates")
+        template_version_folder = version.split(".")[0] + ".x"
+        template_full_path = os.path.join(templates_base_path, self.prefix, template_version_folder,
+                                          "manifest.yml")
+        if os.path.exists(template_full_path):
+            input_manifest_templates = InputManifest.from_file(open(template_full_path))
+        else:
+            input_manifest_templates = InputManifest.from_file(open(os.path.join(templates_base_path, self.prefix,
+                                                                                 "2.x", "manifest.yml")))
```
dblock, line 117: This hard-coding is bound to cause problems. 

1. Use the entire manifest, not just the `templates.ci` part, so remove the default `data`. 
2. Add a default folder for the default manifest and load that instead of 2.x/manifest.yml.

dblock (changes requested):
This is good!

See below to make it a bit simpler.

Write tests for using the `default` manifest vs. using a manifest file that exists.

src/manifests/input_manifest.py
zelinh: @dblock Actually I think we might need this. Otherwise it will output `args: None` in dashboard manifest it generated.

dblock: Make sure we call `to_dict`, which [removes these `None` values](https://github.com/opensearch-project/opensearch-build/blob/794088835319cfd015a107e77f99d3a0b31efad1/src/manifests/manifest.py#L96).

src/manifests_workflow/component.py
```diff
@@ -28,9 +28,4 @@ def branches(self, url: str) -> List[str]:
         return branches
 
     def to_dict(self) -> Any:
-        return Manifest.compact({
-            "name": self.name,
-            "repository": self.git_repo.url,
-            "ref": self.git_repo.ref,
-            "checks": self.checks
-        })
+        return Manifest.compact({"name": self.name, "repository": self.git_repo.url, "ref": self.git_repo.ref, "checks": self.checks})
```
dblock, line 31: I know the linter undoes this formatting, maybe just revert this file since there are no changes.

src/manifests_workflow/input_manifests.py
```diff
@@ -106,30 +111,23 @@ def update(self, min_klass: Union[Type[ComponentOpenSearchMin], Type[ComponentOp
...
-                "image": {
-                    "name": image_map[self.prefix]
-                }
-            },
-            "components": [],
-        }
+        templates_base_path = os.path.join(self.manifests_path(), "templates")
+        template_version_folder = version.split(".")[0] + ".x"
+        template_full_path = os.path.join(templates_base_path, self.prefix, template_version_folder, "manifest.yml")
+        if os.path.exists(template_full_path):
```
dblock, line 117: Avoid duplicating `InputManifest.from_file` logic, this is shorter:

```python
template_path = ...
if not os.path.exists(template_path):
    template_full_path = os.path.join(teplate_base_path, ...)

manifest = InputManifest.from_file(template_path)
```

Rename `input_manifest_template` to `manifest` since it's an actual manifest.

You can probably remove/edit the hard-coded version in the manifest templates to avoid confusion, it's a template, so maybe just say 'version: replace' or something like that.

dblock (changes requested):
I think there's one more problem, the manifests CI workflow is going to pickup these templates. They probably need to move or the folder needs to be excluded somehow.

https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/manifests.yml#L20

src/manifests_workflow/component.py
zelinh: > I think there's one more problem, the manifests CI workflow is going to pickup these templates. They probably need to move or the folder needs to be excluded somehow.
> 
> https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/manifests.yml#L20

This is actually my next step after this PR. I plan to take over this PR to resolve some of the manifest CI failures we have right now. https://github.com/opensearch-project/opensearch-build/pull/2333
I will temporary exclude the templates checks in this PR then.

dblock: Right, I just don't want this PR to break what works today.

src/manifests_workflow/input_manifests.py
zelinh: Updated with new commits. Please help review it. Thanks.

.github/workflows/manifests.yml
```diff
@@ -5,7 +5,8 @@ on:
   push:
   pull_request:
     paths:
-      - manifests/**/*.yml
+      - 'manifests/**/*.yml'
+      - '!manifests/templates/**/'
```
dblock, line 9: That will not trigger the workflow on changes in that path, but will still get the list of all .yml files when it does run.

src/manifests_workflow/input_manifests.py
dblock: See https://github.com/opensearch-project/opensearch-build/pull/2358#discussion_r931498438

.github/workflows/manifests.yml
zelinh: @dblock It may actually won't. Since we executed on manifests file that contains `opensearch` prefix like this `ls manifests/**/opensearch*.yml`. All our templates are called just `manifest.yml`

dblock: Aha! Nice. That works then :)

dblock (approved):

Issue #2361, opened 2022-07-22
Add a new darwin build stage to OpenSearch distribution pipeline


prudhvigodithi:
### Is your feature request related to a problem? Please describe

Part of: https://github.com/opensearch-project/opensearch-build/issues/2216
Add a new darwin build stage to [OpenSearch distribution pipeline](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile).

### Describe the solution you'd like

The new stage, part of pipeline will be restricted to run on MacOS agents, so the platform can be inferred as darwin, with the existing [build_workflow](https://github.com/opensearch-project/opensearch-build/tree/main/src/build_workflow) and [assemble_workflow](https://github.com/opensearch-project/opensearch-build/tree/main/src/assemble_workflow) logic the ouput should be tar artifcat for darwin.

### Describe alternatives you've considered

N/A

### Additional context

N/A

prudhvigodithi:
Closing this issue as the associated PR https://github.com/opensearch-project/opensearch-build/pull/2362 is merged.
Thank you

Issue #2216, opened 2022-06-20
Build test Darwin version of min OpenSearch distribution to support plugin testing on MacOS.


prudhvigodithi:
### Is your feature request related to a problem? Please describe
For k-NN, to run integ tests on Mac is not compatible.
KNN uses lang painless to inject scripts. To test on MacOS right now the options are `INTEG_TEST`, which will bring up a cluster with plain and no modules (mainly lang painless) except [transportModulesFiles](https://github.com/opensearch-project/OpenSearch/blob/main/distribution/archives/build.gradle) 
Second option is to use `ARCHIVE` which will allow direct download of zip and test bringing up the cluster, which looks for darwin zip, which is not available for testing.

https://github.com/opensearch-project/OpenSearch/issues/1676

### Describe the solution you'd like

- [x]  Test the existing [build_workflow](https://github.com/opensearch-project/opensearch-build/tree/main/src/build_workflow) to run the OpenSearch build with platform as `darwin`.
https://github.com/opensearch-project/opensearch-build/issues/2350
- [x] Run the existing [distribution build job](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile) on MacOS ec2 agent which outputs the min distribution of OpenSearch of type Darwin.
https://github.com/opensearch-project/opensearch-build/issues/2361
- [x] Build a Mac ec2 AMI able to execute [build_workflow](https://github.com/opensearch-project/opensearch-build/tree/main/src/build_workflow) and [assemble_workflow](https://github.com/opensearch-project/opensearch-build/tree/main/src/assemble_workflow) that can generate OpenSrearch min distribution.
https://github.com/opensearch-project/opensearch-ci/issues/163
- [x] Automate the Mac jenkins agents creation and add to [Public Jenkins](https://build.ci.opensearch.org/).
https://github.com/opensearch-project/opensearch-ci/issues/157
- [x] Add required missing packages to MacOS agents, that support signing workflow, to sign and upload darwin artifacts
https://github.com/opensearch-project/opensearch-build/issues/2375
- [x] MacOS agents support default `/bin/bash`, remove `nodeProperties` in CDK code and allow jenkins to use default `PATH` part of `AMI`.
https://github.com/opensearch-project/opensearch-ci/issues/174
- [x] Validate the darwin artifacts built using the [distribution-build job](https://build.ci.opensearch.org/job/distribution-build-opensearch/).
[...]

PR #2362, opened 2022-07-22
Add a new darwin build stage to OpenSearch distribution pipeline

prudhvigodithi:
### Description
Add a new darwin build stage to OpenSearch distribution pipeline

### Issues Resolved
Related issue: https://github.com/opensearch-project/opensearch-build/issues/2361
Part of: https://github.com/opensearch-project/opensearch-build/issues/2216


jenkins/opensearch/distribution-build.jenkinsfile
```diff
@@ -140,6 +140,32 @@ pipeline {
                         }
                     }
                 }
+                stage('macos-build-x64-tar') {
```
gaiksaya, line 143: Just say `build-macos-x64-tar`

jenkins/opensearch/distribution-build.jenkinsfile
```diff
@@ -140,6 +140,32 @@ pipeline {
...
+                    agent {
+                        node {
+                            label 'Jenkins-Agent-MacOS12-X64-Mac1Metal-Multi-Host'
+                        }
+                    }
+                    tools {
+                        jdk dockerAgent.javaVersion
+                    }
```
gaiksaya, line 151: Is it something specially needed by macos? Because each docker image has java version specified in manifest args that is used for building the artifact

vars/detectDockerAgent.groovy
```diff
@@ -4,9 +4,18 @@ Map call(Map args = [:]) {
     def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: manifest))
     dockerImage = inputManifest.ci?.image?.name ?: 'opensearchstaging/ci-runner:ci-runner-centos7-v1'
     dockerArgs = inputManifest.ci?.image?.args
+    // Using default javaVersion as jdk-17
+    String javaVersion = 'jdk-17'
```
gaiksaya, line 8: The dockerArgs is the java version. More details in this PR

jenkins/opensearch/distribution-build.jenkinsfile
prudhvigodithi: Hey @gaiksaya, we are not going with docker in mac, there is no docker installed, docker cli for mac is not supported, hence it has to be done this way.

vars/detectDockerAgent.groovy
prudhvigodithi: `dockerArgs` is not only for java version, its all args for the container it can be also as `-p 8080:8080`, so I'm getting the jdk match and extracting the version, else it will take the jdk17.
Thank you

gaiksaya (approved):
Nit: https://github.com/opensearch-project/opensearch-build/pull/2362#discussion_r927985976

vars/detectDockerAgent.groovy
```diff
@@ -4,9 +4,18 @@ Map call(Map args = [:]) {
...
+    java.util.regex.Matcher match = (dockerArgs =~ /jdk-\d+/) 
+    if (match.find()) {
+            def line = match[0]
+           javaVersion = line
```
peterzhuamazon, line 12: Spacing issues.

peterzhuamazon: Can you rename `line` to a better var name.

peterzhuamazon (approved):
Overall great just a few nitpicks.

vars/detectDockerAgent.groovy
prudhvigodithi: Fixed Peter

jenkins/opensearch/distribution-build.jenkinsfile
prudhvigodithi: Fixed Sayali

vars/detectDockerAgent.groovy
```diff
@@ -4,9 +4,18 @@ Map call(Map args = [:]) {
     def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: manifest))
     dockerImage = inputManifest.ci?.image?.name ?: 'opensearchstaging/ci-runner:ci-runner-centos7-v1'
     dockerArgs = inputManifest.ci?.image?.args
+    // Using default javaVersion as openjdk-17
+    String javaVersion = 'openjdk-17'
+    java.util.regex.Matcher jdkMatch = (dockerArgs =~ /openjdk-\d+/) 
+    if (jdkMatch.find()) {
+        def jdkMatchLine = jdkMatch[0]
+        javaVersion = jdkMatchLine
+    }
```
dblock, line 13: Seems complicated :) Match any 1+ digit

prudhvigodithi: Hey @dblock I have added the parsing logic, assuming we add multiple args in future to the manifest file
```
args: -e JAVA_HOME=/opt/java/openjdk-17 -e PATH=<> -p 8080:8080
```

dblock:
The real bug in this code is that if the JDK says “temurin-14” you end up with JDK 17 which will be wrong, but succeed silently. If there's a JDK is specified, and you can't extract the version the code should throw an error.

prudhvigodithi:
> The real bug in this code is that if the JDK says “temurin-14” you end up with JDK 17 which will be wrong, but succeed silently. If there's a JDK is specified, and you can't extract the version the code should throw an error.

I see got ur concern dB, everything we use now is temurin, but in our CI runner
https://github.com/opensearch-project/opensearch-build/blob/main/docker/ci/config/jdk-setup.sh#L44, we point it as openjdk-<version> and in backend [public jenkins](https://build.ci.opensearch.org/) [tools](https://github.com/opensearch-project/opensearch-ci/blob/main/resources/baseJenkins.yaml#L160) is configured as `openjdk-<version>`, in manifest also we have as `args: -e JAVA_HOME=/opt/java/openjdk-<version>` , hence the parsing logic is also looking for `openjdk`.
Thank you

dblock:
> > The real bug in this code is that if the JDK says “temurin-14” you end up with JDK 17 which will be wrong, but succeed silently. If there's a JDK is specified, and you can't extract the version the code should throw an error.
> 
> I see got ur concern dB, everything we use now is temurin, but in our [CI runners](https://github.com/opensearch-project/opensearch-build/blob/main/docker/ci/config/jdk-setup.sh#L44), we point it as `openjdk-<version>` and in backend [public jenkins](https://build.ci.opensearch.org/) [tools](https://github.com/opensearch-project/opensearch-ci/blob/main/resources/baseJenkins.yaml#L160) is configured as `openjdk-<version>`, in manifest also we have as `args: -e JAVA_HOME=/opt/java/openjdk-<version>` , hence the parsing logic is also looking for `openjdk`. Thank you

Just because it works with the current data doesn't mean it will work with new data. Your code fails to produce correct results with anything other than `openjdk-<version>`.

PR #2422, opened 2022-08-04
Temporary fix the manifest workflow

### LINKED ISSUE #2420: [Bug]: Updated input manifests creates old version manifests part of `legacy-manifests/` folder
### Describe the bug

With the change to move the old input manifests to [legacy-manifests folder](url) in build repo, the [auto generate manifest workflow ](https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/versions.yml) creates even the manifests part of the legacy-manifests folder assuming they does not exist.
Sample PR.
https://github.com/opensearch-project/opensearch-build/pull/2389/files

### To reproduce

The workflow PR
https://github.com/opensearch-project/opensearch-build/pull/2389/files

### Expected behavior

The `./manifest.sh update` logic should be modified:
1) Either it should create manifests greater than the version number from the manifests inside the [legacy manifest folder](https://github.com/opensearch-project/opensearch-build/tree/main/legacy-manifests)
2) Logic to compare both manifests and legacy-manifests folder.

### Host / Environment

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

zelinh:

### Description
Temporary fix the manifest workflow to creating new manifests on legacy. Hardcode setting the threshold to be 1.3 and not auto creating new manifests for all version less than that. 

### Issues Resolved
#2420 

### DIFF
--- a/src/manifests_workflow/input_manifests.py
+++ b/src/manifests_workflow/input_manifests.py
@@ -107,6 +107,9 @@ class InputManifests(Manifests):
 
             # generate new manifests
             for release_version in sorted(main_versions.keys() - known_versions):
+                legacy_threshold_version = "1.3"
+                if release_version[:3] < legacy_threshold_version:
+                    continue
                 self.write_manifest(release_version, main_versions[release_version])
                 self.add_to_cron(release_version)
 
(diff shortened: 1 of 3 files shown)


peterzhuamazon (changes requested):
Hi @zelinh,

The issue with this solution is that `1.3` is hardcoded here, and we need to remember about it if we move more versions into the legacy folder over time.

Can you use python to scan the legacy-manifests folder and get the highest version threshold instead of hardcode here?

Thanks.

zelinh:
Closing this PR so far and we would come back to this when we have bandwidth.

Issue #2345, opened 2022-07-15
Release tracker for component repos


zelinh:
### Is your feature request related to a problem? Please describe

When we were launching release, it's hard for us to know if each release component has any new fixes, commits or updates from last release. Right now it would require us to look into each repo and check their commit history on certain branch. 

In addition, If they have any new commits after last release on the same branch, we would also need and enforce them to provide release notes and it would be better if they merge it before we finalize the release candidates. 

There are a lot of back and forth efforts if we found one repo has updates but they didn't backport the changes to the specific branch or provide the release notes on time, which may at the worst case, delay the release. 

We would like to have a generalized report to track any changes on each release plugins since last release so it would be easier for us to work with plugin teams for the release.

### Describe the solution you'd like

Acceptance criteria: A generalized report to track any changes on each release plugins since last time release on the same branch. 

### Describe alternatives you've considered

Keep doing things manually.

bbarani:
[Triaged] We will look in to this issue soon

prudhvigodithi:
My thought is this a perquisite https://github.com/opensearch-project/opensearch-plugins/issues/142, as right now the branch for a specific release is not constant, a plugin source branch for a specific release is an important key here.

Example for 2.2 release `2.x` branch should be the source (https://github.com/opensearch-project/OpenSearch/blob/2.x/buildSrc/version.properties) for all plugins, so a script can be fired to check against 2.x branch, any inconsistencies with branching may not give the accurate results in terms of commits, example if a plugin using `main` for 2.x and other's using `2.x` then the possible solution is a matrix to use [plugin, release_branch] and fire the script.

The other factor that needs to be considered is excluding the `version increment` commit as for sure the version increment will be done moving forward after every release.

Thoughts @bbarani @zelinh @peterzhuamazon ?

prudhvigodithi:
@gaiksaya the idea is to fire a job to follow up if there exists a commit after a specific date (this specific date would be an input to the job) from past release to the date the job will be fired (Assuming code freeze date) and if no release notes update the table in accordingly
Example 
running `./run_releasenotes_check.sh manifests/2.0.0/opensearch-dashboards-2.0.0.yml`
```
# OpenSearch Dashboards Commits & Release Notes info
|           Repo           |   Branch   |Commits|Release Notes|
|--------------------------|------------|-------|-------------|
|OpenSearch-Dashboards     |tags/2.0.0  |       |YES          |
[...]

PR #2438, opened 2022-08-08
Release Notes tracker for component repos

prudhvigodithi:
### Description

- Release tracker for component repos, thats gives information about the most recent commitID after a specific date and checks if there are any associated release notes.
Expected inputs:
1) `GIT_LOG_DATE`: To check if commit exists after a specific date
2) `ADD_COMMENT`: true/false
3) `GIT_ISSUE_NUMBER`: release issue to add the generated table as comment.
4) `GITHUB_TOKEN`: token used to add the comment

- This PR adds code that adheres to existing workflows 

- Execution command: `./run_releasenotes_check.sh manifests/2.2.0/opensearch-2.2.0.yml `
- Create a mark-down table as follows, adding the latest commitID after a certain date, and if contains a release notes with YES/NO/NULL

```
|    Repo    |Branch|CommitID|Release Notes|
|------------|------|--------|-------------|
|OpenSearch  |2.x   |        |NO           |
|common-utils|2.x   |        |NO           |
|security    |main  |        |YES          |
```

- Pending: To update README

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2345

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -24,6 +24,10 @@ atomicwrites = "*"
 validators = "*"
 yamlfix = "*"
 yamllint = "*"
+pytablewriter = "*"
+typed-ast = "*"
+zipp = "*"
+importlib-metadata = "*"
 
 [dev-packages]
 
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@
     - [Build Numbers](#build-numbers)
     - [Latest Distribution Url](#latest-distribution-url)
     - [Testing the Distribution](#testing-the-distribution)
+    - [Checking Release Notes](#checking-release-notes)
     - [Signing Artifacts](#signing-artifacts)
       - [PGP](#pgp)
       - [Windows](#windows)
@@ -136,6 +137,16 @@ Tests the OpenSearch distribution, including integration, backwards-compatibilit
 
 See [src/test_workflow](./src/test_workflow) for more information.
 
+#### Checking Release Notes
+
+Workflow to check if the release notes exists or not and shows the latest commit for OpenSearch and Dashboard distributions.
+
+To run:
+```bash
+./release_notes.sh check manifests/2.2.0/opensearch-2.2.0.yml --date 2022-07-26
+```
+
+See [src/release_notes_workflow](./src/release_notes_workflow) for more information.
 #### Signing Artifacts
 
 For all types of signing within OpenSearch project we use `opensearch-signer-client` (in progress of being open-sourced) which is a wrapper around internal signing system and is only available for authenticated users. The input requires a path to the build manifest or directory containing all the artifacts or a single artifact. 
--- a/src/release_notes_workflow/README.md
+++ b/src/release_notes_workflow/README.md
@@ -1,0 +1,37 @@
+#### Components Release Notes Check 
+
+Pulls the latest code to check if the release notes exists and whether new commits have been made based on user passed argument `--date`. Outputs a formated markdown table as follows.
+
+*Usage*
+```
+./release_notes.sh check manifests/3.0.0/opensearch-3.0.0.yml --date 2022-07-26
+```
+
+*Sample Output*
+```
+#  OpenSearch CommitID(after 2022-07-26) & Release Notes info
+|          Repo           |   Branch   |CommitID|Commit Date|Release Notes|
+|-------------------------|------------|--------|-----------|-------------|
+|OpenSearch               |tags/2.2.0  |b1017fa |2022-08-08 |True         |
+|common-utils             |tags/2.2.0.0|7d53102 |2022-08-04 |False        |
+|job-scheduler            |tags/2.2.0.0|a501307 |2022-08-02 |True         |
+|ml-commons               |tags/2.2.0.0|a7d2695 |2022-08-08 |True         |
+|performance-analyzer     |tags/2.2.0.0|3a75d7d |2022-08-08 |True         |
+|security                 |tags/2.2.0.0|8e9e583 |2022-08-08 |True         |
+|geospatial               |tags/2.2.0.0|a71475a |2022-08-04 |True         |
+|k-NN                     |tags/2.2.0.0|53185a0 |2022-08-04 |True         |
+|cross-cluster-replication|tags/2.2.0.0|14d871a |2022-08-05 |False        |
+```
+
+The workflow uses the following arguments:
+* `--date`: To check if commit exists after a specific date (in format yyyy-mm-dd, example 2022-07-26).
+* `--output`: To dump the output into an `.md` file, example `--output table.md`).
+
+
+The following options are available.
+
+| name               | description                                                             |
+|--------------------|-------------------------------------------------------------------------|
+| --date             | Shows commit after a specific date.                                     |
+| --output           | Saves the table output to user specified file.  		       	       |
+| -v, --verbose      | Show more verbose output.                                               |
--- a/src/release_notes_workflow/release_notes.py
+++ b/src/release_notes_workflow/release_notes.py
@@ -1,0 +1,59 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import logging
+import os
+from typing import List
+
+from pytablewriter import MarkdownTableWriter
+
+from git.git_repository import GitRepository
+from manifests.input_manifest import InputComponentFromSource, InputManifest
+from release_notes_workflow.release_notes_component import ReleaseNotesComponents
+from system.temporary_directory import TemporaryDirectory
+
+
+class ReleaseNotes:
+
+    def __init__(self, manifest: InputManifest, date: str) -> None:
+        self.manifest = manifest
+        self.date = date
+
+    def table(self) -> MarkdownTableWriter:
+        table_result = []
+        for component in self.manifest.components.select():
+            if type(component) is InputComponentFromSource:
+                table_result.append(self.check(component))
+        writer = MarkdownTableWriter(
+            table_name=f" {self.manifest.build.name} CommitID(after {self.date}) & Release Notes info",
+            headers=["Repo", "Branch", "CommitID", "Commit Date", "Release Notes"],
+            value_matrix=table_result
+        )
+        return writer
+
+    def check(self, component: InputComponentFromSource) -> List:
+        results = []
+        with TemporaryDirectory(chdir=True) as work_dir:
+            results.append(component.name)
+            results.append(component.ref)
+            with GitRepository(
+                    component.repository,
+                    component.ref,
+                    os.path.join(work_dir.name, component.name),
+                    component.working_directory,
+            ) as repo:
+                logging.debug(f"Checked out {component.name} into {repo.dir}")
+                release_notes = ReleaseNotesComponents.from_component(component, self.manifest.build.version, repo.dir)
+                commits = repo.log(self.date)
+                if len(commits) > 0:
+                    last_commit = commits[-1]
+                    results.append(last_commit.id)
+                    results.append(last_commit.date)
+                else:
+                    results.append(None)
+                    results.append(None)
+                results.append(release_notes.exists())
+        return results
--- a/src/release_notes_workflow/release_notes_check_args.py
+++ b/src/release_notes_workflow/release_notes_check_args.py
@@ -1,0 +1,49 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import argparse
+import datetime
+import logging
+from typing import IO
+
+
+class ReleaseNotesCheckArgs:
+    action: str
+    manifest: IO
+    date: str
+    output: str
+
+    def __init__(self) -> None:
+        parser = argparse.ArgumentParser(description="Checkout an OpenSearch Bundle and check for CommitID and Release Notes")
+        parser.add_argument("action", choices=["check"], help="Operation to perform.")
+        parser.add_argument("manifest", type=argparse.FileType("r"), help="Manifest file.")
+        parser.add_argument(
+            "-v",
+            "--verbose",
+            help="Show more verbose output.",
+            action="store_const",
+            default=logging.INFO,
+            const=logging.DEBUG,
+            dest="logging_level",
+        )
+        parser.add_argument(
+            "--date",
+            type=lambda s: datetime.datetime.strptime(s, "%Y-%m-%d").date(),
+            dest="date",
+            help="Date to retrieve the commit (in format yyyy-mm-dd, example 2022-07-26)."
+        )
+        parser.add_argument(
+            "--output",
+            help="Output file."
+        )
+        args = parser.parse_args()
+        self.logging_level = args.logging_level
+        self.action = args.action
+        self.manifest = args.manifest
+        self.date = args.date
+        self.output = args.output
+        if self.action == "check" and self.date is None:
+            parser.error("check option requires --date argument")
--- a/src/release_notes_workflow/release_notes_component.py
+++ b/src/release_notes_workflow/release_notes_component.py
@@ -1,0 +1,57 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import os
+from abc import abstractmethod
+
+from manifests.input_manifest import InputComponentFromSource
+
+
+class ReleaseNotesComponent:
+
+    def __init__(self, component: InputComponentFromSource, build_version: str, root: str) -> None:
+        self.component = component
+        self.build_version = build_version
+        self.root = root
+
+    @property
+    @abstractmethod
+    def filename(self) -> str:
+        pass
+
+    @property
+    def path(self) -> str:
+        return os.path.join(self.root, "release-notes")
+
+    def path_exists(self) -> bool:
+        return os.path.exists(self.path)
+
+    def exists(self) -> bool:
+        return self.path_exists() and any(fname.endswith(self.filename) for fname in os.listdir(self.path))
+
+
+class ReleaseNotesOpenSearch(ReleaseNotesComponent):
+
+    @property
+    def filename(self) -> str:
+        return f'.release-notes-{self.build_version}.md'
+
+
+class ReleaseNotesOpenSearchPlugin(ReleaseNotesComponent):
+
+    @property
+    def filename(self) -> str:
+        return f'.release-notes-{self.build_version}.0.md'
+
+
+class ReleaseNotesComponents:
+
+    @classmethod
+    def from_component(self, component: InputComponentFromSource, build_version: str, root: str) -> ReleaseNotesComponent:
+        if component.name == 'OpenSearch' or component.name == 'OpenSearch-Dashboards':
+            return ReleaseNotesOpenSearch(component, build_version, root)
+        else:
+            return ReleaseNotesOpenSearchPlugin(component, build_version, root)
--- a/src/run_releasenotes_check.py
+++ b/src/run_releasenotes_check.py
@@ -1,0 +1,27 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+from manifests.input_manifest import InputManifest
+from release_notes_workflow.release_notes import ReleaseNotes
+from release_notes_workflow.release_notes_check_args import ReleaseNotesCheckArgs
+from system import console
+
+
+def main() -> int:
+    args = ReleaseNotesCheckArgs()
+    console.configure(level=args.logging_level)
+    manifest_file = InputManifest.from_file(args.manifest)
+    release_notes = ReleaseNotes(manifest_file, args.date)
+    if args.action == "check":
+        table_output = release_notes.table()
+        table_output.write_table()
+        if args.output is not None:
+            table_output.dump(args.output)
+    return 0
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/test_run_releasenotes_check.py
+++ b/tests/test_run_releasenotes_check.py
@@ -1,0 +1,59 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import datetime
+import os
+import unittest
+from typing import Any
+from unittest.mock import patch
+
+import pytest
+
+from release_notes_workflow.release_notes_check_args import ReleaseNotesCheckArgs
+from run_releasenotes_check import main
+
+
+class TestRunReleaseNotesCheck(unittest.TestCase):
+
+    OPENSEARCH_MANIFEST = os.path.realpath(
+        os.path.join(
+            os.path.dirname(__file__),
+            "..",
+            "manifests",
+            "templates",
+            "opensearch",
+            "2.x",
+            "manifest.yml"
+        )
+    )
+
+    @pytest.fixture(autouse=True)
+    def _capfd(self, capfd: Any) -> None:
+        self.capfd = capfd
+
+    @patch("argparse._sys.argv", ["run_releasenotes_check.py", "--help"])
+    def test_usage(self) -> None:
+        with self.assertRaises(SystemExit):
+            main()
+
+        out, _ = self.capfd.readouterr()
+        self.assertTrue(out.startswith("usage:"))
+
+    @patch("argparse._sys.argv", ["run_releasenotes_check.py", "check", OPENSEARCH_MANIFEST, "--date", "2022-07-26"])
+    @patch('subprocess.check_call')
+    @patch("run_releasenotes_check.ReleaseNotes")
+    @patch("run_releasenotes_check.main", return_value=0)
+    def test_main(self, *mocks: Any) -> None:
+        self.assertEqual(ReleaseNotesCheckArgs().action, 'check')
+        self.assertEqual(ReleaseNotesCheckArgs().date, datetime.date(2022, 7, 26))
+        self.assertTrue(ReleaseNotesCheckArgs().manifest)
+
+    @patch('subprocess.check_call')
+    @patch("argparse._sys.argv", ["run_releasenotes_check.py", "check", OPENSEARCH_MANIFEST, "--date", "2022-07-26", "--output", "test.md"])
+    @patch("run_releasenotes_check.ReleaseNotes")
+    @patch("run_releasenotes_check.main", return_value=0)
+    def test_main_with_save(self, *mocks: Any) -> None:
+        self.assertEqual(ReleaseNotesCheckArgs().output, "test.md")
--- a/tests/tests_git/test_git_repository.py
+++ b/tests/tests_git/test_git_repository.py
@@ -57,6 +57,26 @@ class TestGitRepository(unittest.TestCase):
         self.repo.output("echo hello")
         mock_subprocess.assert_called_with("echo hello", cwd=self.repo.dir, shell=True)
 
+    @patch("subprocess.check_output", return_value='927de30 2022-07-12'.encode())
+    def test_log(self, mock_subprocess: Any) -> None:
+        log = self.repo.log('2021-10-26')
+        last_commit = log[-1]
+        self.assertEqual(last_commit.id, "927de30")
+        self.assertEqual(last_commit.date, "2022-07-12")
+
+    @patch("subprocess.check_output", return_value=''.encode())
+    def test_log_empty(self, mock_subprocess: Any) -> None:
+        log = self.repo.log('2021-10-26')
+        self.assertEqual(len(log), 0)
+
+    @patch("subprocess.check_output", return_value='927de30 2022-07-12\n787d971 2022-04-28\n989143e 22-04-26\n8b376fc 2022-04-26\n023a2ac 2022-04-26'.encode())
+    def test_log_multiple(self, mock_subprocess: Any) -> None:
+        log = self.repo.log('2022-04-25')
+        self.assertEqual(len(log), 5)
+        last_commit = log[-1]
+        self.assertEqual(last_commit.id, "023a2ac")
+        self.assertEqual(last_commit.date, "2022-04-26")
+
 
 class TestGitRepositoryDir(unittest.TestCase):
     @patch('subprocess.check_call', return_value=0)
--- a/tests/tests_release_notes_workflow/test_release_notes.py
+++ b/tests/tests_release_notes_workflow/test_release_notes.py
@@ -1,0 +1,48 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import os
+import unittest
+from typing import Any
+from unittest.mock import patch
+
+from manifests.input_manifest import InputComponentFromSource, InputManifest
+from release_notes_workflow.release_notes import ReleaseNotes
+
+
+class TestReleaseNotes(unittest.TestCase):
+
+    def setUp(self) -> None:
+        MANIFESTS = os.path.join(
+            os.path.dirname(__file__),
+            "..",
+            "..",
+            "manifests",
+        )
+        OPENSEARCH_MANIFEST = os.path.realpath(os.path.join(MANIFESTS, "templates", "opensearch", "default", "manifest.yml"))
+        self.manifest_file = InputManifest.from_file(open(OPENSEARCH_MANIFEST))
+        self.release_notes = ReleaseNotes(self.manifest_file, "2022-07-26")
+        self.component = InputComponentFromSource({"name": "OpenSearch", "repository": "url", "ref": "ref"})
+
+    @patch("subprocess.check_output", return_value=''.encode())
+    @patch("subprocess.check_call")
+    def test_check(self, *mocks: Any) -> None:
+        self.assertEqual(self.release_notes.check(self.component), ['OpenSearch', 'ref', None, None, False])
+
+    @patch("subprocess.check_output", return_value=''.encode())
+    @patch("subprocess.check_call")
+    def test_check_with_manifest(self, *mocks: Any) -> None:
+        component = next(self.manifest_file.components.select())
+        if type(component) is InputComponentFromSource:
+            self.assertIsInstance(component, InputComponentFromSource)
+            self.assertEqual(self.release_notes.check(component), ['OpenSearch', 'main', None, None, False])
+
+    @patch('release_notes_workflow.release_notes.ReleaseNotes.check', return_value=['OpenSearch', 'main', 'ee26e01', '2022-08-18', False])
+    def test_table(self, *mocks: Any) -> None:
+        table_output = self.release_notes.table()
+        self.assertEqual(table_output._table_name.strip(), 'OpenSearch CommitID(after 2022-07-26) & Release Notes info')
+        self.assertEqual(table_output.headers, ['Repo', 'Branch', 'CommitID', 'Commit Date', 'Release Notes'])
+        self.assertEqual(table_output.value_matrix, [['OpenSearch', 'main', 'ee26e01', '2022-08-18', False]])
--- a/tests/tests_release_notes_workflow/test_releasenotes_check_args.py
+++ b/tests/tests_release_notes_workflow/test_releasenotes_check_args.py
@@ -1,0 +1,59 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import datetime
+import logging
+import os
+import unittest
+from typing import Any
+from unittest.mock import patch
+
+import pytest
+
+from release_notes_workflow.release_notes_check_args import ReleaseNotesCheckArgs
+
+
+class TestReleaseNotesCheckArgs(unittest.TestCase):
+
+    RELEASE_NOTES_CHECK_PY = os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "..", "src", "run_releasenotes_check.py"))
+
+    OPENSEARCH_MANIFEST = os.path.realpath(
+        os.path.join(
+            os.path.dirname(__file__),
+            "..",
+            "..",
+            "manifests",
+            "templates",
+            "opensearch",
+            "1.x",
+            "os-template-1.1.0.yml",
+        )
+    )
+
+    @pytest.fixture(autouse=True)
+    def _capfd(self, capfd: Any) -> None:
+        self.capfd = capfd
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "check", OPENSEARCH_MANIFEST, "--date", '2022-07-26'])
+    def test_manifest(self) -> None:
+        self.assertEqual(ReleaseNotesCheckArgs().manifest.name, TestReleaseNotesCheckArgs.OPENSEARCH_MANIFEST)
+        self.assertEqual(ReleaseNotesCheckArgs().date, datetime.date(2022, 7, 26))
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "check", OPENSEARCH_MANIFEST])
+    def test_manifest_withoutdate(self) -> None:
+        with self.assertRaises(SystemExit) as cm:
+            ReleaseNotesCheckArgs()
+        _, err = self.capfd.readouterr()
+        self.assertTrue("error: check option requires --date argument" in err)
+        self.assertEqual(cm.exception.code, 2)
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "check", OPENSEARCH_MANIFEST, "--date", '2022-07-26', "--verbose"])
+    def test_verbose_true(self) -> None:
+        self.assertTrue(ReleaseNotesCheckArgs().logging_level, logging.DEBUG)
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "check", OPENSEARCH_MANIFEST, "--date", '2022-07-26', "--output", "test.md"])
+    def test_output(self) -> None:
+        self.assertEqual(ReleaseNotesCheckArgs().output, "test.md")
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "3da3cb40cf6d9fb381ad4499ae93f01db922e93f1bed2fd233bb880b2ad422cd"
+            "sha256": "5113bcbe0eef2135817e49ecadae8b4336ab2ca1d9877ed3e8781369b197bf45"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -25,11 +25,11 @@
         },
         "attrs": {
             "hashes": [
-                "sha256:2d27e3784d7a565d36ab851fe94887c5eccd6a463168875832a1be79c82828b4",
-                "sha256:626ba8234211db98e869df76230a137c4c40a12d72445c45d5f5b716f076e2fd"
+                "sha256:29adc2665447e5191d0e7c568fde78b21f9672d344281d0c6e1ab085429b22b6",
+                "sha256:86efa402f67bf2df34f51a335487cf46b1ec130d02b8d39fd248abfd30da551c"
             ],
-            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'",
-            "version": "==21.4.0"
+            "markers": "python_version >= '3.5'",
+            "version": "==22.1.0"
         },
         "cerberus": {
             "hashes": [
@@ -53,6 +53,14 @@
             ],
             "markers": "python_full_version >= '3.6.1'",
             "version": "==3.3.1"
+        },
+        "chardet": {
+            "hashes": [
+                "sha256:0368df2bfd78b5fc20572bb4e9bb7fb53e2c094f60ae9993339e8671d0afb8aa",
+                "sha256:d3e64f022d254183001eccc5db4040520c0f23b1a3f33d6413e099eb7f126557"
+            ],
+            "markers": "python_version >= '3.6'",
+            "version": "==5.0.0"
         },
         "charset-normalizer": {
             "hashes": [
@@ -108,6 +116,14 @@
             "index": "pypi",
             "version": "==4.5.4"
         },
+        "dataproperty": {
+            "hashes": [
+                "sha256:73ccf10f8b123968210438a1a1aa859ea6d5a16b4e1f4d307da7a81b838e79fa",
+                "sha256:a8f29175950f4b2c33a387aa3809130d87b9a8d3b92a916c906c49efdb566b32"
+            ],
+            "markers": "python_version >= '3.6'",
+            "version": "==0.55.0"
+        },
         "decorator": {
             "hashes": [
                 "sha256:637996211036b6385ef91435e4fae22989472f9d571faba8927ba8253acbc330",
@@ -118,10 +134,10 @@
         },
         "distlib": {
             "hashes": [
-                "sha256:6564fe0a8f51e734df6333d08b8b94d4ea8ee6b99b5ed50613f731fd4089f34b",
-                "sha256:e4b58818180336dc9c529bfb9a0b58728ffc09ad92027a3f30b7cd91e3458579"
+                "sha256:a7f75737c70be3b25e2bee06288cec4e4c221de18455b2dd037fe2a795cab2fe",
+                "sha256:b710088c59f06338ca514800ad795a132da19fda270e3ce4affc74abf955a26c"
             ],
-            "version": "==0.3.4"
+            "version": "==0.3.5"
         },
         "distro": {
             "hashes": [
@@ -133,11 +149,11 @@
         },
         "filelock": {
             "hashes": [
-                "sha256:37def7b658813cda163b56fc564cdc75e86d338246458c4c28ae84cabefa2404",
-                "sha256:3a0fd85166ad9dbab54c9aec96737b744106dc5f15c0b09a6744a445299fcf04"
+                "sha256:55447caa666f2198c5b6b13a26d2084d26fa5b115c00d065664b2124680c4edc",
+                "sha256:617eb4e5eedc82fc5f47b6d61e4d11cb837c56cb4544e39081099fa17ad109d4"
             ],
             "markers": "python_version >= '3.7'",
-            "version": "==3.7.1"
+            "version": "==3.8.0"
         },
         "flake8": {
             "hashes": [
@@ -149,11 +165,11 @@
         },
         "identify": {
             "hashes": [
-                "sha256:0dca2ea3e4381c435ef9c33ba100a78a9b40c0bab11189c7cf121f75815efeaa",
-                "sha256:3d11b16f3fe19f52039fb7e39c9c884b21cb1b586988114fbe42671f03de3e82"
+                "sha256:25851c8c1370effb22aaa3c987b30449e9ff0cece408f810ae6ce408fdd20893",
+                "sha256:887e7b91a1be152b0d46bbf072130235a8117392b9f1828446079a816a05ef44"
             ],
             "markers": "python_version >= '3.7'",
-            "version": "==2.5.1"
+            "version": "==2.5.3"
         },
         "idna": {
             "hashes": [
@@ -168,7 +184,7 @@
                 "sha256:637245b8bab2b6502fcbc752cc4b7a6f6243bb02b31c5c26156ad103d3d45670",
                 "sha256:7401a975809ea1fdc658c3aa4f78cc2195a0e019c5cbc4c06122884e9ae80c23"
             ],
-            "markers": "python_version < '3.8'",
+            "index": "pypi",
             "version": "==4.12.0"
         },
         "iniconfig": {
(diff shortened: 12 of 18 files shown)


dblock (changes requested):
This code is a good start, but breaks several patterns in the python code in this repo.

1. Arguments are passed as ENV variables, where they should be passed as `--arguments`. Create an arguments class.
2. It's not object oriented, mixes many concerns, making it quite messy and difficult to extend. Follow the workflow pattern we have, start by adding a class that encapsulates the concept of release notes. 

I don't think this needs to be a Jenkins workflow, consider GHA? The commenting on a GitHub issue can be done in 2 lines with an existing GHA.

prudhvigodithi:
> This code is a good start, but breaks several patterns in the python code in this repo.
> 
> 1. Arguments are passed as ENV variables, where they should be passed as `--arguments`. Create an arguments class.
> 2. It's not object oriented, mixes many concerns, making it quite messy and difficult to extend. Follow the workflow pattern we have, start by adding a class that encapsulates the concept of release notes.
> 
> I don't think this needs to be a Jenkins workflow, consider GHA? The commenting on a GitHub issue can be done in 2 lines with an existing GitHub action.

The reason I went with ENV for few values is I have extended the existing checkout workflow (checkoutArgs), so that way I dont have to create a new args for this minor workflow.
Also we need a jenkins job here as the input is the `DATE` (which us unknown to automate as the release dates could change) to get the latest the commitID after the `DATE` and check if release notes exists or not, with jenkins job its easy to run (by a release manager) during the release lifecycle (code freeze time), to check if the commit exists after a date (passed as an input), followed by to check if there exists a release notes.
Thank you

prudhvigodithi:
> 3\. Python has nice OO ways to manipulate git, I googled https://gitpython.readthedocs.io/en/stable/tutorial.html, get rid of all command shell calls to git.

This will fail with our code setup, which I have tried initially as once we `import gitpython` the library is called as `git`, but this will fail as we already have something called [src/git/](https://github.com/opensearch-project/opensearch-build/tree/main/src/git)
with gh cli its easy to hit the GitHub API as well.

dblock:
> Hey @dblock The reason I went with ENV for few values is I have extended the existing checkout workflow (checkoutArgs), so that way I dont have to create a new args for this minor workflow. 

This is not a good reason, you're just acquiring technical debt.

> Also we need a jenkins job here as the input is the `DATE` (which us unknown to automate as the release dates could change) to get the latest the commitID after the `DATE` and check if release notes exists or not, with jenkins job its easy to run (by a release manager) during the release lifecycle (code freeze time), to check if the commit exists after a date (passed as an input), followed by to check if there exists a release notes.

Ok
 
> `./run_releasenotes_check.sh manifests/2.2.0/opensearch-2.2.0.yml ` with one arg (input manifest) and by ENV the GIT_LOG_DATE, the commitID and release notes can be found, finally if we want to add this info to an issue using `GIT_ISSUE_NUMBER` should we solved by the code, which limits the new arguments class. andThank you

I don't understand this. All our scripts use command line arguments. You're creating something that doesn't. There's no good reason for it.

dblock:
> > 3. Python has nice OO ways to manipulate git, I googled https://gitpython.readthedocs.io/en/stable/tutorial.html, get rid of all command shell calls to git.
> 
> This will fail with our code setup, which I have tried initially as once we `import gitpython` the library is called as `git`, but this will fail as we already have something called [src/git/](https://github.com/opensearch-project/opensearch-build/tree/main/src/git) with gh cli its easy to hit the GitHub API as well. @dblock

Use `from gitpython import ... ` and you don't have this problem. And if you still do, fix it by moving the existing `git` folder to a different one. There's about 100 reasons to use a library. For example, you will get a proper error when something doesn't work. It's easier to mock/stub/test. Don't create more technical debt, please.

prudhvigodithi:
Thanks for the suggestions @dblock 
I'm fine with having separate args for this workflow and agree which makes it clean, but I was doubtful if that will be pushed or not and this workflow mostly is not shared and for specific use case hence went with ENV approach. Secondly `from gitpython import ...` this wont work as well, as we have [src/git/](https://github.com/opensearch-project/opensearch-build/tree/main/src/git) which forces to use this local git, I myself against creating technical debt as well :) 
I dont think so with GitPython the library out there support API calls to GITHUB (create/comment on an issue etc), its good for normal git cli operations, gh cli is more easy and better with less code here, I will explore more on this and update in my next commit.
Thank you 
@bbarani @peterzhuamazon

src/run_releasenotes_check.py
```diff
@@ -0,0 +1,81 @@
...
+        table_name=f"{manifest.build.name} CommitID(after {args.gitlogdate}) & Release Notes info",
+        headers=["Repo", "Branch", "CommitID", "Release Notes"],
+        value_matrix=value_matrix
+    )
+    with TemporaryDirectory(chdir=True) as work_dir:
+        logging.info(f"Checking out into {work_dir.name}")
+        for component in manifest.components.select():
+            logging.info(f"Checking out {component.name}")
+            repo_list = []  # type: List[Any]
+            if type(component) is InputComponentFromSource:
```
dblock, line 30: Add a method `Component#checkout(path)` or at least a `InputComponent#checkout(path)` that returns an instance of `GitRepository`. Depending on the component type it should return OpenSearchGitRepository, OpenSearchPluginGitRepository, OpenSearchDashboardsGitRepository, etc. Each of these should have a `release_notes` property that is implemented differently depending on all this if/then/else logic you have below.  Each one of these things needs tests.

dblock (commented):
There's quite a bit more work to do to make this well organized. 

You have a list of components (already implemented). Each is checked out into a folder (needs to be implemented). Each may have release notes (to be implemented). Don't put all this unrelated logic into a script. If all these are properties of classes then you'll end up with a very simple runner and an easy to test and extend implementation. Give it a try.

src/releasenotes_check_workflow/releasenotes_check_args.py
```diff
@@ -0,0 +1,68 @@
...
+            default=logging.INFO,
+            const=logging.DEBUG,
+            dest="logging_level",
+        )
+        parser.add_argument(
+            "--gitlogdate",
+            type=str,
+            required=True,
+            dest="gitlogdate",
+            help="Date to retrive the commit (in format yyyy-mm-dd, example 2022-07-26)."
```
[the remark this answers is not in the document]
prudhvigodithi: Added dB.

prudhvigodithi:
Hey @dblock I have modified the code to make its separated with classes, methods and using separate args.
Thank you

src/releasenotes_check_workflow/releasenotes_check_args.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+        args = parser.parse_args()
+        self.logging_level = args.logging_level
+        self.manifest = args.manifest
+        self.addcomment = args.addcomment
+        self.gitlogdate = args.gitlogdate
+        self.gitissuenumber = args.gitissuenumber
+        self.gittoken = args.gittoken
+        print(self.addcomment, self.gitissuenumber, self.gittoken)
+        if self.addcomment is True:
```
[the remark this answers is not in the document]
prudhvigodithi: Sure I can explore this as well.

src/releasenotes_check_workflow/releasenotes_check_args.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+        parser.add_argument(
+            "--gitlogdate",
+            type=str,
+            required=True,
+            dest="gitlogdate",
+            help="Date to retrive the commit (in format yyyy-mm-dd, example 2022-07-26)."
+        )
+        parser.add_argument(
+            "--addcomment",
```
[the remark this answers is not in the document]
prudhvigodithi: The workflow outputs the markdown table, it's not always true that we need to add comment to the GH issue, we can even view the table as output on the console (local or jenkins), once `--addcomment` is true, then it will proceed to add a comment. Did I answer your query?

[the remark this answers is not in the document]
dblock: I think I explained myself poorly. 

I think that the feature of adding a comment to a GitHub issue does not belong in python code. It can be done in a 1-liner in the Jenkins job: pipe the output from the python code to a file, then call `gh` with the right arguments.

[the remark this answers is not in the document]
prudhvigodithi: The idea here to make this also work without jenkins as well dB, a user running on local can pass `--addcomment` and followed by the supported git args of choice which should be able to add comment to the desired repo and issue.

run_releasenotes_check.sh
```diff
@@ -0,0 +1,12 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+set -e
+
+DIR="$(dirname "$0")"
+"$DIR/run.sh" "$DIR/src/run_releasenotes_check.py" $@
```
dblock, line 12: I would remove this script, too many scripts at the root. Maybe just call `run.sh` from the Jenkins job?

Otherwise I think it could be called `release_notes.sh`, and have a "check" command line function, so `release_notes.sh check manifests/manifest.yml`. This way it becomes extensible for future commands related to release notes (I'm thinking `gather` and `generate` as future improvements). WDYT?

src/releasenotes_check_workflow/README.md
```diff
@@ -0,0 +1,33 @@
+#### Components Release Notes Check
```
dblock, line 1: Let's rename the folder to `release_notes_workflow`

src/releasenotes_check_workflow/releasenotes_check.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import os
+
+
+class ReleaseNotesCheck:
```
dblock, line 10: While you've moved the check code into a class, you're not taking advantage of OOO. The implementation still has `if component == OpenSearch`, etc.

Refactor:

1. Create a base class called `ReleaseNotes` with an instance method `check`. The initializer takes a component.
2. Add a property called `ReleaseNotes#filename` that defaults to `.release-notes-{buildversion}.md`.
3. Add a method `ReleaseNotes#exists` that checks whether the file exists, don't rely on an exception.
4. Subclass `ReleaseNotes` into `ReleaseNotesOpenSearch` and `ReleaseNotesOpenSearchDashboards` that overrde `filename` to return a different name for those two.

dblock (commented):
See below for next steps.

src/releasenotes_check_workflow/releasenotes_check_args.py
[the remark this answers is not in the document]
dblock: Why? What human is going to run this? I can copy paste the output in an issue if I need to, this is trying to do too much.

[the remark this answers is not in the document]
prudhvigodithi: All our workflows has the ability to use with automation plus execute on local, with this even without jenkins we should be able to add comment to the desired issue (Like run as Job, manual GH workflow trigger, there are many ways we can use this). Hence I would recommend we have this feature `--addcomment`, to either comment on the GH issue or just output to the local.

README.md
```diff
@@ -123,6 +124,16 @@ Tests the OpenSearch distribution, including integration, backwards-compatibilit
 
 See [src/test_workflow](./src/test_workflow) for more information.
 
+#### Components Release Notes Check 
+
+Worflow to check if the release notes exists for the OpenSearch distribution. Uses the argument `gitlogdate` to parse the CommitID after the date and if any commit exists, checks for the release notes.
```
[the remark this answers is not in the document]
prudhvigodithi: `Typo, workflow` thanks will fix that, more detail about the table I have added here dB.
See [src/releasenotes_check_workflow](./src/releasenotes_check_workflow) for more information.

src/releasenotes_check_workflow/releasenotes_check_args.py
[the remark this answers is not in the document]
dblock: I have several objections to this: the biggest one being that the user is asked to supply a GitHub token, aka different from what they are already authenticated as. 

There's a lot of things going on in this PR. How about removing this functionality from this PR for now, and then doing whatever the right thing is on top of it?

peterzhuamazon:
Hi,

This is just a list of improvements for later:
1. Link to actual release notes if exists, easy to track.
2. Adding the ability to generate final release notes so we dont need to use the legacy script anymore.
3. Ability to edit the original comment instead of creating a new comment.

Thanks.

prudhvigodithi:
> Hi,
> 
> This is just a list of improvements for later:
> 
> 1. Link to actual release notes if exists, easy to track.
> 2. Adding the ability to generate final release notes so we dont need to use the legacy script anymore.
> 3. Ability to edit the original comment instead of creating a new comment.
> 
> Thanks.

Hey @peterzhuamazon 
yes 1. is possible, 2. we have to extent this release_notes_workflow to generate final release notes , 3. using gh cli to update, there is an open issue https://github.com/cli/cli/issues/3613 (Should research on some other solution).

dblock:
> Hey @peterzhuamazon yes 1. is possible, 2. we have to extent this release_notes_workflow to generate final release notes , 3. using gh cli to update a comment, there is an open issue [cli/cli#3613](https://github.com/cli/cli/issues/3613) (Should research on some other solution).

Part (3) is probably indicative of us trying to over-engineer something. There's already a tool for adding or editing a comment on GitHub called `gh`. Unix has support for pipes where output of one tool can be piped into another tool. Or a file can be created that becomes input to another tool. Let's ask ourselves what business value we are adding by doing the additional work of taking input parameters such as GitHub token, issue URL, and a flag to add a comment, then shelling out to a tool that does it? Seems to me that just calling an existing tool is a lot simpler.

prudhvigodithi:
> Part (3) is probably indicative of us trying to over-engineer something. There's already a tool for adding or editing a comment on GitHub called `gh`. Unix has support for pipes where output of one tool can be piped into another tool. Or a file can be created that becomes input to another tool. Let's ask ourselves what business value we are adding by doing the additional work of taking input parameters such as GitHub token, issue URL, and a flag to add a comment, then shelling out to a tool that does it? Seems to me that just calling an existing tool is a lot simpler.

Hey @dblock with `gh` tool we cant edit a comment keeping this aside, for creation of comment we need a right github repo URL, issue number and token to authenticate the repo, Initially I have passed via environmental values but inclining with our workflows I have created a new args class `ReleaseNotesCheckArgs` and passed as arguments, these are the inputs expected by gh cli, so what approach you suggest here instead of passing  GitHub token and issue URL? Coming to `--addcomment` :) 
its an extra option provided, use `--addcomment` to proceed with GH issue commenting, so that it supports to execute the workflow with/without  adding comment to github, so you say this extras option is not worth?

src/releasenotes_check_workflow/releasenotes_check_args.py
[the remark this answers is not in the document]
prudhvigodithi: Hey @dblock Github api (using `gh` cli) uses token.
Not multiple things dB, workflow functionality is straight forward, create an info markdown table if  `--addcomment` (true) comment the same markdown table to the gh issue.

dblock:
> > Part (3) is probably indicative of us trying to over-engineer something. There's already a tool for adding or editing a comment on GitHub called `gh`. Unix has support for pipes where output of one tool can be piped into another tool. Or a file can be created that becomes input to another tool. Let's ask ourselves what business value we are adding by doing the additional work of taking input parameters such as GitHub token, issue URL, and a flag to add a comment, then shelling out to a tool that does it? Seems to me that just calling an existing tool is a lot simpler.
> 
> Hey @dblock, for creation of comment we need a right github repo URL, issue number and token to authenticate the repo, Initially I have passed via environmental values but inclining with our workflows I have created a new args class `ReleaseNotesCheckArgs` and passed as arguments, these are the inputs expected by gh cli, so what approach you suggest here instead of passing GitHub token and issue URL? Coming to `--addcomment` :) its an extra option provided, use `--addcomment` to proceed with GH issue commenting, so that it supports to execute the workflow with/without adding comment to github, so you say this extras option is not worth?

I propose not to do this in this tool, not to add `--addcomment` or any related parameters, and use `gh` directly in the Jenkins workflow.

src/release_notes_workflow/releasenotes_check.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import os
+
+
+class ReleaseNotesCheck:
```
dblock, line 10: This still needs refactoring, https://github.com/opensearch-project/opensearch-build/pull/2438#discussion_r941468303

src/release_notes_workflow/releasenotes_gitlog_getcommit.py
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import subprocess
+
+
+class GitLogGetCommit:
```
dblock, line 10: Both methods belong in https://github.com/opensearch-project/opensearch-build/blob/main/src/git/git_repository.py

src/release_notes_workflow/releasenotes_gitlog_getcommit.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @classmethod
+    def gitlog_getcommitID(self, gitrepo: str, date: str) -> str:
```
dblock, line 13: This returns a set of commits after a date, how is it a "commit ID"?

src/release_notes_workflow/releasenotes_gitlog_getcommit.py
```diff
@@ -0,0 +1,22 @@
...
+        gitLogCmd = f'git log --date=local --after={date} --pretty=format:"%h"'
+        gitLogCommitID = subprocess.check_output(gitLogCmd.split(), cwd=gitrepo).decode()
+        return gitLogCommitID
+
+    @classmethod
+    def gitlog_getdate(self, gitrepo: str, date: str) -> str:
```
dblock, line 19: This method is identical to the above minus the format. Implement a method called `commits` that returns a collection of instances of a new class called `GitCommit`, each having properties like `id`. Note that if you use a git library, that's likely already implemented for you. 

When you want the date of the last commit or the id the code should look like so:

```
last_commit = next(repo.commits(after))
# you have last_commit.date and lastcommit.id
```

tests/tests_release_notes_workflow/test_releasenotes_check_functions.py
```diff
@@ -0,0 +1,29 @@
...
+from release_notes_workflow.releasenotes_check import ReleaseNotesCheck
+from release_notes_workflow.releasenotes_gitlog_getcommit import GitLogGetCommit
+
+
+class TestReleaseNotesCheck(unittest.TestCase):
+
+    def test_ReleaseNotesCheck(self) -> None:
+        assert ReleaseNotesCheck.check("OpenSearch", "2.2.2", "/tmp/") == "NULL"
+
+    def test_gitlog_getcommitID(self) -> None:
```
dblock, line 20: A test like this actually checks out code and won't work offline. Needs mocks.

prudhvigodithi: I have updated the test run against a real git repo db.

src/release_notes_workflow/releasenotes_gitlog_getcommit.py
prudhvigodithi: The code should now support  `.commitID` `.commitDate`
https://github.com/opensearch-project/opensearch-build/blob/8e23f2b7da606024053504c1ace23c53b488fbed/src/release_notes_workflow/releasenotes_gitlog_getcommit.py

src/release_notes_workflow/releasenotes_check.py
prudhvigodithi: Just did dB, the logic is still same, we need to segregate OS OSD and rest of all other plugins, as for the plugins release notes added as with extra `.0` please add your thoughts further @dblock.
Thank you

src/release_notes_workflow/releasenotes_gitlog_getcommit.py
prudhvigodithi: The `gitlog` is only required for specific purpose, here to fetch commitID and date from repo log, but the [git_repository.py](https://github.com/opensearch-project/opensearch-build/blob/main/src/git/git_repository.py), is for generic purposes used across other build code. Hence created as a separate methods inside release notes workflow.

src/release_notes_workflow/releasenotes_gitlog_getcommit.py
prudhvigodithi: It will only return one commitID dB, the checkout from `from git.git_repository import GitRepository` only pulls with `--depth 1`

run_releasenotes_check.sh
prudhvigodithi: Added `check` @dblock , can be called now with `./release_notes.sh check manifests/2.2.0/opensearch-2.2.0.yml --date 2022-07-26`.

src/releasenotes_check_workflow/releasenotes_check_args.py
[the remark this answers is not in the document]
prudhvigodithi: Removing the `--addcomment` args from this workflow :)

dblock (changes requested):
It's a bit closer. 

I didn't look at the test code at all.

Pipfile
```diff
@@ -24,7 +24,9 @@ atomicwrites = "*"
 validators = "*"
 yamlfix = "*"
 yamllint = "*"
-
+pytablewriter = "*"
+importlib-metadata = "*"
```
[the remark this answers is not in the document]
prudhvigodithi: pytablewriter is the module used to generate the markdown table.

Pipfile
```diff
@@ -24,7 +24,9 @@ atomicwrites = "*"
...
+typed-ast = "*"
```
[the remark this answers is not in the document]
prudhvigodithi: `importlib-metadata` and `typed-ast` are already existing, but with new change when added `pytablewriter`, they are no longer by default added, hence explicitly added to the Pipfile.

README.md
```diff
@@ -123,6 +124,16 @@ Tests the OpenSearch distribution, including integration, backwards-compatibilit
 
 See [src/test_workflow](./src/test_workflow) for more information.
 
+#### Checking Release Notes
+
+Workflow to check if the release notes exists for OpenSearch and Dashboard distribution. Uses the argument `--date` to parse the CommitID and after the given `--date` if any commit exists checks for the release notes.
```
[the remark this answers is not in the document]
prudhvigodithi: It will only most recent commit @dblock using the existing checkout workflow, which uses `--depth 1`.

src/release_notes_workflow/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+The workflow uses the following arguments:
+* `--date`: To check if commit exists after a specific date (in format yyyy-mm-dd, example 2022-07-26).
+
+
+The following options are available.
+
+| name               | description                                                             |
+|--------------------|-------------------------------------------------------------------------|
+| --date             | Do not delete the temporary working directory on both success or error. |
```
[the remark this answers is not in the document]
prudhvigodithi: Will Update the description, thanks.

src/release_notes_workflow/releasenotes_check_args.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import IO
+
+
+class ReleaseNotesCheckArgs:
+    action: str
+    manifest: IO
+    date: str
+
+    def __init__(self) -> None:
+        parser = argparse.ArgumentParser(description="Checkout an OpenSearch Bundle and check for CommitID and Release Notes")
```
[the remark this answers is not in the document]
prudhvigodithi: Same as above using existing checkout workflow, which uses --depth 1, hence will only show one a single commit and only if that commit is after `--date` args, single its a single commit `CommitID` is more accurate.

src/release_notes_workflow/releasenotes_gitlog.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class GitLog:
+
+    def __init__(self, gitrepo: str, date: str) -> None:
+        self.gitrepo = gitrepo
+        self.date = date
+
+    @property
+    def commitID(self) -> str:
```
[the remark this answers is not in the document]
prudhvigodithi: The code should work dB, if there are more that one commit as well, but using existing checkout workflow, which uses --depth 1 just shows one commit, but the method `commitID`, should return if there are more than one commit.

Pipfile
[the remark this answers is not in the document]
dblock: Why are `importlib-metadata` and `typed-ast` needed? What code uses that?

Pipfile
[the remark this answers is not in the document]
prudhvigodithi: Flake8 requires `importlib-metadata`.

Pipfile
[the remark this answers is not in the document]
prudhvigodithi: Flake8 uses `importlib-metadata` and types `typed-ast` was already part of the `Pipfile.lock`, but with new update this was removed, so added back.
Thank you

src/run_releasenotes_check.py
```diff
@@ -0,0 +1,61 @@
...
+                    gitHistory = GitLog(repo.dir, args.date)
+                    repo_list.append(gitHistory.commitID)
+                    repo_list.append(gitHistory.commitDate)
+                    if args.action == "check":
+                        release_notes = ReleaseNotesCheck(component.name, manifest.build.version, repo.dir)
+                        repo_list.append(release_notes.check())
+                    value_matrix.append(repo_list)
+
+    writer.write_table()
+    writer.dump("table.md")
```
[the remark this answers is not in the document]
prudhvigodithi: Hey @dblock we need this output `table.md`, to make later jenkins uses this --body-file `table.md` to post to GH API. I have added a new arg parameter as `--save`, so a normal run wont generate generate the file, only when ran as example `./release_notes.sh check  manifests/2.2.0/opensearch-2.2.0.yml  --date 2022-07-26  --save`, will output the file.

prudhvigodithi:
Hey @dblock, thanks for the review, I have fixed most of the points addressed. Please review.

Somehow, of a sudden these python tests fails with `Run pipenv run flake8 .` (passes on my local though)
Error
```
    import importlib_metadata
ModuleNotFoundError: No module named 'importlib_metadata'
```
I can see this failed on other [open PR's in build repo](https://github.com/opensearch-project/opensearch-build/pull/2448) as well. Will look into it.

Thank you

prudhvigodithi:
> Somehow, of a sudden these python tests fails with `Run pipenv run flake8 .` (passes on my local though)

Strange, some inconsistencies with the runner tests pass today. However failed previously :) 
https://github.com/opensearch-project/opensearch-build/actions/runs/2853919303/attempts/2
https://github.com/opensearch-project/opensearch-build/actions/runs/2853919303/attempts/1

src/release_notes_workflow/README.md
```diff
@@ -0,0 +1,29 @@
+#### Components Release Notes Check 
+
+Pulls the latest CommitID based on user passed argument `--date` and checks if the release notes exists or not. Outputs a formated markdown table as follows.
```
dblock, line 3: This text is incorrect. It pulls the latest code, then checks whether release notes have been written and whether new commits have been made since date.

src/release_notes_workflow/README.md
```diff
@@ -0,0 +1,29 @@
...
+The workflow uses the following arguments:
+* `--date`: To check if commit exists after a specific date (in format yyyy-mm-dd, example 2022-07-26).
+
+
+The following options are available.
+
+| name               | description                                                             |
+|--------------------|-------------------------------------------------------------------------|
+| --date             | Shows commit after a specific date.                                     |
+| --save             | Saves the table output to table.md.                                     |
```
dblock, line 28: Don’t hard code table.md, add an option `—output table.md`, and fail if that file already exists.

src/release_notes_workflow/component_release_notes.py
```diff
@@ -0,0 +1,72 @@
...
+from manifests.input_manifest import InputComponentFromSource
+
+
+class ComponentReleaseNotes:
+
+    def __init__(self, component: InputComponentFromSource, build_version: str, root: str) -> None:
+        self.component = component
+        self.build_version = build_version
+        self.root = root
+        self.filename
```
dblock, line 18: What’s this? did you mean to assign _filename?

src/release_notes_workflow/component_release_notes.py
```diff
@@ -0,0 +1,72 @@
...
+    def path_exists(self) -> bool:
+        return os.path.exists(self.path)
+
+    def exists(self) -> bool:
+        print(self.filename)
+        return self.path_exists() and any(fname.endswith(self.filename) for fname in os.listdir(self.path))
+
+
+    def check(self) -> str:
+        if self.exists():
```
dblock, line 41: Were you trying to return a third value when path does not exist? Otherwise you don’t need any of this code, just call `exists()` and include true/false in the table.

src/release_notes_workflow/component_release_notes.py
```diff
@@ -0,0 +1,72 @@
...
+        return f'.release-notes-{self.build_version}.0.md'
+
+
+class ComponentsReleaseNotes(ComponentReleaseNotes): 
+
+    @classmethod
+    def from_component(self, component: InputComponentFromSource, build_version: str, root: str) -> ComponentReleaseNotes:
+        if component.name == 'OpenSearch' or component.name == 'OpenSearch-Dashboards':
+            release_notes_os =  ComponentReleaseNotesOpenSearch(component, build_version, root)
+            ComponentReleaseNotes.filename = release_notes_os.filename
```
dblock, line 67: This is wrong. If you need to “calculate” the filename differently in the different component classes create a constructor in each child class and do it in there.

src/release_notes_workflow/release_notes.py
```diff
@@ -0,0 +1,54 @@
...
+                    os.path.join(work_dir.name, component.name),
+                    component.working_directory,
+            ) as repo:
+                logging.debug(f"Checked out {component.name} into {repo.dir}")
+                release_notes = ComponentsReleaseNotes.from_component(component, self.manifest.build.version, repo.dir)
+                commits = repo.log(self.date)
+                if len(commits) > 0:
+                    last_commit = commits[-1]
+                    results.append(last_commit.id)
+                    results.append(last_commit.date)
```
dblock, line 52: Do you need to apprend empty values if len is 0? Otherwise I think the table is broken.

src/release_notes_workflow/release_notes_check_args.py
```diff
@@ -0,0 +1,51 @@
...
+            const=logging.DEBUG,
+            dest="logging_level",
+        )
+        parser.add_argument(
+            "--date",
+            type=lambda s: datetime.datetime.strptime(s, "%Y-%m-%d").date(),
+            dest="date",
+            help="Date to retrive the commit (in format yyyy-mm-dd, example 2022-07-26)."
+        )
+        parser.add_argument(
```
dblock, line 38: Make this —output and take a filename is an argument.

dblock (commented):
See below.

src/release_notes_workflow/component_release_notes.py
prudhvigodithi: I have modified the logic to use property overloading.

src/release_notes_workflow/component_release_notes.py
prudhvigodithi: Modified to just run `exists()`, output is expected to be `True/False`.

src/release_notes_workflow/component_release_notes.py
prudhvigodithi: I have modified the logic to use property overloading.

src/release_notes_workflow/release_notes.py
prudhvigodithi: I have handled to append with `None`. May be if any edge cases, should be handled in future.

src/release_notes_workflow/release_notes_check_args.py
prudhvigodithi: Got it, changed to ` —output` <user passed file>

prudhvigodithi:
Hey @dblock I have addressed the latest comments and added some tests to the new improved code.
Thank you

src/release_notes_workflow/README.md
prudhvigodithi: Got it, changed to ` —output` `user passed file`

dblock (changes requested):
Good. 

The major pieces of feedback is in tests that need to stub execution of git commands instead of doing actual git commands.

tests/test_run_releasenotes_check.py
```diff
@@ -0,0 +1,51 @@
...
+import os
+import unittest
+from typing import Any
+from unittest.mock import patch
+
+import pytest
+
+from run_releasenotes_check import main
+
+gitLogDate = '2022-07-26'
```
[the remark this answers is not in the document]
prudhvigodithi: I will hardcode the date dB.

src/release_notes_workflow/component_release_notes.py
```diff
@@ -0,0 +1,58 @@
...
+
+class ComponentReleaseNotes:
+
+    def __init__(self, component: InputComponentFromSource, build_version: str, root: str) -> None:
+        self.component = component
+        self.build_version = build_version
+        self.root = root
+
+    @property
+    def filename(self) -> str:
```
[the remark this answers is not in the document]
prudhvigodithi: Sure I have added as 
```
    @property
    @abstractmethod
    def filename(self) -> str:
        pass
 ```
 
 So that the function `def exists(self)` can still infer as property.
 @dblock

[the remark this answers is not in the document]
Comment (dblock) [reply src/release_notes_workflow/component_release_notes.py:22]: I am not sure what you mean by `So that the function def exists(self) can still infer as property.`. The implementation of `exists` in the base class calls the overridden versions of `filename` which are implemented in child classes.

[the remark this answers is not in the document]
Comment (prudhvigodithi) [reply src/release_notes_workflow/component_release_notes.py:22]: you get this typical error with mypy as follows:

``` error: Argument 1 to "endswith" of "str" has incompatible type "Callable[[], str]"; expected "Union[str, Tuple[str, ...]]"```

what I'm trying to say, i'm making `filename` in ComponentReleaseNotes as abstract property, which is a read-only property.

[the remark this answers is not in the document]
Comment (dblock) [reply src/release_notes_workflow/component_release_notes.py:22]: Interesting. I think that combination of `@property` and `@abstractmethod` doesn't work. It looks like a bug: https://stackoverflow.com/questions/68763357/strange-behaviour-when-mixing-abstractmethod-classmethod-and-property-decorator, so just leave it out.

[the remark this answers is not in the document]
Comment (prudhvigodithi) [reply src/release_notes_workflow/component_release_notes.py:22]: I just saw this in the doc @dblock https://docs.python.org/3/library/abc.html#abc.abstractproperty adding `@property` with `@abstractmethod` should give a read-only property.

File: tests/tests_git/test_git_repository.py [hunk 30]
```diff
@@ -57,6 +57,16 @@ def test_output(self, mock_subprocess: Any) -> None:
         self.repo.output("echo hello")
         mock_subprocess.assert_called_with("echo hello", cwd=self.repo.dir, shell=True)
 
+    def test_gitlog_getcommitID(self) -> None:
```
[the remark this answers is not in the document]
prudhvigodithi: Thanks @dblock I have done this way mocking.

src/release_notes_workflow/README.md
```diff
@@ -0,0 +1,37 @@
...
+* `--date`: To check if commit exists after a specific date (in format yyyy-mm-dd, example 2022-07-26).
+* `--output`: To dump the output into an `.md` file, example `--output table.md`).
+
+
+The following options are available.
+
+| name               | description                                                             |
+|--------------------|-------------------------------------------------------------------------|
+| --date             | Shows commit after a specific date.                                     |
+| --output             | Saves the table output to user specified file.  |
```
dblock, line 36: Nit: align `|`

tests/test_run_releasenotes_check.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    @patch("argparse._sys.argv", ["run_releasenotes_check.py", "check", OPENSEARCH_MANIFEST, "--date", "2022-07-26"])
+    def test_main(self) -> None:
+        assert main() == 0
+
+    @patch("argparse._sys.argv", ["run_releasenotes_check.py", "check", OPENSEARCH_MANIFEST, "--date", "2022-07-26", "--output", "test.md"])
+    def test_main_with_save(self) -> None:
+        assert main() == 0
+        assert os.path.exists("test.md") is True
```
dblock: This test leaves a file called `test.md` behind, so the second time it runs it may succeed while the actual code may no longer be creating a `test.md`. For this test mock `table_output` and ensure that `table_output.dump("test.md")` is called.

tests/tests_git/test_git_repository.py
```diff
@@ -57,6 +57,13 @@ def test_output(self, mock_subprocess: Any) -> None:
         self.repo.output("echo hello")
         mock_subprocess.assert_called_with("echo hello", cwd=self.repo.dir, shell=True)
 
+    @patch("subprocess.check_output", return_value='927de30 2022-07-12'.encode())
```
dblock, line 60: Add a test for 1) empty output from `check_output`, and 2) multiple lines. In the first case you should get back and empty collection of logs, and in the second multiple items.

tests/tests_release_notes_workflow/test_releasenotes_check_args.py
```diff
@@ -0,0 +1,48 @@
...
+            "..",
+            "manifests",
+            "templates",
+            "opensearch",
+            "1.x",
+            "os-template-1.1.0.yml",
+        )
+    )
+
+    gitLogDate = '2022-07-26'
```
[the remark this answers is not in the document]
dblock: I see you resolved this, but 1) it makes tests hard to read 2) it's still camelCase which is not Python. Change `gitLogDate` below to the actual date "2022-07-26", there's no need for this variable.

tests/tests_release_notes_workflow/test_releasenotes_check_args.py
```diff
@@ -0,0 +1,48 @@
...
+        self.assertEqual(ReleaseNotesCheckArgs().manifest.name, TestReleaseNotesCheckArgs.OPENSEARCH_MANIFEST)
+        self.assertEqual(ReleaseNotesCheckArgs().date, TestReleaseNotesCheckArgs.gitLogDateAssert)
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "check", OPENSEARCH_MANIFEST, "--date", gitLogDate, "--verbose"])
+    def test_verbose_true(self) -> None:
+        self.assertTrue(ReleaseNotesCheckArgs().logging_level, logging.DEBUG)
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "check", OPENSEARCH_MANIFEST, "--date", gitLogDate, "--verbose", "--output", "test.md"])
+    def test_verbose_true_withsave(self) -> None:
+        self.assertTrue(ReleaseNotesCheckArgs().logging_level, logging.DEBUG)
```
dblock, line 48: Add a test where where `check" without a `--date` raises an error.

tests/tests_git/test_git_repository.py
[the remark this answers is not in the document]
prudhvigodithi: Yes I have turned off the wifi and executed the tests, the release notes workflow tests pass without any error.
But I have noticed `TestRunManifests.test_main_update ` failed which us part of manifest workflow ;) (Should fix that in future).
@dblock

tests/tests_release_notes_workflow/test_releasenotes_check_functions.py
```diff
@@ -0,0 +1,65 @@
...
+from release_notes_workflow.release_notes import ReleaseNotes
+
+
+class TestReleaseNotesCheckFunctions(unittest.TestCase):
+
+    @patch('subprocess.check_output', return_value='c3408f34961eb7e8ebc0510306f62238a457dbbc'.encode())
+    @patch("subprocess.check_output", return_value='.release-notes-2.2.0.0.md'.encode())
+    def test_plugin_check(self, *mocks: Any) -> None:
+        component = InputComponentFromSource({"name": "common-utils", "repository": "https://github.com/opensearch-project/common-utils", "ref": "c3408f34961eb7e8ebc0510306f62238a457dbbc"})
+        repo = GitRepository(
```
dblock: I am pretty sure that this calls `__checkout__` [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/git/git_repository.py#L44) and actually checks out code. Put dummy values in `url` and `ref` to ensure that this is correctly mocked.

I think you're missing another `@patch('subprocess.check_call', return_value=0)`.

tests/tests_release_notes_workflow/test_releasenotes_check_functions.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import Any
+from unittest.mock import patch
+
+from git.git_repository import GitRepository
+from manifests.input_manifest import InputComponentFromSource, InputManifest
+from release_notes_workflow.component_release_notes import ComponentsReleaseNotes
+from release_notes_workflow.release_notes import ReleaseNotes
+
+
+class TestReleaseNotesCheckFunctions(unittest.TestCase):
```
dblock, line 18: Create a separate set of tests that map to the file and class structure in `src`.

1. `src/release_notes_workflow/component_release_notes.py` -> `test_component_release_notes.py` with tests for `path`, `exists`, `path_exists`, for each of `ComponentReleaseNotesOpenSearch`, `ComponentReleaseNotesOpenSearchPlugin` and ` ComponentsReleaseNotes#from_component`. Use mocks to avoid all kinds of actual file operations.
2. `src/release_notes_workflow/release_notes.py` -> `test_release_notes.py` with tests for `table` and `check`.

dblock (changes requested):
Getting closer. The big missing piece is that tests lump multiple parts under test in basically two tests in `test_releasenotes_check_function`. That needs to be broken up and unit tests written for each separate class that was added or changed.

tests/tests_release_notes_workflow/test_releasenotes_check_functions.py
dblock: Looks like you fixed that in a subsequent commit.

prudhvigodithi:
Hey @dblock I have addressed the suggested changes, can you review back again?
Thank you

tests/test_run_releasenotes_check.py
```diff
@@ -0,0 +1,52 @@
...
+            "2.x",
+            "manifest.yml"
+        )
+    )
+
+    @patch('subprocess.check_call')
+    @patch("argparse._sys.argv", ["run_releasenotes_check.py", "check", OPENSEARCH_MANIFEST, "--date", "2022-07-26"])
+    @patch("run_releasenotes_check.ReleaseNotes")
+    def test_main(self, *mocks: Any) -> None:
+        main()
```
dblock, line 46: This only tests that `main()` doesn't throw any errors. You need to assert that it loaded the manifest, and checked its release notes.

tests/test_run_releasenotes_check.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @patch('subprocess.check_call')
+    @patch("argparse._sys.argv", ["run_releasenotes_check.py", "check", OPENSEARCH_MANIFEST, "--date", "2022-07-26", "--output", "test.md"])
+    @patch("run_releasenotes_check.ReleaseNotes")
+    def test_main_with_save(self, *mocks: Any) -> None:
+        main()
```
dblock, line 52: This only tests that `main()` doesn't throw any errors. You need to assert that it tried to write the output file, aka that `writer.dump` was called.

tests/tests_git/test_git_repository.py
```diff
@@ -57,6 +57,27 @@ def test_output(self, mock_subprocess: Any) -> None:
...
+    def test_log(self, mock_subprocess: Any) -> None:
+        log = self.repo.log('2021-10-26')
+        last_commit = log[-1]
+        self.assertEqual(last_commit.id, "927de30")
+        self.assertEqual(last_commit.date, "2022-07-12")
+
+    @patch("subprocess.check_output", return_value=''.encode())
+    def test_log_empty(self, mock_subprocess: Any) -> None:
+        log = self.repo.log('2021-10-26')
+        last_commit = log[-1]
```
dblock, line 70: This looks like a bug. I expected `log` to return an empty array if there were no entries.

tests/tests_git/test_git_repository.py
```diff
@@ -57,6 +57,27 @@ def test_output(self, mock_subprocess: Any) -> None:
...
+        self.assertEqual(last_commit.id, None)
+        self.assertEqual(last_commit.date, None)
+
+    @patch("subprocess.check_output", return_value='927de30 2022-07-12\n787d971 2022-04-28\n989143e 22-04-26\n8b376fc 2022-04-26\n023a2ac 2022-04-26'.encode())
+    def test_log_multiple(self, mock_subprocess: Any) -> None:
+        log = self.repo.log('2022-04-25')
+        last_commit = log[-1]
```
dblock, line 77: check the number of entries returned from `.log` match your expectation (I see 5 items)

tests/tests_release_notes_workflow/test_component_release_notes.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_plugin_check(self, *mocks: Any) -> None:
+        component = InputComponentFromSource({"name": "common-utils", "repository": "https://github.com/opensearch-project/common-utils", "ref": "c3408f34961eb7e8ebc0510306f62238a457dbbc"})
+        self.repo = GitRepository(
+            url="https://github.com/opensearch-project/common-utils",
+            ref="c3408f34961eb7e8ebc0510306f62238a457dbbc",
+        )
+        self.assertEqual(self.repo.url, "https://github.com/opensearch-project/common-utils")
+        self.assertEqual(self.repo.ref, "c3408f34961eb7e8ebc0510306f62238a457dbbc")
+        self.assertEqual(self.repo.sha, "c3408f34961eb7e8ebc0510306f62238a457dbbc")
+        release_notes = ComponentsReleaseNotes.from_component(component, "2.2.0", self.repo.dir)
```
dblock, line 31: Test class structure should match the code class structure. So this should be in a `TestComponentReleaseNotesOpenSearch` class and you should be directly creating a `ComponentReleaseNotesPlugin` instead of calling `from_component`. 

The `from_component` method is its own problem, should be in a `TestComponentReleaseNotes` class and have a test that ensures that this method returns the right release notes class based on input.

tests/tests_release_notes_workflow/test_component_release_notes.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_core_check(self, *mocks: Any) -> None:
+        component = InputComponentFromSource({"name": "OpenSearch", "repository": "https://github.com/opensearch-project/OpenSearch", "ref": "a081f2f13e5df63dcc0897ded2154d7eab3f306a"})
+        self.repo = GitRepository(
+            url="https://github.com/opensearch-project/OpenSearch",
+            ref="a081f2f13e5df63dcc0897ded2154d7eab3f306a",
+        )
+        self.assertEqual(self.repo.url, "https://github.com/opensearch-project/OpenSearch")
+        self.assertEqual(self.repo.ref, "a081f2f13e5df63dcc0897ded2154d7eab3f306a")
+        self.assertEqual(self.repo.sha, "a081f2f13e5df63dcc0897ded2154d7eab3f306a")
+        release_notes = ComponentsReleaseNotes.from_component(component, "2.2.0", self.repo.dir)
```
dblock, line 46: Same as above.  Move to a class called `TestComponentReleaseNotesOpenSearchPlugin` (not core).

tests/tests_release_notes_workflow/test_component_release_notes.py
```diff
@@ -0,0 +1,81 @@
...
+        self.assertEqual(release_notes.filename, '.release-notes-2.2.0.md')
+
+    @patch('subprocess.check_call', return_value=boolean)
+    @patch('subprocess.check_output', return_value='a081f2f13e5df63dcc0897ded2154d7eab3f306a'.encode())
+    @patch('release_notes_workflow.component_release_notes.ComponentReleaseNotesOpenSearchPlugin.path_exists')
+    @patch('release_notes_workflow.component_release_notes.ComponentReleaseNotesOpenSearchPlugin.path')
+    @patch('release_notes_workflow.component_release_notes.ComponentReleaseNotesOpenSearchPlugin.exists')
+    @patch('release_notes_workflow.component_release_notes.ComponentReleaseNotesOpenSearchPlugin.check')
+    def test_ComponentReleaseNotesOpenSearchPlugin(self, *mocks: Any) -> None:
```
dblock, line 55: Move to `TestComponentReleaseNotesOpenSearchPlugin` and call it `test_exists`, separate tests for `test_check` and `test_path_exists`.

tests/tests_release_notes_workflow/test_release_notes.py
```diff
@@ -0,0 +1,54 @@
...
+            "..",
+            "manifests",
+        )
+        OPENSEARCH_MANIFEST = os.path.realpath(os.path.join(MANIFESTS, "templates", "opensearch", "default", "manifest.yml"))
+        manifest_file = InputManifest.from_file(open(OPENSEARCH_MANIFEST))
+        release_notes = ReleaseNotes(manifest_file, "2022-07-26")
+        for component in manifest_file.components.select():
+            if type(component) is InputComponentFromSource:
+                self.assertTrue(release_notes.check(component))
+                assert type(release_notes.check(component)) == list
```
dblock, line 35: This doesn't actually test anything. This is supposed to test `ReleaseNotes#check`, you should be calling that method and asserting that it did whatever it was supposed to do.

tests/tests_release_notes_workflow/test_release_notes.py
```diff
@@ -0,0 +1,54 @@
...
+            "..",
+            "..",
+            "manifests",
+        )
+        OPENSEARCH_MANIFEST = os.path.realpath(os.path.join(MANIFESTS, "templates", "opensearch", "default", "manifest.yml"))
+        manifest_file = InputManifest.from_file(open(OPENSEARCH_MANIFEST))
+        release_notes = ReleaseNotes(manifest_file, "2022-07-26")
+        table_output = release_notes.table()
+        table_output.write_table()
+        table_output.dump("test.md")
```
dblock, line 54: This test doesn't test anything other than all these methods can be called.

tests/tests_release_notes_workflow/test_releasenotes_check_args.py
```diff
@@ -0,0 +1,52 @@
...
+    gitLogDateAssert = datetime.datetime.strptime(gitLogDate, "%Y-%m-%d").date()
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "check", OPENSEARCH_MANIFEST, "--date", gitLogDate])
+    def test_manifest(self) -> None:
+        self.assertEqual(ReleaseNotesCheckArgs().manifest.name, TestReleaseNotesCheckArgs.OPENSEARCH_MANIFEST)
+        self.assertEqual(ReleaseNotesCheckArgs().date, TestReleaseNotesCheckArgs.gitLogDateAssert)
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "check", OPENSEARCH_MANIFEST])
+    def test_manifest_withoutdate(self) -> None:
+        self.assertRaises(SystemExit)
```
dblock, line 44: Also need to check that the error message was what you expected.

dblock (changes requested):
There's quite a bit of work to get the tests to actually test what is intended in this code.

tests/test_run_releasenotes_check.py
prudhvigodithi: Added more assertions.

tests/test_run_releasenotes_check.py
prudhvigodithi: Added more assertions.

tests/tests_git/test_git_repository.py
prudhvigodithi: Modifies this to add more assertions that tests `gitlog` with one-commit, no-commit and more-commits.

tests/tests_git/test_git_repository.py
prudhvigodithi: Modifies this to add more assertions that tests `gitlog` with one-commit, no-commit and more-commits.

tests/tests_release_notes_workflow/test_component_release_notes.py
prudhvigodithi: Modifies the tests.

tests/tests_release_notes_workflow/test_component_release_notes.py
prudhvigodithi: Modified the tests.

tests/tests_release_notes_workflow/test_component_release_notes.py
prudhvigodithi: Modified the tests.

tests/tests_release_notes_workflow/test_release_notes.py
prudhvigodithi: Modified the test.

tests/tests_release_notes_workflow/test_releasenotes_check_args.py
prudhvigodithi: Assertions with unit tests, `SystemExit` does not have `message` attribute.
Error:
```AttributeError: 'SystemExit' object has no attribute 'message'```
However I have added `self.assertEqual(cm.exception.code, 2)`

File: tests/tests_release_notes_workflow/test_release_notes.py (hunk 45 above)
Comment (prudhvigodithi) [reply tests/tests_release_notes_workflow/test_release_notes.py]: Modifies the tests.

**Review by dblock — CHANGES_REQUESTED** [review]
1. Fix CI on Windows, needs an `os.path.join` instead of a hard-coded `/`.
2. See below.

File: src/release_notes_workflow/release_notes_component.py [hunk 48]
```diff
@@ -0,0 +1,52 @@
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
prudhvigodithi: Thanks dB, this was an issue with my editor.

tests/tests_release_notes_workflow/test_component_release_notes.py
```diff
@@ -0,0 +1,80 @@
...
+    def setUp(self) -> None:
+        self.component_release_notes = ComponentReleaseNotesOpenSearchPlugin(MagicMock(), "2.2.0", "path")
+
+    def test_filename(self) -> None:
+        self.assertEqual(self.component_release_notes.filename, ".release-notes-2.2.0.0.md")
+
+
+class TestComponentsReleaseNotes(unittest.TestCase):
+
+    def test_from_component(self) -> None:
```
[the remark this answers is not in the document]
prudhvigodithi: Just pushed this in my latest commit.

tests/tests_release_notes_workflow/test_release_notes.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @patch("subprocess.check_output", return_value=''.encode())
+    @patch("subprocess.check_call")
+    def test_check(self, *mocks: Any) -> None:
+        self.assertEqual(self.release_notes.check(self.component), ['OpenSearch', 'ref', None, None, False])
+
+    @patch("subprocess.check_output", return_value=''.encode())
+    @patch("subprocess.check_call")
+    def test_check_with_manifest(self, *mocks: Any) -> None:
+        for component in self.manifest_file.components.select():
```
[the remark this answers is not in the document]
prudhvigodithi: Got it, just pushed this change in my latest commit.

tests/tests_release_notes_workflow/test_releasenotes_check_args.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "--help"])
+    def test_usage(self) -> None:
+        with self.assertRaises(SystemExit):
+            main()
+
+        out, _ = self.capfd.readouterr()
+        self.assertTrue(out.startswith("usage:"))
+
+    @patch("argparse._sys.argv", [RELEASE_NOTES_CHECK_PY, "check", OPENSEARCH_MANIFEST, "--date", gitLogDate])
```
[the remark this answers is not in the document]
prudhvigodithi: Since the `gitLogDate` and `gitLogDateAssert` are used at multiple places I have done that way, but just pushed the change in my latest commit.

dblock (approved):

Issue #2410, opened 2022-08-03
[Enhancement] Display running status for each build and for each type of distribution components


zelinh:
### Is your feature request related to a problem? Please describe

Following to #2325.
We want to add a overall running status on the top of our input manifest and the running status of each component for all the distribution types we supported (e.g. tarball/rpm/windows(not yet supported)/...)

### Describe the solution you'd like

Looking to use our current groovy libraries to implement based on the input manifest [this PR](https://github.com/opensearch-project/opensearch-build/pull/2390) generated.

zelinh:
According to @bshien, additional info will be needed for the manifest file. 

General build:
- Result
- Start time
- Duration/End time
- Associated test numbers for the builds
Integ/BWC test info:
- result of overall test(what the result is on the associated test in Jenkins
- For the individual tests: security enabled?, distribution, test type(integ or bwc), architecture, tests executed, tests passed, tests failed, test start time, test end time,
For the performance tests,
- result of the overall test
- version, architecture, security enabled?

zelinh:
Close this as this has been completed. More enhancements can be related to #2512

Issue #2297, opened 2022-07-06
[Meta] Create UI dashboards to view the build, test and release stats


bbarani:
### Is your feature request related to a problem? Please describe

We currently do not have any dashboards to view the status of builds, tests and releases for OpenSearch and OpenSearch dashboard artifacts.

### Describe the solution you'd like

**Build, release and test Dashboards  (OpenSearch and OpenSearch Dashboards)**

- Integrate and display build time, PR’s, CVE’s, integration / BWC tests results corresponding to every distribution build(s) for all supported versions
- Link Performance results stats for distribution builds
- Display the performance stats along with highlighting anomalies for  distribution build(s) for all supported versions
- Add ability to download snapshot artifacts generated through distribution build process directly from the dashboard page 

### Architectural Diagram
![Untitled drawio (1)](https://user-images.githubusercontent.com/44730413/188994033-8922370c-e544-4237-a579-9ba6427618c4.png)
.jpg)

### Implementation

- [x] [Reduce API calls on main page from 30+ to 1](https://github.com/opensearch-project/opensearch-build/issues/2391)
- [x] [Use commits.yml to fetch commits per build/PR](https://github.com/opensearch-project/opensearch-build/issues/2392)
  - https://github.com/opensearch-project/opensearch-build/issues/2410
- [x] [Create Commits page](https://github.com/opensearch-project/opensearch-build/issues/2393)
- [x] [Create Integ test page](https://github.com/opensearch-project/opensearch-build/issues/2394)
- [ ] [Create BWC test page](https://github.com/opensearch-project/opensearch-build/issues/2395)
- [x] [Create Perf test page](https://github.com/opensearch-project/opensearch-build/issues/2396)
- [x] [Create CVE page](https://github.com/opensearch-project/opensearch-build/issues/2397)
- [ ] Start time display on dashboard

#### P0 (POC) :
- [x] Users can view current status of distribution build along with build stats
- [x] Users can download the distribution build artifacts directly from the dashboard
- [x] Users can view PR details corresponding to a build
- [x] Users can view CVE information corresponding to a build
- [x]  Users can view integration test stats corresponding to a build
- [x] Users can view performance test stats corresponding to a build
- [ ] Users can view BWC test stats corresponding to a build

bbarani:
This is the subset of https://github.com/opensearch-project/opensearch-build/issues/2080

bbarani:
[...]

PR #2473, opened 2022-08-17
Jenkins library to generate buildInfo.yml

zelinh:

### Description
Add the new jenkins library to generate buildInfo.yml used for our self-service dashboards. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2410


jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -75,16 +75,18 @@ pipeline {
...
             steps {
                 script {
-                    getRepositoryCommit(
-                        componentName: "${COMPONENT_NAME}",
-                        inputManifest: "manifests/${INPUT_MANIFEST}",
-                        outputFile: "commits.yml"
+                    buildInfoYaml(
+                        componentName: COMPONENT_NAME,
+                        inputManifest: "manifests/$INPUT_MANIFEST",
+                        outputFile: "buildInfo.yml",
```
dblock, line 81: Can we rename this to something not containing camelCase, we will land on case-insensitive file systems. Maybe just `build.yml`?

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -132,6 +134,17 @@ pipeline {
                                                 string(name: 'AGENT_LABEL', value: AGENT_X64)
                                             ]
 
+                                        unstash "buildInfo_manifest"
+                                        buildInfoYaml(
+                                            componentName: COMPONENT_NAME,
+                                            inputManifest: "buildInfo.yml",
```
dblock, line 140: Input = output?

vars/buildInfoYaml.groovy
```diff
@@ -0,0 +1,69 @@
...
+        echo("stage is $stageField")
+        echo("status is $status")
+        inputManifest.results.("$stageField".toString()) = "$status"
+    }
+    else if (args.stage == "COMPLETE") {
+        inputManifest.build.status = status
+        inputManifest.results.duration = currentBuild.duration
+        inputManifest.results.startTimestamp = currentBuild.startTimeInMillis
+    }
+    writeYaml(file: outputFile, data: inputManifest, overwrite: true)
```
dblock, line 67: This set of `ifs` feels very hard-coded. Use `args.stage` to identify the key that needs to be updated and set it to `status`.

dblock (commented):
This feels very verbose and complicated. 

From what I understand you're trying to save status of the build. So maybe the library function should be `status`, and it should stash/unstash a `build.yml`, then update it? I also believe things like "current stage" can be obtained from ENV and may not need to be passed around.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
zelinh: @dblock This is actually taking `buildInfo.yml` as input since we initialize it in the previous stage.

vars/buildInfoYaml.groovy
zelinh: I will try to refractor these sets to make it more straightforward.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
zelinh: I am thinking the main purpose of this yaml file is to provide the build information (e.g. each stage status, start time & duration, the job ids of triggering tests) from each workflow. `build.yml` maybe a little confusing with the build manifest. WDYT @dblock

zelinh:
> I also believe things like "current stage" can be obtained from ENV and may not need to be passed around, which should cut the logic here by half

The stage parameter passing into the function will also be used as the name variable in the `buildInfo.yml` file. That why I wish to name it instead of just using the stage name from the distribution workflow since a lot of them are more like a sentence.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
dblock: Isn't this all already provided by exporting a jenkins job history?
https://stackoverflow.com/questions/52866480/export-build-history-of-jenkins-job

zelinh:
Had some offline discussion with @dblock , I will be make following changes:
- `job.yml` will be more related to what this final result yaml file contains and will replace my current name `BuildInfo.yml`.
- Current I checkout the commits for each components when I initialized the yaml file, this will be wrong if any plugin commits anything into their repo. We should be getting these commit IDs from the assemble manifest (e.g. [this sample assemble manifest](https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.3.5/5951/linux/x64/tar/dist/opensearch/manifest.yml)). 
- Current `buildInfoYaml` jenkins method contains too much hardcoded structure and could be improved. 
- Future enhancements could be made to these entire workflow similar to https://github.com/opensearch-project/opensearch-build/blob/main/src/jenkins/Messages.groovy

vars/buildInfoYaml.groovy
```diff
@@ -0,0 +1,54 @@
...
+        }
+    } else {
+        echo ("Components parameter is null")
+        inputManifest.components.each { component ->
+            componentsList.add(component.name)
+        }
+    }
+    echo (componentsList.toString())
+
+    if (args.stage == "START") {
```
dblock, line 23: The `start` and `complete` should just be separate functions. Like `buildInfoCreate`, `buildInfoUpdate` and `buildInfoUpload`.

zelinh: After I discussed with @peterzhuamazon, we cut the the scope of this PR to exclude the status for individual stages. I think having separate groovy function can make it more complicate and hard to track.

vars/buildInfoYaml.groovy
```diff
@@ -0,0 +1,62 @@
...
+            if (componentsList.contains(component.name)) {
+                // Convert ref from branch to commit
+                dir(component.name) {
+                    checkout([$class: 'GitSCM', branches: [[name: component.ref]],
+                              userRemoteConfigs: [[url: component.repository]]])
+                    def commitID = sh(
+                            script: "git rev-parse HEAD",
+                            returnStdout: true
+                    ).trim()
+                    component.ref = commitID
```
peterzhuamazon, line 46: We should put this to the start and the complete stage.
If failed, we still have a commitid to check.
If complete, update the commitid again just in case.

jenkins/opensearch/distribution-build.jenkinsfile
```diff
@@ -75,16 +75,16 @@ pipeline {
...
                 script {
-                    getRepositoryCommit(
-                        componentName: "${COMPONENT_NAME}",
-                        inputManifest: "manifests/${INPUT_MANIFEST}",
-                        outputFile: "commits.yml"
+                    buildInfoYaml(
+                        componentName: COMPONENT_NAME,
+                        inputManifest: "manifests/$INPUT_MANIFEST",
+                        status: "NOT_STARTED",
+                        stage: "START"
```
peterzhuamazon, line 82: Using `Initialize_stage` better than `START` here, not to be confused with `STARTED/NOT_STARTED`.

jenkins/opensearch/distribution-build.jenkinsfile
```diff
@@ -594,6 +618,20 @@ pipeline {
         }
     }
     post {
+        always {
+            node(AGENT_X64) {
+                script {
+                    buildInfoYaml(
+                        componentName: COMPONENT_NAME,
+                        status: currentBuild.result,
+                        stage: "COMPLETE"
```
peterzhuamazon, line 627: Using `Finalize_stage` better than `COMPLETE` here, not to be confused with `STARTED/NOT_STARTED`.

vars/buildInfoYaml.groovy
```diff
@@ -0,0 +1,62 @@
...
+        }
+        sourceyml.build.status = status
+    } else {
+        stageField = args.stage
+        echo("stage is $stageField")
+        echo("status is $status")
+        sourceyml.results.("$stageField".toString()) = status
+    }
+    sourceyml.results.duration = currentBuild.duration
+    sourceyml.results.startTimestamp = currentBuild.startTimeInMillis
```
peterzhuamazon, line 58: Update duration during finalized stage and update starttime during initialized stage.

vars/buildInfoYaml.groovy
zelinh: Actually, i realized this part is in the `post always` stage. This would be executed no matter the job fails or succeeds.

peterzhuamazon (approved):

vars/buildInfoYaml.groovy
```diff
@@ -0,0 +1,62 @@
...
+        }
+    } else {
+        echo ("Components parameter is null")
+        sourceyml.components.each { component ->
+            componentsList.add(component.name)
+        }
+    }
+    echo (componentsList.toString())
+
+    if (args.stage == "INITIALIZE_STAGE") {
```
dblock, line 30: I still find this big state machine quite problematic. I won't block this PR for it, but I think the biggest bug here is that your workflow will silently produce an incorrect YML if you forget to call the function with `INITIALIZE_STAGE`.

I would extract the initialization into its own function in which you do not `unstash` an existing build info and remove the `catch` in unstashing it - that should never happen unless there's a bug in the workflow.

vars/buildInfoYaml.groovy
```diff
@@ -0,0 +1,62 @@
...
+        echo("Initiate the build info yaml file.")
+        sourceyml.build.status = "IN_PROGRESS"
+        sourceyml.build.number = "${BUILD_NUMBER}"
+        sourceyml.results = [:]
+        sourceyml.results.startTimestamp = currentBuild.startTimeInMillis
+    } else if (args.stage == "FINALIZE_STAGE") {
+        sourceyml.components.each { component ->
+            if (componentsList.contains(component.name)) {
```
dblock, line 38: The value obtained here will be wrong. Let's say your ref is 2.x, it's almost certain that there will be a new commit between the time you build and the time you run this code. We already produce the commit ID in build output manifest (e.g. [here](https://github.com/opensearch-project/opensearch-build/blob/main/tests/data/opensearch-build-2.0.0-rc1.yml#L2465)), use that.

dblock (dismissed):
The finalize stage produces wrong commit IDs, so I think that warrants a must have.

vars/buildInfoYaml.groovy
zelinh: Actually what i was thinking is this build manifest or bundle manifest would only output if the build or assemble succeed. If the building stages failed before, we won't have this certain manifest. The scope of this PR is to display the head commit of each components no matter the job succeed or fail and the commit ID it contains will serve for informative purpose in the [dashboard](https://github.com/opensearch-project/opensearch-build/issues/2297).

dblock: I think yo can use `--lock`, https://github.com/opensearch-project/opensearch-build/tree/main/src/build_workflow#avoiding-rebuilds.

```
./build.sh --lock $MANIFEST # generates opensearch-1.2.0.yml.lock
```

The refs in this manifest are commit IDs instead of any refs.

zelinh: after i going through, this looks like different than what i expect.
- running ./build.sh --lock manifests/2.2.0/opensearch-2.2.0.yml will not actually build & generate the lock file. it would only generate the lock file and return. https://github.com/opensearch-project/opensearch-build/blob/7b61ab48515cf8a1e55c63bbe2971685708e3fbf/src/run_build.py#L28-L39
- If we then adding this line separate from the actual process of build & assemble , its logic is pretty much the same as what I was doing in my PR: https://github.com/opensearch-project/opensearch-build/blob/7b61ab48515cf8a1e55c63bbe2971685708e3fbf/src/manifests/input_manifest.py#L241-L244 and still have the same concern of wrong data if people push when building.
- I think the best approach would be adding this lock process into the actual building process instead of simply returning after generating the lock file. While I think that’s out of the scope of this PR. I could raise an issue to for requesting this feature. What do you think?

dblock: I think that’s better.

bbarani: @dblock Are you ok with merging this PR and opening a new issue to address the gaps with future PR? Currently this is created to address a POC but we will definitely enhance it further when we productionize it.

dblock: I'm ok with merging this PR. @zelinh please open an issue describing your suggestion of adding the lock process into the actual building process, and consuming the stable manifest instead of this code.

zelinh: @dblock Yes. Sure I will create an issue for enhancement feature request. We will be working on the improvement. Thanks.

Issue #2306, opened 2022-07-07
[WIN M1] WIN (ZIP/MSI) Build/Assemble Process 


peterzhuamazon:
|Tasks(ZIP) |**Tasks(MSI)** |Estimate   |Status(ZIP)    |Status(MSI)    |Notes  |
|---    |---    |---    |---    |---    |---    |
|Re-use the existing build process to generate the  OpenSearch/Dashboards min + all of the plugins artifacts for WIN package to  use |Same   |0  |Completed  |Completed  |   |
|The artifacts should be built with ~~LINUX~~ Windows platform  specified, ~~as we will cross-compile WIN binary on LINUX then package with  WINDOWS JDK~~.  |Same   |1  |Completed    |Completed   |~~This is still in debate as we can techinically build WIN  on Windows machine, but there are a lot of things to setup just so Jenkins  can run Python on Windows Agent.~~ We are able to run shell scripts natively on Windows agent  |
|We already have "zip" supported for "—distribution“ parameter, but needs to check whether it is already combined with `--platform windows`. |We do not have "exe" support for "—distribution“ yet. However, this is different from "RPM" as we do not need min artifact to be a exe. The min artifact can be zip and the final product in assemble can be exe.  |2  |Completed   |Completed   |As for "exe" we need to discuss whether a  standard exe is enough, or do we want to invest into Windows official  installer "msi".  |
|We already have "--distribution" param available in assemble workflow, just need to verify existing functions of "ZIP".    |We already have "--distribution" param available  in assemble workflow, but no support for "EXE" redirection. Need to  add a child class supporting the new distribution.   |2  |Completed   |Completed     |   |
|The generation code should pull the artifacts from the build workflow to a temporary location  |Same   |1  |Completed   |Completed    |   |
|The code will compile the components and also call existing install function to install  plugins on min artifacts   |Same   |1  |Completed  |Completed   |  ETA: 2022/09/16 |
|After installation, the code will execute a tool or  utility to wrap all the content into corresponding distribution format |Same   |1  |Completed  |Completed    | **20220819 Note:** Plugin compilation currently have some issues with the build scripts, the compilation itself seems ok at least on things like common-utils. <br/><br/> ETA: 2022/09/16|
|The code will move the final distribution artifact from the temp location to dist folder   |Same   |1  |Completed  |Completed    |  ETA: 2022/09/07 |

[...]

PR #2483, opened 2022-08-19
Add Windows Min Artifacts Building on Jenkins 

peterzhuamazon:
### Description
Add Windows Min Artifacts Building on Jenkins 

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2306


src/system/temporary_directory.py
```diff
@@ -17,9 +16,21 @@
...
-        func(path)
+    logging.debug(f"excvalue {excvalue}")
+    logging.debug(f"func {func.__name__}")
+    if func in (os.rmdir, os.remove, os.unlink):
+        os.chmod(path, stat.S_IRWXU | stat.S_IRWXG | stat.S_IRWXO | stat.S_IWRITE | stat.S_IREAD)  # 0777 nix* / +wr windows
+        try_total = 3
+        for try_count in range(try_total):  # Re-run func to force deletion especially on windows
+            try:
+                logging.debug(f'Try count: {try_count + 1}/{try_total}')
+                func_result = None
```
dblock, line 27: Unnecessary assign here.

src/system/temporary_directory.py
```diff
@@ -17,9 +16,21 @@
...
 def g__handleRemoveReadonly(func: FunctionType, path: str, exc: Any) -> Any:
     excvalue = exc[1]
-    if func in (os.rmdir, os.remove, os.unlink) and excvalue.errno == errno.EACCES:
-        os.chmod(path, stat.S_IRWXU | stat.S_IRWXG | stat.S_IRWXO)  # 0777
-        func(path)
+    logging.debug(f"excvalue {excvalue}")
+    logging.debug(f"func {func.__name__}")
+    if func in (os.rmdir, os.remove, os.unlink):
+        os.chmod(path, stat.S_IRWXU | stat.S_IRWXG | stat.S_IRWXO | stat.S_IWRITE | stat.S_IREAD)  # 0777 nix* / +wr windows
+        try_total = 3
```
dblock, line 23: Rename to `retry_count`

src/system/temporary_directory.py
```diff
@@ -17,9 +16,21 @@
...
+        for try_count in range(try_total):  # Re-run func to force deletion especially on windows
```
dblock, line 24: Rename to `retry`

src/system/temporary_directory.py
```diff
@@ -17,9 +16,21 @@
...
+            try:
+                logging.debug(f'Try count: {try_count + 1}/{try_total}')
+                func_result = None
+                func_result = func(path)
+                if func_result is None:
+                    break
+            except Exception as ex:
+                logging.debug(func_result)
```
dblock, line 32: Do `logging.warn` of the actual error inside `ex`.

I would remove `logging.debug`'s.

dblock (commented):
This is fine. 

I think we should raise the bar on the retry implementation and take a separate function from https://stackoverflow.com/questions/42521549/retry-function-in-python for example.

peterzhuamazon:
Hi @dblock I agree with many points you raised but not agree with putting the retries into another function.
The `func(path)` is specifically to `onerror` and requires a specific `path` param.
If we want to make a function that is generalized for everything we need to write complicated code just to take `x` number of params from different functions which is not very suitable.

src/system/temporary_directory.py
peterzhuamazon: removed.

src/system/temporary_directory.py
peterzhuamazon: renamed.

src/system/temporary_directory.py
peterzhuamazon: renamed

src/system/temporary_directory.py
peterzhuamazon: changed to warn.

zelinh (approved):

PR #2513, opened 2022-08-25
add performance-analyzer to manifests

kaushalmahi12:
Signed-off-by: Kaushal Kumar <kshkmr@amazon.com>

### Description
Add performance-analyzer to manifests.

### Issues Resolved

### DIFF
--- a/manifests/2.3.0/opensearch-2.3.0.yml
+++ b/manifests/2.3.0/opensearch-2.3.0.yml
@@ -25,3 +25,9 @@ components:
   - name: security
     repository: https://github.com/opensearch-project/security.git
     ref: 2.x
+  - name: performance-analyzer
+    repository: https://github.com/opensearch-project/performance-analyzer.git
+    ref: 2.2
+    checks:
+      - gradle:properties:version
+      - gradle:dependencies:opensearch.version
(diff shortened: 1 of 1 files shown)


manifests/2.3.0/opensearch-2.3.0.yml
```diff
@@ -25,3 +25,9 @@ components:
   - name: security
     repository: https://github.com/opensearch-project/security.git
     ref: 2.x
+  - name: performance-analyzer
+    repository: https://github.com/opensearch-project/performance-analyzer.git
+    ref: 2.2
```
zelinh, line 30: Hi @kaushalmahi12, the reason python test failed is because you would need single quote for this floating type of reference. 
Also since you are modifying on the 2.3 manifest, this shouldn't be on 2.2 branch as currently i see its version is 2.2.1. 
For this PR, this is designated to be the main branch as the version bump [here](https://github.com/opensearch-project/performance-analyzer/pull/258) on the main branch.

qreshi: Ideally I think Performance Analyzer should create a `2.x` branch from their current `main` and specify that here in the 2.3.0 manifest. This is consistent with the branching strategy agreed upon for OpenSearch and it'll avoid having to change this again when their `main` inevitably upgrades to `3.0`

kaushalmahi12: Hi folk, I will point to 2.3 branch in that case.

peterzhuamazon (approved):

PR #2520, opened 2022-08-28
Added exclude patterns to the version increment workflow.

prudhvigodithi:
Signed-off-by: prudhvigodithi <pgodithi@amazon.com>

### Description
Added exclude patterns to the version increment workflow.
This exclude pattern required for `geospatial` and `notifications` plugins as they are not part of 1.x version and released as part of 2.x.

### Issues Resolved

### DIFF
--- a/.github/workflows/increment-plugin-versions.yml
+++ b/.github/workflows/increment-plugin-versions.yml
@@ -40,6 +40,10 @@ jobs:
           - {repo: sql}
         branch:
           - '2.2'
+          - '1.3'
+        exclude:
+          - {repo: geospatial, branch: '1.3'}
+          - {repo: notifications, branch: '1.3'}
     steps:
       - name: Check out OpenSearch repo
         uses: actions/checkout@v3


dblock:
What is the reason to exclude those? (Add to the PR description).

prudhvigodithi:
> What is the reason to exclude those? (Add to the PR description).

Sure @dblock thanks, updated the description, will create an issue and more details towards it, the reason I have added to draft is still looking for better solution, right now the branches like `1.3`, `2.2` added are hardcoded and will be added/removed based of future releases, rather I'm just thinking how I can add this dynamically or as an input for workflow execution, so that the execution will only be for the right version and during the right time, rather than run the version increment always for all the listed branches.

prudhvigodithi:
> Sure @dblock thanks, updated the description, the reason I have added to draft is still looking for better solution, right now the branches like `1.3`, `2.2` added are hardcoded and will be added/removed based of future releases, rather I'm just thinking how I can add this dynamically or as an input for workflow execution, so that the execution will only be for the right version and during the right time, rather than run the version increment always for all the listed branches.

lets go with this PR, its fine and one way good, to keep the version increment consistent for all release branches, example if the release is 1.3.5, when this workflow is executed, not only 1.3 it will also auto increment the branch 2.2.
@dblock @gaiksaya @peterzhuamazon @bb

gaiksaya (approved):

Issue #2558, opened 2022-09-06
Jenkinsfiles to have an input `BUILD_BRANCH`.


prudhvigodithi:
### Is your feature request related to a problem? Please describe

Jenkinsfiles to have an input `BUILD_BRANCH`. Once the [branching strategy](https://github.com/opensearch-project/opensearch-build/issues/2549) in place, Jenkinsfiles should be able to checkout the right branch as per release to make sure release specific code is executed.

### Describe the solution you'd like

- Add a new input `BUILD_BRANCH`.
- Add a new stage that checks out the build repo with specific `BUILD_BRANCH`.
- Stash the repo files.
- Unstash in every required stage to make sure the stage uses the right code specific to `BUILD_BRANCH`

### Additional context

More useful info on stash and unstash https://www.youtube.com/watch?v=0GQZzLlnUws

prudhvigodithi:
PR converted to draft https://github.com/opensearch-project/opensearch-build/pull/2559, until a solution is finalized that satisfies the version/branching requirement with OpenSearch and OpenSearch Dashboards version being offset
More details: https://github.com/opensearch-project/opensearch-build/issues/2549

bbarani:
Can we also close this issue until we finalize a solution @prudhvigodithi

Issue #2549, opened 2022-09-02
Branching Strategy Implementation


prudhvigodithi:
### Is your feature request related to a problem? Please describe

Coming from the initial [[Proposal] Branching strategy for build](https://github.com/opensearch-project/opensearch-build/issues/1975) and the solution [proposed](https://github.com/opensearch-project/opensearch-build/issues/1975#issuecomment-1145862692) as part of this proposal would be the closest implementation for the build repo.

To summarize the solution:

- The `main` branch is used for active development and targets the upcoming release version.
- A `<MAJOR_VERSION>.x` branch is cut for every major release version, example as `2.x`, `3.x`. This is inclined with the Core branching and with planned release cycles.  Considering the next major release is `3.0.0`, the `main` would have active development, new features and everything compatible for `3.x` and above. A `2.x` branch should have everything related to `2.x` series. Now when the next upcoming release is `4.0.0`, `main` would shift everything compatible for `4.x` and above and `3.x` branch will be added to the repo.
- During distribution jobs execution, for `2.2.1` release, the pipelines/jobs would use an input `BRANCH_NAME` and use the `2.x` and reference.

**NOTE**: The on going activity to [separate out the Jenkins shared libraries ](https://github.com/opensearch-project/opensearch-build/issues/2489) would indirectly influence this branching strategy. The jenkinsfiles should be modified to fetch the shared libs from external repo, example https://github.com/prudhvigodithi/opensearch-build-pipelines/blob/main/opensearch/distribution-build.jenkinsfile#L1-L4. With this, the jenkins shared libraries and its associated tests need not be part of this branching implementation.

There is other Open issue related to this topic to server the entire build repo as framework with right versioning, which can be later fetch during runtime using the right version.

- [ ] https://github.com/opensearch-project/opensearch-build/issues/2490

### Describe the solution you'd like

- [ ] The Jenkinsfiles cannot be branched, hence the parking place for all the jenkinsfiles would be the `main` branch. The public Jenkins jobs will use Jenkinsfiles fetching from `main` branch.
- [ ] Add an input `BUILD_BRANCH` to the Jenkinsfiles so that jobs can clone the [build repo](https://github.com/opensearch-project/opensearch-build) with right `BUILD_BRANCH` used for the release.
https://github.com/opensearch-project/opensearch-build/issues/2558
[...]

Issue #1606, opened 2022-02-09
Create a function for a jenkins pipeline with defaults and basic configuration parameters


peternied:
#### Note; this work should not start until all design tasks in https://github.com/opensearch-project/opensearch-build/issues/1234 have been completed as they have implications on the acceptance criteria of these tasks.

See [standard release pipeline](https://github.com/peternied/opensearch-build/blob/e2dcfd49cef494b0c273bce40bd532cf8099a45d/vars/standardReleasePipeline.groovy) function from the proof of concept.

Acceptance Criteria:
- The standard release pipeline should contain common code to be inherited by client's jenkins files

PR #2559, opened 2022-09-06
Jenkinsfile change to support branching

prudhvigodithi:
Signed-off-by: prudhvigodithi <pgodithi@amazon.com>

### Description
Jenkinsfile change to support branching, starting with daily used jenkins jobs of OpenSearch and OpenSearch dashboard.

### Issues Resolved
Part of: https://github.com/opensearch-project/opensearch-build/issues/2549
Related: https://github.com/opensearch-project/opensearch-build/issues/2558


gaiksaya:
Hi @prudhvigodithi, can you check this issue? https://github.com/opensearch-project/opensearch-build/issues/1606
We can put this functionality into a common pipeline which then can be consumed by all JenkinsFiles

peterzhuamazon:
There are too many stash unstash and I wonder is there a way to load a common lib like @gaiksaya suggested.

prudhvigodithi:
The problem I see is git checkout, yes can be added to shared library, still that shared library function has to be called in jenkinsfile, but I dont see a way we can escape the unstash line, as jenkins expects for every stage to have that unstash method so that it can re-use those cloned files, will do some more research here.
@peterzhuamazon @gaiksaya 
Thank you

prudhvigodithi:
Even to load the shared library we need to add a stage/line to checkout and stash, I dont see a huge difference here to add 2 line in the jenkinsfile itself that does the git checkout and stash the files, also to unstash I dont see any other way to declare globally other then to have `unstash` part of each stage.

jenkins/opensearch-dashboards/bwc-test.jenkinsfile
```diff
@@ -20,20 +22,40 @@ pipeline {
             description: 'The build manifest URL, e.g. https://ci.opensearch.org/ci/dbc/distribution-build-opensearch-dashboards/2.0.0/98/linux/x64/builds/opensearch-dashboards/manifest.yml.',
             trim: true
         )
+        string(
+            name: 'BUILD_BRANCH',
+            description: "Build repo branch to use.",
+            defaultValue: "main",
+            trim: true
+        )
```
dblock, line 30: This feels very wrong. This job itself lives in this repo in a branch. If I am running the job out branch X, I would not expect the job to check out build tools from branch Y.

dblock:
I think you want to model this after GitHub actions. The .yml jobs for GHA live in a branch (e.g. 1.2.3) and run from that branch (1.2.3) checking out source code by default from the same branch (1.2.3). This it's never a parameter to the job, the only time you care about which branch to use is when you import the job into Jenkins.

jenkins/opensearch-dashboards/bwc-test.jenkinsfile
prudhvigodithi: This is done for the fact Jenkinsfiles cannot be branched, they has to reside in `main` and jenkins job git SCM configuration uses single branch to pull the jenkinsfile, once jenkinsfile is pulled, depending on the branch passed as input using `BUILD_BRANCH`, then the checkout happens with right branch and all the stages are executed as per the flow.

prudhvigodithi:
> I think you want to model this after GitHub actions. The .yml jobs for GHA live in a branch (e.g. 1.2.3) and run from that branch (1.2.3) checking out source code by default from the same branch (1.2.3). This it's never a parameter to the job, the only time you care about which branch to use is when you import the job into Jenkins.

Hey @dblock, assume the build repo is branched as `main`, `2.x` and `3.x`, now jenkinsfiles can only be on `main`, since the job configuration need to pull the jenkinsfiles from a fixed branch (which has be configured inside the job settings), now we have to build workflow code part of `2.x` (we are releasing `2.3.0`), the job clones the jenkinsfiles from `main` and switch the branch to `2.`x using the `BUILD_BRANCH` parameter, this way the workspace now has the right `2.x` release code to run the build and assemble.

gaiksaya:
I think what is causing the confusion is job configuration (one time thing manually for now) is where we state from where will the related JenkinsFile be pulled up. 
See the snippet below:
image
Need to check if that `branch` is configurable via code.

prudhvigodithi:
We can go with this PR until the [Branching Strategy Implementation](https://github.com/opensearch-project/opensearch-build/issues/2549) is finalized, I dont see any harm that would break the existing distribution functionality. This input value `BUILD_BRANCH` would be basic requirement to fetch the right branch code.
Since the release [2.3.0](https://github.com/opensearch-project/opensearch-build/issues/2447) is ongoing waiting to get a RC candidate for 2.3 before we can merge this.

dblock:
I think this is the wrong implementation for an otherwise good idea. As an example, your `distribution-build.jenkinsfile` comes from `main`, but you're trying to specify `1.2.3` as a build branch. The problem is that it's the branch in _the same repo_, so your .jenkinsfile may or may not work with version `1.2.3` at all. 

I think you want to extract the python tooling into a new repo, then hard-code the branch to use from that repo in the manifests, not take a parameter in build. This would make it reproducible and configurable in 1 place.

prudhvigodithi:
Hey @dblock I like the idea to separate the python tooling and pull the right tooling from manifest using the branch (or version), but its just not python tooling, we have some scripts here as well https://github.com/opensearch-project/opensearch-build/tree/main/scripts that needs to be focused for branching.

> The problem is that it's the branch in _the same repo_, so your .jenkinsfile may or may not work with version `1.2.3` at all.

@dblock Jenkinsfile cannot be branched, hence the are only part of `main` , shared libraries are already being separated out and will be imported by version, jenkinsfile will switch the branch and execute the stages as per release. So ideally same .jenkinsfile will be used for 1.x and 2.x, i mean if there is a drastic change in jenkinsfile, even with separation there are chances pipeline fails for past releases as well ;).

How about we get this merged, observe with `BUILD_BRANCH` input and then if required consider for separating the python tooling or even other scripts if required?

If no then I will close this PR and start focusing on python workflow separation and consider branching(or versioning) on those separated repos. :)  Once done then some versioning has to be done for those scripts as well https://github.com/opensearch-project/opensearch-build/tree/main/scripts 
@bbarani @peterzhuamazon @gaiksaya 
Thank you

dblock:
I recommend closing and focusing on branching those new repos, because this is a step in the wrong direction.

prudhvigodithi:
Closing this PR. More details https://github.com/opensearch-project/opensearch-build/issues/2549#issuecomment-1244334846, https://github.com/opensearch-project/opensearch-build/issues/2549#issuecomment-1245859146
Thank you

Issue #2601, opened 2022-09-09
Automate the branch entry to version increment workflow.


prudhvigodithi:
### Is your feature request related to a problem? Please describe

Coming from the comment https://github.com/opensearch-project/opensearch-build/issues/1375#issuecomment-1241267951
Exiting [version increment workflow](https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/increment-plugin-versions.yml), works by a branch input passed to matrix array, once the branch is the array, release manager can trigger the workflow with single click, but still the branch entry is a manual process that has to be done by a release manager. This involves raising a PR to add a new release branch and get it approved before we run it.

### Describe the solution you'd like

Extend the existing manifest workflow that raises an [AUTO] PR with new version by adding new version manifests and updating the [check-for-build.jenkinsfile](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/check-for-build.jenkinsfile). Sample PR https://github.com/opensearch-project/opensearch-build/pull/2599.
Add a logic to update the [version increment workflow](https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/increment-plugin-versions.yml) to have the new `<MAJOR_VERSION.x>` branch entry.

prudhvigodithi:
Closing this issue as the related [PR](https://github.com/opensearch-project/opensearch-build/pull/2602), to add the branch entry is merged.
Thank you

Issue #1375, opened 2021-12-17
OpenSearch: Automate increment to next development iteration


dblock:
### Is your feature request related to a problem? Please describe

We begin incrementing versions _after_ we decide that we want a release. This takes a day and makes it that nightly distribution builds do not include full bundle until T-14 days to release (#1140). Instead, prepare the project to make the next release after the previous release is done, so that we can save that day and have release candidates always available for the next development iteration.

After [automatic tagging](https://github.com/opensearch-project/opensearch-build/issues/378) has happened, increment the version for the next development iteration. For example:

- when we release 1.2.1, increment version to 1.2.2 on the 1.2 branch
- when we release 1.3.0, increment version to 1.4.0 on 1.x branch and 1.3.1 on 1.3 branch
- when we release 2.0, increment version to 3.0 on main

x 15 plugins

### Describe the solution you'd like

Every project should have a workflow that notices new tags, and increments versions accordingly and makes a PR. 

- On OpenSearch this change is a bit involved because it needs custom work in bcwVersions.
- In plugins it's mostly build.gradle, but there are plugins with their own weird custom steps

### Acceptance Criteria
- [x] POC Concept/Solution to Automate the version increment to next development iteration
https://github.com/opensearch-project/opensearch-build/issues/2215
- [x] Implement Automated workflow that will increment the version for plugin and take care of all other code changes required to pass the CI.
https://github.com/opensearch-project/opensearch-build/issues/2223
Pr: https://github.com/opensearch-project/opensearch-build/pull/2291
- [x] PR is merged upon review and ready to use.

### Additional Acceptance Criteria
Follow [Core Branching Strategy](https://github.com/opensearch-project/.github/blob/main/RELEASING.md#plugin-branching) (Ensure 1.x and 2.x branches).
https://github.com/opensearch-project/opensearch-plugins/issues/142

#### OpenSearch 
[Solution Proposed](https://github.com/opensearch-project/opensearch-build/issues/1375#issuecomment-1159476686)
Gradle project: Staging to add gradle tasks (`setVersion` and `versionIncrement`) that support version increment automation for supported versions and add/update `gradle.properties` file to the project.
- [x] [opensearch-plugins](https://github.com/opensearch-project/opensearch-plugins/issues/155)
Pr's: https://github.com/opensearch-project/opensearch-plugin-template-java/pull/32
- [x] [alerting](https://github.com/opensearch-project/alerting/issues/482)
Pr's:
[...]

PR #2602, opened 2022-09-09
Automate the branch entry to version increment workflow

prudhvigodithi:
### Description
Automate the branch entry to version increment workflow

### Issues Resolved
Part of: https://github.com/opensearch-project/opensearch-build/issues/1375
Solves: https://github.com/opensearch-project/opensearch-build/issues/2601

### DIFF
--- a/src/manifests_workflow/input_manifests.py
+++ b/src/manifests_workflow/input_manifests.py
@@ -173,26 +173,29 @@ class InputManifests(Manifests):
         logging.info(f"Wrote {jenkinsfile}")
 
     def add_to_versionincrement_workflow(self, version: str) -> None:
-        logging.info("Adding new version the version increment workflow")
         versionincrement_workflow_file = self.versionincrement_workflow()
         yaml = ruamel.yaml.YAML()
-        yaml.explicit_start = True
-        yaml.preserve_quotes = True
+        yaml.explicit_start = True  # type: ignore
+        yaml.preserve_quotes = True  # type: ignore
 
         with open(versionincrement_workflow_file) as f:
             data = yaml.load(f)
 
-        version_entry = version.split(".")[0] + ".x"
-        for branch in data["jobs"]["plugin-version-increment-sync"]["strategy"]["matrix"]["branch"]:
-            if version_entry in branch:
-                print(f"{versionincrement_workflow_file} already contains an entry for {version_entry}")
-                return None
+        version_entry = []
+        major_version_entry = version.split(".")[0] + ".x"
+        minor_version_entry = version.rsplit(".", 1)[0]
+        if minor_version_entry not in data["jobs"]["plugin-version-increment-sync"]["strategy"]["matrix"]["branch"]:
+            print(f"Adding {minor_version_entry} to {versionincrement_workflow_file}")
+            version_entry.append(minor_version_entry)
+        if major_version_entry not in data["jobs"]["plugin-version-increment-sync"]["strategy"]["matrix"]["branch"]:
+            print(f"Adding {major_version_entry} to {versionincrement_workflow_file}")
+            version_entry.append(major_version_entry)
 
-        branch_list = list(data["jobs"]["plugin-version-increment-sync"]["strategy"]["matrix"]["branch"])
-        branch_list.append(version_entry)
-        data["jobs"]["plugin-version-increment-sync"]["strategy"]["matrix"]["branch"] = branch_list
-        yaml.indent(mapping=2, sequence=4, offset=2)
-        with open(versionincrement_workflow_file, 'w') as f:
-            yaml.dump(data, f)
-
-        logging.info("Added new version the version increment workflow")
+        if version_entry:
+            branch_list = list(data["jobs"]["plugin-version-increment-sync"]["strategy"]["matrix"]["branch"])
+            branch_list.extend(version_entry)
+            data["jobs"]["plugin-version-increment-sync"]["strategy"]["matrix"]["branch"] = branch_list
+            yaml.indent(mapping=2, sequence=4, offset=2)
+            with open(versionincrement_workflow_file, 'w') as f:
+                yaml.dump(data, f)
+            logging.info("Added new version to the version increment workflow")
--- a/src/manifests_workflow/input_manifests.py
+++ b/src/manifests_workflow/input_manifests.py
@@ -11,6 +11,8 @@ import re
 from abc import abstractmethod
 from typing import Dict, List, Type, Union
 
+import ruamel.yaml
+
 from manifests.input_manifest import InputComponents, InputManifest
 from manifests.manifests import Manifests
 from manifests_workflow.component_opensearch import ComponentOpenSearch
@@ -30,6 +32,10 @@ class InputManifests(Manifests):
         return os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "..", "manifests"))
 
     @classmethod
+    def workflows_path(self) -> str:
+        return os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "..", ".github", "workflows"))
+
+    @classmethod
     def legacy_manifests_path(self) -> str:
         return os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "..", "legacy-manifests"))
 
@@ -40,6 +46,10 @@ class InputManifests(Manifests):
     @classmethod
     def cron_jenkinsfile(self) -> str:
         return os.path.join(self.jenkins_path(), "check-for-build.jenkinsfile")
+
+    @classmethod
+    def versionincrement_workflow(self) -> str:
+        return os.path.join(self.workflows_path(), "increment-plugin-versions.yml")
 
     @classmethod
     def files(self, name: str) -> List:
@@ -114,6 +124,7 @@ class InputManifests(Manifests):
             for release_version in sorted(main_versions.keys() - known_versions):
                 self.write_manifest(release_version, main_versions[release_version])
                 self.add_to_cron(release_version)
+                self.add_to_versionincrement_workflow(release_version)
 
     def create_manifest(self, version: str, components: List = []) -> InputManifest:
         templates_base_path = os.path.join(self.manifests_path(), "templates")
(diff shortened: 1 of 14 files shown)


prudhvigodithi:
Strange this works on my local 
```
pipenv run mypy .
Success: no issues found in 323 source files
```

src/manifests_workflow/input_manifests.py
```diff
@@ -11,6 +11,8 @@
 from abc import abstractmethod
 from typing import Dict, List, Type, Union
 
+import ruamel.yaml
```
peterzhuamazon, line 14: What is the reason to use ruamel.yaml without using the default yaml module?

peterzhuamazon:
@prudhvigodithi can you explain what you are trying to do here?
I see you are trying to load a yaml and check the matrix.
While figuring out what is missing and add new entries, but not very clear without the data. Thanks.

src/manifests_workflow/input_manifests.py
prudhvigodithi: Hey @peterzhuamazon there is no default YAML module, the comparison I did was between [PyYAML](https://pyyaml.org/wiki/PyYAMLDocumentation) and [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) and I see `ruamel.yaml` is working better when dealing with yaml parsing.
1) Using `ruamel.yaml` it's easy to override the methods like `explicit_start`, `preserve_quotes` and right `indent` which honors `yamllint` and gh  workflow `.yml` rules.
2) `ruamel.yaml` also does not change the [on](https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/increment-plugin-versions.yml#L4) key to boolean (default the library considers `on` as boolean) which is not easy with PyYAML, need some extra logic to create constructer and override methods.
https://stackoverflow.com/questions/36463531/pyyaml-automatically-converting-certain-keys-to-boolean-values
3) [---](https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/increment-plugin-versions.yml#L1), this is also preserved with `ruamel.yaml` easily compared to `PyYAML`, without this the GH and yamllint flags it as error.

prudhvigodithi:
> @prudhvigodithi can you explain what you are trying to do here? I see you are trying to load a yaml and check the matrix. While figuring out what is missing and add new entries, but not very clear without the data. Thanks.

So @peterzhuamazon in order to trigger the [version-increment workflow](https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/increment-plugin-versions.yml), we the right [branch entries](https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/increment-plugin-versions.yml#L41), having these branches, for every branch the a job is created and version increment is performed, now every release one has to raise a [PR](https://github.com/opensearch-project/opensearch-build/pull/2532/files) to add this branch entry.

With this PR this branch entry with be auto updated as part of the manifest workflow [[AUTO] Updated input manifests](https://github.com/opensearch-project/opensearch-build/pull/2599) PR, so next time we dont need to manually raise a PR for `2.x`, `3.x` etc.

peterzhuamazon:
> 

I see what you mean so if there is any new branches it will update the versions-increment yml workflow so we dont need to manually raise PR.

Thanks.

src/manifests_workflow/input_manifests.py
peterzhuamazon: Since we are running our python code on win/macos now can you also confirm these two can be installed correctly on win/macos? We have cases before where some modules are not compatible on those platforms.

Thanks.

peterzhuamazon: And yes I mean the PyYAML module as it is widely used.

prudhvigodithi: Got it, I see windows python tests pass, but I will confirm if this package has any restrictions to windows.
Thanks @peterzhuamazon

prudhvigodithi: Mac it works with as all my tests and development are from mac (with jdk14)
from website i see its as [os independent](https://pypi.org/project/ruamel.yaml.cmd/)
Some more useful details: https://snyk.io/advisor/python/ruamel-yaml

gaiksaya: @prudhvigodithi Do you think the default PyYaml should be replaced by this library?
If yes can you create an issue since you already have good points noted above?
Thanks!

src/manifests_workflow/input_manifests.py
```diff
@@ -160,3 +171,28 @@ def add_to_cron(self, version: str) -> None:
             f.write(data)
 
         logging.info(f"Wrote {jenkinsfile}")
+
+    def add_to_versionincrement_workflow(self, version: str) -> None:
+        logging.info("Adding new version the version increment workflow")
```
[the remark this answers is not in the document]
prudhvigodithi: Thanks @gaiksaya, updated the sentence.

src/manifests_workflow/input_manifests.py
```diff
@@ -160,3 +171,28 @@ def add_to_cron(self, version: str) -> None:
...
+        yaml = ruamel.yaml.YAML()
+        yaml.explicit_start = True  # type: ignore
+        yaml.preserve_quotes = True  # type: ignore
+
+        with open(versionincrement_workflow_file) as f:
+            data = yaml.load(f)
+
+        version_entry = version.split(".")[0] + ".x"
+        for branch in data["jobs"]["plugin-version-increment-sync"]["strategy"]["matrix"]["branch"]:
+            if version_entry in branch:
```
gaiksaya, line 187: Why only check for `.x` branches?

peterzhuamazon: I dont think it is easy to check non-.x branch right?
We can definitely add a section to keep bumping tho.
But .x is the source of truth on the latest version.

src/manifests_workflow/input_manifests.py
prudhvigodithi: Yes I can @gaiksaya, some helpful info
cookiecutter migration to `ruamel.yaml` https://github.com/cookiecutter/cookiecutter/issues/557
`ruamel.yaml` difference with `PyYAML`: https://yaml.readthedocs.io/en/latest/pyyaml.html#

src/manifests_workflow/input_manifests.py
prudhvigodithi: True, also we want to only version increment the `.x` branches, the plugin teams will anyhow use the `.x` as source of truth to cut the `MAJOR_VERSION.x` branch, by then its already incremented and we can run this workflow on crontab (or with some trigger), that keep checking the `.x` branch, compare with core and increment the version, this way the version is already incremented when the plugin team is ready to cut a release branch.
@gaiksaya

gaiksaya: I see. In case of patch releases however which track `major.minor` branch (not the `.x` branch) I believe we would need to add it manually then for now?

prudhvigodithi: Good point @gaiksaya let me handle the case for `major.minor` branch.

prudhvigodithi: Hey @gaiksaya @peterzhuamazon the change should now handle the `major.minor` branch, can you please review back again?

peterzhuamazon (approved):

prudhvigodithi:
Thanks @peterzhuamazon, I have also just removed ([commit](https://github.com/opensearch-project/opensearch-build/compare/d3964dc786153c22ad11db709020ff935dcc6ab2..76b6280c0982628b9cb058b3af79aee11ec919d9)) `- '2.3'` from the entry as we are already on '2.4', just an overkill to run again for all `2.3` branches

dblock:
I’m late to this PR, but it needs more work even if it seems to work. Please followup / open issues accordingly.

1. What is the ruyaml business here? It looks like there’s now more than one way to read and write YML, please consolidate into one.
2. There are no tests that ensure that the correct values were added to the workflow in the various cases of major/minor/etc., just that the add method was called.

tests/tests_manifests_workflow/test_input_manifests.py
```diff
@@ -122,5 +127,17 @@ def test_add_to_cron(self, mock_open: MagicMock) -> None:
...
+    def test_versionincrement_workflow(self) -> None:
+        self.assertEqual(
+            InputManifests.versionincrement_workflow(),
+            os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "..", ".github", "workflows", "increment-plugin-versions.yml"))
+        )
+
+    @patch("builtins.open", new_callable=mock_open)
+    @patch("manifests_workflow.input_manifests.InputManifests.add_to_versionincrement_workflow")
+    def test_add_to_versionincrement_workflow(self, *mocks: Any) -> None:
+        input_manifests = InputManifests("test")
```
dblock, line 139: This test doesn’t test anything.

src/manifests_workflow/input_manifests.py
```diff
@@ -160,3 +171,31 @@ def add_to_cron(self, version: str) -> None:
...
+    def add_to_versionincrement_workflow(self, version: str) -> None:
+        versionincrement_workflow_file = self.versionincrement_workflow()
+        yaml = ruamel.yaml.YAML()
+        yaml.explicit_start = True  # type: ignore
+        yaml.preserve_quotes = True  # type: ignore
+
+        with open(versionincrement_workflow_file) as f:
+            data = yaml.load(f)
+
+        version_entry = []
```
dblock, line 184: This part is not tested.

src/manifests_workflow/input_manifests.py
dblock, line 14: The code now reads and writes YML files in two different ways. We need to have one.

prudhvigodithi:
Hey @dblock for the new YAML library `ruamel.yaml` I have have added useful information https://github.com/opensearch-project/opensearch-build/pull/2602#discussion_r968867762, please check. I can create an issue to migrate to `ruamel.yaml`

src/manifests_workflow/input_manifests.py
prudhvigodithi: Let me add this as well, I will raise a new PR

prudhvigodithi:
Created a new issue to use `ruamel.yaml` https://github.com/opensearch-project/opensearch-build/issues/2614
@dblock @peterzhuamazon @gaiksaya @bbarani

PR #2636, opened 2022-09-16
Restricting git issue creation to OpenSearch

rishabh6788:
Signed-off-by: Rishabh Singh <sngri@amazon.com>

### Description
As per discussion restricting github issue creation to one component on component failure during distribution build. Once we are sure that the solution is stable will enable for all components. 

### DIFF
--- a/vars/createGithubIssue.groovy
+++ b/vars/createGithubIssue.groovy
@@ -27,7 +27,10 @@ def call(Map args = [:]){
             if (failedComponents.contains(component.name)) {
                 println("Component ${component.name} failed, creating github issue")
                 compIndex = failedComponents.indexOf(component.name)
-                create_issue(component.name, component.repository, currentVersion, failureMessages[compIndex])
+                if (component.name == "OpenSearch"){
+                    create_issue(component.name, component.repository, currentVersion, failureMessages[compIndex])
+                }
+                // create_issue(component.name, component.repository, currentVersion, failureMessages[compIndex])
                 sleep(time:3,unit:"SECONDS")
             }
         }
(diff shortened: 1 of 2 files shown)


vars/createGithubIssue.groovy
```diff
@@ -27,7 +27,10 @@ def call(Map args = [:]){
             if (failedComponents.contains(component.name)) {
                 println("Component ${component.name} failed, creating github issue")
                 compIndex = failedComponents.indexOf(component.name)
-                create_issue(component.name, component.repository, currentVersion, failureMessages[compIndex])
+                if (component.name == "OpenSearch"){
```
prudhvigodithi, line 30: Hey @rishabh6788 thanks for the PR, this will only create issue for a component `OpenSearch`, we might need for other components as well, the ask is this (correct me if i'm wrong), if a component would not want to create an issue upon failure that specific component has to be excluded, the component can `OpenSearch` or any other plugin as well.

gaiksaya: @prudhvigodithi I dont think we should do that. A component is added to the manifest only when they are ready. IF its failing to build, it is good to have an issue open irrespective of wanting it or not.

prudhvigodithi: I'm fine with it, there was an ask from plugin team if there is a way to exclude from creating a ticket, but this PR does not solve that, it only creates issue for `OpenSearch` which is what I was saying @gaiksaya 
@bbarani

gaiksaya: I believe we told them that this wont be happening since we want the issues to be created if the build failed. If the component is not ready just dont add them in the manifest then until you are.

prudhvigodithi: Again as I said, i'm ok that way, just conveying what I heard and reviewing the PR in accordingly ;)

gaiksaya: Wondering if we should keep the failure for some another component. Did OpenSearch ever fail? I saw security failing a bunch of times

rishabh6788: @gaiksaya @prudhvigodithi as discussed removed the condition and enabled issue creation for all components.

prudhvigodithi (approved):

Issue #2066, opened 2022-05-02
Add support for multiple clusters for integration tests (CCR)


dblock:
### Is your feature request related to a problem? Please describe

Coming from https://github.com/opensearch-project/cross-cluster-replication/issues/383#issuecomment-1114747305, CCR needs a multi-node cluster for replication tests. Because of the different scenarios under test with security, CCR ITs create local clusters with specific configurations for each test case. An analysis put that number of different configurations at 7 clusters. This approach was discussed on [this PR](https://github.com/opensearch-project/OpenSearch/pull/1930) where I updated the language around recommendations/practices for testing in OpenSearch.

### Describe the solution you'd like

Need to find a path forward to test at least basic CCR replication functionality for the complete distribution. This could be a "good enough" multi-node cluster available to all plugins to run integration tests.

### Additional context

https://github.com/opensearch-project/opensearch-build/issues/1499
https://github.com/opensearch-project/opensearch-build/issues/1428

prudhvigodithi:
[Triage] Adding @ankitkala, any reason why we cant re-use the performance testing setup that is out there for integration testing?
@peterzhuamazon @gaiksaya

ankitkala:
I think both `LocalTestCluster` and `PerfTestCluster` can be used.  Primary issue is that we only create one test cluster for running the integration tests during the distro build. 
Cross cluster replication requires at least 2 clusters to run the performance tests.

bbarani:
@peterzhuamazon @gaiksaya What would it take to support multiple clusters for CCR integration test? I think we can specifically spin up multiple clusters just for CCR tests as we are running parallely now.

peterzhuamazon:
There are several big changes needs to be done in test framework.
As of now it is assuming only one cluster so discovery is not needed to be configured.
This is not a small task and would require some code diving before we can tell the efforts. Thanks.

dblock:
Practically speaking I would introduce test configuration settings that define the cluster configuration, e.g.:

```yaml
  - name: cross-cluster-replication
    integ-test:
      topology: multi-cluster
      test-configs:
        - with-security
        - without-security
```

or something more elaborate

```yaml
  - name: cross-cluster-replication
    integ-test:
      topology: 
        - name: cluster1
           nodes: 3
        - name: cluster2
           ...
      test-configs:
        - with-security
        - without-security
        - 
```

[...]

PR #2663, opened 2022-09-27
Added support to run integTests, in either single-cluster or multi-cluster mode.

monu-aws:
This feature will allow opensearch-build test workflow to launch single or multiple cluster so that plugins which require multiple clusters can run their tests.

### Description

Parent issue https://github.com/opensearch-project/opensearch-build/issues/2066
issue on CCR repo https://github.com/opensearch-project/cross-cluster-replication/issues/532

PR on CCR repo https://github.com/opensearch-project/cross-cluster-replication/pull/543

This can be set using topology in manifest.
```
  - name: cross-cluster-replication
    integ-test:
      topology: single-cluster
      test-configs:
        - with-security
        - without-security
```
OR
  ```
- name: cross-cluster-replication
    integ-test:
      topology: multi-cluster
      test-configs:
        - with-security
        - without-security
```

Signed-off-by: Monu Singh <msnghgw@amazon.com>

### DIFF
--- a/src/manifests/test_manifest.py
+++ b/src/manifests/test_manifest.py
@@ -61,6 +61,7 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
                         "type": "dict",
                         "schema": {
                             "build-dependencies": {"type": "list"},
+                            "topology": {"type": "string","required": False },
                             "test-configs": {"type": "list", "allowed": ["with-security", "without-security"]},
                             "additional-cluster-configs": {"type": "dict"},
                         },
@@ -137,3 +138,4 @@ TestManifest.VERSIONS = {"1.0": TestManifest}
 
 TestComponent.__test__ = False  # type: ignore[attr-defined]
 TestManifest.__test__ = False  # type: ignore[attr-defined]
+
--- a/src/test_workflow/integ_test/integ_test_suite.py
+++ b/src/test_workflow/integ_test/integ_test_suite.py
@@ -97,6 +97,39 @@ class IntegTestSuite(abc.ABC):
             logging.info(f"{script} does not exist. Skipping integ tests for {self.component.name}")
             return 0
 
+
+    def multi_execute_integtest_sh(self, endpoint1: str, port1:int , endpoint2:str, port2: int, security: bool, test_config: str) -> int:
+            print("integ Test suite opensearch Execute integtest")
+        # here
+            script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
+            if os.path.exists(script):
+                cmd = f"{script} -m {endpoint1} -n {port1} -x {endpoint2} -y {port2}  -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
+                self.repo_work_dir = os.path.join(
+                    self.repo.dir, self.test_config.working_directory) if self.test_config.working_directory is not None else self.repo.dir
+                (status, stdout, stderr) = execute(cmd, self.repo_work_dir, True, False)
+
+                test_result_data = TestResultData(
+                    self.component.name,
+                    test_config,
+                    status,
+                    stdout,
+                    stderr,
+                    self.test_artifact_files
+                )
+
+                self.save_logs.save_test_result_data(test_result_data)
+                print("Inside integ test suite")
+                if stderr:
+                    logging.info("Integration test run failed for component " + self.component.name)
+                    logging.info(stderr)
+                return status
+            else:
+                logging.info(f"{script} does not exist. Skipping integ tests for {self.component.name}")
+                return 0
+
+
+
+	
     def is_security_enabled(self, config: str) -> bool:
         if config in ["with-security", "without-security"]:
             return True if config == "with-security" else False
--- a/src/test_workflow/integ_test/integ_test_suite_opensearch.py
+++ b/src/test_workflow/integ_test/integ_test_suite_opensearch.py
@@ -71,6 +71,52 @@ class IntegTestSuiteOpenSearch(IntegTestSuite):
         if "additional-cluster-configs" in self.test_config.integ_test.keys():
             self.additional_cluster_config = self.test_config.integ_test.get("additional-cluster-configs")
             logging.info(f"Additional config found: {self.additional_cluster_config}")
+
+
+        if "topology" in self.test_config.integ_test.keys():
+            self.topology = self.test_config.integ_test.get("topology")
+            logging.info(f"topology config found: {self.topology}")
+            if(self.topology == "multi-cluster"):
+                        logging.info(f"Will run multi test cluster")
+                        if self.additional_cluster_config is None:
+                          self.additional_cluster_config={"cluster.name":"opensearch"}
+                        self.additional_cluster_config['cluster.name']="opensearch"
+                        self.additional_cluster_config['http.port']=9200
+                        logging.info(f"Additional config found: {self.additional_cluster_config}")
+                        print("Creating cluster 1")
+
+                        with LocalTestCluster.create(
+                            self.dependency_installer,
+                            self.work_dir,
+                            self.component.name,
+                            self.additional_cluster_config,
+                            self.bundle_manifest,
+                            security,
+                            config,
+                            self.test_recorder,
+                             9200,
+                        ) as (endpoint1, port1):
+                          print("Local cluster1 returned : ", endpoint1, " " ,port1)
+                          self.additional_cluster_config['cluster.name']="follower"
+                          self.additional_cluster_config['http.port']=9201
+                          logging.info(f"Additional config found: {self.additional_cluster_config}")
+                          with LocalTestCluster.create(
+                              self.dependency_installer,
+                              self.work_dir,
+                              self.component.name,
+                              self.additional_cluster_config,
+                              self.bundle_manifest,
+                              security,
+                              config,
+                              self.test_recorder,
+                              9201
+                          ) as (endpoint2 ,port2):
+                              print("Local cluster returned : ", endpoint2," ",port2)
+                              os.chdir(self.work_dir)
+                              self.pretty_print_message("Running integration tests for " + self.component.name)
+                              return self.multi_execute_integtest_sh(endpoint1, port1,endpoint2, port2, security, config)
+
+
         with LocalTestCluster.create(
             self.dependency_installer,
             self.work_dir,
@@ -79,7 +125,7 @@ class IntegTestSuiteOpenSearch(IntegTestSuite):
             self.bundle_manifest,
             security,
             config,
-            self.test_recorder,
+            self.test_recorder
         ) as (endpoint, port):
             self.pretty_print_message("Running integration tests for " + self.component.name)
             os.chdir(self.work_dir)
--- a/src/test_workflow/integ_test/local_test_cluster.py
+++ b/src/test_workflow/integ_test/local_test_cluster.py
@@ -34,6 +34,7 @@ class LocalTestCluster(TestCluster):
         security_enabled: bool,
         component_test_config: str,
         test_recorder: TestRecorder,
+		xport=9200
     ) -> None:
         super().__init__(
             work_dir,
@@ -41,9 +42,11 @@ class LocalTestCluster(TestCluster):
             component_test_config,
             security_enabled,
             additional_cluster_config,
-            test_recorder.local_cluster_logs
+            test_recorder.local_cluster_logs,
+			xport
         )
 
+        self.xport=xport
         self.manifest = bundle_manifest
         self.dependency_installer = dependency_installer
 
@@ -53,7 +56,8 @@ class LocalTestCluster(TestCluster):
             self.additional_cluster_config,
             self.security_enabled,
             self.dependency_installer,
-            self.work_dir
+            self.work_dir,
+			self.xport
         )
 
     @property
@@ -66,4 +70,4 @@ class LocalTestCluster(TestCluster):
 
     @property
     def port(self) -> int:
-        return 9200
+        return self.xport
--- a/src/test_workflow/integ_test/service_opensearch.py
+++ b/src/test_workflow/integ_test/service_opensearch.py
@@ -31,10 +31,11 @@ class ServiceOpenSearch(Service):
         additional_config: dict,
         security_enabled: bool,
         dependency_installer: DependencyInstaller,
-        work_dir: str
+        work_dir: str,
+		xport=9200
     ) -> None:
         super().__init__(work_dir, version, distribution, security_enabled, additional_config, dependency_installer)
-
+        self.xport=xport
         self.dist = Distributions.get_distribution("opensearch", distribution, version, work_dir)
         self.dependency_installer = dependency_installer
         self.install_dir = self.dist.install_dir
@@ -70,7 +71,7 @@ class ServiceOpenSearch(Service):
             yamlfile.write(yaml.dump(additional_config))
 
     def port(self) -> int:
-        return 9200
+        return self.xport
 
     def check_service_response_text(self, response_text: str) -> bool:
         return ('"status":"green"' in response_text) or ('"status":"yellow"' in response_text)
--- a/src/test_workflow/test_cluster.py
+++ b/src/test_workflow/test_cluster.py
@@ -36,9 +36,10 @@ class TestCluster(abc.ABC):
         component_test_config: str,
         security_enabled: bool,
         additional_cluster_config: dict,
-        save_logs: LogRecorder
+        save_logs: LogRecorder,
+		xport=9200
     ) -> None:
-        self.work_dir = os.path.join(work_dir, "local-test-cluster")
+        self.work_dir = os.path.join(work_dir, "local-test-cluster"+f"{xport}")
         self.component_name = component_name
         self.component_test_config = component_test_config
         self.security_enabled = security_enabled
@@ -60,6 +61,7 @@ class TestCluster(abc.ABC):
             cluster.start()
             yield cluster.endpoint, cluster.port
         finally:
+            print("Terminating service")
             cluster.terminate()
 
     def start(self) -> None:

### CODE REVIEW

File: src/test_workflow/integ_test/integ_test_suite.py [hunk 1]
```diff
@@ -97,6 +97,31 @@ def execute_integtest_sh(self, endpoint: str, port: int, security: bool, test_co
             logging.info(f"{script} does not exist. Skipping integ tests for {self.component.name}")
             return 0
 
+    def multi_execute_integtest_sh(self, endpoint1: str, port1: int, endpoint2: str, port2: int, security: bool, test_config: str) -> int:
```
peterzhuamazon, line 100: Are we assuming there is only 2 endpoints?

src/test_workflow/integ_test/local_test_cluster.py
```diff
@@ -34,16 +34,19 @@ def __init__(
...
     ) -> None:
         super().__init__(
             work_dir,
             component_name,
             component_test_config,
             security_enabled,
             additional_cluster_config,
-            test_recorder.local_cluster_logs
+            test_recorder.local_cluster_logs,
+            xport
```
peterzhuamazon, line 46: Can we have a better naming convention than `xport`?

peterzhuamazon (changes requested):
Hi @monu-aws,

Seems like the code is assuming there is maximum 2 clusters?
It is better not to hardcode as 9200 vs 9201 but be flexible.

I suggest we set a max number of cluster available and a max number of nodes available in the test manifest, tho the latter can be implemented at a later stage.

Cluster can start with 9200, 9210, 9220......
While the nodes in the cluster can be (9201, 9202...), (9211, 9212...), (9221, 9222...).

And `multi_execute_integtest_sh` is having way too many similar/same code compares to `execute_integtest_sh`. I suggest we combine the both and introduce new params to resolve # of endpoints, # of ports, etc.

Also, there is no tests involved here. 

Thanks.

peterzhuamazon:
I convert this PR back to draft before we clear it up.

Thanks.

src/test_workflow/integ_test/integ_test_suite.py
monu-aws: Yes, CCR is the only plugin requiring more than one cluster. Since no other plugin uses 

Creating n number of clusters will be an overkill, since we have to refactor a lot of things, the parameters have to be passed dynamically. Since no other plugin/team is using this we are going ahead with just 2 clusters. If you would like, we can rename it to cross_cluster (topology) as we need 2 clusters.

src/test_workflow/integ_test/integ_test_suite_opensearch.py
```diff
@@ -71,6 +71,43 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
...
+                        security,
+                        config,
+                        self.test_recorder,
+                        9201
+                    ) as (endpoint2, port2):
+                        os.chdir(self.work_dir)
+                        self.pretty_print_message("Running integration tests for " + self.component.name)
+                        cluster_endpoints = [{"endpoint": endpoint1, "port": port1}, {"endpoint": endpoint2, "port": port2}]
+                        return self.multi_execute_integtest_sh(cluster_endpoints, security, config)
         with LocalTestCluster.create(
```
dblock, line 111: Looks like this cluster is still created even if topology is specified?

dblock (changes requested):
Add a new class called Topology that abstracts single and multi-cluster instead of the big `if`. Then in test suite write:

```python
with Topology.create(self.topology):
 ....
```

src/test_workflow/integ_test/integ_test_suite.py
peterzhuamazon: We should not assume CCR is the only plugin therefore hardcode the numbers.
Dashboards will soon to use the same functionalities with more than 2 multinode required.

src/test_workflow/integ_test/integ_test_suite_opensearch.py
monu-aws: no, the function returns after the tests are completed.

dblock: Oh I see. It's hard to read. See my refactoring ask below.

src/test_workflow/integ_test/topology.py
```diff
@@ -0,0 +1,58 @@
...
+from test_workflow.dependency_installer import DependencyInstaller
+from test_workflow.integ_test.local_test_cluster import LocalTestCluster
+from test_workflow.test_recorder.test_recorder import TestRecorder
+
+
+class Topology:
+
+    @classmethod
+    @contextmanager
+    def create(
```
dblock, line 22: So a Topology is something that has a collection of clusters? Add an initializer to `Topology` that holds `clusters`, `endpoints_list` and `original_config` fields, it will make this code a lot easier to read and to test.

src/manifests/test_manifest.py
```diff
@@ -62,6 +62,7 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
                         "type": "dict",
                         "schema": {
                             "build-dependencies": {"type": "list"},
+                            "topology": {"type": "integer", "required": False},
```
dblock, line 65: I recommend future-proofing this:

```
topology:
    nodes: 2
```

src/test_workflow/integ_test/integ_test_suite_opensearch.py
monu-aws: Made relevant class to address this.

src/test_workflow/integ_test/integ_test_suite.py
monu-aws: Added support to create multiple clusters by specifying the number in topology.

src/test_workflow/integ_test/local_test_cluster.py
monu-aws: Renamed

src/manifests/test_manifest.py
ankitkala: +1 . Let's finalize the topology definition even if we're not planning to add support for multi-node clusters. 
Let's also standardize how we'll pass these clusters endpoints as generic params to the `integtest.sh`. 

Maybe define topology something like: 
```
topology: 
    cluster1:
        data_nodes: 1
        master_nodes: 0
    cluster2:
        data_nodes: 1
        master_nodes: 0
```
Since we only support create single node clusters as of now, maybe let's add an assertion for that. 


For cluster endpoints, let's pass it as a json maybe. Somthing like: 
```
{
    "cluster1": {
        "data_nodes": [
             {
                 "endpoint": "localhost",
                 "http_port": 9200,
                 "transport_port": 9300
              },
             {
                 "endpoint": "localhost",
                 "http_port": 9201,
                 "transport_port": 9301
              }
         ]
        "master_nodes": [
             {
                 "endpoint": "localhost",
                 "http_port": 9202,
                 "transport_port": 9302
              }
        ]
    },
    "cluster2": {
        "data_nodes": [
             {
                 "endpoint": "localhost",
                 "http_port": 9203,
                 "transport_port": 9303
              },
             {
                 "endpoint": "localhost",
                 "http_port": 9204,
                 "transport_port": 9304
              }
         ]
        "master_nodes": [
             {
                 "endpoint": "localhost",
                 "http_port": 9205,
                 "transport_port": 9305
              }
        ]
    }
}

```

monu-aws: Sure, I will work on finalizing the Topology and the way endpoints are passed.

src/test_workflow/integ_test/topology.py
monu-aws: Adding this support

src/test_workflow/integ_test/topology.py
```diff
@@ -0,0 +1,79 @@
...
+        number_of_cluster: int,
+        dependency_installer: DependencyInstaller,
+        work_dir: Path,
+        component_name: str,
+        additional_cluster_config: dict,
+        bundle_manifest: BundleManifest,
+        security_enabled: bool,
+        component_test_config: str,
+        test_recorder: TestRecorder
+    ) -> None:
```
dblock, line 32: Most of these parameters aren't needed for a topology to exist, but are only needed for executing tests, so it should be broken up. 

With the refactor suggested above by @kotwanikunal you will be getting a `InputManifest.TestTopology` from `manifest....topology` with a bunch of information such as the number of nodes. So this class should only take that as input. Another method `create` would start the cluster and manage its shutdown. And another method will execute tests, taking as input whatever it needs to execute.

src/test_workflow/integ_test/topology.py
monu-aws: Done

src/manifests/test_manifest.py
monu-aws: Added this support

monu-aws (commented):
- Created warpper for multi_execute_integtest_sh
- Added support to support for multi cluster( more than 2 too)
- Created Topology class
- Test for Topology class written

src/manifests/test_manifest.py
```diff
@@ -62,6 +62,19 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
...
+                            "topology": {
+                                "type": "list", "required": False,
+                                "schema": {
+                                    "type": "dict",
+                                    "schema": {
+                                        "cluster_name": {"required": True, "type": "string"},
+                                        "data_nodes": {"type": "integer", "required": True},
+                                        "master_nodes": {"type": "integer"}
+                                    }
+
```
dblock, line 74: Remove empty lines.

src/manifests/test_manifest.py
```diff
@@ -62,6 +62,19 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
...
                         "schema": {
                             "build-dependencies": {"type": "list"},
+                            "topology": {
+                                "type": "list", "required": False,
+                                "schema": {
+                                    "type": "dict",
+                                    "schema": {
+                                        "cluster_name": {"required": True, "type": "string"},
+                                        "data_nodes": {"type": "integer", "required": True},
+                                        "master_nodes": {"type": "integer"}
```
dblock, line 72: The master terminology is deprecated in 3.0, any way we can make this `cluster_manager_nodes` now?

src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -73,14 +74,20 @@ def __init__(
     def execute_tests(self) -> TestComponentResults:
         pass
 
-    def execute_integtest_sh(self, endpoint: str, port: int, security: bool, test_config: str) -> int:
+    def multi_execute_integtest_sh(self, cluster_endpoints: list, security: bool, test_config: str) -> int:
         script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
         if os.path.exists(script):
-            cmd = f"{script} -b {endpoint} -p {port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
+            if len(cluster_endpoints) == 1:
+                single_data_node = cluster_endpoints[0]['data_nodes'][0]
```
dblock, line 81: This should be a list of first class citizens of type `ClusterEndpoint`, and the code would say `next(next(cluster_endpoints).data_nodes)` or `cluster_endpoints[0].data_nodes[0]`.

src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -73,14 +74,20 @@ def __init__(
     def execute_tests(self) -> TestComponentResults:
         pass
 
-    def execute_integtest_sh(self, endpoint: str, port: int, security: bool, test_config: str) -> int:
+    def multi_execute_integtest_sh(self, cluster_endpoints: list, security: bool, test_config: str) -> int:
```
dblock, line 77: `cluster_endpoints` should be a `List` of `ClusterEndpoint`.

src/test_workflow/integ_test/integ_test_suite_opensearch.py
```diff
@@ -72,19 +73,26 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
         if "additional-cluster-configs" in self.test_config.integ_test.keys():
             self.additional_cluster_config = self.test_config.integ_test.get("additional-cluster-configs")
             logging.info(f"Additional config found: {self.additional_cluster_config}")
-        with LocalTestCluster.create(
+        if self.additional_cluster_config is None:
+            self.additional_cluster_config = {"cluster.name": "opensearch1"}
+        if "topology" in self.test_config.integ_test.keys():
+            self.topology = self.test_config.integ_test.get("topology")
+        else:
+            self.topology = [{'cluster_name': 'cluster1', 'data_nodes': 1, 'master_nodes': 0}]
```
dblock, line 81: The `self.topology` property should return an instance of a class similar to `InputManifest.*` classes and all JSON manipulation self-contained in there. For example https://github.com/opensearch-project/opensearch-build/blob/main/src/manifests/input_manifest.py#L126.

dblock (changes requested):
This remains non-object-oriented for about half of this.

1. The test manifest should return a first class `TestManifest.Topology` instance with all the properties as first class citizens, such as data or cluster manager nodes.
2. A running test Topology takes (1) as input and boots itself. This is almost done here.
3. Need tests for all these classes separately.

Please also fix DCO and make sure everything is green.

src/manifests/test_manifest.py
monu-aws: Sure, made the changes.

src/manifests/test_manifest.py
monu-aws: Done

src/manifests/test_manifest.py
```diff
@@ -117,12 +130,21 @@ def __create__(self, data: dict) -> 'TestComponent':
         return TestComponent(data)
 
 
+class TestComponentTopology:
+    def __init__(self, data: dict):
+        if data is not None:
+            self.cluster_configs = data
+        else:
+            self.cluster_configs = [{'cluster_name': 'cluster1', 'data_nodes': 1, 'manager_nodes': 0}]  # type: ignore[assignment]
```
dblock, line 138: Continue with the class hierarchy just like for `TestComponentTopology`. Add a `ClusterConfig` class that has properties such as `cluster_name` and `data_nodes`, so that the caller doesn't have to deal with JSON.

src/test_workflow/integ_test/topology.py
```diff
@@ -0,0 +1,109 @@
...
+        self,
+        topology_config: list,
+        dependency_installer: DependencyInstaller,
+        work_dir: Path,
+        component_name: str,
+        additional_cluster_config: dict,
+        bundle_manifest: BundleManifest,
+        security_enabled: bool,
+        component_test_config: str,
+        test_recorder: TestRecorder
```
dblock, line 31: Half of these fields do not belong in `Topology`. Things that are properties of a topology (e.g. topology configuration) should be parameters to `init`, and things that are needed to `create_clusters` should be parameters to `create` and then `create_clusters`.

src/test_workflow/integ_test/topology.py
```diff
@@ -0,0 +1,109 @@
...
+        current_work_dir_num = 1
+        for cluster_config in self.topology_config:
+
+            data_nodes: list = []
+            manager_nodes: list = []
+            data_nodes_endpoints: list = []
+            manager_nodes_endpoints: list = []
+
+#             Create data nodes
+            for i in range(cluster_config['data_nodes']):
```
dblock, line 68: For each data node you create a cluster? I thought that a cluster is a set of data nodes.

src/test_workflow/integ_test/topology.py
```diff
@@ -0,0 +1,109 @@
...
+                current_port = current_port + 1
+                current_work_dir_num = current_work_dir_num + 1
+
+#           Create manager nodes
+#           for i in range (cluster_config['manager_nodes']):
+            self.clusters.append({"cluster_name": cluster_config['cluster_name'], "data_nodes": data_nodes, "manager_nodes": manager_nodes})
+
+            endpoints_list.append(ClusterEndpoint(cluster_config['cluster_name'], data_nodes_endpoints, manager_nodes_endpoints))
+            current_transport_port = current_transport_port + 1
+        return endpoints_list
```
dblock, line 94: This code is very hard to understand (and probably has many bugs). As far as I can read, a cluster is a set of nodes of different type (manager, data). Each node has an endpoint. Each one of these things needs to become a class, like `TestCluster` (already exists) and `Node` with `DataNode` and `ManagerNode` as child classes. This loop would live inside `TestCluster.create(cluster_config)`, iterate over `cluster_config.data_nodes` and call `Node.create(config)` for example, then add the node to `TestCluster.nodes`. The function should return the cluster. If one needs the list of endpoints, call `TestCluster.endpoints` that can iterate over its nodes to collect endpoints.

src/test_workflow/integ_test/topology.py
```diff
@@ -0,0 +1,109 @@
...
+                    os.path.join(self.work_dir, f'{current_work_dir_num}'),
+                    self.component_name,
+                    original_config,
+                    self.bundle_manifest,
+                    self.security_enabled,
+                    self.component_test_config,
+                    self.test_recorder,
+                    cluster_port)
+                data_nodes.append(cluster)
+                data_nodes_endpoints.append({"endpoint": cluster.endpoint, "port": cluster.port, "transport": current_transport_port})
```
dblock, line 83: This is a Node, and needs to be a class of its own with `endpoint`, `port` and `transport` properties.

dblock (approved):
I'd be ok merging this assuming all tests pass, and if you take on refactoring it as a quick followup, because it's still pretty far from proper object-oriented design. Every "thing" that we have here: a test cluster, a node, a configuration, need to be classes with properties and represent the "thing". Then you can assemble the creation of the cluster from these classes.

src/manifests/test_manifest.py
```diff
@@ -62,6 +62,19 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
...
                         "schema": {
                             "build-dependencies": {"type": "list"},
+                            "topology": {
+                                "type": "list", "required": False,
+                                "schema": {
+                                    "type": "dict",
+                                    "schema": {
+                                        "cluster_name": {"required": True, "type": "string"},
+                                        "data_nodes": {"type": "integer", "required": True},
+                                        "manager_nodes": {"type": "integer"}
```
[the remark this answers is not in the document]
monu-aws: Done

src/manifests/test_manifest.py
```diff
@@ -117,12 +130,30 @@ def __create__(self, data: dict) -> 'TestComponent':
...
+
+
+class TestComponentTopology:
+    def __init__(self, data: dict):
+        if data is not None:
+            self.cluster_configs = []
+            for cluster_config_data in data:
+                self.cluster_configs.append(ClusterConfig(cluster_config_data))
+        else:
+            self.cluster_configs = [ClusterConfig({'cluster_name': 'cluster1', 'data_nodes': 1, 'cluster_manager_nodes': 0})]  # type: ignore[assignment]
```
dblock, line 147: This is the wrong place to provide a default, it should be in `ClusterConfig#__init__` (see above).

src/manifests/test_manifest.py
```diff
@@ -117,12 +130,30 @@ def __create__(self, data: dict) -> 'TestComponent':
         return TestComponent(data)
 
 
+class ClusterConfig:
+    def __init__(self, cluster_config_data: dict):
```
dblock, line 134: Make `cluster_config_data` just `data` and default it to null. 

Then 

```
self.cluster_name = data['cluster_name'] if data else 'cluster1'
...
```

src/manifests/test_manifest.py
```diff
@@ -117,12 +130,30 @@ def __create__(self, data: dict) -> 'TestComponent':
...
+        self.cluster_name = cluster_config_data['cluster_name']
+        self.data_nodes = cluster_config_data['data_nodes']
+        self.cluster_manager_nodes = cluster_config_data['cluster_manager_nodes']
+
+
+class TestComponentTopology:
+    def __init__(self, data: dict):
+        if data is not None:
```
dblock, line 142: Add a class called `ClusterConfigs` that is a collection and takes `data` as an argument and does the iteration over the configs.

src/test_workflow/integ_test/topology.py
```diff
@@ -0,0 +1,182 @@
...
+    def destroy(self) -> None:
+        for topology_cluster in self.clusters:
+            for datanode in topology_cluster.data_nodes:
+                datanode.cluster.terminate()
+            for managernode in topology_cluster.cluster_manager_nodes:
+                managernode.cluster.terminate()
+        self.clusters = []
+
+
+class DataNode:
```
dblock, line 102: Move these out of this file or make them inner classes of `Topology`.

dblock (commented):
I'll keep doing CRs and make further suggestions of refactoring, but still ok to merge on 🍏 .

src/manifests/test_manifest.py
```diff
@@ -117,12 +130,31 @@ def __create__(self, data: dict) -> 'TestComponent':
         return TestComponent(data)
 
 
+class ClusterConfig:
+    def __init__(self, cluster_config_data: dict):
+        self.cluster_name = cluster_config_data['cluster_name']
+        self.data_nodes = cluster_config_data['data_nodes']
+        self.cluster_manager_nodes = cluster_config_data['cluster_manager_nodes'] if "cluster_manager_nodes" in cluster_config_data.keys() else 0
+        assert self.cluster_manager_nodes == 0, "Cluster manager nodes not supported ,use value 0 or skip this parma"
```
dblock, line 138: Please fix typos, space before/after ,

src/manifests/test_manifest.py
```diff
@@ -117,12 +130,31 @@ def __create__(self, data: dict) -> 'TestComponent':
...
+            self.cluster_configs = [ClusterConfig({'cluster_name': 'cluster1', 'data_nodes': 1, 'cluster_manager_nodes': 0})]  # type: ignore[assignment]
+
+
 class TestComponent(Component):
     def __init__(self, data: dict) -> None:
         super().__init__(data)
         self.working_directory = data.get("working-directory", None)
         self.integ_test = data.get("integ-test", None)
         self.bwc_test = data.get("bwc-test", None)
+        self.topology = TestComponentTopology(self.integ_test.get("topology"))
```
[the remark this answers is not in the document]
monu-aws: Done

src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -73,14 +75,26 @@ def __init__(
...
         script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
+
+        def custom_node_endpoint_encoder(node_endpoint: NodeEndpoint) -> dict:
+            return {"endpoint": node_endpoint.endpoint, "port": node_endpoint.port, "transport": node_endpoint.transport}
         if os.path.exists(script):
-            cmd = f"bash {script} -b {endpoint} -p {port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
+            if len(cluster_endpoints) == 1:
+                single_data_node = cluster_endpoints[0].data_nodes[0]
+                cmd = cmd = f"bash {script} -b {single_data_node.endpoint} -p {single_data_node.port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
```
peterzhuamazon, line 86: `cmd = cmd =`?

src/manifests/test_manifest.py
```diff
@@ -62,6 +62,19 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
...
+                                    "type": "dict",
+                                    "schema": {
+                                        "cluster_name": {"required": True, "type": "string"},
+                                        "data_nodes": {"type": "integer", "required": True},
+                                        "cluster_manager_nodes": {"type": "integer"}
+                                    }
+
+                                }
+
+                            },
```
peterzhuamazon, line 77: Does this support older version of test manifest?
Could you provide example of the new manifest as part of the pytest?
Thanks

src/test_workflow/integ_test/topology.py
```diff
@@ -0,0 +1,187 @@
...
+        test_recorder: TestRecorder,
+        dependency_installer: DependencyInstaller,
+        work_dir: Path
+    ):
+        self.cluster_name = cluster_name
+        self.data_nodes: list = []
+        self.cluster_manager_nodes: list = []
+        additional_cluster_config['cluster.name'] = "opensearch" + cluster_config.cluster_name
+
+        for i in range(cluster_config.data_nodes):
```
peterzhuamazon, line 162: Is there any limitation on the number of nodes to be added?
Ex: 9200+100 will be 9300 so I assume the limit is 99?
Thanks.

peterzhuamazon:
Thanks @monu-aws for the changes on the PR.

Some questions in the comments here.

Also, what would be the behavior if user want to test dashboards?

Is that limit to 1 OS 1 OSD or allowing multiOS + 1 OSD?

As the topology is only for opensearch it seems.

Thanks.

src/test_workflow/integ_test/topology.py
monu-aws: yes, total number of nodes is 100

src/test_workflow/integ_test/integ_test_suite.py
monu-aws: Corrected this

src/manifests/test_manifest.py
monu-aws: Yes older version of test_manifest is supported, the Topology config is optional and defaults to one data node is not configured.

src/test_workflow/integ_test/topology.py
peterzhuamazon: Hi @monu-aws do you have a limit already setup in your code?
Or you are expecting people to have good intention and dont add more than 100 nodes?

monu-aws: Added this assertion.

peterzhuamazon (approved):

peterzhuamazon:
Hi @monu-aws please fix DCO so we can merge this.
Great work!

Thanks.

monu-aws:
Requesting merge

peterzhuamazon (approved):

peterzhuamazon:
Hi @monu-aws we will merge this right now.
Appreciate all the work and being patient with the review.

Thanks.

dblock:
@monu-aws @peterzhuamazon Can we please fast-follow and fix the code from my comments such as https://github.com/opensearch-project/opensearch-build/pull/2663/files#r997163120, not necessarily all together?

peterzhuamazon:
@monu-aws this breaks the ci workflow here:
https://github.com/opensearch-project/opensearch-build/actions/runs/3381138172/jobs/5615730409

```

$ ./ci.sh manifests/1.3.2/opensearch-dashboards-1.3.2-test.yml --snapshot
Installing dependencies from Pipfile.lock (b36c9c)...
Running ./src/run_ci.py manifests/1.3.2/opensearch-dashboards-1.3.2-test.yml --snapshot ...
  File "./src/run_ci.py", line 25, in <module>
    sys.exit(main())
  File "./src/run_ci.py", line 20, in main
    CiManifests.from_file(args.manifest, args).check()
  File "/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/opensearch-build/src/ci_workflow/ci_manifests.py", line 28, in from_file
    return cls.__klass(file.name)(file, args)
  File "/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/opensearch-build/src/ci_workflow/ci_test_manifest.py", line 19, in __init__
    super().__init__(TestManifest.from_file(file), args)
  File "/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/opensearch-build/src/manifests/manifest.py", line 36, in from_file
    return loader(yml)
  File "/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/opensearch-build/src/manifests/test_manifest.py", line 98, in __init__
    self.components = TestComponents(data.get("components", []))  # type: ignore[assignment]
  File "/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/opensearch-build/src/manifests/component_manifest.py", line 45, in __init__
    super().__init__(map(create_component, data))
  File "/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/opensearch-build/src/manifests/component_manifest.py", line 44, in <lambda>
    create_component: Callable[[Any], Tuple[str, ComponentType]] = lambda component: (component["name"], self.__create__(component))
  File "/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/opensearch-build/src/manifests/test_manifest.py", line 130, in __create__
    return TestComponent(data)
  File "/Users/zhujiaxi/amazonworks/sourcecodes/peterzhuamazon/opensearch-build/src/manifests/test_manifest.py", line 161, in __init__
    self.topology = TestComponentTopology(self.integ_test.get("topology", None))
AttributeError: 'NoneType' object has no attribute 'get'
```

peterzhuamazon:
1.3.1 test manifest success but 1.3.2 failed:

```
$ diff -bi manifests/1.3.1/opensearch-dashboards-1.3.1-test.yml manifests/1.3.2/opensearch-dashboards-1.3.2-test.yml
6c6
<     name: opensearchstaging/ci-runner:ci-runner-rockylinux8-opensearch-dashboards-integtest-v1
---
>     name: opensearchstaging/ci-runner:ci-runner-rockylinux8-opensearch-dashboards-integtest-v2
7a8,12
>   - name: OpenSearch-Dashboards
>     bwc-test:
>       test-configs:
>         - with-security
>         - without-security
```

peterzhuamazon:
Fixed in https://github.com/opensearch-project/opensearch-build/pull/2842.

PR #2704, opened 2022-10-06
Opensearch security plugin windows install certs demo config and add batch install script

peterzhuamazon:
### Description
Add windows batch file and resolve security demo config temporarily.
Follow up: https://github.com/opensearch-project/security/issues/2148

### Issues Resolved
https://github.com/opensearch-project/security/issues/2122


peternied:
Quoting DB from https://github.com/opensearch-project/opensearch-build/pull/2061
> Shouldn't you be committing this [scripts/components/security/install.sh] into the security plugin repo instead in the right branch?

Work around looks good to me - please work with @scrawfor99 if you'd like help staging things between here and the security repository

peterzhuamazon:
> Quoting DB from #2061
> 
> > Shouldn't you be committing this [scripts/components/security/install.sh] into the security plugin repo instead in the right branch?
> 
> Work around looks good to me - please work with @scrawfor99 if you'd like help staging things between here and the security repository

Hi,
Glad you bring up this question. As this PR is a completely different situation compares to the PR you raised in #2061.

I would suggest you to check these categories within the build repo code base a bit more and it should help you understand the reasoning better. Here are some few points that might interest you.

**1. Check script finder class on how the install.sh being allocated.**
(Yes, install.sh will not be allocated from the component repo directly)

**2. Check dist and bundle code within assemble workflow to understand why it cannot be saved in plugin repositories. Understand the differences between build.sh and install.sh and how are they being processed.**
(Build requires all components to be cloned thus scripts can exist in separate repos, but install will allocate only the artifacts on the build machines without the need to allocate other repos, or it will be a heavy operation.)

Additionally if you have time available I would suggest you contribute the functionalities to call install.sh from separate repositories if you would like this feature to be added. Bare in mind, this is a heavy operation and as of current state it is not required to be used for the assemble workflow.

Thanks.

peterzhuamazon:
> 
> Work around looks good to me - please work with @scrawfor99 if you'd like help staging things between here and the security repository

Sure will work with @scrawfor99 when needed.
Thanks.

prudhvigodithi (approved):

peterzhuamazon:
Will force merge for now as the groovy test failure is not related to my code in this PR.
More information here: https://github.com/opensearch-project/opensearch-build-libraries/issues/7#issuecomment-1273835143

Thanks.

Issue #2712, opened 2022-10-10
Validate the artifact(s) download URL's


bbarani:
Automate the validation of download Url's for OpenSearch and Opensearch-Dashboards during Opensearch releases.

-Verifies if the Tar, Rpm and Yum distributions can be downloaded from the Opensearch website.

Divyaasm:
Closed by https://github.com/opensearch-project/opensearch-build/pull/2725

Issue #2524, opened 2022-08-29
Add automated validation during releases


bbarani:
### Is your feature request related to a problem? Please describe

We are currently doing manual validation for verifying the artifacts, download links and signatures during release process. This is taking additional time and adding manual overhead during release process.

### Describe the solution you'd like

Automate the below validation and verification process during OpenSearch releases,

- Validate the OpenSearch distribution artifacts on DockerHub and ECR
- Validate the artifact(s) download URL's 
- Validate the artifact(s) signatures
- Validate native plugin installation
- Verify the artifacts in Maven

gaiksaya:
Proposal:

I believe it is better to have a validation framework (something similar to https://github.com/opensearch-project/opensearch-build/blob/main/src/sign_workflow/signer.py) with abstract methods like download, install, start_cluster, run_apis, etc.
This will allow us to have a uniform execution mechanism across different platforms like docker, rpm, tarballs, windows, macos, etc. 
This framework then can be wrapped by a jenkins build library which can be called in CI workflows to run validation after each release.

PR #2725, opened 2022-10-12
Add Validation Framework

Divyaasm:
### Description
Validation of Download Url's for OpenSearch.
GitHub issue :https://github.com/opensearch-project/opensearch-build/issues/2712

### Issues Resolved
Automated the downloads of OpenSearch artifacts related to Linux platform.


gaiksaya:
Hi @Divyaasm ,

Please check the approach posted here: https://github.com/opensearch-project/opensearch-build/issues/2524#issuecomment-1275426246
Thanks!

src/validation_workflow/download_utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+import requests
+
+
+class DownloadUtils:
+    @staticmethod
+    def is_url_valid(url: str) -> int:
+        response = requests.head(url)
+        if response.status_code == 200:
+            return 'content-length' in response.headers
```
gaiksaya, line 17: Why are we returning the 'content-length' here?

src/validation_workflow/yum.py
```diff
@@ -0,0 +1,26 @@
...
+
+class Yum(Validation, DownloadUtils):
+    @classmethod
+    def get_architectures(cls) -> list:
+        pass
+
+    @classmethod
+    def download_urls(cls, projects: list, version: str, platform: str) -> None:
+        for project in projects:
+            url = f"https://artifacts.opensearch.org/releases/bundle/{project}/2.x/{project}-2.x.repo"
```
gaiksaya, line 20: This is won't be true everytime. For example, for 1.x version this static link will point to `1.x`. Calculate it using the user entered `version` value. For example, for 3.0.0 version the url will be `3.x`
cc: @peterzhuamazon

src/validation_workflow/yum.py
```diff
@@ -0,0 +1,26 @@
...
+            if Yum.is_url_valid(url):
+                print(f"Valid URL - {url}")
```
gaiksaya, line 22: Not sure if we should just log the statements and return true instead of printing them out.

gaiksaya (changes requested):
Please have proper PR title like"Add validation framework".
Also link the actually Github issue in the description.

src/validation_workflow/yum.py
Divyaasm: I will work on the changes requested Sayali.Thanks

prudhvigodithi: Something like this https://github.com/opensearch-project/opensearch-build/blob/main/src/release_notes_workflow/release_notes.py#L49

src/validation_workflow/tar.py
```diff
@@ -0,0 +1,27 @@
...
+# compatible open source license.
+
+from .download_utils import DownloadUtils
+from .validation import Validation
+
+
+class Tar(Validation, DownloadUtils):
+    @classmethod
+    def get_architectures(cls) -> list:
+        return ["x64", "arm64"]
```
peterzhuamazon, line 15: I see this has been returned every .py which does not seems right.
If this is the same for every child class this should exist in the parent class.

Also this does not seems to belongs to a classmethod but more like a property.

src/validation_workflow/validation_args.py
```diff
@@ -0,0 +1,51 @@
...
+from .rpm import Rpm
+from .tar import Tar
+from .yum import Yum
+
+
+class ValidationArgs:
+    SUPPORTED_DISTRIBUTIONS = ["tar", "yum", "rpm"]
+    SUPPORTED_PLATFORMS = ["linux"]
+    DISTRIBUTION_MAP = {"tar": Tar, "yum": Yum, "rpm": Rpm}
+
```
peterzhuamazon, line 19: A better practice is to write them into multiple lines instead of one line, for better visual or better maintenance.

src/validation_workflow/validation_args.py
```diff
@@ -0,0 +1,51 @@
...
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import argparse
+
+from .rpm import Rpm
+from .tar import Tar
+from .yum import Yum
```
peterzhuamazon, line 12: Instead of writing relative path it is better to write the full path of these libs.
Also it is better to give a name with full contaxt of these classes.

Example: `ValidationRpm`, `ValidationTar`.

peterzhuamazon (changes requested):

peterzhuamazon:
A good start @Divyaasm but I think we can improve it even better 😄 
Please see my above comments on some tweaks.

Thanks.

src/validation_workflow/tar.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @classmethod
+    def download_urls(cls, projects: list, version: str, platform: str) -> None:
+        for project in projects:
+            for package_type in cls.get_architectures():
+                url = f"https://artifacts.opensearch.org/releases/bundle/{project}/{version}/{project}-{version}-{platform}-{package_type}.tar.gz"
+                if Tar.is_url_valid(url):
+                    print(f" Valid URL - {url}")
```
gaiksaya, line 23: Make the return type as boolean. True if successful, else false along with logging. We want the workflow to fail if the validation fails instead of throwing false positive and just logging.

prudhvigodithi:
Hey @Divyaasm some info on tests with existing workflows:

Tests to assert the `main()` method: https://github.com/opensearch-project/opensearch-build/blob/main/tests/test_run_releasenotes_check.py#L46-L60

Tests per class which is used my `main()` : [release_notes.py](https://github.com/opensearch-project/opensearch-build/blob/main/src/release_notes_workflow/release_notes.py) [run_releasenotes_check.py](https://github.com/opensearch-project/opensearch-build/blob/main/src/run_releasenotes_check.py)

tests/tests_run_validation.py
```diff
@@ -0,0 +1,56 @@
...
+        self.assertEqual(ValidationArgs().distribution, "tar")
+        self.assertNotEqual(ValidationArgs().distribution, "rpm")
+
+    @patch("argparse._sys.argv", ["run_validation.py", "2.1.0", "--distribution", "rpm", "--platform", "linux"])
+    @patch("src.validation_workflow.validation_rpm.ValidationRpm.download_urls", return_value=True)
+    @patch("run_validation.main", return_value=0)
+    def test_main_rpm(self, mock_tar: Mock, *mocks: Any) -> None:
+        self.assertEqual(ValidationArgs().version, "2.1.0")
+        self.assertEqual(ValidationArgs().distribution, "rpm")
+        self.assertNotEqual(ValidationArgs().distribution, "yum")
```
gaiksaya, line 48: rpm and yum are same thing. I believe you want to assert that its not equal to `tar` which is default

gaiksaya (changes requested):
Good work overall. Need some minor improvements.

tests/tests_run_validation.py
Divyaasm: Thanks for the suggestions Sayali, I will work on the changes.

src/validation_workflow/validation.py
```diff
@@ -0,0 +1,25 @@
...
+import argparse
+from abc import ABC, abstractmethod
+
+
+class Validation(ABC):
+    url = "https://artifacts.opensearch.org/releases/bundle/"
+
+    def __init__(self) -> None:
+        parser = argparse.ArgumentParser(description="Download Artifacts.")
+        parser.add_argument("version", help="Enter Version.")
```
zelinh, line 18: Why we have another argument parser in this class? Should all be handled in the `ValidationArgs()` and passing that `args` assigned in `run_validation` within all classes?

Divyaasm: I haven't called the base class in my workflow by creating an object,So I have left it to default version. I will pass all the arguments from ValidationArgs() to __init__ if the base class object is utilized in further development of the code or can set it to default without passing and arguments and without declaring any new variables.

prudhvigodithi:
Hey @Divyaasm thanks for PR/changes, 

I would suggest to include [TemporaryDirectory](https://github.com/opensearch-project/opensearch-build/blob/main/src/system/temporary_directory.py), in your logic to download the files, else they would be on the user working directory and would never be deleted, example https://github.com/opensearch-project/opensearch-build/blob/main/src/release_notes_workflow/release_notes.py#L40, with this it would download all the artifacts in a temporary directly and will be removed once the execution is done. 

Can you also add a small README file?

src/validation_workflow/download_utils.py
```diff
@@ -0,0 +1,28 @@
...
+    @staticmethod
+    def is_url_valid(url: str) -> int:
+        response = requests.head(url)
+        if response.status_code == 200:
+            return 1
+        else:
+            return 0
+
+    @staticmethod
+    def download(url: str, tmp_dir: TemporaryDirectory) -> bool:
```
gaiksaya, line 24: nit: rename to download_artifacts
Just download is bit confusing

src/validation_workflow/download_utils.py
```diff
@@ -0,0 +1,28 @@
...
+        response = requests.get(url, stream=True)  # get() method sends a GET request to the url
+        path = tmp_dir.name + "/" + url.split("/")[-1]
+        val = bool(open(path, "wb").write(response.content))  # writes the contents from the response object into temporary directory file name fetched from the end of the url
```
gaiksaya, line 27: Are these comments required? If yes please move it above the usage or remove it entirely. Also recommending to use better variables names instead of `val`

src/validation_workflow/validation.py
```diff
@@ -0,0 +1,24 @@
...
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+
+from abc import ABC, abstractmethod
+
+
+class Validation(ABC):
+    url = "https://artifacts.opensearch.org/releases/bundle/"
```
gaiksaya, line 13: nit: base_url sounds more appropriate

src/validation_workflow/validation_args.py
```diff
@@ -0,0 +1,74 @@
...
+    DISTRIBUTION_MAP = {
+        "tar": ValidationTar,
+        "yum": ValidationYum,
+        "rpm": ValidationRpm
+    }
+
+    version: str
+
+    def __init__(self) -> None:
+        parser = argparse.ArgumentParser(description="Download Artifacts.")
```
gaiksaya, line 36: Can you make the description more generic as the same framework is going to be used for multiple other platforms?

src/validation_workflow/validation_rpm.py
```diff
@@ -0,0 +1,28 @@
...
+    def download_artifacts(self, projects: list, version: str, platform: str, architectures: list) -> bool:
+        tmp_dir = TemporaryDirectory()
+        for project in projects:
+            for architecture in architectures:
+                url = f"{self.url}{project}/{version}/{project}-{version}-{platform}-{architecture}.rpm"
+                if ValidationRpm.is_url_valid(url) and ValidationRpm.download(url, tmp_dir):
+                    logging.info(f" Valid URL - {url} and Download Successful !")
+                else:
+                    logging.info(f"Invalid URL - {url}")
+                    raise Exception("Invalid url - check version")
```
gaiksaya, line 27: Not everytime the issue might be in version, sometimes the artifact won't exist at all. 
Just `invalid url` with url provided works.

src/validation_workflow/validation_tar.py
```diff
@@ -0,0 +1,28 @@
...
+    def download_artifacts(self, projects: list, version: str, platform: str, architectures: list) -> bool:
+        tmp_dir = TemporaryDirectory()
+        for project in projects:
+            for architecture in architectures:
+                url = f"{self.url}{project}/{version}/{project}-{version}-{platform}-{architecture}.tar.gz"
+                if ValidationTar.is_url_valid(url) and ValidationTar.download(url, tmp_dir):
+                    logging.info(f" Valid URL - {url} and Download Successful !")
+                else:
+                    logging.info(f"Invalid URL - {url}")
+                    raise Exception("Invalid url - check version")
```
gaiksaya, line 27: Same as above, change all for all.

gaiksaya (approved):
Added some nits.

src/validation_workflow/download_utils.py
Divyaasm: This is a different function in downloadutils.py which is called by the download_artifacts method in tar, rpm and yum.

src/validation_workflow/validation.py
Divyaasm: I made the change.

src/validation_workflow/validation_rpm.py
Divyaasm: Made the changes

src/validation_workflow/validation_tar.py
Divyaasm: Sure, Thanks.

src/validation_workflow/download_utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class DownloadUtils:
+    @staticmethod
+    def is_url_valid(url: str) -> int:
+        response = requests.head(url)
+        if response.status_code == 200:
+            return 1
+        else:
+            return 0
```
peterzhuamazon, line 21: Is there a benefit using 1/0 instead of True/False?
This is also opposite compares to normal programming pattern where 1 is false and 0 is true normally.

Plus, the `else` block can be completely removed in this case.

src/validation_workflow/download_utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @staticmethod
+    def download(url: str, tmp_dir: TemporaryDirectory) -> bool:
+        # This method writes the contents from the response object into temporary directory file name fetched from the end of the url.
+        response = requests.get(url, stream=True)
+        path = tmp_dir.name + "/" + url.split("/")[-1]
```
peterzhuamazon, line 27: This is not a good implementation and I would rather use os.join because `/` is only working on nix* and would cause trouble on windows (even tho windows now can work with `/` it is still not ideal here.)

peterzhuamazon: Also, if you want to get the file name, use something like os.basename or similar, do not try to split the url and get the last item.

src/validation_workflow/validation.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Validation(ABC):
+    base_url = "https://artifacts.opensearch.org/releases/bundle/"
+
+    def __init__(self, version: str, distribution: str, platform: str, projects: list) -> None:
+        self.version = version
+        self.distribution = distribution
+        self.platform = platform
+        self. projects = projects
```
peterzhuamazon, line 19: Typo, which would cause the code not running at all.

src/validation_workflow/validation.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @classmethod
```
peterzhuamazon, line 21: Why is this a classmethod after all?

src/validation_workflow/validation_args.py
```diff
@@ -0,0 +1,74 @@
...
+    SUPPORTED_PLATFORMS = ["linux"]
+    SUPPORTED_ARCHITECTURES = [
+        "x64",
+        "arm64",
+    ]
+    DISTRIBUTION_MAP = {
+        "tar": ValidationTar,
+        "yum": ValidationYum,
+        "rpm": ValidationRpm
+    }
```
peterzhuamazon, line 31: We should not need one list and one map just to rewrite all the dist again.

The design can be using 1 map max:

Example:

```

    DISTRIBUTION_MAP = {
        "tar": ValidationTar,
        "yum": ValidationYum,
        "rpm": ValidationRpm
    }
```

Then the platform/arch information can be distributed directly in the validation<> class.

src/validation_workflow/validation_rpm.py
```diff
@@ -0,0 +1,28 @@
...
+class ValidationRpm(Validation, DownloadUtils):
+    tmp_dir = TemporaryDirectory()
+
+    @classmethod
+    def download_artifacts(self, projects: list, version: str, platform: str, architectures: list) -> bool:
+        for project in projects:
+            for architecture in architectures:
+                url = f"{self.base_url}{project}/{version}/{project}-{version}-{platform}-{architecture}.rpm"
+                if ValidationRpm.is_url_valid(url) and ValidationRpm.download(url, self.tmp_dir):
+                    logging.info(f" Valid URL - {url} and Download Successful !")
```
peterzhuamazon, line 24: Why is there a leading space?

src/validation_workflow/validation_tar.py
```diff
@@ -0,0 +1,28 @@
...
+class ValidationTar(Validation, DownloadUtils):
+    tmp_dir = TemporaryDirectory()
+
+    @classmethod
+    def download_artifacts(self, projects: list, version: str, platform: str, architectures: list) -> bool:
+        for project in projects:
+            for architecture in architectures:
+                url = f"{self.base_url}{project}/{version}/{project}-{version}-{platform}-{architecture}.tar.gz"
+                if ValidationTar.is_url_valid(url) and ValidationTar.download(url, self.tmp_dir):
+                    logging.info(f" Valid URL - {url} and Download Successful !")
```
peterzhuamazon, line 24: Same as above.

src/validation_workflow/validation_yum.py
```diff
@@ -0,0 +1,27 @@
...
+
+class ValidationYum(Validation, DownloadUtils):
+    tmp_dir = TemporaryDirectory()
+
+    @classmethod
+    def download_artifacts(self, projects: list, version: str, platform: str, architectures: list) -> bool:
+        for project in projects:
+            url = f"{self.base_url}{project}/{version[0:1]}.x/{project}-{version[0:1]}.x.repo"
+            if ValidationYum.is_url_valid(url) and ValidationYum.download(url, self.tmp_dir):
+                logging.info(f" Valid URL - {url} and Download Successful !")
```
peterzhuamazon, line 23: Same as above.

src/validation_workflow/validation_rpm.py
```diff
@@ -0,0 +1,28 @@
...
+
+import logging
+
+from system.temporary_directory import TemporaryDirectory
+from validation_workflow.download_utils import DownloadUtils
+from validation_workflow.validation import Validation
+
+
+class ValidationRpm(Validation, DownloadUtils):
+    tmp_dir = TemporaryDirectory()
```
peterzhuamazon, line 16: If this is a temp dir that all the test requires, why not just put it back to the parent class?

src/validation_workflow/validation_tar.py
peterzhuamazon, line 24: All of these implementation are extremely similar, means the core code should exist in parent, and child should just pass the different information.

peterzhuamazon (changes requested):
Hi @Divyaasm good start but there are some parts that I question the implementation.

Could you please take a look at these comments I raised.

Thanks.

src/validation_workflow/validation.py
peterzhuamazon: This also means you are missing test cases because the test pass but this code is invalid.
Or it means there are variables you did not use.

src/validation_workflow/validation_args.py
Divyaasm: Sure Peter, I will use a single dictionary for storing distributions and mapping the distributions to respective classes. Could you please explain what `platform/arch information can be distributed directly in the validation<> class.`
in your comment means.Do you recommend that I declare these lists where they are actually implemented without declaring them in validationargs.py so I need not pass them as arguments to the function.

src/validation_workflow/validation_tar.py
Divyaasm: I was initially suggested to have only abstract methods in the base class so I have implemented it this way. I can actually change the core code to base class and pass the required information from the child class. Can I know what you would recommend!

Divyaasm: I'm expecting to leave the method abstract in base class  as Docker  also extends the same method which has a different implementation all together.So, had to implement it independently for each distribution even though the code is reused.

src/validation_workflow/validation.py
```diff
@@ -0,0 +1,24 @@
...
+    @abstractmethod
+    def download_artifacts(self, projects: list, version: str, platform: str, architectures: list) -> bool:
```
jordarlu, line 23: Greetings, @Divyaasm , I am working on a dependant issue/PR to validation docker artifact and want to update you that I will need only the 'version' argument to Pull and Spin-up containers by docker-compose; hence, if you can make other arguments in your script as _optional_, that will help!  Thanks!

src/validation_workflow/validation.py
Divyaasm: Removing class method threw me an error `Missing positional argument "self" in call to "download_artifacts" of "Validation"` so I let it stay Peter.

src/validation_workflow/validation.py
Divyaasm: I haven't used the init method in base class for implementation.So tried removing the init as I don't need any class properties in particular.

src/validation_workflow/validation.py
gaiksaya: +1 I believe we can make platform, architectures optional

src/validation_workflow/validation_rpm.py
```diff
@@ -0,0 +1,26 @@
...
+from validation_workflow.download_utils import DownloadUtils
+from validation_workflow.validation import Validation
+
+
+class ValidationRpm(Validation, DownloadUtils):
+
+    @classmethod
+    def download_artifacts(self, projects: list, version: str) -> bool:
+        for project in projects:
+            for architecture in ["x64", "arm64"]:
```
peterzhuamazon, line 19: Why is there still hardcoded architecture here?
Cant they be saved in the Validation<> class?

src/validation_workflow/validation_tar.py
```diff
@@ -0,0 +1,26 @@
...
+from validation_workflow.download_utils import DownloadUtils
+from validation_workflow.validation import Validation
+
+
+class ValidationTar(Validation, DownloadUtils):
+
+    @classmethod
+    def download_artifacts(self, projects: list, version: str) -> bool:
+        for project in projects:
+            for architecture in ["x64", "arm64"]:
```
peterzhuamazon, line 19: Same as above

src/validation_workflow/validation_rpm.py
peterzhuamazon, line 19: One nit: define `[]` as an actual map instead of calling it on the fly.
```
architecture_map=[]
for architecture in architecture_map
```

peterzhuamazon (approved):
Thanks @Divyaasm, one nitpik from above.

PR #2761, opened 2022-10-18
Add windows integTest to OpenSearch Clusters 

peterzhuamazon:
### Description
Add windows integTest to OpenSearch Clusters 

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2306


tests/tests_test_workflow/test_integ_workflow/integ_test/test_distribution_zip.py
```diff
@@ -0,0 +1,54 @@
...
+import os
+import unittest
+from unittest.mock import MagicMock, Mock, patch
+
+from test_workflow.integ_test.distribution_zip import DistributionZip
+
+
+class TestDistributionZip(unittest.TestCase):
+
+    def setUp(self) -> None:
```
dblock, line 17: There's a bit of mixing of OpenSearch and Dashboards here. Split into 3 test classes in this file `TestDistributionZip` for common logic, and `TestDistributionZipOpenSearch` and `TestDistributionZipOpenSearchDashboards`?

src/test_workflow/integ_test/distribution_zip.py
```diff
@@ -0,0 +1,43 @@
...
+    @property
+    def config_dir(self) -> str:
+        return os.path.join(self.install_dir, "config")
+
+    def install(self, bundle_name: str) -> None:
+        logging.info(f"Installing {bundle_name} in {self.install_dir}")
+        with ZipFile(bundle_name, "r") as zip:
+            zip.extractall(self.work_dir)
+
+    @property
```
dblock, line 33: The right way to do this is inheritance: create `DistributionZipOpenSearch` and `DistributionZipOpenSearchDashboards` and pick which instance to create in a `create` class method. Then the first one would implement `start_cmd` returning one filename and the second the other.

dblock (changes requested):
Refactor the command line `if`.

src/test_workflow/integ_test/distribution_zip.py
peterzhuamazon: Since the scope of this is way bigger than adding windows integTest support.

tests/tests_test_workflow/test_integ_workflow/integ_test/test_distribution_zip.py
peterzhuamazon: Hi DB,

I think it is too complicated that way.
OS and OSD have exactly the same way of extracting artifact as well as triggering the start cmd to certain degree.

If this is more component adding up we can definitely separate them into different classes.
This is a exact one to one copy from the tar tests as well.

Thanks.

peterzhuamazon:
After talking to DB seems like it is better to:
```
Add a product selector, a product class, 2 product subclasses to run product specific start up cmd.

```

dblock (approved):

gaiksaya:
Were adding zips / binary files intended? https://github.com/opensearch-project/opensearch-build/tree/main/tests/tests_test_workflow/test_integ_workflow/integ_test/data/artifacts/dist/

peterzhuamazon:
> Were adding zips / binary files intended? https://github.com/opensearch-project/opensearch-build/tree/main/tests/tests_test_workflow/test_integ_workflow/integ_test/data/artifacts/dist/

Yes that is a dummy file to test.

PR #2850, opened 2022-11-04
Fix signing the artifacts for .asc signatures

gaiksaya:
### Description
Currently when we say we want `.asc` signature, we only change the extension of the signature file which it is still a binary. This PR fixes the issue where an actual enamored  signature is generated based on the binary signature.
This PR also changes the default signature_type from `.asc` to `.sig`.

Follow up, need to make changes in jenkinsFiles that assumes default signature as `.asc`

### Issues Resolved
resolves https://github.com/opensearch-project/opensearch-build/issues/368

### DIFF
--- a/src/sign_workflow/sign_args.py
+++ b/src/sign_workflow/sign_args.py
@@ -26,7 +26,7 @@ class SignArgs:
         parser.add_argument("target", type=Path, help="Path to local manifest file or artifact directory")
         parser.add_argument("-c", "--component", type=str, nargs='*', dest="components", help="Component or components to sign")
         parser.add_argument("--type", help="Artifact type")
-        parser.add_argument("--sigtype", choices=self.ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of signature file.", default=".asc")
+        parser.add_argument("--sigtype", choices=self.ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of signature file.", default=".sig")
         parser.add_argument("--platform", choices=self.ACCEPTED_PLATFORM, help="Distribution platform.", default="linux")
         parser.add_argument(
             "-v",
--- a/src/sign_workflow/signer_pgp.py
+++ b/src/sign_workflow/signer_pgp.py
@@ -8,7 +8,7 @@
 
 import os
 from pathlib import Path
-
+import logging
 from sign_workflow.signer import Signer
 
 """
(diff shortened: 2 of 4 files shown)


prudhvigodithi:
Hey @gaiksaya I'm ok with this PR, as the default output a `.sig` file, its good we convert to proper `.asc` file instead of just renaming the file to `.asc`, however should jenkinsfile changes (especially the maven publish) should go along with this PR?
Thank you 
@peterzhuamazon @zelinh

prudhvigodithi:
Also just add @gaiksaya once the function is executed `__convert_to_asc` that outputs a `.asc` file, can you confirm by running validation 
```
gpg --verify <>.tar.gz.asc <>.tar.gz`
```
Aslo we still need to import global .asc key right using `gpg --import GLOBAL.asc`?

gaiksaya:
> Hey @gaiksaya I'm ok with this PR, as the default output a `.sig` file, its good we convert to proper `.asc` file instead of just renaming the file to `.asc`, however should jenkinsfile changes (especially the maven publish) should go along with this PR? Thank you @peterzhuamazon @zelinh

Let me add those files as well.

gaiksaya:
> Also just add @gaiksaya once the function is executed `__convert_to_asc` that outputs a `.asc` file, can you confirm by running validation
> 
> ```
> gpg --verify <>.tar.gz.asc <>.tar.gz`
> ```
> 
> Aslo we still need to import global .asc key right using `gpg --import GLOBAL.asc`?

Hey! the verification part is taken care by parent method here https://github.com/opensearch-project/opensearch-build/pull/2850/files#diff-b97cb5a1ac1ecef8c9c134f702092d1ba8818099a722cc5da21d73f8ded3e018R25

gaiksaya:
> Let me add those files as well.

Done @prudhvigodithi

zelinh (commented):
Overall this looks good to me. Just one thing do we need to change the groovy lib `signArtifacts` to accommodate this change?

src/sign_workflow/signer_pgp.py
```diff
@@ -33,7 +34,7 @@ def is_valid_file_type(self, file_name: str) -> bool:
 
     def sign(self, artifact: str, basepath: Path, signature_type: str) -> None:
         filename = os.path.join(basepath, artifact)
-        signature_file = filename + signature_type
+        signature_file = filename + ".sig"
```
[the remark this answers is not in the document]
gaiksaya: we cannot make that as class variable as the `filename` is not static and it is calculated from method params. 
Also, I have instead named the other file as asc_signature_file. It is just a variable name.

prudhvigodithi:
> Overall this looks good to me. Just one thing do we need to change the groovy lib `signArtifacts` to accommodate this change?

Hey @zelinh you mean this [signArtifacts](https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/signArtifacts.groovy)? adding @peterzhuamazon

gaiksaya:
We do not need to change anything on jenkins library side as it takes all the args from jenkins and gives it to python code as it is 
Example: https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/signArtifacts.groovy#L132

prudhvigodithi (approved):

src/sign_workflow/sign_args.py
```diff
@@ -26,7 +26,7 @@ def __init__(self) -> None:
         parser.add_argument("target", type=Path, help="Path to local manifest file or artifact directory")
         parser.add_argument("-c", "--component", type=str, nargs='*', dest="components", help="Component or components to sign")
         parser.add_argument("--type", help="Artifact type")
-        parser.add_argument("--sigtype", choices=self.ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of signature file.", default=".asc")
+        parser.add_argument("--sigtype", choices=self.ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of signature file.", default=".sig")
```
peterzhuamazon, line 29: So sigtype is still kinda useless here because you hard code it to .sig and convert to .asc later.

gaiksaya: The acceptable types are sig and asc only https://github.com/opensearch-project/opensearch-build/pull/2850/files#diff-57b5e80836a33ceb7cbcdfb92fba22572d31fc761d817039e05119fa531c54f5R15

If we need more types in future for pgp we need to refactor the code. However, the upstream (opensearch-signer-client) only supports binary signatures so this hardcoding seems okay to me for now

gaiksaya: See the if condition here: https://github.com/opensearch-project/opensearch-build/pull/2850/files#diff-b97cb5a1ac1ecef8c9c134f702092d1ba8818099a722cc5da21d73f8ded3e018R49

PR #2856, opened 2022-11-06
Make input manifest ci checks accept commit id as refs

peterzhuamazon:

### Description
Make input manifest ci checks accept commit id as refs

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2649

### DIFF
--- a/src/ci_workflow/ci_check_list.py
+++ b/src/ci_workflow/ci_check_list.py
@@ -14,6 +14,7 @@ from ci_workflow.ci_target import CiTarget
 class CiCheckList(ABC):
     def __init__(self, component: Any, target: CiTarget) -> None:
         self.component = component
+        self.component_ref_is_commit = True if len(self.component.ref) == 40 and int(self.component.ref, 16) else False
         self.target = target
 
     @abstractmethod
--- a/src/ci_workflow/ci_check_list_source_ref.py
+++ b/src/ci_workflow/ci_check_list_source_ref.py
@@ -7,8 +7,10 @@
 
 import logging
 import subprocess
+import os
 
 from ci_workflow.ci_check_list import CiCheckList
+from git.git_repository import GitRepository
 
 
 class CiCheckListSourceRef(CiCheckList):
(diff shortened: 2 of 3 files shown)


src/ci_workflow/ci_check_list.py
```diff
@@ -14,6 +15,11 @@
 class CiCheckList(ABC):
     def __init__(self, component: Any, target: CiTarget) -> None:
         self.component = component
+        try:
+            self.component_ref_is_commit = True if len(self.component.ref) == 40 and int(self.component.ref, 16) else False
```
zelinh, line 19: I think the length check and hexa-decimal int check don't mean the ref exists in the repo. Maybe we say this is a valid sha1 commit here, but we would check if this commit ref exists in the component repo with the `check()` method below.

peterzhuamazon: The check starts on line 36 of src/ci_workflow/ci_check_list_source_ref.py.
The check here is to ensure it is sha1 before it reach the check stage.

peterzhuamazon: Talked to @zelinh we will change component_ref_is_commit to component_ref_is_sha1 to reduce confusion before reaching the actual check stage.

zelinh (approved):
LGTM. Had offline discussion with @peterzhuamazon. This would be performed when checks in input manifest are absent.

dblock (commented):
This should have been implemented differently by adding two subclasses of `CiCheckListSourceRef`, one that has a ref as SHA1 and another that doesn't, instead of all these `ifs` and creating the right instance thereof accordingly. Care to refactor this @zelinh?

peterzhuamazon:
> This should have been implemented differently by adding two subclasses of `CiCheckListSourceRef`, one that has a ref as SHA1 and another that doesn't, instead of all these `ifs` and creating the right instance thereof accordingly. Care to refactor this @zelinh?

This is a temp measure to unblock 2.4.0.
We will refactor it later on.

Thanks.

Issue #2749, opened 2022-10-17
[Feature] Make it easy to create/consume massive changes without breaking major version


peternied:

saratvemulapalli:
Just had a chat with @peternied. Looks like the question mostly is around the feature of supporting feature branches for the build.

peternied:
> why are breaking changes on 3.x causing problems in 2.x,1.x in plugins

Our main branches are pointing to 3.0.0 , so we default to accept work on main and then backport it to 2.x  , etc...

> Which makes sense, but the breaking changes will only go in Majors, so shouldn’t the changes be limited to 3.0 and not backport them ?

True, but the change is so disruptive that we are still working on handling it with no firm ETA.

---

> Hopefully this is something we solve with extensions.

I think this kind of break is a classic case for a mono-repo, its easy to measure impact of a change if you can see all of it at once before merging.

In distributed environments being really careful about breaking interfaces is the name of the game.  In this case there are direct dependencies passed via public interfaces and internal systems.

Extensions is a much healthy place for managing these interfaces in a more abstract way.

_Recapped from an offline conversation with @saratvemulapalli_

cliu123:
@bbarani @peternied Is it possible to have a build including OpenSearch core and all plugins for every PR of OpenSearch core before merging to surface up the potential failures in plugins earlier? Breaking changes really hold off plugins and make development on main impossible, and this could happen more frequently as more plugins start building 3.0.
Cc: @dblock @nknize

cliu123:
@opensearch-project/engineering-effectiveness  Is it possible to have a build including OpenSearch core and all plugins for every PR of OpenSearch core before merging to surface up the potential failures in plugins earlier? Breaking changes really hold off plugins and make development on main impossible, and this could happen more frequently as more plugins start building 3.0.
Cc: @dblock @nknize

reta:
@cliu123 @saratvemulapalli @peternied I believe we should think beyond just core plugins but the ecosystem as a whole. By releasing milestones (3.0.0-M1, 3.0.0-M2, ... ) we allow not only core plugins to catch up with the changes (depending on SNAPSHOT is always unstable), but allow external plugin authors to gradually prepate for what is coming. With these regards, we should not only rely on the CI/build infrastructure but keep in mind that it is not available for external plugins.

cliu123:
[...]

PR #2881, opened 2022-11-10
Add latest url update toggle on distribution workflows

peterzhuamazon:

### Description
Add latest url update toggle on distribution workflows

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2649


zelinh (approved):
As offline discussion with @peterzhuamazon, the use case for this param will be when we want to pushing only OpenSearch min to maven but we want the `latest` index still pointing to the last bundle including all plugins.

gaiksaya:
@peterzhuamazon  Does this solve https://github.com/opensearch-project/opensearch-build/issues/2749 ? Would take input from there before proceeding

prudhvigodithi (approved):

peterzhuamazon:
> @peterzhuamazon Does this solve #2749 ? Would take input from there before proceeding

Thanks @gaiksaya I replied to that issue and will merge this for now.
We can always revert this as it is a very simple change.
The default option runs exactly like before.

PR #2910, opened 2022-11-16
Add custom import map plugin to 3.0.0 OSD

junqiu-lei:
Signed-off-by: Junqiu Lei <junqiu@amazon.com>

### Description
Add custom import map plugin to OSD 3.0.0 with correct plugin path.

### Issues Resolved
https://github.com/opensearch-project/dashboards-maps/issues/83

### DIFF
--- a/manifests/3.0.0/opensearch-dashboards-3.0.0.yml
+++ b/manifests/3.0.0/opensearch-dashboards-3.0.0.yml
@@ -27,3 +27,7 @@ components:
   - name: searchRelevanceDashboards
     repository: https://github.com/opensearch-project/dashboards-search-relevance.git
     ref: main
+  - name: customImportMapDashboards
+    repository: https://github.com/opensearch-project/dashboards-maps.git
+    working_directory: custom_import_map
+    ref: main
--- a/scripts/components/customImportMapDashboards/build.sh
+++ b/scripts/components/customImportMapDashboards/build.sh
@@ -69,7 +69,7 @@ fi
 
 NVM_CMD="source $NVM_DIR/nvm.sh && nvm use"
 if [ "$PLATFORM" = "windows" ]; then
-    NVM_CMD="volta install node@`cat ../../../../OpenSearch-Dashboards/.nvmrc` && volta install yarn@`jq -r '.engines.yarn' ../../../../OpenSearch-Dashboards/package.json`"
+    NVM_CMD="volta install node@`cat ../../OpenSearch-Dashboards/.nvmrc` && volta install yarn@`jq -r '.engines.yarn' ../../OpenSearch-Dashboards/package.json`"
 fi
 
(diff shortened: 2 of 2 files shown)


peterzhuamazon:
Hi @junqiu-lei would 2.4.0+ versions fail on this?
Thanks.

junqiu-lei:
@peterzhuamazon Since 2.4.0 branch cut out, in my understand, this change will only effect to OSD builds that using main branch(current 3.0.0 and future versions)

gaiksaya:
Hi @junqiu-lei ,

The question here is if we are to release 2.x or 1.x version in future which are based on respective branches, will this affect those builds?
Also @peterzhuamazon What do you think about moving this script to the respective repo rather than maintaining here which uses same scripts for all branches? In that way the right script will be checkout for right version of the component

junqiu-lei:
@gaiksaya Thanks for explanation. Another potential solution I am thinking is that we can add a conditional check for the version in the script to determine use the right path. The custom import map plugin started release from 2.2.0, so except 2.2.x and 2.3.x and 2.4.x use old path, others versions should use the new path.

gaiksaya:
> @gaiksaya Thanks for explanation. Another potential solution I am thinking is that we can add a conditional check for the version in the script to determine use the right path. The custom import map plugin started release from 2.2.0, so except 2.2.x and 2.3.x and 2.4.x use old path, others versions should use the new path.

That sounds kind of hacky to me. Overtime if would become difficult to maintain as versions go on increasing.
@dblock What do you think?

dblock:
> > @gaiksaya Thanks for explanation. Another potential solution I am thinking is that we can add a conditional check for the version in the script to determine use the right path. The custom import map plugin started release from 2.2.0, so except 2.2.x and 2.3.x and 2.4.x use old path, others versions should use the new path.
> 
> That sounds kind of hacky to me. Overtime if would become difficult to maintain as versions go on increasing. @dblock @prudhvigodithi What do you think?

Move the build script into your repo and all its branches under `scripts/`, then delete it here.

zelinh (commented):
@junqiu-lei Hi the build workflow will firstly look for the build script from the current location in our build repo and if it can't find it here, it would look for it in the respective component repo (logic from [here](https://github.com/opensearch-project/opensearch-build/blob/4a3914af5219091bda277e370661e8443aa3436c/src/paths/script_finder.py#L59)). 
As our build repo currently only maintains on main branch, it's hard for us to control any specific `build.sh` changes for different major versions. So we would recommend you to have this build script maintained in your repo so that you could update it specifically in different branches. 
Once it's created there in your component repo, make sure to remove it from here in build repo.

junqiu-lei:
Thanks team's suggestions, today I will talk with @peterzhuamazon for the workable solution.

PR #2913, opened 2022-11-16
Improve output on gradle check timeout

andrross:
The result ends up being reported as 'null' when this script times out. Also the output never directly indicates that a timeout occurred, so this commit makes the output more explicit.

---

### DIFF
--- a/scripts/gradle/gradle-check.sh
+++ b/scripts/gradle/gradle-check.sh
@@ -53,19 +53,21 @@ if [ -z "$QUEUE_URL" ] || [ "$QUEUE_URL" != "null" ]; then
             RUNNING=$(curl -s -XGET ${WORKFLOW_URL}api/json | jq --raw-output .building)
         done
 
-        echo "Complete the run, checking results now......"
-        RESULT=$(curl -s -XGET ${WORKFLOW_URL}api/json | jq --raw-output .result)
+        if [ "$RUNNING" = "true" ]; then
+            echo "Timed out"
+            RESULT="TIMEOUT"
+        else
+            echo "Complete the run, checking results now......"
+            RESULT=$(curl -s -XGET ${WORKFLOW_URL}api/json | jq --raw-output .result)
+        fi
 
     fi
 fi
 
 echo "Please check jenkins url for logs: $WORKFLOW_URL"
-
+echo "Result: $RESULT"
 if [ "$RESULT" == "SUCCESS" ] || [ "$RESULT" == "UNSTABLE" ]; then
-    echo "Result: $RESULT"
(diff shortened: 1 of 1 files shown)


kotwanikunal (approved):

peterzhuamazon (changes requested):
The issue with this approach is, it is possible that we have running state = false, while seconds exceeds timeout value.

You would end up with a success test, but still failed due to timeout, as seconds will continue to count.

running=false
seconds=7336
timeout=7200
results=success
return=timeout

That is why my initial approach is to accurately compute the number of timepass, so I can control when to add up the time instead of letting it compute by itself.

Thanks.

peterzhuamazon:
> > I think we can definitely improve the output, but no need to change the structure of the time calculation.
> 
> @peterzhuamazon Good call out. I'm happy to change this back to the way it was, but let me know what you think of the [updated approach](https://github.com/opensearch-project/opensearch-build/compare/0c9f33da7b4036f6967f61ffb429a6026dcb3146..c73182a2ccb44f5d9acf5e2cd66aaf28e69fad26) I just published. It should fix the issue you mentioned. The reason I changed the time calculation is so that the script will print the accurate elapsed time instead of always being 30 seconds ahead.

I think this still needs some testing, as there are cases where timeout reached but then running state changed to false right after, still not safe.

I dont think the 30seconds ahead count is really a big issue tho.
Let's just merge the results changes without touching the time calculation for now.

Thanks,
Peter

scripts/gradle/gradle-check.sh
```diff
@@ -53,19 +53,22 @@ if [ -z "$QUEUE_URL" ] || [ "$QUEUE_URL" != "null" ]; then
...
 fi
 
 echo "Please check jenkins url for logs: $WORKFLOW_URL"
-
+echo "Result: $RESULT"
 if [ "$RESULT" == "SUCCESS" ] || [ "$RESULT" == "UNSTABLE" ]; then
-    echo "Result: $RESULT"
     echo "Get codeCoverage.xml" && curl -SLO ${WORKFLOW_URL}artifact/codeCoverage.xml
-    echo 0
+    exit 0
```
andrross, line 71: I'm assuming `echo 0` was a typo here, though it made no functional difference because the script would exit with a code of 0 anyway.

peterzhuamazon: It is not a typo it is just to print 0 to tell user it is downloading correctly.
Tho that is really a debug msg that kinda leftover.
We can delete this if needed.

peterzhuamazon: We can remove this line no need to reform the exit status to 0.

peterzhuamazon (approved):

PR #2919, opened 2022-11-17
Update manifest workflow to add build_platform

### LINKED ISSUE #2912: [Bug]: Cron created by manifest workflow doesn't have build_platform parameter
### Describe the bug

Our manifest workflow will help us create input manifest and add the build job to our cron once we have a new version from OS or OSD core. e.g. #2894 

However since we start to build for different platform distribution, we would need to add build_platform parameters to trigger the build workflow which our current manifest workflow doesn't have. We would need to manually add the param to the cron in order to trigger it. #2911 

### To reproduce

When we have a new OS or OSD version, run `./manifest.sh update`

### Expected behavior

When the workflow raise the PR, it should also add this param to the cron. For different job type(OS or OSD), it would also add different platform parameter. e.g. for OS it should be `BUILD_PLATFORM=linux macos windows` but for OSD it's `BUILD_PLATFORM=linux windows`

### Host / Environment

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

zelinh:

### Description
Update manifest workflow to add build_platform to cron.

### Issues Resolved
#2912 

### DIFF
--- a/src/manifests_workflow/input_manifests.py
+++ b/src/manifests_workflow/input_manifests.py
@@ -164,7 +164,12 @@ class InputManifests(Manifests):
         with open(jenkinsfile, "r") as f:
             data = f.read()
 
-        cron_entry = f"H 1 * * * %INPUT_MANIFEST={version}/{self.prefix}-{version}.yml;TARGET_JOB_NAME=distribution-build-{self.prefix}\n"
+        build_platform_dict = {"opensearch": "linux macos windows", "opensearch-dashboards": "linux windows"}
+        build_platform = build_platform_dict.get(self.prefix, "linux macos windows")
+
+        cron_entry = f"H 1 * * * %INPUT_MANIFEST={version}/{self.prefix}-{version}.yml;" \
+                     f"TARGET_JOB_NAME=distribution-build-{self.prefix};" \
+                     f"BUILD_PLATFORM={build_platform}\n"
 
         if cron_entry in data:
             raise ValueError(f"{jenkinsfile} already contains an entry for {self.prefix} {version}")
--- a/tests/tests_manifests_workflow/test_input_manifests.py
+++ b/tests/tests_manifests_workflow/test_input_manifests.py
@@ -125,7 +125,8 @@ class TestInputManifests(unittest.TestCase):
         mock_open.assert_has_calls([call(InputManifests.cron_jenkinsfile(), 'w')])
         mock_open.assert_has_calls([call(InputManifests.cron_jenkinsfile(), 'r')])
         mock_open().write.assert_called_once_with(
-            f"parameterizedCron '''\n{' ' * 12}H 1 * * * %INPUT_MANIFEST=0.1.2/test-0.1.2.yml;TARGET_JOB_NAME=distribution-build-test\n"
+            f"parameterizedCron '''\n{' ' * 12}H 1 * * * %INPUT_MANIFEST=0.1.2/test-0.1.2.yml;"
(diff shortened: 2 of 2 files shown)


src/manifests_workflow/input_manifests.py
```diff
@@ -164,7 +164,12 @@ def add_to_cron(self, version: str) -> None:
         with open(jenkinsfile, "r") as f:
             data = f.read()
 
-        cron_entry = f"H 1 * * * %INPUT_MANIFEST={version}/{self.prefix}-{version}.yml;TARGET_JOB_NAME=distribution-build-{self.prefix}\n"
+        build_platform_dict = {"opensearch": "linux macos windows", "opensearch-dashboards": "linux windows"}
+        build_platform = build_platform_dict.get(self.prefix, "linux macos windows")
```
prudhvigodithi, line 168: NIT: Hey @zelinh, thanks for the PR,  can we get set the default to just `linux` ?

gaiksaya: +1 I believe this will fail for 1.x version where windows is not supported yet

zelinh: @prudhvigodithi hi, as the discussion in this issue https://github.com/opensearch-project/opensearch-build/issues/2912, I think having three as default would be the target in long run. We are also planning to support windows for 1.x series but haven't decided which version to start. No strong objective though since this default won't be used in most cases. I would update this to `linux`.

src/manifests_workflow/input_manifests.py
```diff
@@ -164,7 +164,12 @@ def add_to_cron(self, version: str) -> None:
         with open(jenkinsfile, "r") as f:
             data = f.read()
 
-        cron_entry = f"H 1 * * * %INPUT_MANIFEST={version}/{self.prefix}-{version}.yml;TARGET_JOB_NAME=distribution-build-{self.prefix}\n"
+        build_platform_dict = {"opensearch": "linux macos windows", "opensearch-dashboards": "linux windows"}
```
dblock, line 167: Make this a CONSTANT?

zelinh: Hey @dblock I change this dict to a class method so I believe this would be better for maintenance or we could move it to the top of this class as a fixed variable. Let me know what do you think? Thanks.

dblock: There's nothing dynamic about this map, so since it's a constant, the convention is to declare it as such, e.g. https://github.com/opensearch-project/opensearch-build/blob/e6c5ce66ab7d702bf347da011a83129c3e5aeb82/src/test_workflow/integ_test/integ_test_runners.py#L17

zelinh: Oh I see. Yeah that makes more sense. Updating it. Thanks.

src/manifests_workflow/input_manifests.py
```diff
@@ -164,7 +169,11 @@ def add_to_cron(self, version: str) -> None:
         with open(jenkinsfile, "r") as f:
             data = f.read()
 
-        cron_entry = f"H 1 * * * %INPUT_MANIFEST={version}/{self.prefix}-{version}.yml;TARGET_JOB_NAME=distribution-build-{self.prefix}\n"
+        build_platform = self.BUILD_PLATFORM.get(self.prefix, "linux")
+
+        cron_entry = f"H 1 * * * %INPUT_MANIFEST={version}/{self.prefix}-{version}.yml;" \
```
dblock, line 174: If I were to nitpick, I would make this an array and call `join` with `;\n` to avoid `\`.

dblock (approved):

PR #2926, opened 2022-11-19
Update build cron for 2.5.0 to every hour

bowenlan-amzn:
Signed-off-by: bowenlan-amzn <bowenlan23@gmail.com>

### Description
Update the cron schedule of build workflow for 2.5.0 to every hour

### Issues Resolved

### DIFF
--- a/jenkins/check-for-build.jenkinsfile
+++ b/jenkins/check-for-build.jenkinsfile
@@ -14,10 +14,10 @@ pipeline {
     }
     triggers {
         parameterizedCron '''
-            H 1 * * * %INPUT_MANIFEST=2.5.0/opensearch-dashboards-2.5.0.yml;TARGET_JOB_NAME=distribution-build-opensearch-dashboards;BUILD_PLATFORM=linux windows
+            H/60 * * * %INPUT_MANIFEST=2.5.0/opensearch-2.5.0.yml;TARGET_JOB_NAME=distribution-build-opensearch;BUILD_PLATFORM=linux macos windows
+            H/60 * * * %INPUT_MANIFEST=2.5.0/opensearch-dashboards-2.5.0.yml;TARGET_JOB_NAME=distribution-build-opensearch-dashboards;BUILD_PLATFORM=linux windows
             H 1 * * * %INPUT_MANIFEST=2.4.1/opensearch-dashboards-2.4.1.yml;TARGET_JOB_NAME=distribution-build-opensearch-dashboards;BUILD_PLATFORM=linux windows
             H 1 * * * %INPUT_MANIFEST=2.4.1/opensearch-2.4.1.yml;TARGET_JOB_NAME=distribution-build-opensearch;BUILD_PLATFORM=linux macos windows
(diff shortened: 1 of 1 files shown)


peterzhuamazon:
Hi @gaiksaya @prudhvigodithi I have discussed with @bowenlan-amzn offline that they want to start building 2.5.0 right now in a more frequent fashion.

Please add you thoughts on whether this is fine for now.

Thanks.

prudhvigodithi:
Hey @peterzhuamazon and @bowenlan-amzn LGTM.

prudhvigodithi (approved):

zelinh (commented):
I'm good with increasing the frequency for the builds on 2.5.0 but just notice we only have core and CU plugin right now. Prob would need to add those plugins that are being actively developed to the manifest as well.

gaiksaya (approved):

peterzhuamazon (approved):

peterzhuamazon:
> I'm good with increasing the frequency for the builds on 2.5.0 but just notice we only have core and CU plugin right now. Prob would need to add those plugins that are being actively developed to the manifest as well.

Yeah, right now @bowenlan-amzn is waiting for the new CU maven to work on their plugins.
I would suggest Bowen add this line as well for a more frequent Core+CU deployment that does not impact the artifacts.

```
            H/60 * * * %INPUT_MANIFEST=2.5.0/opensearch-2.5.0.yml;TARGET_JOB_NAME=distribution-build-opensearch;BUILD_PLATFORM=linux;COMPONENTS=OpenSearch common-utils;UPDATE_LATEST_URL=false
```

This probably need to make changes here so maven can be constantly updated without interrupting actual artifacts creation.
https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/check-for-build.jenkinsfile#L89-L93

Thanks.

gaiksaya:
I would keep keep default to `true` and work on https://github.com/opensearch-project/opensearch-build/issues/2749

Posting some suggestions for anyone to pick up:
Currently, the snapshots are published only after the build is successful https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L152-L171
I believe if we want to unblock dependency on entire build, the publishing to snapshot should be a part of this stage https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L178-L218

peterzhuamazon:
> I would keep keep default to `true` and work on #2749
> 
> Posting some suggestions for anyone to pick up: Currently, the snapshots are published only after the build is successful https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L152-L171 I believe if we want to unblock dependency on entire build, the publishing to snapshot should be a part of this stage https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L178-L218

That is also fine I just want to unblock @bowenlan-amzn right now just in case other failures block core/CU maven updates.

Thanks.

PR #3003, opened 2022-12-13
chore: fix rename references

derek-ho:
Signed-off-by: Derek Ho <dxho@amazon.com>

### Description
_Describe what this change achieves._

### Issues Resolved

### DIFF
--- a/manifests/1.3.7/opensearch-dashboards-1.3.7.yml
+++ b/manifests/1.3.7/opensearch-dashboards-1.3.7.yml
@@ -38,6 +38,6 @@ components:
     repository: https://github.com/opensearch-project/index-management-dashboards-plugin
     ref: tags/1.3.7.0
   - name: reportsDashboards
-    repository: https://github.com/opensearch-project/dashboards-reports.git
+    repository: https://github.com/opensearch-project/reporting.git
     working_directory: dashboards-reports
     ref: tags/1.3.7.0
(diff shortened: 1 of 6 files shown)


dblock:
Are these required? I believe the references are stable.

zelinh (commented):
Seems like the CI still fails on dashboards-reporting since it's not accessible. Could you take care of it as well. @derek-ho

derek-ho:
@zelinh @gaiksaya I think I fixed all of it and removed an extra reference in the 3.0.0 manifest, please review, thanks!

manifests/1.3.7/opensearch-dashboards-1.3.7.yml
```diff
@@ -28,7 +28,7 @@ components:
     working_directory: gantt-chart
     ref: tags/1.3.7.0
   - name: observabilityDashboards
-    repository: https://github.com/opensearch-project/dashboards-observability.git
+    repository: https://github.com/opensearch-project/observability.git
```
gaiksaya, line 31: Since this is public? Should we be pointing the repo to https://github.com/opensearch-project/dashboards-observability

zelinh: If we change it to dashboards-observability, the manifest check would be failing as we haven't created tags for that repo. I think we would need some discussions for how to modify our manifests accordance to these repo name changes.

derek-ho: @gaiksaya @zelinh +1 to what Zelin is saying I don't think it's right to use `dashboards-observability` in this case, or at least if it should be changed then all of the manifest files should be changed in my opinion. Please let me know of the result of those discussions, I can modify the PR as necessary or let me know if I can be of any help in those discussions. Thanks!

zelinh (approved):
I'm good with this PR. Any other concerns before I merge this? @gaiksaya @prudhvigodithi @peterzhuamazon

gaiksaya (approved):

prudhvigodithi (approved):

Issue #2464, opened 2022-08-16
[Bug]: Local test cluster logs for integ-test generated files are empty


gaiksaya:
### Describe the bug

The test recorder records logs for integ-test as below:
1. Local cluster logs:

- stdout: https://ci.opensearch.org/ci/dbc/integ-test/2.1.0/5740/linux/x64/tar/test-results/1/integ-test/alerting/with-security/local-cluster-logs/stdout.txt
- stderr: https://ci.opensearch.org/ci/dbc/integ-test/2.1.0/5740/linux/x64/tar/test-results/1/integ-test/alerting/with-security/local-cluster-logs/stderr.txt

2. Plugin actual integ-test logs:

- stdout:  https://ci.opensearch.org/ci/dbc/integ-test/2.1.0/5740/linux/x64/tar/test-results/1/integ-test/alerting/with-security/test-results/stdout.txt
- stderr:https://ci.opensearch.org/ci/dbc/integ-test/2.1.0/5740/linux/x64/tar/test-results/1/integ-test/alerting/with-security/test-results/stderr.txt
- <component>.yml : https://ci.opensearch.org/ci/dbc/integ-test/2.1.0/5740/linux/x64/tar/test-results/1/integ-test/alerting/with-security/test-results/alerting.yml

However, the local cluster logs do not have any data in them. See above links

### To reproduce

Click the above urls or try running integ-test locally and see that local-test cluster logs are empty

### Expected behavior

The txt files should have data about the cluster.

jordarlu:
@rishabh6788 commented how to fix this issue on https://github.com/opensearch-project/opensearch-build/pull/3016#issuecomment-1368177810

bbarani:
@gaiksaya Can we close this issue since we have merged the fix?

gaiksaya:
Yes the issue is fixed as seen in recent integ-test run https://ci.opensearch.org/ci/dbc/integ-test/1.3.8/7019/linux/x64/tar/test-results/3754/integ-test/alerting/with-security/local-cluster-logs/stdout.txt

PR #3016, opened 2022-12-16
Changes to Process file in System workflow

Divyaasm:
Signed-off by: Divya Madala divyaasm@amazon.com 

### Description
Changes to process.py to fix local test cluster logs and a fix to access the stdout data from the subprocess command.

### Issues Resolved


src/system/process.py
```diff
@@ -78,10 +80,14 @@ def pid(self) -> int:
 
     @property
     def stdout_data(self) -> Any:
+        if self.started:
+            self.stdout.seek(0)
```
dblock, line 84: This will break `if not self.stdout`.

dblock (changes requested):
Could you please explain why this works?

- sleep after a process starts seems unnecessary
- rewinding data makes `stdout_data` always re-read everything, is that the intention?

tianleh:
Could you please add more details about the motivation for this change in the PR description? @Divyaasm

src/system/process.py
```diff
@@ -34,6 +35,7 @@ def start(self, command: str, cwd: str) -> None:
             stdout=self.stdout,
             stderr=self.stderr,
         )
+        time.sleep(1)
```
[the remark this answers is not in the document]
Divyaasm: Hi @dblock,

I see that the output of the subprocess is not redirected to stdout which is leaving the file empty.So tried by giving it some time to write into stdout before we read it in stdout_data, even a time lapse of 0.1 works here.

And the file pointer is at the end of the file when we tried reading it in stdout_data, so got it to starting using seek().

src/system/process.py
Divyaasm: seek(0) can only be called before we terminate the process. So this throws an error if we try to access stdout_data after terminating the process.So I tried to return an empty string in that case.

src/system/process.py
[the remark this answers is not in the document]
peterzhuamazon: I think instead of adding a delay of fixed 1 second, you can try to check if there is data in stdout before reading.

If someone use an extremely slow server such as t3.small or worse instance, the assumption of 0.1/1 second might not be valid anymore.

It is generally better to do a periodic check in my opinion on the data availability here.

src/system/process.py
peterzhuamazon: If there is already data in stdout, and you do not add `seek(0)`, what would be the outcome of it?

peterzhuamazon:
Hi @Divyaasm add some comment on @dblock concerns.

Thanks.

src/system/process.py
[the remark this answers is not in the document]
dblock: I am not sure what problem we're trying to solve in this code. If the process runs and never has any output, then waiting for output will loop indefinitely.  

This method starts a process. It should not have any other side effects.

dblock (changes requested):
@Divyaasm Now I am confused to what the actual problem you're trying to solve is. I think you might want to reproduce the issue in a test, then we can see what the best way to fix it is?

Divyaasm:
> @Divyaasm Now I am confused to what the actual problem you're trying to solve is. When are these logs empty? I think you might want to reproduce the issue in a test, then we can see what the best way to fix it is?

Hi @dblock, I tried to add some print statements for testing in the start() function and tried to show the data in stdout in different cases by creating an object to the Process class and passing an "ls" command to the subprocess.I have added images for reference to depict the issue.

Screen Shot 2022-12-27 at 10 47 33 AM

Output:

Screen Shot 2022-12-27 at 10 49 12 AM

src/system/process.py
Divyaasm: I see an empty string while reading stdout or stderr, if I do not add seek(0).

dblock:
@Divyaasm Makes sense. Next, write a unit test that reproduces the failure that you have in those screenshots.

Divyaasm:
> @Divyaasm Makes sense. Next, write a unit test that reproduces the failure that you have in those screenshots.
> 
> I think what's missing is waiting for the process to finish, no? Once it's finished you have all the data in stdout/stderr.

Yes @dblock, giving some wait time redirects the data into stdout/stderr. I will further add additional test-cases for the changed code. Thanks

dblock:
> > @Divyaasm Makes sense. Next, write a unit test that reproduces the failure that you have in those screenshots.
> > I think what's missing is waiting for the process to finish, no? Once it's finished you have all the data in stdout/stderr.
> 
> Yes @dblock, giving some wait time redirects the data into stdout/stderr. I will further add additional test-cases for the changed code. Thanks

In your use case, does the process finish by the time you need the stdout output? If so, would using a more reliable way to wait for the process to do so, instead of waiting some number of seconds, work?

Divyaasm:
> > > @Divyaasm Makes sense. Next, write a unit test that reproduces the failure that you have in those screenshots.
> > > I think what's missing is waiting for the process to finish, no? Once it's finished you have all the data in stdout/stderr.
> > 
> > 
> > Yes @dblock, giving some wait time redirects the data into stdout/stderr. I will further add additional test-cases for the changed code. Thanks
> 
> In your use case, does the process finish by the time you need the stdout output? If so, would using a more reliable way to wait for the process to do so, instead of waiting some number of seconds, work?

We can do that by using wait() function or passing timeout value to subprocess command, but ultimately they lead to waiting for a number seconds to let the process complete.

rishabh6788:
`self.__stdout_data__ = self.stdout.read()` wouldn't work as `self.stdout` is basically reference to a file. To read the content of `self.stdout` can you please replace the current code [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/system/process.py#L57-L59) with below and try?

```
self.__stdout_data__ = open(self.stdout.name, 'r').read()
self.stdout.close()
self.stdout = None
```

Do the same for `stderr` as well. 

There should be no need to add sleep or wait. 
@Divyaasm @dblock @peterzhuamazon

jordarlu:
> `self.__stdout_data__ = self.stdout.read()` wouldn't work as `self.stdout` is basically reference to a file. To read the content of `self.stdout` can you please replace the current code [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/system/process.py#L57-L59) with below and try?
> 
> ```
> self.__stdout_data__ = open(self.stdout.name, 'r').read()
> self.stdout.close()
> self.stdout = None
> ```
> 
> Do the same for `stderr` as well.
> 
> There should be no need to add sleep or wait. @Divyaasm @dblock @peterzhuamazon

Thank you, @rishabh6788 , you nailed it!!
I tested the integ-test local cluster log empty issue with the recommendation you mentioned above and it worked!
@Divyaasm can further confirm from her end, but I think your recommendation fixed https://github.com/opensearch-project/opensearch-build/issues/2464
I will cancel my PR https://github.com/opensearch-project/opensearch-build/pull/3034. Thanks!!

Divyaasm:
Hi @peterzhuamazon,
Could you please take a look why a windows testcase failed due to the changes made in the file. I see ubuntu and mac-os successfully ran all the testcases.

Divyaasm:
I see the test cluster logs are now generated. And to access the data in stdout and stderr from user side at any instance of time using time.sleep() on the user end fetches the values from stdout_data and stderr_data

rishabh6788:
@dblock @peterzhuamazon There is an issue with reading `tempfile.NamedTemporaryFile` in windows https://bugs.python.org/issue14243. The work around is to set `delete=False` while calling `tempfile.NamedTemporaryFile()` method and take care of the deletion manually. 
I have added the changes in our `process.py` file to fix the empty cluster logs file issue, please review and provide your comments. 
I have ignored testing for file's existence before removing as in our we are actually creating temp files at the start of the process and are also testing their existence condition as well. Let me know if you have any better way to handle deletion. 
@Divyaasm @jordarlu

peterzhuamazon:
> @dblock @peterzhuamazon There is an issue with reading `tempfile.NamedTemporaryFile` in windows https://bugs.python.org/issue14243. The work around is to set `delete=False` while calling `tempfile.NamedTemporaryFile()` method and take care of the deletion manually. I have added the changes in our `process.py` file to fix the empty cluster logs file issue, please review and provide your comments. I have ignored testing for file's existence before removing as in our we are actually creating temp files at the start of the process and are also testing their existence condition as well. Let me know if you have any better way to handle deletion. @Divyaasm @jordarlu

Thanks @rishabh6788 seems like a better approach than my original deleting 3 times on Windows method.
Probably we are utilize the same for this: https://github.com/opensearch-project/opensearch-build/blob/main/src/system/temporary_directory.py#L39

peterzhuamazon (approved):
I am good with this approach, @dblock please let me know your thought.

Thanks.

rishabh6788:
@dblock are you okay with the proposed update?

Issue #3030, opened 2022-12-23
Fix to Process.py from system workflow having empty logs in stdout and stderr 


Divyaasm:
process.py to run and track subprocess commands runs the command in the background but does not store output and errors  into respective stdout and strerr properties.

rishabh6788:
Please provide related code references where you encountered that the logs are not getting written to stdout/stderr. @Divyaasm

dblock:
I am guessing this relates to https://github.com/opensearch-project/opensearch-build/pull/3016

PR #3034, opened 2022-12-27
Fix for #2464 - Local test cluster logs for integ-test generated files are empty 

jordarlu:
Signed-off-by: Jeff Lu <chunglu@amazon.com>

### Description
This is the PR to address issue #2464

### Issues Resolved
At current, when running the integ_test, the local cluster logs are not logged with any data as described in #2464.
It looks to be the original design on test_cluster.py to record only after the process is terminated by using Process.py.

In order to have stdout and stderr(if happend) written in the local cluster logs for better visibility of local cluster process, this PR will add the stdout and stderr taken from the step where the cluster is started in the local cluster logs.

### DIFF
--- a/src/test_workflow/integ_test/service_opensearch.py
+++ b/src/test_workflow/integ_test/service_opensearch.py
@@ -24,8 +24,8 @@ class ServiceOpenSearch(Service):
     install_dir: str
     opensearch_yml_dir: str
     security_plugin_dir: str
-    trans_stdout: str
-    trans_stderr: str
+    transport_stdout: str
+    transport_stderr: str
 
     def __init__(
         self,
@@ -36,8 +36,6 @@ class ServiceOpenSearch(Service):
         dependency_installer: DependencyInstaller,
         work_dir: str,
         cluster_port: int = 9200,
-        trans_stdout: str = "",
-        trans_stderr: str = ""
     ) -> None:
         super().__init__(work_dir, version, distribution, security_enabled, additional_config, dependency_installer)
         self.cluster_port = cluster_port
@@ -60,8 +58,8 @@ class ServiceOpenSearch(Service):
         self.process_handler.start(self.dist.start_cmd, self.install_dir)
         logging.info(f"Started OpenSearch with parent PID {self.process_handler.pid}")
 
-        ServiceOpenSearch.trans_stdout = self.process_handler.stdout_data
-        ServiceOpenSearch.trans_stderr = self.process_handler.stderr_data
+        ServiceOpenSearch.transport_stdout = self.process_handler.stdout_data
+        ServiceOpenSearch.transport_stderr = self.process_handler.stderr_data
 
     def uninstall(self) -> None:
         self.dist.uninstall()
--- a/src/test_workflow/integ_test/service_opensearch.py
+++ b/src/test_workflow/integ_test/service_opensearch.py
@@ -24,6 +24,8 @@ class ServiceOpenSearch(Service):
     install_dir: str
     opensearch_yml_dir: str
     security_plugin_dir: str
+    trans_stdout: str
+    trans_stderr: str
 
     def __init__(
         self,
@@ -33,7 +35,9 @@ class ServiceOpenSearch(Service):
         security_enabled: bool,
         dependency_installer: DependencyInstaller,
         work_dir: str,
-        cluster_port: int = 9200
+        cluster_port: int = 9200,
+        trans_stdout: str = "",
+        trans_stderr: str = ""
     ) -> None:
         super().__init__(work_dir, version, distribution, security_enabled, additional_config, dependency_installer)
         self.cluster_port = cluster_port
@@ -55,6 +59,9 @@ class ServiceOpenSearch(Service):
 
         self.process_handler.start(self.dist.start_cmd, self.install_dir)
         logging.info(f"Started OpenSearch with parent PID {self.process_handler.pid}")
+
+        ServiceOpenSearch.trans_stdout = self.process_handler.stdout_data
+        ServiceOpenSearch.trans_stderr = self.process_handler.stderr_data
 
     def uninstall(self) -> None:
         self.dist.uninstall()
--- a/src/test_workflow/integ_test/service_opensearch.py
+++ b/src/test_workflow/integ_test/service_opensearch.py
@@ -24,8 +24,6 @@ class ServiceOpenSearch(Service):
     install_dir: str
     opensearch_yml_dir: str
     security_plugin_dir: str
-    transport_stdout: str
-    transport_stderr: str
 
     def __init__(
         self,
@@ -35,7 +33,7 @@ class ServiceOpenSearch(Service):
         security_enabled: bool,
         dependency_installer: DependencyInstaller,
         work_dir: str,
-        cluster_port: int = 9200,
+        cluster_port: int = 9200
     ) -> None:
         super().__init__(work_dir, version, distribution, security_enabled, additional_config, dependency_installer)
         self.cluster_port = cluster_port
@@ -57,9 +55,6 @@ class ServiceOpenSearch(Service):
 
         self.process_handler.start(self.dist.start_cmd, self.install_dir)
         logging.info(f"Started OpenSearch with parent PID {self.process_handler.pid}")
-
-        ServiceOpenSearch.transport_stdout = self.process_handler.stdout_data
-        ServiceOpenSearch.transport_stderr = self.process_handler.stderr_data
 
     def uninstall(self) -> None:
         self.dist.uninstall()
--- a/src/test_workflow/test_cluster.py
+++ b/src/test_workflow/test_cluster.py
@@ -48,8 +48,8 @@ class TestCluster(abc.ABC):
         self.save_logs = save_logs
         self.all_services = []
         self.termination_result = None
-        self.start_stdout = ""
-        self.start_stderr = ""
+        self.start_stdout_data = ""
+        self.start_stderr_data = ""
 
     @classmethod
     @contextmanager
@@ -82,8 +82,8 @@ class TestCluster(abc.ABC):
         for service in self.all_services:
             service.wait_for_service()
 
-        self.start_stdout = self.service.process_handler.stdout_data
-        self.start_stderr = self.service.process_handler.stderr_data
+        self.start_stdout_data = self.service.process_handler.stdout_data
+        self.start_stderr_data = self.service.process_handler.stderr_data
 
     def terminate(self) -> None:
         if self.service:
@@ -103,8 +103,8 @@ class TestCluster(abc.ABC):
             self.component_name,
             self.component_test_config,
             termination_result.return_code,
-            self.start_stdout + termination_result.stdout_data,
-            self.start_stderr + termination_result.stderr_data,
+            self.start_stdout_data + termination_result.stdout_data,
+            self.start_stderr_data + termination_result.stderr_data,
             termination_result.log_files
         )
 
--- a/src/test_workflow/test_cluster.py
+++ b/src/test_workflow/test_cluster.py
@@ -11,8 +11,10 @@ from contextlib import contextmanager
 from typing import Any, Generator, List, Tuple
 
 from test_workflow.integ_test.service import Service
+from test_workflow.integ_test.service_opensearch import ServiceOpenSearch
 from test_workflow.integ_test.service_termination_result import ServiceTerminationResult
 from test_workflow.test_recorder.log_recorder import LogRecorder
+from test_workflow.test_recorder.test_recorder import TestRecorder
 from test_workflow.test_recorder.test_result_data import TestResultData
 
 
@@ -37,7 +39,7 @@ class TestCluster(abc.ABC):
         component_test_config: str,
         security_enabled: bool,
         additional_cluster_config: dict,
-        save_logs: LogRecorder,
+        test_recorder: TestRecorder,
         cluster_port: int = 9200
     ) -> None:
         self.work_dir = os.path.join(work_dir, "local-test-cluster")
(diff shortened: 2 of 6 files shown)


jordarlu:
There is a Process.py issue that is currently trcked on #3030 and #3016 
This PR #3034 can be merged only after the Process.py issue is resolved.

jordarlu:
The step I tested at my dev-desktop : 
1.  to replace the src/system/process.py on dev-desktop by taking the file from #3016 (currently under review) temporaryly 
2.  run the following at dev-desktop
`./test.sh integ-test manifests/2.4.1/opensearch-2.4.1-test.yml --component ml-commons --test-run-id 1 --paths opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.4.1/6667/linux/x64/tar/`

(In order to get the result faster, the above command is to run integ-test on only 1 component, the ml-commons, instead of all)
it takes about 10 minutes to run , and the following is to show the local_test_cluster is with the data after the PR  : 

```
dev-dsk-chunglu-2c-2d46bbcc % ./test.sh integ-test manifests/2.4.1/opensearch-2.4.1-test.yml --component ml-commons --test-run-id 1 --paths opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.4.1/6667/linux/x64/tar/
Installing dependencies from Pipfile.lock (b36c9c)...
Running ./src/run_integ_test.py manifests/2.4.1/opensearch-2.4.1-test.yml --component ml-commons --test-run-id 1 --paths opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.4.1/6667/linux/x64/tar/ ...
2022-12-27 21:10:34 INFO     Loading /local/home/chunglu/opensearch-build-git/opensearch-build/manifests/2.4.1/opensearch-2.4.1-test.yml
.....
.....
2022-12-27 21:18:23 INFO     | ml-commons           | with-security        | PASS  |    0 |

(22-12-27 21:18:33) <0> [~/opensearch-build-git/opensearch-build]  
dev-dsk-chunglu-2c-2d46bbcc % ls -l ./test-results/1/integ-test/ml-commons/with-security/local-cluster-logs 
total 4
-rw-r--r-- 1 chunglu amazon    0 Dec 27 21:18 stderr.txt
-rw-r--r-- 1 chunglu amazon 2237 Dec 27 21:18 stdout.txt

(22-12-27 21:18:46) <0> [~/opensearch-build-git/opensearch-build]  
dev-dsk-chunglu-2c-2d46bbcc % 
```

src/test_workflow/integ_test/service_opensearch.py
```diff
@@ -33,7 +35,9 @@ def __init__(
         security_enabled: bool,
         dependency_installer: DependencyInstaller,
         work_dir: str,
-        cluster_port: int = 9200
+        cluster_port: int = 9200,
+        trans_stdout: str = "",
+        trans_stderr: str = ""
```
peterzhuamazon, line 40: Why is the value initialize to empty?

src/test_workflow/integ_test/service_opensearch.py
```diff
@@ -24,6 +24,8 @@ class ServiceOpenSearch(Service):
     install_dir: str
     opensearch_yml_dir: str
     security_plugin_dir: str
+    trans_stdout: str
+    trans_stderr: str
```
peterzhuamazon, line 28: Does this mean `transport_stdout` and `transport_stderr`?
If so, please use full name here or it will be very confusing later on.
Thanks.

src/test_workflow/test_cluster.py
```diff
@@ -98,8 +101,8 @@ def __save_test_result_data(self, termination_result: ServiceTerminationResult)
             self.component_name,
             self.component_test_config,
             termination_result.return_code,
-            termination_result.stdout_data,
-            termination_result.stderr_data,
+            ServiceOpenSearch.trans_stdout + termination_result.stdout_data,
+            ServiceOpenSearch.trans_stderr + termination_result.stderr_data,
```
peterzhuamazon, line 105: Is this supposed to be `ServiceOpenSearch.trans_*` or an instance of the `ServiceOpenSearch` class?

tests/tests_test_workflow/test_integ_workflow/integ_test/test_local_test_cluster.py
```diff
@@ -32,6 +33,9 @@ def setUp(self) -> None:
         self.dependency_installer = None
         self.test_recorder = None
 
+        ServiceOpenSearch.trans_stdout = ""
+        ServiceOpenSearch.trans_stderr = ""
```
peterzhuamazon, line 37: Same as above, why is it initialized to empty string?
If you have a new instance of the class, it should default to empty or None isnt it?

src/test_workflow/test_cluster.py
```diff
@@ -11,8 +11,10 @@
 from typing import Any, Generator, List, Tuple
 
 from test_workflow.integ_test.service import Service
+from test_workflow.integ_test.service_opensearch import ServiceOpenSearch
```
peterzhuamazon, line 14: I see you also change OSD above but here you only set SeviceOpenSearch, do you also need to set up ServiceOpenSearchDashboards?

Thanks.

peterzhuamazon (changes requested):
Hi @jordarlu,

A few confusions on the PR.

Please take a look and let me know what you think. Thanks.

src/test_workflow/integ_test/service_opensearch.py
jordarlu: sorry, It was left over from my other earlier attempts,, I have them removed from my latest update. Thank you for noticing that..

src/test_workflow/integ_test/service_opensearch.py
jordarlu: I got it, and it makes sense, I have got them changed to transport_stdout and transport_stderr. Thanks for the guidence.

src/test_workflow/test_cluster.py
jordarlu: What I tried to do was to set-up two new class-level arguments in class ServiceOpenSearch, and then, assign the stdout and stderr when the start() method is triggered. 
Because they are class-level arguments, I can access them by doing like ServiceOpenSearch.trans_stdout.

I actually tried using instence level arguments and applying setter and getter. It did work but I couldn't find the way to Pass the mypy test :(  ; hence, I turned to use class-level arguments instead..  maybe you foresee any issue by doing this way? Thanks!

tests/tests_test_workflow/test_integ_workflow/integ_test/test_local_test_cluster.py
jordarlu: That is another good point you mentioned. 
Since it is the test validation on local_test_cluster.py, I think I can actually add some test strings on them. I have added test string and updated the assertion expectation accordingly in the latest update. Hope that works better. Thanks!

src/test_workflow/test_cluster.py
jordarlu: mm.. I am not certain if there are the same expectation to ServiceOpenSearchDashboards tho, maybe I can check with @gaiksaya when she is back to work?

But if you are referring to the update I made on [src/test_workflow/integ_test/local_test_cluster_opensearch_dashboards.py](https://github.com/opensearch-project/opensearch-build/pull/3034/files#diff-be5ff0e69b8c84d6471f8ca99a9af4ad6093ca25142ae6ac1f9ba7ab5aa5fdce), the reson behind it is I touched the parent class, the TestCluster; hence, I need to upddate both OS and OSD accordinly to prevent error on validation. Thanks!!

jordarlu:
> Hi @jordarlu,
> 
> A few confusions on the PR.
> 
> Please take a look and let me know what you think. Thanks.

Thanks a lot for reviewing the PR and providing me great comments duing your PTO, @peterzhuamazon.. I really appreciate it...  I made some updated and response accordingly, hope that remove few doubts ... thanks!!

src/test_workflow/integ_test/service_opensearch.py
```diff
@@ -56,6 +58,9 @@ def start(self) -> None:
         self.process_handler.start(self.dist.start_cmd, self.install_dir)
         logging.info(f"Started OpenSearch with parent PID {self.process_handler.pid}")
 
+        ServiceOpenSearch.transport_stdout = self.process_handler.stdout_data
+        ServiceOpenSearch.transport_stderr = self.process_handler.stderr_data
```
dblock, line 62: This sets a global value, should be on the instance, `self`.

dblock (changes requested):
I think there's some confusion in where the output needs to live. With this change you can't have multiple instances of `ServiceOpenSeaerch`.

src/test_workflow/integ_test/service_opensearch.py
jordarlu: > Member

Makes sense, @dblock .. indeed, it will not be able to support multiple instances ... let me go back to work on the instence level solution. Thank you!!
As for the test case, I think #3034 may not be a bug but enhancement ( adding local-cluster process log when starting the cluster ) , let me also think how to improve the test cases for it ... thanks again!

jordarlu:
Hi, @peterzhuamazon and @dblock 
Looks like we just need to alter test_cluster.py to add the stdout & stderr taken from the start() into the result generated at the terminate(). 

here is the test result on my dev-desktop. 
```
dev-dsk-chunglu-2c-2d46bbcc % ls -l ./test-results/1/integ-test/ml-commons/with-security/local-cluster-logs 
total 36
-rw-r--r-- 1 chunglu amazon   844 Dec 29 04:41 stderr.txt
-rw-r--r-- 1 chunglu amazon 32278 Dec 29 04:41 stdout.txt

(22-12-29 5:33:08) <0> [~/opensearch-build-git/opensearch-build]  
dev-dsk-chunglu-2c-2d46bbcc % ls -l ./test-results/1/integ-test/ml-commons/with-security/test-results 
total 16
-rw-r--r-- 1 chunglu amazon  104 Dec 29 04:41 ml-commons.yml
-rw-r--r-- 1 chunglu amazon 4159 Dec 29 04:41 stderr.txt
-rw-r--r-- 1 chunglu amazon 1867 Dec 29 04:41 stdout.txt
```
hope it works better this time, and if any comments, thanks a lot !!

src/test_workflow/test_cluster.py
```diff
@@ -48,6 +48,8 @@ def __init__(
         self.save_logs = save_logs
         self.all_services = []
         self.termination_result = None
+        self.start_stdout = ""
```
dblock, line 51: Semantically this is incorrect, the default stdout/stderr is `None`. Set it to that?

src/test_workflow/test_cluster.py
```diff
@@ -80,6 +82,9 @@ def start(self) -> None:
         for service in self.all_services:
             service.wait_for_service()
 
+        self.start_stdout = self.service.process_handler.stdout_data
```
dblock, line 85: Should this be called `stdout_data` instead? Why is it `start`?

dblock (changes requested):
I am a bit confused to what this fixes. Care to write a test?

From the code with the change it looks like 1) a process starts 2) we read some data at a random time 3) the process is terminated 4) we save data when it started + data when it ended. This should be the same as just reading data when it ended.

src/test_workflow/test_cluster.py
jordarlu: Hi, @dblock , thanks for the asking..because later at L106-107, I want to concatenate two strings, so, I

jordarlu:
> 

Hi, @dblock , I think the expectation was as you mentioned. The local_cluster stdout & stderr outputs were supposed to be written into the local_cluster logs after the process was terminated; however, as described at #2464 , local cluster logs were empty becasue the stdout and stderr were empty after the Terminate().
I found that the stdout and stderr were logged at Start(), so, I proposed to add them into the local cluster log at L106-107.

A possible test for this, maybe I can just assert if the stdout + stderr generated by process_handler is Not empty ? because they are empty currently. 
thanks!

src/test_workflow/test_cluster.py
jordarlu: thanks for the asking again..
I was thinking of using the argument names that can represent in which stage that the data was generated.
On L51, 52, I set the arguments to be start_stdout, and start_stderr as the outputs generated at the stage start().
Will you suggest changing to start_stdout_data and start_stderr_data? Maybe that will be less confusion? Thanks!

jordarlu: Updated with start_stdout_data and start_stderr_data. thanks

dblock:
@jordarlu was the process terminated or did it finish successfully? 

I don't think this is the right fix. If we're launching a process, we need to be waiting for it, and potentially reading stdout continuously until it's done. It's possible that by reading some output right after the process starts you get some data, but I don't think you've fixed the bug.

jordarlu:
> @jordarlu was the process terminated or did it finish successfully?
> 
> I don't think this is the right fix. If we're launching a process, we need to be waiting for it, and potentially reading stdout continuously until it's done. It's possible that by reading some output right after the process starts you get some data, but I don't think you've fixed the bug.

Hi, @dblock , the process was finished successfully, and the process_handler.stdout and .stderr were just empty at that time.

jordarlu:
> > @jordarlu was the process terminated or did it finish successfully?
> > I don't think this is the right fix. If we're launching a process, we need to be waiting for it, and potentially reading stdout continuously until it's done. It's possible that by reading some output right after the process starts you get some data, but I don't think you've fixed the bug.
> 
> Hi, @dblock , the process was finished successfully, and the process_handler.stdout and .stderr were just empty at that time.

Thanks, @dblock , let me keep looking into it to see why the Terminate() outpout/stdout&stderr was not logged properly.

jordarlu:
@rishabh6788 has found where the issue and how to fix it. His recommendation is at https://github.com/opensearch-project/opensearch-build/pull/3016#issuecomment-1368177810

Issue #2883, opened 2022-11-11
Validation workflow for Tar,Rpm and Yum distributions


Divyaasm:
The workflow creates a mechanism to automatically validate the tar, rpm and yum distributions.

Divyaasm:
Closed by https://github.com/opensearch-project/opensearch-build/pull/3038

Issue #2759, opened 2022-10-18
Validate the OpenSearch distribution artifacts on DockerHub and ECR


gaiksaya:
Acceptance Criteria:

DockerHub and ECR images need below validations:
* Pull the image
* Validate docker digest/SHA/layers
* Use docker-compose to bring up the cluster
* Run basic apis against the above cluster

jordarlu:
After reviewing the objective with the team, the step#2 'Validate docker digest/SHA/layers' shall cover the following  : 
1. Validating the digest between 'opensearchstaging/opensearch' and 'opensearchproject/opensearch'
2. Validating the digest between 'opensearchstaging/opensearch-dashboards and opensearchproject/opensearch-dashboards'

The expectation is the digest of 'opensearchproject' matching with 'opensearchstaging'

jordarlu:
Close issue as PR was merged. Thanks

PR #3038, opened 2022-12-29
Validation Workflow for Docker and Tar,Rpm,Yum distributions

Divyaasm:
Signed-off-by: Divya Madala <divyaasm@amazon.com>

### Description
PR to adress the issues https://github.com/opensearch-project/opensearch-build/issues/2883 (Tar,Rpm,Yum) and https://github.com/opensearch-project/opensearch-build/issues/2759 (Docker)

### Issues Resolved
Created a validation mechanism to automatically download, install and validate OS and OSD artifacts for Docker, Tar, Rpm, Yum  


src/run_validation.py
```diff
@@ -10,18 +10,30 @@
...
 def main() -> int:
-
     args = ValidationArgs()
 
+    DISTRIBUTION_MAP = {
+        "tar": ValidationTar,
+        "yum": ValidationYum,
+        "rpm": ValidationRpm
+    }
+
```
rishabh6788, line 27: Nit: `Validate` instead of `Validation`

src/validation_workflow/validation_rpm.py
```diff
@@ -25,3 +29,47 @@ def download_artifacts(self, projects: list, version: str) -> bool:
                     logging.info(f"Invalid URL - {url}")
                     raise Exception(f"Invalid url - {url}")
         return True
+
+    @classmethod
+    def installation(self, args: ValidationArgs) -> bool:
```
rishabh6788, line 34: since the methods `installation`, `start_cluster`, `validation` and `stop_cluster` have to be present for each validation type, wouldn't it be a good idea to make them abstract and make sure they are implemented for all the future distribution types?

src/validation_workflow/api_test.py
```diff
@@ -0,0 +1,33 @@
...
+
+import requests
+
+apiHeaders_auth = {"Authorization": "Basic YWRtaW46YWRtaW4="}  # default user/pass "admin/admin" in Base64 format
+apiHeaders_accept = {"Accept": "*/*"}
+apiHeaders_content_type = {"Content-Type": "application/json"}
+apiHeaders = {}
+apiHeaders.update(apiHeaders_auth)
+apiHeaders.update(apiHeaders_accept)
+apiHeaders.update(apiHeaders_content_type)
```
dblock, line 19: Please don't use globals.

src/validation_workflow/validation_rpm.py
dblock: Another way of saying the same, all `@classmethod` should become instance methods and override common implementations in the parent `Validation` class.

dblock (commented):
The class hierarchy needs some work.

src/validation_workflow/api_test.py
jordarlu: Thanks, @dblock , let me work on it and update!

prudhvigodithi:
Hey @Divyaasm @jordarlu can we organize as sub folders (of distribution type) under main folder `validation_workflow`, the more distributions in future the more messy it looks and hard to debug the code, it would be easier to update/debug/add if sorted as sub folders.
Thank you

src/validation_workflow/validation_docker.py
```diff
@@ -0,0 +1,95 @@
...
+from validation_workflow.api_test_cases import TestCases
+from validation_workflow.check_docker_daemon import DockerDaemonRunning
+from validation_workflow.cleanup_docker import CleanupDocker
+from validation_workflow.get_local_docker_imageId import GetDockerImageId
+from validation_workflow.inspect_docker_image import InspectDockerImage
+from validation_workflow.spinup_docker_container import RunDocker
+from validation_workflow.validation_args import ValidationArgs
+
+
+class ValidationDocker():
```
prudhvigodithi, line 20: I can see there are already added abstract methods in file `src/validation_workflow/validation.py`, but not extending the class `Validation`,  (should be as `ValidationDocker(Validation)`)
This is done for the file `src/validation_workflow/validation_rpm.py`, can we follow the same please?

src/run_validation.py
Divyaasm: Wiil change it Rishabh, Thanks

src/validation_workflow/validation_rpm.py
Divyaasm: So does that mean we need to directly declare classmethods in parent class instead of making them abstract and override them in child classes when required.

src/validation_workflow/validation_docker.py
Divyaasm: > Hey @Divyaasm @jordarlu can we organize as sub folders (of distribution type) under main folder `validation_workflow`, the more distributions in future the more messy it looks and hard to debug the code, it would be easier to update/debug/add if sorted as sub folders. We can still keep the generic code out under `validation_workflow`, example for files like `src/validation_workflow/validation.py`, `src/validation_workflow/validation_args.py` but with files `validation_docker.py`, `spinup_docker_container.py` can be inside `src/validation_workflow/docker` folder. Thank you

Sure Prudhvi, that might look good altogether. We will work on it. Thanks

src/validation_workflow/validation_rpm.py
dblock: A class method is a method on the class itself, where the class acts as a namespace. You can call those as `Klass.method`. You almost never need those. An instance method is just a method that can be overridden in a child class. An abstract method just means that all classes in the class hierarchy will have this method, but the implementations are very different, so we don't need to provide a common implementation in the base class.

gaiksaya: Agree to this. That was the plan. All the future distributions too need to have those methods. Override or pass, depends on the respective distribution.

src/validation_workflow/concrete_validation.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+from validation_workflow.validation import Validation
+from validation_workflow.validation_args import ValidationArgs
+
+
+class ConcreteValidation(Validation):
```
gaiksaya, line 12: Why do we need this class when Validation already has all abstract methods that can be implemented by all distributions?
https://github.com/opensearch-project/opensearch-build/pull/3038/files#diff-a318e15081b06176764de2b9bd492c632948b9398c4de3b9fc000145d2c18e89R32-R49

src/validation_workflow/docker/Pipfile
```diff
@@ -0,0 +1,12 @@
...
+verify_ssl = true
+name = "pypi"
+
+[packages]
+install = "*"
+
+[dev-packages]
+
+[requires]
+python_version = "3.11"
```
gaiksaya, line 12: Why different Pipfile? Better to maintain and use the main one to add your dependencies. Similar goes for piplock.

src/validation_workflow/docker/check_docker_daemon.py
```diff
@@ -0,0 +1,27 @@
...
+from typing import Any
+
+"""
+This class is to check if the client (Mac or Linux) has the docker daemon running.
+If the docker daemon is not running, the called shell script will try to bring up the docker daemon.
+It retun 0 if the script runs without error.
+"""
+
+
+class DockerDaemonRunning:
```
gaiksaya, line 19: I dont think we need a different class to check if daemon is running. Add it as one of the validation steps in docker class.

src/validation_workflow/docker/cleanup_docker.py
```diff
@@ -0,0 +1,34 @@
...
+from subprocess import PIPE
+
+yml_template = 'docker-compose.yml'
+
+"""
+This class is used to stop&remove docker containers created at step 3, and remove docker-compose.yml in /tmp/.
+"""
+
+
+class CleanupDocker:
```
gaiksaya, line 19: I dont think even you need a separate class just to clean up. It can be a method under docker  class as well.

src/validation_workflow/validation_args.py
```diff
@@ -7,34 +7,32 @@
...
 
 
 class ValidationArgs:
-    SUPPORTED_PLATFORMS = ["linux"]
-    DISTRIBUTION_MAP = {
-        "tar": ValidationTar,
-        "yum": ValidationYum,
-        "rpm": ValidationRpm
-    }
+    SUPPORTED_PLATFORMS = ["linux", "docker"]
```
gaiksaya, line 14: docker should be a distribution, not a platform

src/run_validation.py
```diff
@@ -10,19 +10,31 @@
...
 
     console.configure(level=args.logging_level)
     logging.getLogger("urllib3").setLevel(logging.WARNING)
 
-    dist = args.DISTRIBUTION_MAP.get(args.distribution, None)
-    dist.download_artifacts(projects=args.projects, version=args.version)
-
-    return 0
+    dist = None
+    if args.platform == 'docker':
```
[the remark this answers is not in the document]
jordarlu: Thanks for the comment, @gaiksaya. The reason we placed the docker as a platform was because what it shown on opensearch.org/downloads.html where it defines Linux, Docker, Windows, and FreeBSD as Platform. We can certainly update it/docker to be one of the distribution.

src/validation_workflow/concrete_validation.py
jordarlu: Hi, @gaiksaya , thanks for asking. There was an error on MyPy when we instantiate the object of the ABC class (validation.py) directly. With a bit research, looks like the ABC class is not meant to be used to create objects directly; hence, we create a concreate subclass of validatioon this way. 
I think mypy has more strick checking; hence, we proposed the concreate class here..
Also, an interesting finding we would love to show you ... if there are only two arguments in RUNNER dictionary, MyPy does not complain .. but if there are more than two, mypy will fail for that reason ...

src/validation_workflow/docker/check_docker_daemon.py
jordarlu: Thanks for the comment @gaiksaya .
While testing on my local MAC, I wanted to make sure the docker daemon is running before proceeding further; hende, I added a step to make sure that; and if it is not running, the daemon will be called up.  I am happy to remove this extra step if you would like. thanks!

src/validation_workflow/docker/cleanup_docker.py
jordarlu: oh.. sure .. I can certainly add cleanup into other class as a method. I was trying to seperate each function out to be an individual class. Maybe this is not a good practice to do ,. LOL

jordarlu:
Thanks for all the comments, @gaiksaya .. I replied on most of them, and we can make all the necessary changes accordingly.

src/validation_workflow/docker/spinup_docker_container.py
```diff
@@ -0,0 +1,62 @@
...
+from system.temporary_directory import TemporaryDirectory
+
+"""
+This class is to spin up docker containers with the OS and OSD images specificed earlier.
+It uses docker-compose to bring up docker containers.
+It returns True if containers are Up.
+"""
+
+
+class RunDocker:
```
gaiksaya, line 22: I don't think we need a class to spin up the docker containers right?
Why not put all of this under `install` and `start_cluster` here https://github.com/opensearch-project/opensearch-build/pull/3038/files#diff-968cdbdd15c8ab34a3bc9aa41df188abc7e459e0fc56f6affc36be763b78b898R48-R54

jordarlu: Thanks, @gaiksaya .. Because there are many lines of code to spin up the docker containers, and I am afraid the validation_docker.py will become too many lines of code; hence, I have it seperated from the validation_docker.py. And, sure, I can place it as a method in the validation_docker.py as well.

gaiksaya: That's okay I think! Also would suggest to use better naming convention for variables. OS1_name  does not give me much information on what that is. Use complete name if possible.

src/validation_workflow/validation.py
```diff
@@ -7,15 +7,44 @@
...
+        self.args = args
+
+    def run(self) -> Any:
+        try:
+            if self.download_artifacts(self.args):
+                if self.installation(self.args):
+                    if self.start_cluster(self.args):
+                        if self.validation(self.args):
+                            if self.cleanup(self.args):
+                                return True
```
zelinh, line 28: Can we use `&` here instead of nested `if`?

src/run_validation.py
```diff
@@ -10,19 +10,31 @@
...
+        dist = "yum"
+    elif args.distribution == "rpm":
+        dist = "rpm"
+
+    test_result = ValidationTestRunner.dispatch(args, dist).run()
+
+    if test_result:
+        return True
+    else:
+        return False
```
zelinh, line 37: Can we just `return test_result` here?

src/validation_workflow/validation.py
zelinh: or just `return self.download_artifacts(self.args) && self.installation(self.args) & ...`

jordarlu: Thanks a lot for the feedback, @zelinh .. indeed, we want to return True only when all of those sub processes returned True.. we can consolidate them with && indeed. Let us update it in the next commit.

src/run_validation.py
jordarlu: Thanks for the feedback, @zelinh ...yes, we should just return test_result directly as it will be either True or False.

src/validation_workflow/api_test_cases.py
```diff
@@ -0,0 +1,46 @@
...
+
+from validation_workflow.api_request import ApiTest
+from validation_workflow.validation_args import ValidationArgs
+
+'''
+This class is the collection of test cases to run.
+'''
+
+
+class TestCases:
```
gaiksaya, line 20: Maybe name this as ApiTestCases? Its more clear.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -0,0 +1,203 @@
...
+
+            self.local_image_OS_id = self.get_image_id(_OS_image_name, self.args.version)
+            self.local_image_OSD_id = self.get_image_id(_OSD_image_name, self.args.version)
+            logging.info('the OS image ID is : ' + self.local_image_OS_id)
+            logging.info('the OSD image ID is : ' + self.local_image_OSD_id + '\n\n')
+
+            return True
+
+        else:
+            raise Exception('docker_daemon check is fail. Exiting the validation.')
```
gaiksaya, line 49: Recommending rewording: `Docker Daemon is not running. Exiting the docker validation.` Also I believe we should hard fail at this point as validation won't go ahead so maybe `return 1`

src/validation_workflow/docker/validation_docker.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def installation(self) -> bool:
+        ''' Pass this method for docker as no installation required in docker'''
```
gaiksaya, line 52: All the method descriptions should go above and before the method name

src/validation_workflow/docker/validation_docker.py
```diff
@@ -0,0 +1,203 @@
...
+        return True
+
+    def start_cluster(self) -> bool:
+        ''' Pass this method for docker and combine it with the following method because we want to Pass the digest validation in docker first before spinning up the docker container'''
```
gaiksaya, line 56: Same as above

src/validation_workflow/docker/validation_docker.py
```diff
@@ -0,0 +1,203 @@
...
+            return_code, self._target_yml_file = self.run_container(
+                self.args.OS_image,
+                self.args.OSD_image,
+                self.args.version,
+                'opensearch-node1-test',
+                'opensearch-node2-test',
+                'opensearch-dashboards-test',
+            )
+            if (return_code):
+                logging.info('Cluster is running now')
```
gaiksaya, line 77: Why not put this under `start_cluter` method as that is actually what you are doing here?

src/validation_workflow/docker/validation_docker.py
```diff
@@ -0,0 +1,203 @@
...
+                # STEP 4 . OS, OSD API validation
+                _test_result, _counter = TestCases().test_cases()
+
+                if (_test_result):
+                    logging.info(f'All tests Pass : {_counter}')
+                    return True
+                else:
+                    raise Exception(f'Not all tests Pass : {_counter}')
+            else:
+                raise Exception('The container is fail to run. Exiting the validation.')
```
gaiksaya, line 95: Rewording: `'The container failed to start. Exiting the validation.'`

src/validation_workflow/docker/validation_docker.py
```diff
@@ -0,0 +1,203 @@
...
+        # remove docker-compose.yml from the tmp folder
+        try:
+            os.remove(self._target_yml_file)
+        except OSError as e:
+            print("Error: %s - %s." % (e.filename, e.strerror))
+
+        return('returncode=0' in (str(result)))
+
+    def is_container_daemon_running(self) -> Any:
+        check_docker_bash = "./src/validation_workflow/docker/check_docker_daemon.sh"
```
gaiksaya, line 134: Use path.join to connect paths everywhere else these will fail on windows

src/validation_workflow/docker/validation_docker.py
```diff
@@ -0,0 +1,203 @@
...
+        result = subprocess.call([check_docker_bash])
+        return (result)
+
+    def run_container(self, OpenSearch_image: str, OpenSearchDashboard_image: str, version: str, OpenSearchNode1_name: str, OpenSearchNode2_name: str, OpenSearchDashboard_name: str) -> Any:
+        # replace the placeholder in template docker-compose.yaml which is taken from sample in https://opensearch.org/docs/latest/opensearch/install/docker/#sample-docker-composeyml
+        yml_template = 'docker-compose.yml'
+        yml_template_location = './src/validation_workflow/docker/'
```
gaiksaya, line 141: Use path.join to form paths

src/validation_workflow/docker/validation_docker.py
```diff
@@ -0,0 +1,203 @@
...
+        if (result_inspect.returncode == 0):
+            logging.info('Image exists at local : ' + result_inspect.stdout + ' : ' + image_name + ":" + image_version)
+            logging.info('removing the local image and pulling a fresh one from DockerHub')
+            result_remove = subprocess.run(local_remove, shell=True, stdout=PIPE, stderr=PIPE, universal_newlines=True)
+            if (result_remove.returncode == 0):
+                logging.info('Image is removed at local : ' + image_name + ":" + image_version)
+            else:
+                raise Exception('Image removal fail : return code ' + str(result_remove.returncode))
+        else:
+            logging.info('Image does not exist at local, proceed with Pull from the dockerHub')
```
gaiksaya, line 186: Can you check if you force pull instead of removing and then checking locally? Force pull will override the local image as well

src/validation_workflow/rpm/validation_rpm.py
```diff
@@ -0,0 +1,76 @@
...
+            raise Exception('Failed to install Opensearch')
+        return True
+
+    def start_cluster(self) -> bool:
+        try:
+            for project in self.args.projects:
+                os.system(f'sudo systemctl enable {project}')
+                os.system(f'sudo systemctl start {project}')
+                time.sleep(20)
+                os.system(f'sudo systemctl status {project}')
```
gaiksaya, line 56: Why not use `src/system/execute.py` instead of os.system?

src/validation_workflow/rpm/validation_rpm.py
```diff
@@ -0,0 +1,76 @@
...
+        except:
+            raise Exception('Failed to Start Cluster')
+        return True
```
gaiksaya, line 59: Will this hard fail and exit or just throw an exception and continue the flow?

src/validation_workflow/validation_args.py
```diff
@@ -43,25 +41,33 @@ def __init__(self) -> None:
             "--platform",
             type=str,
             choices=self.SUPPORTED_PLATFORMS,
-            help="Platform to validate.",
+            help="(optional) Platform to validate.",
             default="linux"
         )
         parser.add_argument(
             "--stgosbuild",
```
gaiksaya, line 48: We only have 1 build number. Prod does not have a build number as such here. To make it more clear we can just use `--os-build-number` and `osd-build-number` Keep the description as is it but stgosbuild does not sound right

src/validation_workflow/validation_args.py
```diff
@@ -43,25 +41,33 @@ def __init__(self) -> None:
...
             "--stgosdbuild",
             type=str,
             required=False,
-            help="The opensearchstaging OpenSearchDashboard image build number if required, for example : 4104\n",
+            help="(optional) The opensearchstaging OpenSearchDashboard image build number if required, for example : 4104\n",
             default="",
             dest="stgosdbuild",
         )
+        parser.add_argument(
+            "--dockersource",
```
gaiksaya, line 64: rewording recommendation: `--docker-source`

tests/tests_validation_workflow/test_api_test_cases.py
```diff
@@ -0,0 +1,32 @@
...
+    @patch('validation_workflow.api_test_cases.ValidationArgs')
+    @patch.object(ApiTest, 'api_get')
+    def test_test_cases(self, mock_api_get: Mock, mock_validation_args: Mock) -> None:
+        mock_validation_args.return_value.stg_tag.return_value = '1.0.0.1000'
+        mock_api_get.return_value = (200, 'green')
+        test_cases = TestCases()
+        result = test_cases.test_cases()
+
+        self.assertEqual(result[1], 'There are 2/4 test cases Pass')  # type: ignore
+        mock_api_get.assert_has_calls([call(), call(), call(), call()])
```
gaiksaya, line 26: Not sure what are we asserting here. Can you please explain? Maybe I am missing out something.

validation.sh
```diff
@@ -11,4 +11,8 @@
 set -e
 
 DIR="$(dirname "$0")"
-"$DIR/run.sh" "$DIR/src/run_validation.py" $@
+if [  -n "$(uname -a | grep Ubuntu)" ]; then
+    "sudo" "$DIR/run.sh" "$DIR/src/run_validation.py" $@
```
gaiksaya, line 15: I don't think you need to do this. @peterzhuamazon  Can add more but entire code base runs as it is irrespective of sudo. Example: https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/distribution_rpm.py#L29-L32

gaiksaya (changes requested):

tests/tests_validation_workflow/test_api_test_cases.py
jordarlu: Thanks, @gaiksaya .. currently we defined 4 test cases in api_test_cases.py, and this assertion does two things : 1) making sure there will be 4 calls, one for each test case; 2), with test data retunrned, we expected 2 fail and 2 pass, so, we assert the exact string 'There are 2/4 test cases Pass' shown up.

src/validation_workflow/validation_args.py
jordarlu: Thanks, @gaiksaya , sounds good, let us update it on next commit.

src/validation_workflow/validation_args.py
jordarlu: Thanks, @gaiksaya , sounds good, let us update it at next commit.

tests/tests_validation_workflow/test_api_test_cases.py
gaiksaya: In the case you can assert the number of calls :) will make its easier for line number 26

src/validation_workflow/docker/validation_docker.py
jordarlu: Thanks, @gaiksaya , got it, let us update it at next commit.

src/validation_workflow/api_test_cases.py
jordarlu: Thanks, @gaiksaya , sounds good, let us update it at next commit.

src/validation_workflow/docker/validation_docker.py
jordarlu: Thanks, @gaiksaya , sounds good, let us update it at next commit.

src/validation_workflow/docker/validation_docker.py
jordarlu: Thanks, @gaiksaya , sounds good, let us update it at next commit.

src/validation_workflow/docker/validation_docker.py
jordarlu: Thanks, @gaiksaya , sounds good, let us update it at next commit.

src/validation_workflow/rpm/validation_rpm.py
Divyaasm: This exits from the flow and raises exception Sayali.

src/validation_workflow/docker/validation_docker.py
jordarlu: Thanks for the question, @gaiksaya. The reason is because we want to validate the digest before spining up the cluster,; so, I had to pass the 'start_cluster' and have both of them in the 'validation'. Hope that is fine.

src/validation_workflow/docker/validation_docker.py
jordarlu: Thanks, @gaiksaya , sounds good, let us update it at next commit.

src/validation_workflow/docker/validation_docker.py
jordarlu: Thanks, @gaiksaya , makes sense... better than just using '+' , let us update it at next commit.

src/validation_workflow/docker/validation_docker.py
gaiksaya: Does the order matter? I believe you can declare and call `start_cluster` anytime right??

jordarlu: Looks like it does, @gaiksaya ..  we set the following at validation.py
```
          try:
            return self.download_artifacts() and self.installation() and self.start_cluster() and self.validation() and self.cleanup()
```
so, it goes to the next method only if the curent returns True.  :(

tests/tests_validation_workflow/test_api_test_cases.py
jordarlu: LOL // indeed, indeed ... why I did not think of that .. LOL

src/validation_workflow/docker/validation_docker.py
jordarlu: Thanks, @gaiksaya , let me use '-f' to simplify the code.. thanks again... I now recall @prudhvigodithi said the same recommendation in the dry-run as well.. :)

src/validation_workflow/yum/validation_yum.py
```diff
@@ -0,0 +1,76 @@
...
+    def installation(self) -> bool:
+        try:
+            os.chdir(self.tmp_dir.path)
+            os.system('sudo rpm --import https://artifacts.opensearch.org/publickeys/opensearch.pgp')
+            for project in self.args.projects:
+                os.system(f'sudo yum remove {project} -y')
+                logging.info('Removed previous versions of Opensearch')
+                urllink = f"{self.base_url}{project}/{self.args.version[0:1]}.x/{project}-{self.args.version[0:1]}.x.repo -o /etc/yum.repos.d/{project}-{self.args.version[0:1]}.x.repo"
+                os.system(f'sudo curl -SL {urllink}')
+                os.system(f"sudo yum install '{project}-{self.args.version}' -y")
```
gaiksaya, line 45: Same as above `Why not use src/system/execute.py instead of os.system?`

gaiksaya (changes requested):
Would recommend to use https://github.com/opensearch-project/opensearch-build/blob/main/src/system/execute.py for all commands as it logs all the commands that you are running along with failures.

src/validation_workflow/yum/validation_yum.py
Divyaasm: Replaced os.system implementations with execute.py Sayali. Thanks.

peterzhuamazon (changes requested):
A lot of the code is assuming the run is happening only on nix* env.
It is hardcoding `/` everywhere and it is probably better if we could use os.path related ways to join paths instead.

src/validation_workflow/docker/check_docker_daemon.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+if (! docker stats --no-stream 2>/dev/null); then
```
peterzhuamazon, line 9: We should completely remove this script.
We should just do a simple check to see if docker is online, if not, exit.
There is no reason to keep waiting infinitely and even try to start docker for user.

validation.sh
peterzhuamazon, line 15: Just completely remove the sudo part from this code.
Simply echo this line so inform user that if their user account does not have access to docker, they need to run the workflow in sudo.

This is not a problem in our docker image as it already assume the runner socket, and the runner would not have this issue because their user is already added to the docker group.

See RPM/DEB example (just for reference): https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/distribution_rpm.py#L30

peterzhuamazon: Testing without sudo and just adding the user to docker group:
```

Running ./src/run_validation.py --version 1.3.8 --distribution docker --os_build_number 7019 --osd_build_number 5411 ...
CONTAINER ID   NAME      CPU %     MEM USAGE / LIMIT   MEM %     NET I/O   BLOCK I/O   PIDS

Docker started
2023-02-03 23:05:23 INFO     docker_daemon is running
2023-02-03 23:05:23 INFO     Image does not exist at local, proceed with Pull from the dockerhub
2023-02-03 23:05:23 INFO     Pulling opensearchproject/opensearch:1.3.8 from the dockerhub
2023-02-03 23:05:32 INFO     Image is pulled at local : sha256:4460662780001ed00650637d774570dd3cfb64be1afc1461b6620fc6dc625240
 : opensearchproject/opensearch:1.3.8
2023-02-03 23:05:33 INFO     Image does not exist at local, proceed with Pull from the dockerhub
2023-02-03 23:05:33 INFO     Pulling opensearchproject/opensearch-dashboards:1.3.8 from the dockerhub
2023-02-03 23:05:46 INFO     Image is pulled at local : sha256:96d6ab746fd71a16b6f984e101ece24158c030b51bcfff0375d393cb39ff4014
 : opensearchproject/opensearch-dashboards:1.3.8
2023-02-03 23:05:46 INFO     the OS image ID is : sha256:4460662780001ed00650637d774570dd3cfb64be1afc1461b6620fc6dc625240

2023-02-03 23:05:46 INFO     the OSD image ID is : sha256:96d6ab746fd71a16b6f984e101ece24158c030b51bcfff0375d393cb39ff4014



2023-02-03 23:05:46 INFO     Fetching access token from dockerHub
2023-02-03 23:05:46 INFO     Fetching mainfest from DockerHub
2023-02-03 23:05:47 INFO     sha256:4460662780001ed00650637d774570dd3cfb64be1afc1461b6620fc6dc625240 --> DockerHub image digest opensearchstaging/opensearch:1.3.8.7019
2023-02-03 23:05:47 INFO     Fetching mainfest from local
2023-02-03 23:05:47 INFO     sha256:4460662780001ed00650637d774570dd3cfb64be1afc1461b6620fc6dc625240 --> local image digest opensearchproject/opensearch:1.3.8
2023-02-03 23:05:47 INFO     Fetching access token from dockerHub
2023-02-03 23:05:47 INFO     Fetching mainfest from DockerHub
2023-02-03 23:05:47 INFO     sha256:96d6ab746fd71a16b6f984e101ece24158c030b51bcfff0375d393cb39ff4014 --> DockerHub image digest opensearchstaging/opensearch-dashboards:1.3.8.5411
2023-02-03 23:05:47 INFO     Fetching mainfest from local
2023-02-03 23:05:47 INFO     sha256:96d6ab746fd71a16b6f984e101ece24158c030b51bcfff0375d393cb39ff4014 --> local image digest opensearchproject/opensearch-dashboards:1.3.8
2023-02-03 23:05:47 INFO     Image digest is validated


Changing "OS_IMAGE_PLACEHOLDER" to "opensearchproject/opensearch:1.3.8" in /tmp/tmp42fa8n__/docker-compose.yml
Changing "OSD_IMAGE_PLACEHOLDER" to "opensearchproject/opensearch-dashboards:1.3.8" in /tmp/tmp42fa8n__/docker-compose.yml
Changing "OS1_NAME" to "opensearch-node1-test" in /tmp/tmp42fa8n__/docker-compose.yml
Changing "OS2_NAME" to "opensearch-node2-test" in /tmp/tmp42fa8n__/docker-compose.yml
Changing "OSD_NAME_PLACEHOLDER" to "opensearch-dashboards-test" in /tmp/tmp42fa8n__/docker-compose.yml
2023-02-03 23:05:47 INFO     final test_result = False
```

peterzhuamazon: Added my comments.

jordarlu: Thank you, @peterzhuamazon , let us update the script accordingly.

src/validation_workflow/docker/check_docker_daemon.sh
jordarlu: Thanks for the comment, @peterzhuamazon .. I was thinking we could help user to bring up the docker daemon on the background, but sure I could just exit the script if the daemon was not running at the time.

src/run_validation.py
```diff
@@ -10,19 +10,28 @@
...
-    return 0
+    dist = None
+    if args.distribution == 'docker':
+        dist = "docker"
+    elif args.distribution == "tar":
+        dist = "tar"
+    elif args.distribution == "yum":
+        dist = "yum"
+    elif args.distribution == "rpm":
+        dist = "rpm"
```
peterzhuamazon, line 30: Why cant dist being directly assigned with args.distribution since they are the same value anyway?

src/validation_workflow/api_test_cases.py
```diff
@@ -0,0 +1,51 @@
...
+            status_code, response_text = ApiTest(str(request_url)).api_get()
+            logging.info(f"\nStatus_code ->{status_code} \nresponse_text ->{response_text}")
+            time.sleep(3)
+
+            if status_code == success_status_code and (not validate_string or validate_string in response_text):
+                pass_counter += 1
+            else:
+                fail_counter += 1
+
+        return (fail_counter == 0, "There are " + str(pass_counter) + "/" + str(pass_counter + fail_counter) + " test cases Pass")
```
peterzhuamazon, line 51: Avoid using `/` but replace with `os.path.join` or `os.path.sep` or similar, so it can be used outside of nix* env.

src/validation_workflow/docker/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
+---
+version: '3'
```
peterzhuamazon, line 2: No need to add another docker-compose file as in `docker/release` we already have compose file for different versions. 

The current one you include wont work on 1.x anyway.

src/validation_workflow/docker/inspect_docker_image.py
```diff
@@ -0,0 +1,66 @@
...
+        api_url = self.registry_url + self.image_name.replace("opensearchproject", "opensearchstaging") + "/manifests/" + self.image_tag
+        access_token = response_dict['token']
+
+        # set up all necessary token ( if you use VPN, it may slow down the response from dockerHub )
+        headersToken = {
+            "Authorization": f'{self.access_type} {access_token}',
+            "Accept": "application/vnd.docker.distribution.manifest.v2+json",
+            "Content-Type": "application/json;charset=UTF-8"
+        }
+
```
peterzhuamazon, line 53: Where is the token coming from?

peterzhuamazon: I mean to say which user? Are you hardcoding the token or you have a way to allow users enter that, or you can retrieve after user logging in?

src/validation_workflow/docker/validation_docker.py
```diff
@@ -0,0 +1,202 @@
...
+            logging.info(f'Overwriting the local image by pulling a fresh {image_name}:{image_version} from {self.args.docker_source}')
+        else:
+            logging.info(f'Image does not exist at local, proceed with Pull from the {self.args.docker_source}')
+
+        return (self.pull_image(image_name, image_version))
+
+    def pull_image(self, image_name: str, image_version: str) -> str:
+
+        dockerHub_pull = "docker pull " + image_name + ":" + image_version
+        local_inspect = "docker image inspect -f '{{ .Id }}' " + image_name + ":" + image_version
```
peterzhuamazon, line 192: Why not use `f` string but trying to `+` everything?

src/validation_workflow/api_test_cases.py
```diff
@@ -0,0 +1,51 @@
...
+'''
+This class is the collection of test cases to run.
+'''
+
+
+class ApiTestCases:
+
+    def __init__(self) -> None:
+        self.OS_image_version = ValidationArgs().stg_tag('OS').split(' ')[0]
+        self.OSD_image_version = ValidationArgs().stg_tag('OSD').split(' ')[0]
```
peterzhuamazon, line 24: Is there any reason to use OS/OSD instead of opensearch/dashboards? OS can be very confused with `os` packages, or similar, and this is not a good practice to use acronyms as var names.a

peterzhuamazon (changes requested):
Hi @Divyaasm posted some issues that I quickly observed, thanks.

src/validation_workflow/api_test_cases.py
peterzhuamazon: This applied to any similar issues in this PR.

src/validation_workflow/docker/validation_docker.py
peterzhuamazon: This applied to any similar issues in this PR.

src/run_validation.py
Divyaasm: Will take care of this Peter, thanks

src/validation_workflow/api_test_cases.py
jordarlu: Thanks, @peterzhuamazon, we will use the fullname to be more clear. Was trying to avoid using long argument name.

src/validation_workflow/api_test_cases.py
jordarlu: Thanks for the comment, @peterzhuamazon, this is a return string to show how many cases pass over the total test cases, I guess it won't be matter per OS, correct?

src/validation_workflow/docker/docker-compose.yml
jordarlu: Thanks for notifying that, @peterzhuamazon, I took the docker-compose.yml from https://opensearch.org/samples/docker-compose.yml and put some placeholders in the file to be replaced with the arguments where taken from the user input. Also, placeholder for the container_name which avoid the possible name conflict if the user also have the same container name existing at the time the script is running; hence, I wanted to have this edited docker-compose.yml to be used for validation specifically. 

Good point you mentioned that 'won't work on 1.x anyway'.. I wasn't aware of it, does that mean the docker-compose.xml provided at opensearch.org actually does not work on 1.x either? should we notify the documentation team about that as well?

src/validation_workflow/docker/inspect_docker_image.py
jordarlu: Thanks for the asking, @peterzhuamazon , at L39-41, we retrieve the token at everytime the script is run ( if I am not mistaken, the token expired in 1 hour); and the L31-32 defines the token_url to be used in API call.

src/validation_workflow/docker/validation_docker.py
jordarlu: Thanks, @peterzhuamazon. Let us use f-strings in those cases..

jordarlu: Thanks, @peterzhuamazon. Let us use f-strings in those cases..

src/validation_workflow/docker/docker-compose.yml
jordarlu: Hi, @peterzhuamazon , I added another line to replace 'initial_cluster_manager_nodes' with 'initial_master_nodes' in docker-compose.xml to get it work on ver1.x, Thanks again for noticing me that the template earlier did not work on 1.x. I did not find it mentioned on https://opensearch.org/ tho ... maybe the documentation team should note that there just to be clear to the users.

src/validation_workflow/api_test_cases.py
gaiksaya: This is just a print statement not a path so we are good

jordarlu: thanks, @gaiksaya

src/validation_workflow/docker/docker-compose.yml
gaiksaya: Is it possible to just retrieve the docker-compose file on the go? The reason being we won't know what is shipped to customer and we would need to maintain docker-compose at multiple places. Dont worry about container name, let if fail if one already exists. You can grab the latest docker-compose yml from website and just replace the version.

jordarlu: Thanks for the idea, @gaiksaya .. I think so, I can just fetch the file from https://opensearch.org/samples/docker-compose.yml at everytime script run and make some necessary replacement. Do you think we can do that as an enhancement later on?

gaiksaya: Sure! Make sure you create an issue and have a fast follow up later.
Thanks!

jordarlu: Yes, so far, I collected 3 items to enhance at https://github.com/opensearch-project/opensearch-build/issues/3205  :) thanks.

jordarlu: Hi. @gaiksaya , I got the script updated to fetch the docker-compose.xml on the go, and also removed the local file. thanks! https://github.com/opensearch-project/opensearch-build/pull/3038/commits/cfe77608ed8fc09022387d6ee8b5f87f754172e4

gaiksaya:
@Divyaasm There are 15 commits in this PR, can you rebase and bring it to 1? Its better to get in a clean commit. 
Thanks!

Divyaasm:
> @Divyaasm There are 15 commits in this PR, can you rebase and bring it to 1? Its better to get in a clean commit. Thanks!

Sure Sayali, thanks

validation.sh
peterzhuamazon: You dont need to update the script, just ask user to make sure docker can run, and they would do the necessary later.

src/validation_workflow/api_test_cases.py
peterzhuamazon: I am not saying this string only, see my comment I am talking about if there is any path that is trying this, we should aviod, also why not use `f` string but using `+` here?

src/validation_workflow/docker/docker-compose.yml
peterzhuamazon: Like I mentioned in the comment why retrieve the file from website when our repo in `docker/release` already have the necessary files?

peterzhuamazon (dismissed):
Hi,

I am still blocking this PR as 2 key issues have still not resolved:
1. Still no confirming if any hardcoded `/` is change to os.path.join/os.path.sep.
2. With `docker/release` folder already providing the necessary compose file, still trying to retrieve from website and modify live.

I hope you can understand as we need to keep the code high standard.

Thanks

src/validation_workflow/docker/docker-compose.yml
gaiksaya: Because we want to validate what customer uses not what we have in the repo.

src/validation_workflow/api_test_cases.py
gaiksaya: Everywhere else os.path.join has been used https://github.com/opensearch-project/opensearch-build/pull/3038/files#diff-968cdbdd15c8ab34a3bc9aa41df188abc7e459e0fc56f6affc36be763b78b898R144

Using f string is a nit. I dont think PR should be blocked for a print statement.

jordarlu: thanks, @peterzhuamazon and @gaiksaya, and Yes, I just double check all the places that I need to concatenate path and file, I have them by using the os.path.join.

Divyaasm: Yeah Peter, we have used os.path.join omitting usage of "/" in all the scenarios. And have encapusulated the direct url links in f-string

bbarani:
> Hi,
> 
> I am still blocking this PR as 2 key issues have still not resolved:
> 
> 1. Still no confirming if any hardcoded `/` is change to os.path.join/os.path.sep.
> 2. With `docker/release` folder already providing the necessary compose file, still trying to retrieve from website and modify live.
> 
> I hope you can understand as we need to keep the code high standard.
> 
> Thanks

@peterzhuamazon Do you think the above lists are hard blockers? if not, I would recommend to merge this PR and create issues for the above open issues to iterate faster.

src/validation_workflow/docker/docker-compose.yml
jordarlu: I went to check if any existing code that is using the /docker/release/dockercomposefiles/xx but did not find any; hence, I thought it would be better Not to link the docker-validation to it either as well. ( keeping no dependancy with those twp yml files as current )
By fetching the docker-compose.yml from the example file provided by https://opensearch.org/ looks a good suggestion to me because it is the file brocasted to all in official, and we don't need to maintain it either; hence, I switch to use it. Hope that make sense... thanks!

peterzhuamazon:
> > Hi,
> > I am still blocking this PR as 2 key issues have still not resolved:
> > 
> > 1. Still no confirming if any hardcoded `/` is change to os.path.join/os.path.sep.
> > 2. With `docker/release` folder already providing the necessary compose file, still trying to retrieve from website and modify live.
> > 
> > I hope you can understand as we need to keep the code high standard.
> > Thanks
> 
> @peterzhuamazon Do you think the above lists of issues are hard blockers? if not, I would recommend to merge this PR and create issues for the above open issues to iterate faster.

Hi @bbarani I will work with @Divyaasm and @jordarlu on this.
Thanks.

peterzhuamazon:
Hi, I have talked to @Divyaasm and @jordarlu and we have discussed these points:

1. @Divyaasm and @jordarlu will confirm that no file/folder path is hardcoding with `/`, and are all using `os.path.join/os.path.sep` to handle on different platforms.
2. @Divyaasm and @jordarlu will check if there is any more strings can be handled by `f` string, tho this is not a hard blocker.
3. The docker compose file part will be changed to sync files from our build repo `docker/release/dockercomposefiles` https://github.com/opensearch-project/opensearch-build/tree/main/docker/release/dockercomposefiles. @jordarlu will make sure only the version / tag get replaced during the validation, instead of hardcoding certain parameters on the fly.
4. @Divyaasm and @jordarlu will create new issues on handling cleanups properly during the run, especially when `sigint/sigexit/sigkill` was called during the run. As of now, the validation workflow is assuming no interruption during the run, and would not properly cleanup process if such event happens per https://github.com/opensearch-project/opensearch-build/blob/c17530445633b6eb8ff869f3015665235e009960/src/validation_workflow/validation.py#L23.


In conclusion, 1 and 3 above is hardblockers and @Divyaasm / @jordarlu will resolve them before this PR get merged.
2 and 4 are soft blockers that new issues will be created so they can be fixed later on.

Thanks.

gaiksaya:
> The docker compose file part will be changed to sync files from our build repo docker/release/dockercomposefiles https://github.com/opensearch-project/opensearch-build/tree/main/docker/release/dockercomposefiles. @jordarlu will make sure only the version / tag get replaced during the validation, instead of hardcoding certain parameters on the fly.

I do not agree with point 3. We need to validate what we ship to the customer and not out connivence of replacing values.
Maintaining 2 docker-composes is not feasible and what we have in this repo does not matter but what we ship to customer does. Adding @dblock @prudhvigodithi @rishabh6788 to get more inputs.

Divyaasm:
> Hi, I have talked to @Divyaasm and @jordarlu and we have discussed these points:
> 
> 1. @Divyaasm and @jordarlu will confirm that no file/folder path is hardcoding with `/`, and are all using `os.path.join/os.path.sep` to handle on different platforms.
> 2. @Divyaasm and @jordarlu will check if there is any more strings can be handled by `f` string, tho this is not a hard blocker.
> 3. The docker compose file part will be changed to sync files from our build repo `docker/release/dockercomposefiles` https://github.com/opensearch-project/opensearch-build/tree/main/docker/release/dockercomposefiles. @jordarlu will make sure only the version / tag get replaced during the validation, instead of hardcoding certain parameters on the fly.
> 4. @Divyaasm and @jordarlu will create new issues on handling cleanups properly during the run, especially when `sigint/sigexit/sigkill` was called during the run. As of now, the validation workflow is assuming no interruption during the run, and would not properly cleanup process if such event happens per https://github.com/opensearch-project/opensearch-build/blob/c17530445633b6eb8ff869f3015665235e009960/src/validation_workflow/validation.py#L23
>    .
> 
> In conclusion, 1 and 3 above is hardblockers and @Divyaasm / @jordarlu will resolve them before this PR get merged. 2 and 4 are soft blockers that new issues will be created so they can be fixed later on.
> 
> Thanks.

Regarding point 1, @jordarlu and I confirm that we haven't hardcoded "/" in any part of the code. Thanks

peterzhuamazon:
> Hi, I have talked to @Divyaasm and @jordarlu and we have discussed these points:
> 
>     1. @Divyaasm and @jordarlu will confirm that no file/folder path is hardcoding with `/`, and are all using `os.path.join/os.path.sep` to handle on different platforms.
> 
>     2. @Divyaasm and @jordarlu will check if there is any more strings can be handled by `f` string, tho this is not a hard blocker.
> 
>     3. The docker compose file part will be changed to sync files from our build repo `docker/release/dockercomposefiles` https://github.com/opensearch-project/opensearch-build/tree/main/docker/release/dockercomposefiles. @jordarlu will make sure only the version / tag get replaced during the validation, instead of hardcoding certain parameters on the fly.
> 
>     4. @Divyaasm and @jordarlu will create new issues on handling cleanups properly during the run, especially when `sigint/sigexit/sigkill` was called during the run. As of now, the validation workflow is assuming no interruption during the run, and would not properly cleanup process if such event happens per https://github.com/opensearch-project/opensearch-build/blob/c17530445633b6eb8ff869f3015665235e009960/src/validation_workflow/validation.py#L23
>        .
> 
> 
> In conclusion, 1 and 3 above is hardblockers and @Divyaasm / @jordarlu will resolve them before this PR get merged. 2 and 4 are soft blockers that new issues will be created so they can be fixed later on.
> 
> Thanks.



> > The docker compose file part will be changed to sync files from our build repo docker/release/dockercomposefiles https://github.com/opensearch-project/opensearch-build/tree/main/docker/release/dockercomposefiles. @jordarlu will make sure only the version / tag get replaced during the validation, instead of hardcoding certain parameters on the fly.
> 
> I do not agree with point 3. We need to validate what we ship to the customer and not out connivence of replacing values. Maintaining 2 docker-composes is not feasible and what we have in this repo does not matter but what we ship to customer does. Adding @dblock @prudhvigodithi @rishabh6788 to get more inputs.

I disagree with that as the official website is not providing the correct docker compose files based on versions.
I have opened the issue to them and there is no drive in fixing it in the website.
https://github.com/opensearch-project/project-website/issues/1245

It is very bad practice to replace parameter on the fly for 1.x based on 2.x docker compose file provided on the website.
If there is any changes on the parameter, instead of changing the docker compose files, we have to change the hardcoded parameter in python code.

Thanks.

gaiksaya:
>I disagree with that as the official website is not providing the correct docker compose files based on versions.
I have opened the issue to them and it surely needs some attraction.
https://github.com/opensearch-project/project-website/issues/1245

It is very bad practice to replace parameter on the fly for 1.x based on 2.x docker compose file provided on website. If there is any changes on the parameter, instead of changing the docker compose files, we have to change the hardcoded parameter in python code.

Isn't customer doing the same? Adding the required values as per version? I believe we can go with this change for now and once docker-compose per version is added we can modify the code to pull the same. My point is we should test what we ship to the customer rather than having our own set of files. Any change in website provided file will be detected sooner before reaching the customer.

peterzhuamazon:
> > I disagree with that as the official website is not providing the correct docker compose files based on versions.
> > I have opened the issue to them and it surely needs some attraction.
> > [opensearch-project/project-website#1245](https://github.com/opensearch-project/project-website/issues/1245)
> 
> > It is very bad practice to replace parameter on the fly for 1.x based on 2.x docker compose file provided on website. If there is any changes on the parameter, instead of changing the docker compose files, we have to change the hardcoded parameter in python code.
> 
> Isn't customer doing the same? Adding the required values as per version? I believe we can go with this change for now and once docker-compose per version is added we can modify the code to pull the same. My point is we should test what we ship to the customer rather than having our own set of files. Any change in website provided file will be detected sooner before reaching the customer.

That is the point we have already test the ones in our repo.
However, for the ones on the website, it is being changed multiple times in the past.
And it only runs on 2.x, not able to run on 1.x as of now.

jordarlu:
Thanks for the discussion and we have come to the agreement that we want to use the docker-compose file at the build repo for now. 
In the future, if we are certain that the OpenSearch website has the proper docker-compose files per OpenSearch ver 1.x and 2.x, we will review again and decide if we want to go for using the file from the website.

@peterzhuamazon ,and @gaiksaya , if you could give one more review on the latest commit please .. thanks!

prudhvigodithi:
Hey @jordarlu I have seen the entire logic is driven with version as `version[0]` assuming the docker-compose files will have the version with single digit `2` [image: opensearchproject/opensearch:2](https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/dockercomposefiles/docker-compose-2.x.yml), if that changes from the compose file end the validation breaks, I'm fine to move forward keeping an eye on this.
@gaiksaya @peterzhuamazon 

Also yes we can start with validating with docker files from the build repo and once the website changes are reflected right with right versions then we can change the logic to fetch the docker-compose files directly from the opensearch.org website, adding @JeffH-AWS

prudhvigodithi (approved):

Issue #2290, opened 2022-06-30
Add signature verification to windows signing workflow


zelinh:
### Is your feature request related to a problem? Please describe

Coming from https://github.com/opensearch-project/opensearch-build/pull/2156#pullrequestreview-1025505369.

We should add windows signature verification from command line. Right now our signing workflow has functionality to sign the windows distribution but it doesn't have method to verify the signature. We can only validate the signature manually for now. 

### Describe the solution you'd like

Acceptance criteria: python [windows signer](https://github.com/opensearch-project/opensearch-build/blob/main/src/sign_workflow/signer_windows.py) has built in method to verify the signature like what we had for PGP signer. 
https://github.com/opensearch-project/opensearch-build/blob/700c80d60b463a8ed41f09d562709afb9d0a55f0/src/sign_workflow/signer_pgp.py#L49-L51

### Acceptance Criteria 

- This should Support all windows artifacts and validate all types of supported windows signatures.
- Integrate the windows artifact validation with validation workflow.

prudhvigodithi:
[Triaged] This is a valid issue, the solution should be able to verify both windows and nuget packages.

zelinh:
> @zelinh Can we also expand this story to add signature validation to TAR artifacts as well?

We actually have GPG signature verification for tarball artifacts immediately after we sign it which implemented here.
https://github.com/opensearch-project/opensearch-build/blob/356b2a2397795065a818fe32224fcac82c54eca7/src/sign_workflow/signer_pgp.py#L49-L51
For windows verification, it's quite different so we don't have it for now.

bbarani:
Lets make sure to add validation for all Windows distribution types that we are planning to support (i.e. .exe, .zip, .msi, .msix etc..)

gaiksaya:
The above PR adds the verification step using `osslsigncode` tool. The tool is currently installed only on release clients docker image. We need to install tool on other docker images if required. Currently windows signer is not used by distribution.

Issue #2965, opened 2022-12-05
Create publishToNuGet library


gaiksaya:
The dot net artifacts like the net client is published on [Nuget](https://www.nuget.org).
Add jenkins library that publishes the artifacts under  https://www.nuget.org/profiles/opensearchproject namespace

Acceptance Criteria:
- [x] The library should sign the given artifacts
- [x] Publish to NuGet automatically (without any human intervention)

gaiksaya:
As of today we have https://github.com/opensearch-project/opensearch-net client that publishes the artifacts on nuget.
Got some information from @Xtansia that all clients in that repository will be released at the same time.

gaiksaya:
Some information on how to publish to nuget using command line and api key https://learn.microsoft.com/en-us/nuget/nuget-org/publish-a-package#push-by-using-a-command-line

gaiksaya:
The security approved signing process for Nuget was causing issues for this automation. Currently, we were creating `.dll` files, signing them and then packing them to `.nupkg` files that were finally released to the customers. However, following #1234 we are trying to avoid building anything on jenkin and signing can only happen on jenkins for now.

We tested the following theory:

Create `.nupkg` packages and provide them as release candidates via GHA. The above library will do the following:
1. Download them and Convert them to `.zip` (just a rename)
2. Unzip them and sign the `.dll` files
3. Zip them again 
4. Rename to `.nupkg`

@Xtansia helped us test this theory and the packages worked as expected. We will be going forward with this solution.
Thanks!

cc: @wbeckler @VachaShah @bbarani @Yury-Fridlyand

gaiksaya:
There was a change in plan. Unzipping and packing the artifacts back again gave a number of permission issues on various files. It felt a lot of risky and kind of hacky, so we are making an exception for dotnet clients. We will be building, signing and releasing all on same platform using jenkins. 
See the above PR in review. Thanks!

PR #3118, opened 2023-01-19
Add support to overwrite an existing artifact for signing process and add verification for windows

gaiksaya:

### Description
- Currently the signer workflow does not allow the existing artifact to be overwritten. This change adds a functionality to add that option. The current use case is windows `.dll` artifacts. The same artifact needs to be overwritten in same exact location to proceed with further packaging. Defaults to `False` so the earlier is not changing just that the check can be disabled at will.
- Also adds verification of signed artifact for windows which was missing until now.
- Removes `.zip` as supported artifact in windows signing as it is not supported by backend code. 
- Adds `osssigncode` and `dotnet` installations on release-clients docker file. 

Depends on and the Backend PR: https://github.com/opensearch-project/opensearch-signer-client/pull/11

Next steps:
- Update jenkins library to accommodate this change PR: https://github.com/opensearch-project/opensearch-build-libraries/pull/115
- Publish new docker images with sigtool installed
   - release client docker image

### Issues Resolved
needed for signing of #2965 
closes https://github.com/opensearch-project/opensearch-build/issues/2290

### DIFF
--- a/src/sign_workflow/sign_args.py
+++ b/src/sign_workflow/sign_args.py
@@ -20,14 +20,46 @@ class SignArgs:
     type: str
     sigtype: str
     platform: str
+    overwrite: bool
 
     def __init__(self) -> None:
         parser = argparse.ArgumentParser(description="Sign artifacts")
-        parser.add_argument("target", type=Path, help="Path to local manifest file or artifact directory")
-        parser.add_argument("-c", "--component", type=str, nargs='*', dest="components", help="Component or components to sign")
-        parser.add_argument("--type", help="Artifact type")
-        parser.add_argument("--sigtype", choices=self.ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of signature file.", default=".sig")
-        parser.add_argument("--platform", choices=self.ACCEPTED_PLATFORM, help="Distribution platform.", default="linux")
+        parser.add_argument(
+            "target",
+            type=Path,
+            help="Path to local manifest file or artifact directory"
+        )
+        parser.add_argument(
+            "-c",
+            "--component",
+            type=str,
+            nargs='*',
+            dest="components",
+            help="Component or components to sign"
+        )
+        parser.add_argument(
+            "--type",
+            help="Artifact type"
+        )
+        parser.add_argument(
+            "--sigtype",
+            choices=self.ACCEPTED_SIGNATURE_FILE_TYPES,
+            help="Type of signature file.",
+            default=".sig"
+        )
+        parser.add_argument(
+            "--platform",
+            choices=self.ACCEPTED_PLATFORM,
+            help="Distribution platform.",
+            default="linux"
+        )
+        parser.add_argument(
+            "--overwrite",
+            action="store_true",
+            help="Overwrite existing artifacts or signature files.",
+            default=False,
+            dest="overwrite"
+        )
         parser.add_argument(
             "-v",
             "--verbose",
@@ -45,3 +77,4 @@ class SignArgs:
         self.sigtype = args.sigtype
         self.components = args.components
         self.platform = args.platform
+        self.overwrite = args.overwrite
--- a/src/sign_workflow/sign_artifacts.py
+++ b/src/sign_workflow/sign_artifacts.py
@@ -25,13 +25,13 @@ class SignArtifacts:
     platform: str
     signer: Signer
 
-    def __init__(self, target: Path, components: List[str], artifact_type: str, signature_type: str, platform: str) -> None:
+    def __init__(self, target: Path, components: List[str], artifact_type: str, signature_type: str, platform: str, overwrite: bool) -> None:
         self.target = target
         self.components = components
         self.artifact_type = artifact_type
         self.signature_type = signature_type
         self.platform = platform
-        self.signer = Signers.create(platform)
+        self.signer = Signers.create(platform, overwrite)
 
     @abstractmethod
     def __sign__(self) -> None:
@@ -57,9 +57,9 @@ class SignArtifacts:
             return SignArtifactsExistingArtifactFile
 
     @classmethod
-    def from_path(self, path: Path, components: List[str], artifact_type: str, signature_type: str, platform: str) -> Any:
+    def from_path(self, path: Path, components: List[str], artifact_type: str, signature_type: str, platform: str, overwrite: bool) -> Any:
         klass = self.__signer_class__(path)
-        return klass(path, components, artifact_type, signature_type, platform)
+        return klass(path, components, artifact_type, signature_type, platform, overwrite)
 
 
 class SignWithBuildManifest(SignArtifacts):
--- a/src/sign_workflow/signer_windows.py
+++ b/src/sign_workflow/signer_windows.py
@@ -18,11 +18,13 @@ The signed artifacts will be found in the subfolder called signed under the orig
 
 
 class SignerWindows(Signer):
-
-    ACCEPTED_FILE_TYPES = [".msi", ".exe", ".dll", ".sys", ".ps1", ".psm1", ".psd1", ".cat", ".zip"]
+    ACCEPTED_FILE_TYPES = [".msi", ".exe", ".dll", ".sys", ".ps1", ".psm1", ".psd1", ".cat"]
 
     def generate_signature_and_verify(self, artifact: str, basepath: Path, signature_type: str) -> None:
+        filename = os.path.join(basepath, artifact)
+        signed_filename = filename if self.overwrite else os.path.join(basepath, "signed_" + artifact)
         self.sign(artifact, basepath, signature_type)
+        self.verify(signed_filename)
 
     def is_valid_file_type(self, file_name: str) -> bool:
         return any(
@@ -31,21 +33,20 @@ class SignerWindows(Signer):
 
     def sign(self, artifact: str, basepath: Path, signature_type: str) -> None:
         filename = os.path.join(basepath, artifact)
-        signed_prefix = "signed_"
-        signature_file = os.path.join(basepath, signed_prefix + artifact)
-        self.__remove_existing_signature__(signature_file)
+        signed_filename = filename if self.overwrite else os.path.join(basepath, "signed_" + artifact)
         signing_cmd = [
             "./opensearch-signer-client",
             "-i",
             filename,
             "-o",
-            signature_file,
+            signed_filename,
             "-p",
-            "windows",
+            "windows"
+            " -r",
+            str(self.overwrite)
         ]
         self.git_repo.execute(" ".join(signing_cmd))
-        signed_folder = os.path.join(basepath, "signed")
-        if not os.path.exists(signed_folder):
-            os.mkdir(signed_folder)
-        signed_location = os.path.join(signed_folder, artifact)
-        os.rename(signature_file, signed_location)
+
+    def verify(self, filename: str) -> None:
+        verify_cmd = ["osslsigncode", "verify", "-in", filename]
+        self.git_repo.execute(" ".join(verify_cmd))
--- a/src/run_sign.py
+++ b/src/run_sign.py
@@ -23,7 +23,8 @@ def main() -> int:
         components=args.components,
         artifact_type=args.type,
         signature_type=args.sigtype,
-        platform=args.platform
+        platform=args.platform,
+        overwrite=args.overwrite
     )
 
     sign.sign()
--- a/src/sign_workflow/signer.py
+++ b/src/sign_workflow/signer.py
@@ -17,8 +17,10 @@ from git.git_repository import GitRepository
 
 class Signer(ABC):
     git_repo: GitRepository
+    overwrite: bool
 
-    def __init__(self) -> None:
+    def __init__(self, overwrite: bool) -> None:
+        self.overwrite = overwrite
         self.git_repo = GitRepository(self.get_repo_url(), "HEAD", working_subdirectory="src")
         self.git_repo.execute("./bootstrap")
         self.git_repo.execute("rm config.cfg")
--- a/src/sign_workflow/signer_pgp.py
+++ b/src/sign_workflow/signer_pgp.py
@@ -19,7 +19,6 @@ The signed artifacts will be found in the same location as the original artifact
 
 
 class SignerPGP(Signer):
-
     ACCEPTED_FILE_TYPES = [".zip", ".jar", ".war", ".pom", ".module", ".tar.gz", ".whl", ".crate", ".rpm", ".deb"]
 
     def generate_signature_and_verify(self, artifact: str, basepath: Path, signature_type: str) -> None:
@@ -35,7 +34,9 @@ class SignerPGP(Signer):
     def sign(self, artifact: str, basepath: Path, signature_type: str) -> None:
         filename = os.path.join(basepath, artifact)
         signature_file = filename + ".sig"
-        self.__remove_existing_signature__(signature_file)
+        if not self.overwrite:
+            self.__remove_existing_signature__(signature_file)
+
         signing_cmd = [
             "./opensearch-signer-client",
             "-i",
@@ -44,9 +45,11 @@ class SignerPGP(Signer):
             signature_file,
             "-p",
             "pgp",
+            "-r",
+            str(self.overwrite)
         ]
         self.git_repo.execute(" ".join(signing_cmd))
-        if(signature_type == ".asc"):
+        if (signature_type == ".asc"):
             self.__convert_to_asc(filename, signature_file)
 
     def __convert_to_asc(self, filename: str, signature_file: str) -> None:
--- a/src/sign_workflow/signers.py
+++ b/src/sign_workflow/signers.py
@@ -26,6 +26,6 @@ class Signers:
         return klass  # type: ignore[return-value]
 
     @classmethod
-    def create(cls, platform: str) -> Signer:
+    def create(cls, platform: str, overwrite: bool) -> Signer:
         klass = cls.from_platform(platform)
-        return klass()  # type: ignore[no-any-return, operator]
+        return klass(overwrite)  # type: ignore[no-any-return, operator]
--- a/tests/tests_sign_workflow/test_sign_args.py
+++ b/tests/tests_sign_workflow/test_sign_args.py
@@ -56,6 +56,14 @@ class TestSignArgs(unittest.TestCase):
     def test_platform_default(self) -> None:
         self.assertEqual(SignArgs().platform, "linux")
 
+    @patch("argparse._sys.argv", [SIGN_PY, OPENSEARCH_MANIFEST])
+    def test_overwrite_default(self) -> None:
+        self.assertEqual(SignArgs().overwrite, False)
+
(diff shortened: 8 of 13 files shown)


zelinh (commented):
I'm not sure about the use case for this one. Current logic for signing windows artifact is to create a new subfolder within the original file path and generate the signed artifact with the same name, at the mean time original unsigned file will stay in the same location. https://github.com/opensearch-project/opensearch-build/blob/2ffb0cbf0531a9a34b04fc0abaa988938e4336e1/src/sign_workflow/signer_windows.py#L33-L51
Does it have to be in the same location for your use case? Or could we add a single step to replace the old file with the new one from the `signed` folder?

gaiksaya:
@zelinh 
> I'm not sure about the use case for this one.

See the PR description. The default behavior is not changing just the check is being disabled at will. 
Thanks!

src/sign_workflow/signer_windows.py
```diff
@@ -42,6 +42,8 @@ def sign(self, artifact: str, basepath: Path, signature_type: str) -> None:
             signature_file,
```
zelinh, line 42: Sorry I might be missing something. I don't think this would work as you expected. We are still outputing the file we want to sign with a prefix file name. Are you trying to remove the old one and replace with the new signed one?

gaiksaya: No! so if you run with below command:
```
./sign.sh -i opensearch.tar.gz -o opensearch.tar.gz.sig -r True
```
with current set up if `opensearch.tar.gz.sig` exists in the workspace, the workflow will fail. Which I have kept as default behavior.
However, for windows artifacts example `.dll` files or any windows files, the backend code signs the actual artifact rather than generating a signature. In that case we want to be able to replace the original artifact with signed one. 
I know that you can just place that signed artifacts somewhere else too, but for above use case, I need the artifact in the exact same place so that I can package the signed dll with other bunch of files into final `.nupkg` package that can be published.

Hope this helps.

gaiksaya: >Current logic for signing windows artifact is to create a new subfolder within the original file path and generate the signed artifact with the same name, at the mean time original unsigned file will stay in the same location.

I think I missed that. Thanks! Let me check this logic as well.

src/sign_workflow/sign_args.py
```diff
@@ -28,6 +30,7 @@ def __init__(self) -> None:
         parser.add_argument("--type", help="Artifact type")
         parser.add_argument("--sigtype", choices=self.ACCEPTED_SIGNATURE_FILE_TYPES, help="Type of signature file.", default=".sig")
         parser.add_argument("--platform", choices=self.ACCEPTED_PLATFORM, help="Distribution platform.", default="linux")
+        parser.add_argument("-r", "--allow_output_override", choices=self.ACCEPTED_OVERRIDE_VALUES, help="Boolean to allow output file to override existing file", default="False")
```
peterzhuamazon, line 33: Suggest using dashes instead of underscore for these arguments.

src/sign_workflow/sign_artifacts.py
```diff
@@ -23,14 +23,16 @@ class SignArtifacts:
     artifact_type: str
     signature_type: str
     platform: str
+    allow_output_override: str
     signer: Signer
 
-    def __init__(self, target: Path, components: List[str], artifact_type: str, signature_type: str, platform: str) -> None:
+    def __init__(self, target: Path, components: List[str], artifact_type: str, signature_type: str, platform: str, allow_output_override: str) -> None:
```
peterzhuamazon, line 29: Why is string used here for the arg value?
I would expect True/False to be bool, or yes/no to be str for the value.

src/sign_workflow/sign_args.py
peterzhuamazon: The argument can be dash while the internal var name can be underscore.

peterzhuamazon: And I would remove the short version of the param as it is confusing.
Or add short term for all existing ones.

peterzhuamazon (changes requested):
So `allow_output_override` is a param that already available for use?
You just add a param to call it?

Thanks.

gaiksaya:
> So `allow_output_override` is a param that already available for use? You just add a param to call it?
> 
> Thanks.

The dependent PR adding that functionality is raised. See the PR description

src/sign_workflow/sign_args.py
gaiksaya: I had dashes before. Had to change to underscore because of https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/signArtifacts.groovy#L162. Groovy does not allow `-` so libtesters were giving issues. Will see if I can acomodate something else

src/sign_workflow/sign_artifacts.py
gaiksaya: Because in the underlying client is again shell and then python. So it's Groovy -> shell -> python -> shell -> python end to end workflow during which even a 'False' boolean is considered as True

src/sign_workflow/sign_args.py
peterzhuamazon: Sure thanks.

src/sign_workflow/sign_artifacts.py
peterzhuamazon: Got it, thanks.

peterzhuamazon:
> > So `allow_output_override` is a param that already available for use? You just add a param to call it?
> > Thanks.
> 
> The dependent PR adding that functionality is raised. See the PR description

What I mean to say is, in the signer client PR you did not add the actual `override` function.
Is that meaning the existing call already have override function, you just need a way to call it?

Thanks.

gaiksaya:
> What I mean to say is, in the signer client PR you did not add the actual `override` function. Is that meaning the existing call already have override function, you just need a way to call it?
> 
> Thanks.

The backend code does not care if you overwrite the artifact or not. It was the signer client that I had the check. So created a PR to support this in client

src/sign_workflow/signer_windows.py
gaiksaya: Refactored the logic to support this.

src/sign_workflow/signer_windows.py
```diff
@@ -21,31 +22,46 @@ class SignerWindows(Signer):
...
+                signature_file,
+                "-p",
+                "windows"
+            ]
+            self.git_repo.execute(" ".join(signing_cmd))
+            signed_folder = os.path.join(basepath, "signed")
+            if not os.path.exists(signed_folder):
+                os.mkdir(signed_folder)
+            signed_location = os.path.join(signed_folder, artifact)
+            os.rename(signature_file, signed_location)
```
peterzhuamazon, line 67: Instead of writing two blocks can we just do something similar to this?
https://github.com/opensearch-project/opensearch-build/pull/1726/files#diff-52d601c8af5bd5f8c66790da6f0e84de498460c18a30989e80634b33aba688c3

gaiksaya: It has whole another logic of creating a subfolder called `signed` by default. Should we change that behavior too?
I think we should and keep it like pgp signing where the generated signature is written in the same location as the original artifact. Let me know what you think.

peterzhuamazon: I just feel that the if statement is just for the `overrite=True or False`, which does not reward such a big behavior change on the matter.

If we can have a specific flag on such behavior changes it would be better.

Not a hard blocker for me now, let me know.

gaiksaya: The filename of the output file also varies in both statements. Hence keeping it that way. The default behavior is also weird. Not sure if we are actually using it anywhere today but surely would like to change it in future.

gaiksaya: Adding @zelinh to the conversation to get an idea why the old approach was taken.

zelinh: I see the logic here. If the signer is actually implemented to support doing this, I am fine with this logic. The reason that I created another `signed` subfolder is the signer doesn't support to override before.

gaiksaya: Got it so now since it supports override we can remove this logic completely right?

gaiksaya: I refactored the logic to support overwrite as well as just add `signed_` prefix if the overwrite is disabled. Let me know if the overwrite should be enabled for windows by default.
@prudhvigodithi @peterzhuamazon @zelinh

src/sign_workflow/sign_args.py
```diff
@@ -14,12 +14,14 @@
 class SignArgs:
     ACCEPTED_SIGNATURE_FILE_TYPES = [".sig", ".asc"]
     ACCEPTED_PLATFORM = ["linux", "windows"]
+    ACCEPTED_OVERWRITE_VALUES = ["True", "False"]
```
dblock, line 17: This should be a plain Boolean arg without a value and used with just `—overwrite`

src/sign_workflow/sign_args.py
```diff
@@ -14,12 +14,14 @@
...
 
     target: Path
     components: List[str]
     type: str
     sigtype: str
     platform: str
+    overwrite_existing: str
```
dblock, line 25: Boolean

dblock (changes requested):
Call this —overwrite and make sure it’s a Boolean flag. I also think it’s being passed around a lot, maybe there’s a way to refactor this with classes?

src/sign_workflow/sign_args.py
gaiksaya: done!

src/sign_workflow/sign_args.py
gaiksaya: Changed everything to boolean.

gaiksaya:
Hi @dblock @peterzhuamazon Can you please review the changes?
Need to get this for another PR.
Thanks!

docker/ci/dockerfiles/current/build.centos7.opensearch.x64.arm64.dockerfile
```diff
@@ -97,6 +97,10 @@ RUN pip3 install pip==21.3.1
 RUN pip3 install cmake==3.21.3
 RUN pip3 install awscli==1.22.12
 
+# Install osslsigncode tool
+RUN yum clean all && wget https://download-ib01.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/o/osslsigncode-2.2-1.el8.x86_64.rpm && \
+    yum install -y osslsigncode-2.2-1.el8.x86_64.rpm 
+
```
peterzhuamazon, line 103: We should avoid using wget if not need to, use `curl -SLO` instead.
Also why are we adding this tool specifically to the OS build image?

Thanks.

peterzhuamazon: Also this would not work as you are only installing the X64 version of the rpm, the ARM64 will fail after all.

peterzhuamazon (changes requested):

docker/ci/dockerfiles/current/build.centos7.opensearch.x64.arm64.dockerfile
gaiksaya: I can remove it from build image if it is not required now. I dont know what image we are using for signing today. https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/promotion/promote-artifacts.jenkinsfile#L45

Any idea?

gaiksaya: Made the necessary changes. Please review. Thanks!

src/sign_workflow/sign_args.py
```diff
@@ -20,14 +20,46 @@ class SignArgs:
...
+        parser.add_argument(
+            "--platform",
+            choices=self.ACCEPTED_PLATFORM,
+            help="Distribution platform.",
+            default="linux"
+        )
+        parser.add_argument(
+            "--overwrite",
+            action="store_true",
+            help="Boolean to allow output file to overwrite existing file",
```
dblock, line 59: Nit: make `help` consistent everywhere here, add missing periods at the end.

src/sign_workflow/sign_artifacts.py
```diff
@@ -23,14 +23,16 @@ class SignArtifacts:
     artifact_type: str
     signature_type: str
     platform: str
+    overwrite: bool
     signer: Signer
 
-    def __init__(self, target: Path, components: List[str], artifact_type: str, signature_type: str, platform: str) -> None:
+    def __init__(self, target: Path, components: List[str], artifact_type: str, signature_type: str, platform: str, overwrite: bool) -> None:
```
dblock, line 29: I think this one doesn't need to take `overwrite`. It will be passed into `sign(... overwrite)` later, it's not a property of a signer I think.

src/sign_workflow/sign_artifacts.py
```diff
@@ -42,10 +44,10 @@ def sign(self) -> None:
         logging.info("Done.")
 
     def __sign_artifacts__(self, artifacts: List[str], basepath: Path) -> None:
```
dblock, line 46: This virtual method should take `overwrite`.

src/sign_workflow/sign_artifacts.py
```diff
@@ -42,10 +44,10 @@ def sign(self) -> None:
...
-        self.signer.sign_artifacts(artifacts, basepath, self.signature_type)
+        self.signer.sign_artifacts(artifacts, basepath, self.signature_type, self.overwrite)
 
     def __sign_artifact__(self, artifact: str, basepath: Path) -> None:
```
dblock, line 49: This should take `overwrite`.

src/sign_workflow/sign_artifacts.py
```diff
@@ -57,9 +59,9 @@ def __signer_class__(self, path: Path) -> Type[Any]:
             return SignArtifactsExistingArtifactFile
 
     @classmethod
-    def from_path(self, path: Path, components: List[str], artifact_type: str, signature_type: str, platform: str) -> Any:
+    def from_path(self, path: Path, components: List[str], artifact_type: str, signature_type: str, platform: str, overwrite: bool) -> Any:
```
dblock, line 62: This should not take `overwrite`. The signer it returns is the same, whether we are trying to overwrite on signing or not.

dblock (changes requested):
You are still passing `overwrite` to instances of `SignArtifacts`, which I think is incorrect. I tried to point out where. I think a signer is the same (doesn't know about overwrite/don't overwrite), but when one calls `sign`, we pass the `overwrite` option into that and then the signer behaves differently (overwrites/doesn't overwrite).

_or_ the signer knows about whether it's an overwriting signer, and then you don't need to pass it around at all, and do `self.overwrite` in the right place to switch behavior. 

Right now you have both.

docker/ci/dockerfiles/current/build.centos7.opensearch.x64.arm64.dockerfile
```diff
@@ -97,6 +97,10 @@ RUN pip3 install pip==21.3.1
 RUN pip3 install cmake==3.21.3
 RUN pip3 install awscli==1.22.12
 
+# Install osslsigncode tool
+RUN arch=`uname -m` && yum clean all && curl -SLO https://download-ib01.fedoraproject.org/pub/epel/8/Everything/$arch/Packages/o/osslsigncode-2.2-1.el8.$arch.rpm && \
```
peterzhuamazon, line 101: This is not necessarily right as centos7 should not use rhel8 rpms.

gaiksaya: What do you suggest?

gaiksaya:
@dblock Regarding 
> or the signer knows about whether it's an overwriting signer, and then you don't need to pass it around at all, and do self.overwrite in the right place to switch behavior.

What would be the entry point here? Meaning how do I pass `args.overwrite` to the Signer class? Only `SignArtifacts.from_path` and `.sign` are called at the start
https://github.com/opensearch-project/opensearch-build/blob/main/src/run_sign.py#L21-L29

gaiksaya:
> You are still passing `overwrite` to instances of `SignArtifacts`, which I think is incorrect. I tried to point out where. I think a signer is the same (doesn't know about overwrite/don't overwrite), but when one calls `sign`, we pass the `overwrite` option into that and then the signer behaves differently (overwrites/doesn't overwrite).
> 
> _or_ the signer knows about whether it's an overwriting signer, and then you don't need to pass it around at all, and do `self.overwrite` in the right place to switch behavior.
> 
> Right now you have both.

I think the signer will need to know about overwrite because few other methods like `verify` depend on it. My question here is how would we pass the `overwrite` to signer in this case. Because the only point of entry is this https://github.com/opensearch-project/opensearch-build/pull/3118/files#diff-12026d6fa48dfc2ea7e9d91ee545ea373ebbdf957579abefb15390c63b8b50f9R21-R30

docker/ci/dockerfiles/current/build.centos7.opensearch.x64.arm64.dockerfile
gaiksaya: I did not find the package for centos 7 https://download-ib01.fedoraproject.org/pub/centos/7/

gaiksaya:
@dblock I refactored the code based on your comments. Please review! Thanks!

docker/ci/dockerfiles/current/build.centos7.opensearch.x64.arm64.dockerfile
peterzhuamazon: You are installing a rpm compiled for EL8 on EL7.
Though it might seems fine it is not officially supported.
Ex: https://access.redhat.com/articles/54483

Since we are doing the production here I recommend just do the compilation directly in dockerfile.
Or just get an additional image on Ubuntu2004, just for verify the signatures as it supports Ubuntu official repo.

Thanks.

gaiksaya: I reverted the change to build docker file. However for release clients I kept it for now. I tried compiling using the instructions here https://github.com/mtrojnar/osslsigncode#building but there were multiple build failures. We can use this image to sign and promote as well.

peterzhuamazon (changes requested):
This needs to be merged and built 1st:
* https://github.com/opensearch-project/opensearch-build/pull/3165

dblock (approved):
Better. As a practice you should rebase these long PRs down to 1 clean commit, so we don't have a lot of noise in squash merge description.

gaiksaya:
> Better. As a practice you should rebase these long PRs down to 1 clean commit, so we don't have a lot of noise in squash merge description.

Yeah will do that now. Was waiting if there were more changes to make. Thanks!

peterzhuamazon (approved):

dblock:
@gaiksaya if you have time, try rewriting this the other way around: not storing `overwrite` as a property of `Signer`. I think it will be a lot simpler.

PR #3159, opened 2023-01-31
Fix PA log path

dzane17:
Signed-off-by: David Zane <davizane@amazon.com>

### Description
Users are reporting large file sizes at `/usr/share/opensearch/logs/performance-analyzer.log` due to the log not rotating properly.

### Fix
Update PA log filepath in opensearch-build from `$OPENSEARCH_HOME/logs/performance-analyzer.log` to `/tmp/PerformanceAnalyzer.log` to allow for rotation by PA-RCA (https://github.com/opensearch-project/performance-analyzer-rca/blob/main/config/log4j2.xml#L7).

### Issues Resolved
Fixes https://github.com/opensearch-project/performance-analyzer/issues/226


sruti1312 (approved):

kiranprakash154 (approved):

khushbr (approved):

kaushalmahi12 (approved):

peterzhuamazon:
Please PR to this repo as well:
https://github.com/opensearch-project/docker-images/blob/main/2.x/bin/opensearch-docker-entrypoint.sh

Thanks.

dzane17:
docker-images repo PR: https://github.com/opensearch-project/docker-images/pull/13

gaiksaya:
Is this only affecting docker or any other distribution as well?

dzane17:
Seems like only docker is affected. Everyone in the [Github issue](https://github.com/opensearch-project/performance-analyzer/issues/226) is using docker and I see no reference of `/performance-analyzer.log` anywhere else in the codebase.

prudhvigodithi:
Hey do we have to even save the log of the PA agent, do users see the agent log now and then? I see this is being moved to /tmp and so I assume its not that useful, moving to /tmp can cause some implications when running `docker save <image>` if a user decided to save the running image and the /tmp file is too large (sample related [issue](https://github.com/moby/moby/issues/4396)), how about we remove the log generation for PA if its not useful?
@peterzhuamazon @gaiksaya @bbarani

khushbr:
> Hey do we have to even save the log of the PA agent, do users see the agent log now and then? I see this is being moved to /tmp and so I assume its not that useful, moving to /tmp can cause some implications when running `docker save <image>` if a user decided to save the running image and the /tmp file is too large (sample related [issue](https://github.com/moby/moby/issues/4396)), how about we remove the log generation for PA if its not useful? @peterzhuamazon @gaiksaya @bbarani @dzane17

The logs are certainly useful for debugging and monitoring purposes. Additionally, removing an application log file is never recommended. The docker image save issue should be addressed separately, and the solution should be generic.

prudhvigodithi:
Hey @khushbr if the logs are important then adding to /tmp is not recommended as there is no guarantee that it will be preserved, then the solution should come from the user who is using the docker to mount the /logs folder using the docker-compose file and configure auto rotation as per their own policy to rotate the logs, this log rotation coming from the product (OpenSearch) is not ideal.
@dblock can you add your thoughts as well.

prudhvigodithi:
The other way to handle this to use `logrotate` cli against the `/usr/share/opensearch/logs/performance-analyzer.log` that identifies the size and compress the log file into `gz` format and with timestamp added while creating a new file  `/usr/share/opensearch/logs/performance-analyzer.log` with fresh log, this is the right way to handle the log rotation as it saves the old log for user reference.

kkhatua:
I'm inclined to agree with @prudhvigodithi  on this. App logs should not go in tmp, even if they are being rotated. We've had cases where the kernel has deleted /tmp files after a specific time window, leading to nasty behaviour. 
@dzane17 , @khushbr 
We should look at another way of doing this, including size-based rotation and compress the old ones.

dzane17:
Looks like we already have time & size based rotation on `/tmp/PerformanceAnalyzer.log` [here](https://github.com/opensearch-project/performance-analyzer-rca/blame/main/config/log4j2.xml#L7
). Can we simply rotate `/usr/share/opensearch/logs/performance-analyzer.log` path instead and disregard the change in this PR?

Also, remove other references to `/PerformanceAnalyzer.log` in opensearch-project.
https://github.com/search?q=org%3Aopensearch-project+PerformanceAnalyzer.log&type=code

peterzhuamazon:
> Looks like we already have time & size based rotation on `/tmp/PerformanceAnalyzer.log` [here](https://github.com/opensearch-project/performance-analyzer-rca/blame/main/config/log4j2.xml#L7). Can we simply rotate `/usr/share/opensearch/logs/performance-analyzer.log` path instead and disregard the change in this PR?
> 
> Also, remove other references to `/PerformanceAnalyzer.log` in opensearch-project. https://github.com/search?q=org%3Aopensearch-project+PerformanceAnalyzer.log&type=code

That would probably impact other distributions such as tar/deb/rpm.

bbarani:
@dzane17 @peterzhuamazon Whats the next step?

PR #3165, opened 2023-02-01
Add osslsigncode compilation on CentOS7 Docker Image

peterzhuamazon:

### Description
Add osslsigncode compilation on CentOS7 Docker Image

### Issues Resolved
#3118 
See this comment:
https://github.com/opensearch-project/opensearch-build/pull/3118#discussion_r1091326973

This resolves the issue where the osslsigncode tool only compile on Ubuntu not on CentOS7.
And also update to the current latest stable 2.5 release, compares to EPEL 8 having 2.2 release only, with openssl1.1.1g.
https://download-ib01.fedoraproject.org/pub/epel/8/Everything/$arch/Packages/o/osslsigncode-2.2-1.el8.$arch.rpm

Thanks.

### DIFF
--- a/docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
+++ b/docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
@@ -91,7 +91,7 @@ RUN pip3 install twine cmake==3.24.1.1
 ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib:/usr/local/lib64:/usr/lib
 RUN yum install -y curl libcurl-devel libfaketime && yum remove -y openssl-devel perl-core pcre-devel && yum clean all && \
     mkdir -p /tmp/openssl && cd /tmp/openssl && \
-    curl -sSL -o- https://ftp.openssl.org/source/openssl-1.1.1g.tar.gz | tar -xz --strip-components 1 && \
+    curl -sSL -o- https://www.openssl.org/source/openssl-1.1.1g.tar.gz | tar -xz --strip-components 1 && \
     ./config --prefix=/usr --openssldir=/etc/ssl --libdir=lib shared zlib-dynamic && make && make install && \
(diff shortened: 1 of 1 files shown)


gaiksaya (approved):
Thanks for taking care of this.
Will wait for building a new image based on this dockerfile as I have few more packages to add to this image so all can go in 1 build

docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
```diff
@@ -85,7 +85,20 @@ RUN ln -sfn /usr/local/bin/python3.7 /usr/bin/python3 && \
     pip3 install pipenv && pipenv --version && pip install --upgrade pip
 
 # Install twine
-RUN pip3 install twine
+RUN pip3 install twine cmake==3.24.1.1
+
+# Install openssl1.1.1
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib:/usr/local/lib64:/usr/lib
+RUN yum install -y curl libcurl-devel libfaketime perl-core pcre-devel && yum remove -y openssl-devel && yum clean all && \
```
gaiksaya, line 92: We are installing `openssl-devel` on line 30 and then removing here. Is that intended?

peterzhuamazon: That is exactly intended as you need openssl-devel (1.0.2) for Python37 compilation.
Only when python37 compiled and installed then pip3 can be used for cmake3 installation.
Only when cmake3 installed you can then compile openssl 1.1.1 (while also removing the old 1.0.2), in order for the osslsigncode 2.5 to compile correctly.

Thanks.

gaiksaya: Got it. Thanks!

peterzhuamazon: All of this hassle is due to CentOS7 ship with openssl 1.0.2, not 1.1.1.

Issue #2885, opened 2022-11-12
[BUG] Add Datasources Keystore for SQL integTest on test clusters


peterzhuamazon:
Hi,

This command is required for 2.4.0 + SQL to pass integTest:

```
curl -SLO https://github.com/opensearch-project/sql/blob/2.4/integ-test/src/test/resources/catalog/catalog.json
echo y | bin/opensearch-keystore add-file plugins.query.federation.datasources.config catalog.json
```

Temp solution:
https://github.com/peterzhuamazon/opensearch-build/blob/sql-240-test/src/test_workflow/integ_test/service_opensearch.py#L50-L67

@vamsi-amazon also confirms this change in 2.4.0 integTest runs.

Thanks.

prudhvigodithi:
Hey @peterzhuamazon @vamsi-amazon, please check the following points.

- Keystore are used to save sensitive information, I dont see anything sensitive in catalog.json, can you please tell you do we need to add it to the keystore?
- Also the added temp solution https://github.com/peterzhuamazon/opensearch-build/blob/sql-240-test/src/test_workflow/integ_test/service_opensearch.py#L50-L67, just adds `list` to the keystore which does not specifically list the added `plugins.query.federation.datasources.config`, the right test would be to use `show`.
- Adding this setting to the `service_opensearch.py` for sql would not be generic. Different repos could have different Keystore values per version.

@dblock @gaiksaya @bbarani Please add your thoughts.

Thank you

peterzhuamazon:
> * https://github.com/peterzhuamazon/opensearch-build/blob/sql-240-test/src/test_workflow/integ_test/service_opensearch.py#L50-L67

Hi @prudhvigodithi 
* I proposed this before that keystore is added through SQL plugin build.gradle.
* You can ignore the `list` command as that is just a quick check on whether the keystore is imported or not.

vamsi-amazon:
@prudhvigodithi  catalog.json can consist of accessKey, secretKey, username and password. So that's the reason we are taking it through keystore.

zelinh:
Closing this issue as this has been fixed. By adding the customized integtest script in the SQL repo (https://github.com/opensearch-project/sql/pull/1332), now the local integ test I run are passing for 2.6.0 in 2.x branch. 
Way to reproduce the passing tests:
Add a temp test manifest for 2.6.0 and run below. 
```
[...]

PR #3187, opened 2023-02-08
Add catalog.json when starting integ test

zelinh:
### Description
This would be used for the integ tests on SQL plugins on tarball & windows specifically for v2.4.0 and later.
Running command
```
./test.sh integ-test manifests/2.5.0/opensearch-2.5.0-test.yml --component sql --paths opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.5.0/6976/linux/x64/tar
```
```
2023-02-08 00:13:41 INFO     | sql                  | with-security        | PASS  |
2023-02-08 00:13:41 INFO     | sql                  | without-security     | PASS  |
```

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2885

### DIFF
--- a/src/test_workflow/integ_test/service_opensearch.py
+++ b/src/test_workflow/integ_test/service_opensearch.py
@@ -7,6 +7,7 @@
 
 import logging
 import os
+import subprocess
 
 import requests
 import yaml
@@ -22,7 +23,7 @@ class ServiceOpenSearch(Service):
     dist: Distribution
     dependency_installer: DependencyInstaller
     install_dir: str
-    opensearch_yml_dir: str
+    opensearch_yml_path: str
     security_plugin_dir: str
 
     def __init__(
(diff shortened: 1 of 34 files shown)


src/test_workflow/integ_test/service_opensearch.py
```diff
@@ -47,6 +48,16 @@ def start(self) -> None:
...
 
+        # Additional for sql
+        print("start catalog")
+        self.catalog_json_path = os.path.join(self.install_dir, "catalog.json")
+        os.system(f"curl -SL https://raw.githubusercontent.com/opensearch-project/sql/2.4/integ-test/src/test/resources/catalog/catalog.json -o {self.catalog_json_path}")
+        subprocess.check_call(
+            f"echo y | {self.dist.keystore_cmd} add-file plugins.query.federation.datasources.config catalog.json",
+            shell=True,
+            cwd=self.install_dir,
+        )
```
gaiksaya, line 59: Can we do this is generic way? More plugins would need some or other things for their integration test. I believe SQL should add this via their gradle tasks (if possible).
Also what about previous versions (1.x series) that do not need this?
Something like this https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.5.0/opensearch-2.5.0-test.yml#L16-L17 I believe is the right way to go forward

gaiksaya (changes requested):
Adding @prudhvigodithi @dblock to recommend better solution to this request.

zelinh:
Closing this PR as the issue should have been resolved.

PR #3265, opened 2023-03-03
Changes to Tar, Rpm and Yum distributions in the validation workflow

Divyaasm:
### Description
Changes to Yum and Rpm.
1. Removed 'sudo system enable ' from rpm and yum distributions
2. Added code to remove open-search at the end of rpm and yum validation.
3. Constrained download_artifacts method to download the artifacts based on architecture  and eventually validate them later, earlier artifacts related to both x64 and arm64 used to get downloaded.
4. Added a statement to print the type of exception occurred if the workflow fails.

### Issues Resolved

### DIFF
--- a/src/validation_workflow/rpm/validation_rpm.py
+++ b/src/validation_workflow/rpm/validation_rpm.py
@@ -25,15 +25,9 @@ class ValidateRpm(Validation, DownloadUtils):
         self.tmp_dir = TemporaryDirectory()
 
     def download_artifacts(self) -> bool:
-        architectures = ["x64", "arm64"]
         for project in self.args.projects:
-            for architecture in architectures:
-                url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{architecture}.rpm"
-                if ValidateRpm.is_url_valid(url) and ValidateRpm.download(url, self.tmp_dir):
-                    logging.info(f"Valid URL - {url} and Download Successful !")
-                else:
-                    logging.info(f"Invalid URL - {url}")
-                    raise Exception(f"Invalid url - {url}")
+            url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{self.args.arch}.rpm"
+            self.check_url(url)
         return True
 
     def installation(self) -> bool:
@@ -50,7 +44,6 @@ class ValidateRpm(Validation, DownloadUtils):
     def start_cluster(self) -> bool:
         try:
             for project in self.args.projects:
-                execute(f'sudo systemctl enable {project}', ".")
                 execute(f'sudo systemctl start {project}', ".")
                 time.sleep(20)
                 (stdout, stderr, status) = execute(f'sudo systemctl status {project}', ".")
@@ -75,6 +68,7 @@ class ValidateRpm(Validation, DownloadUtils):
         try:
             for project in self.args.projects:
                 execute(f'sudo systemctl stop {project}', ".")
-        except:
-            raise Exception('Failed to Stop Cluster')
+                execute(f'sudo yum remove {project} -y', ".")
+        except Exception as e:
+            raise Exception(f'Exception occurred either while attempting to stop cluster or removing OpenSearch/OpenSearch-Dashboards. {str(e)}')
         return True
--- a/src/validation_workflow/tar/validation_tar.py
+++ b/src/validation_workflow/tar/validation_tar.py
@@ -28,15 +28,9 @@ class ValidateTar(Validation, DownloadUtils):
         self.osd_process = Process()
 
     def download_artifacts(self) -> bool:
-        architectures = ["x64", "arm64"]
         for project in self.args.projects:
-            for architecture in architectures:
-                url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{architecture}.tar.gz"
-                if ValidateTar.is_url_valid(url) and ValidateTar.download(url, self.tmp_dir):
-                    logging.info(f"Valid URL - {url} and Download Successful !")
-                else:
-                    logging.info(f"Invalid URL - {url}")
-                    raise Exception(f"Invalid url - {url}")
+            url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{self.args.arch}.tar.gz"
+            self.check_url(url)
         return True
 
     def installation(self) -> bool:
@@ -72,5 +66,5 @@ class ValidateTar(Validation, DownloadUtils):
             self.os_process.terminate()
             self.osd_process.terminate()
         except:
-            raise Exception('Failed to Stop Cluster')
+            raise Exception('Failed to terminate the processes that started OS and OSD')
         return True
--- a/src/validation_workflow/validation.py
+++ b/src/validation_workflow/validation.py
@@ -6,6 +6,7 @@
 # compatible open source license.
 
 
+import logging
 from abc import ABC, abstractmethod
 from typing import Any
 
@@ -18,11 +19,18 @@ class Validation(ABC):
         super().__init__()
         self.args = args
 
+    def check_url(self, url: str) -> bool:
+        if self.is_url_valid(url) and self.download(url, self.tmp_dir):  # type: ignore
+            logging.info(f"Valid URL - {url} and Download Successful !")
+            return True
+        else:
+            raise Exception(f"Invalid url - {url}")
+
     def run(self) -> Any:
         try:
             return self.download_artifacts() and self.installation() and self.start_cluster() and self.validation() and self.cleanup()
-        except Exception:
-            return False
+        except Exception as e:
+            raise Exception(f'An error occurred while running the validation tests: {str(e)}')
 
     @abstractmethod
     def download_artifacts(self) -> bool:
--- a/src/validation_workflow/yum/validation_yum.py
+++ b/src/validation_workflow/yum/validation_yum.py
@@ -26,11 +26,7 @@ class ValidateYum(Validation, DownloadUtils):
     def download_artifacts(self) -> bool:
         for project in self.args.projects:
             url = f"{self.base_url}{project}/{self.args.version[0:1]}.x/{project}-{self.args.version[0:1]}.x.repo"
-            if ValidateYum.is_url_valid(url) and ValidateYum.download(url, self.tmp_dir):
-                logging.info(f"Valid URL - {url} and Download Successful !")
-            else:
-                logging.info(f"Invalid URL - {url}")
-                raise Exception(f"Invalid url - {url}")
+            self.check_url(url)
         return True
 
     def installation(self) -> bool:
@@ -49,7 +45,6 @@ class ValidateYum(Validation, DownloadUtils):
     def start_cluster(self) -> bool:
         try:
             for project in self.args.projects:
-                execute(f'sudo systemctl enable {project}', ".")
                 execute(f'sudo systemctl start {project}', ".")
                 time.sleep(20)
                 execute(f'sudo systemctl status {project}', ".")
@@ -69,6 +64,7 @@ class ValidateYum(Validation, DownloadUtils):
         try:
             for project in self.args.projects:
                 execute(f'sudo systemctl stop {project}', ".")
-        except:
-            raise Exception('Failed to Stop Cluster')
+                execute(f'sudo yum remove {project} -y', ".")
+        except Exception as e:
+            raise Exception(f'Exception occurred either while attempting to stop cluster or removing OpenSearch/OpenSearch-Dashboards. {str(e)}')
         return True
--- a/tests/tests_validation_workflow/test_validation_rpm.py
+++ b/tests/tests_validation_workflow/test_validation_rpm.py
@@ -15,8 +15,9 @@ class TestValidationRpm(unittest.TestCase):
 
     @patch("validation_workflow.download_utils.DownloadUtils.is_url_valid", return_value=True)
     @patch("validation_workflow.download_utils.DownloadUtils.download", return_value=True)
+    @patch("validation_workflow.validation.Validation.check_url", return_value=True)
     @patch('validation_workflow.rpm.validation_rpm.ValidationArgs')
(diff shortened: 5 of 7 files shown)


src/validation_workflow/tar/validation_tar.py
```diff
@@ -28,15 +28,13 @@ def __init__(self, args: ValidationArgs) -> None:
...
-        architectures = ["x64", "arm64"]
         for project in self.args.projects:
-            for architecture in architectures:
-                url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{architecture}.tar.gz"
-                if ValidateTar.is_url_valid(url) and ValidateTar.download(url, self.tmp_dir):
-                    logging.info(f"Valid URL - {url} and Download Successful !")
-                else:
-                    logging.info(f"Invalid URL - {url}")
-                    raise Exception(f"Invalid url - {url}")
+            url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{self.args.arch}.tar.gz"
```
peterzhuamazon, line 32: Where is the self.args.arch coming from, could you point me to the code?

peterzhuamazon (changes requested):
Hi @Divyaasm some comment here thanks.

src/validation_workflow/validation.py
```diff
@@ -22,7 +22,7 @@ def run(self) -> Any:
         try:
             return self.download_artifacts() and self.installation() and self.start_cluster() and self.validation() and self.cleanup()
         except Exception:
-            return False
+            traceback.print_exc()
```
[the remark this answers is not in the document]
Divyaasm: Just wanted to print out the line in the code which failed during the validation.

src/validation_workflow/yum/validation_yum.py
```diff
@@ -69,6 +68,7 @@ def cleanup(self) -> bool:
         try:
             for project in self.args.projects:
                 execute(f'sudo systemctl stop {project}', ".")
+                execute(f'sudo yum remove {project} -y', ".")
         except:
             raise Exception('Failed to Stop Cluster')
```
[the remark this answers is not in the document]
Divyaasm: Sure Peter, Thanks

src/validation_workflow/yum/validation_yum.py
```diff
@@ -69,6 +68,7 @@ def cleanup(self) -> bool:
         try:
             for project in self.args.projects:
                 execute(f'sudo systemctl stop {project}', ".")
+                execute(f'sudo yum remove {project} -y', ".")
```
peterzhuamazon, line 71: Not really keen on the idea to run everything in `.`, I would rather we have some fixed working dir, such as the root of the build repo.

cc: @gaiksaya thoughts?

peterzhuamazon (commented):
One more.

src/validation_workflow/tar/validation_tar.py
Divyaasm: That's from the user input initialized in the init function could be either x64 or arm64 on which the workflow runs
https://github.com/opensearch-project/opensearch-build/blob/6191810fc47f8aa903a99ac6688da065d01bef70/src/validation_workflow/tar/validation_tar.py#LL23C16-L23C16

src/validation_workflow/yum/validation_yum.py
gaiksaya: Since this is for validation only I am okay to remove everything installed as a part of the package. Do you have any concerns regarding it?

Divyaasm: Hi @gaiksaya, I have addressed the issue to remove all opensearch installations before ending the workflow. I think Peter is talking about "." given as current working directory parameter to execute module instead of a specific path which appears in the logging statements for example `Executing "sudo systemctl stop opensearch" in .` in the output log. Not sure of the path where I need to run all the commands.

src/validation_workflow/rpm/validation_rpm.py
```diff
@@ -25,15 +25,13 @@ def __init__(self, args: ValidationArgs) -> None:
...
-        architectures = ["x64", "arm64"]
         for project in self.args.projects:
-            for architecture in architectures:
-                url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{architecture}.rpm"
-                if ValidateRpm.is_url_valid(url) and ValidateRpm.download(url, self.tmp_dir):
-                    logging.info(f"Valid URL - {url} and Download Successful !")
-                else:
-                    logging.info(f"Invalid URL - {url}")
-                    raise Exception(f"Invalid url - {url}")
+            url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{self.args.arch}.rpm"
```
gaiksaya, line 29: If the url does not end in `/` this will fail right? Maybe add a check or add the `/` and trim the `self.base_url`

src/validation_workflow/rpm/validation_rpm.py
```diff
@@ -75,6 +72,7 @@ def cleanup(self) -> bool:
         try:
             for project in self.args.projects:
                 execute(f'sudo systemctl stop {project}', ".")
+                execute(f'sudo yum remove {project} -y', ".")
         except:
-            raise Exception('Failed to Stop Cluster')
+            raise Exception('Exception occurred either while attempting to stop cluster or while removing opensearch')
```
gaiksaya, line 77: This can be OS or OSD so keep the message generic `{project}` would be the right parameter here

src/validation_workflow/tar/validation_tar.py
```diff
@@ -28,15 +28,13 @@ def __init__(self, args: ValidationArgs) -> None:
...
+            if ValidateTar.is_url_valid(url) and ValidateTar.download(url, self.tmp_dir):
+                logging.info(f"Valid URL - {url} and Download Successful !")
+            else:
+                logging.info(f"Invalid URL - {url}")
+                raise Exception(f"Invalid url - {url}")
```
gaiksaya, line 37: Do we need to log and throw exception both here?

src/validation_workflow/validation.py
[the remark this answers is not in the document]
gaiksaya: Remove traceback. Not a good practice to add this in prod code. throw an exception instead

src/validation_workflow/rpm/validation_rpm.py
Divyaasm: Don't think so @gaiksaya . self.base_url is predefined [here](https://github.com/opensearch-project/opensearch-build/blob/4a6e2f2118ff7381a9709b32ae75711ea3f41494/src/validation_workflow/tar/validation_tar.py#L25)
And the string format is as per download url's for opensearch, this basically fails only when the given version does not exist.

src/validation_workflow/rpm/validation_rpm.py
Divyaasm: Sure Sayali, will change it

src/validation_workflow/rpm/validation_rpm.py
```diff
@@ -25,15 +25,13 @@ def __init__(self, args: ValidationArgs) -> None:
...
-                    logging.info(f"Valid URL - {url} and Download Successful !")
-                else:
-                    logging.info(f"Invalid URL - {url}")
-                    raise Exception(f"Invalid url - {url}")
+            url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{self.args.arch}.rpm"
+            if ValidateRpm.is_url_valid(url) and ValidateRpm.download(url, self.tmp_dir):
+                logging.info(f"Valid URL - {url} and Download Successful !")
+            else:
+                logging.info(f"Invalid URL - {url}")
+                raise Exception(f"Invalid url - {url}")
```
dblock, line 34: Add a `check!` method in the parent class of `ValidateRpm` and `ValidateTar` that does this to avoid duplication.

src/validation_workflow/yum/validation_yum.py
```diff
@@ -69,6 +68,7 @@ def cleanup(self) -> bool:
...
         except:
-            raise Exception('Failed to Stop Cluster')
+            raise Exception('Exception occurred either while attempting to stop cluster or while removing opensearch')
```
dblock, line 73: opensearch -> OpenSearch

Add a period to the full sentence.

src/validation_workflow/rpm/validation_rpm.py
Divyaasm: > There are no tests to this code change, which is concerning.
> 
> * Refactor validation checks that are identical.
> * Add tests.

Thanks @dblock for reviewing the PR,
-Just the tarball, rpm and yum distributions use the same implementation for validating url's but not docker, So I had the the methods is_url_valid(url) and download(url, self.tmp_dir) in a seperate parent class and called them individually, I can actually have the complete code block in Function like` Check` to avoid duplication.

- So far the code changes in the PR are to edit the logging statements but does not alter function-calls or the return type of value, So this didn't not require any updates to the tests. I'll update them accordingly along with the code changes if needed.

src/validation_workflow/rpm/validation_rpm.py
```diff
@@ -75,6 +68,7 @@ def cleanup(self) -> bool:
         try:
             for project in self.args.projects:
                 execute(f'sudo systemctl stop {project}', ".")
-        except:
-            raise Exception('Failed to Stop Cluster')
+                execute(f'sudo yum remove {project} -y', ".")
+        except Exception as e:
+            raise Exception(f'Exception occurred either while attempting to stop cluster or removing OpenSearch {str(e)}')
```
gaiksaya, line 73: This can be dashboards too right? Either add {project} at class level to show actually what failed or make the message genetic saying `Exception occurred either while attempting to stop cluster or removing OpenSearch/OpenSearch-Dashboards`

gaiksaya (approved):
Just a small nit!

Divyaasm:
Hi @dblock @peterzhuamazon, could you please help in reviewing the addressed changes.
Thanks

dblock (changes requested):
This is a lot better!

I think you can delete more code.

src/validation_workflow/validation.py
```diff
@@ -18,11 +19,18 @@ def __init__(self, args: ValidationArgs) -> None:
         super().__init__()
         self.args = args
 
+    def check_url(self, url: str) -> bool:
+        if self.is_url_valid(url) and self.download(url, self.tmp_dir):  # type: ignore
```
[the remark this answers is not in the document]
Divyaasm: Thanks for the review @dblock,
Yes,` is_url_valid` does a HEAD request, and `download` downloads a file in a sequential manner at this point.

Right now in the line `if self.is_url_valid(url) and self.download(url, self.tmp_dir):` since `and` operator short circuits the download method is not called, it is only called when the url is valid and the validation part completely relies on the downloaded artifact as the ultimate goal is to validate this artifact for a respective distribution.

I thinks it's good to have a double verification before we download the artifact, and I can call the two methods seperately inside the check_url method if I'm not wrong and return a different exception when failed in each case as you said , let me know
Thanks.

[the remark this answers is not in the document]
dblock: Why do you need to double verify? In which case will download fail when HEAD succeeds?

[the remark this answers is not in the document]
Divyaasm: Makes sense @dblock , can I go by implementing only download(url, self.tmp_dir) and ignore is_url_valid(url). This gets the required job done.

src/validation_workflow/tar/validation_tar.py
```diff
@@ -28,15 +28,9 @@ def __init__(self, args: ValidationArgs) -> None:
...
         for project in self.args.projects:
-            for architecture in architectures:
-                url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{architecture}.tar.gz"
-                if ValidateTar.is_url_valid(url) and ValidateTar.download(url, self.tmp_dir):
-                    logging.info(f"Valid URL - {url} and Download Successful !")
-                else:
-                    logging.info(f"Invalid URL - {url}")
-                    raise Exception(f"Invalid url - {url}")
+            url = f"{self.base_url}{project}/{self.args.version}/{project}-{self.args.version}-linux-{self.args.arch}.tar.gz"
+            self.check_url(url)
```
dblock, line 33: The method here is `check_url` whereas the calling function is called `download_artifacts`. So is it trying to download or check artifacts?

src/validation_workflow/tar/validation_tar.py
```diff
@@ -72,5 +66,5 @@ def cleanup(self) -> bool:
             self.os_process.terminate()
             self.osd_process.terminate()
         except:
-            raise Exception('Failed to Stop Cluster')
+            raise Exception('Failed to terminate the processes that started OS and OSD')
```
dblock, line 69: We spell out `OpenSearch` and `OpenSearch Dashboards` in other exceptions, and finish sentences with `.`. Can we try to please make errors consistent?

Divyaasm: Sure @dblock, I'll update all the output messages to stay in format.

src/validation_workflow/validation.py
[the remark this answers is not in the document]
Divyaasm: Th idea of validation workflow is to verify that the OS and OSD artifacts from the OpenSearch official site can be successfully downloaded and to validate them as in the [documentation](https://opensearch.org/docs/latest/install-and-configure/install-opensearch/tar/) , mainly used to validate the artifacts for a particular version during releases and is independent of distribution and architecture. The base for the mechanism are the files that are downloaded from the official site.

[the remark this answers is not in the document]
dblock: I see. The method names and the location of downloading is confusing then. There should be a `download` step that is clearly doing that.

src/validation_workflow/tar/validation_tar.py
dblock: Ok. Add a clear step to do the download.

Divyaasm:
Closing as the all the changes were addressed in the PR #3993

PR #3269, opened 2023-03-07
Revert "Add dotnet to clients' release docker ci image"

gaiksaya:
Reverts opensearch-project/opensearch-build#3266

PR #3274, opened 2023-03-07
Add dotnet compilation for arm64 specifically

### LINKED ISSUE #3141: Onboard opensearch-net to new release process

[opensearch-net](https://github.com/opensearch-project/opensearch-net) is still using a manual release process. It needs onboarding to a new automated one using one-click release.

Adapt to #1234, Depends on #2965 

Continue using manual approach

### Additional context

_No response_

peterzhuamazon:
### Description
Add dotnet compilation for arm64 specifically

### Issues Resolved
#3141 #3269

### DIFF
--- a/docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
+++ b/docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
@@ -32,20 +32,20 @@ RUN yum install -y @development zlib-devel bzip2 bzip2-devel readline-devel sqli
 # Add Yarn dependencies
 RUN yum groupinstall -y "Development Tools" && yum clean all && rm -rf /var/cache/yum/*
 
-# Install dotnet
+# Installing dotnet
 ARG DOT_NET_LIST="6.0"
 RUN rpm -Uvh https://packages.microsoft.com/config/centos/7/packages-microsoft-prod.rpm && \
     for dotnet_version in $DOT_NET_LIST; do yum install -y dotnet-sdk-$dotnet_version; done
 
 RUN if [[ `uname -m` = 'aarch64' ]]; then mkdir -p aarch64-builds && cd aarch64-builds && \
-    echo "Install higher version of libstdc++ from Anaconda" && \
+    echo "Installing higher version of libstdc++ from Anaconda" && \
     curl -SL https://repo.anaconda.com/pkgs/main/linux-aarch64/libstdcxx-devel_linux-aarch64-11.2.0-h1234567_1.tar.bz2 -o libstdxxx-devel.tar.br2 && \
     tar --strip-components 2 -xjvf libstdxxx-devel.tar.br2 aarch64-conda-linux-gnu/lib64/libstdc++.so.6.0.29 && mv -v libstdc++.so.6.0.29 /lib64 && \
     ln -sfn /lib64/libstdc++.so.6.0.29 /lib64/libstdc++.so.6 && \
-    echo "Install glibc 2.18" && \
+    echo "Installing glibc 2.18" && \
     curl -SLO https://ftp.gnu.org/gnu/glibc/glibc-2.18.tar.gz && tar -xzvf glibc-2.18.tar.gz && cd glibc-2.18 && mkdir -p build && cd build && \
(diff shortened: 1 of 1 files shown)


docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
```diff
@@ -26,18 +26,34 @@ RUN yum clean all && yum-config-manager --add-repo https://cli.github.com/packag
...
-
-
 # Add Python37 dependencies
 RUN yum install -y @development zlib-devel bzip2 bzip2-devel readline-devel sqlite sqlite-devel openssl-devel xz xz-devel libffi-devel findutils
 
 # Add Yarn dependencies
 RUN yum groupinstall -y "Development Tools" && yum clean all && rm -rf /var/cache/yum/*
 
+# Install dotnet
+ARG DOT_NET_LIST="5.0 6.0"
```
gaiksaya, line 36: Just keep 6.0 in the list. Confirmed with the team we only need 6.0 for build.

docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
```diff
@@ -26,18 +26,34 @@ RUN yum clean all && yum-config-manager --add-repo https://cli.github.com/packag
...
+    tar --strip-components 2 -xjvf libstdxxx-devel.tar.br2 aarch64-conda-linux-gnu/lib64/libstdc++.so.6.0.29 && mv -v libstdc++.so.6.0.29 /lib64 && \
+    ln -sfn /lib64/libstdc++.so.6.0.29 /lib64/libstdc++.so.6 && \
+    echo "Install glibc 2.18" && \
+    curl -SLO https://ftp.gnu.org/gnu/glibc/glibc-2.18.tar.gz && tar -xzvf glibc-2.18.tar.gz && cd glibc-2.18 && mkdir -p build && cd build && \
+    ../configure --prefix=/usr && make && make install && cd ../../ && \
+    echo "Install libicu 53+" && \
+    rpm -e --nodeps libicu && \
+    curl -SLO https://github.com/unicode-org/icu/releases/download/release-53-2/icu4c-53_2-src.tgz && tar -xzvf icu4c-53_2-src.tgz && cd icu && \
+    cd source && ./configure --prefix=/usr && make && make install && \
+    cd ../../../ && rm -rf aarch64-builds; fi
```
gaiksaya, line 52: Would it be better if we add dotnet.sh just like yq and jdk in config folder here: https://github.com/opensearch-project/opensearch-build/tree/main/docker/ci/config

gaiksaya (changes requested):

docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
gaiksaya: Looks like the issue is only on centos arm64 so this is an edge case. Just a nit: echo "Installing ...." rather than install/  Also are you sure the `cd ../../..` wont cause an issue? Should we do something like NVM_DIR and give absolute paths?

peterzhuamazon: white_check_mark
eyes
raised_hands





We do not need a script for this case.

This step is specifically for centos7 arm64.

In centos7 it has the glibc 2.17, that is why we compile knn specifically on centos7.

The dotnet team requires a version above the lowest version on centos7 arm64 specifically.

You can see I replaced the stdcxx, then replace icu

Because those requirements are on Ubuntu arm64, not centos7 arm64.

gaiksaya (approved):

docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
peterzhuamazon: The `cd ../../../ ` wont have any issues because they are only going through dirs created by the new block of scripts.

gaiksaya:
Image is building https://build.ci.opensearch.org/job/docker-build/3158/

PR #3283, opened 2023-03-08
Trigger all tests and validations asynchronously for build pipelines

gaiksaya:
### Description
CUrrently the build pipeline takes forever to complete as it waits for subsequent  test workflows to complete as well. 
This change triggers the tests and validation workflows asynchronously and does not wait for them to complete making build independent of tests. 

Ref: https://www.jenkins.io/doc/pipeline/steps/pipeline-build-step/

### Issues Resolved

### DIFF
--- a/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
+++ b/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
@@ -200,7 +200,7 @@ pipeline {
                                         def integTestResults =
                                             build job: INTEG_TEST_JOB_NAME,
                                             propagate: false,
-                                            wait: true,
+                                            wait: false,
                                             parameters: [
                                                 string(name: 'TEST_MANIFEST', value: TEST_MANIFEST),
                                                 string(name: 'BUILD_MANIFEST_URL', value: buildManifestUrl),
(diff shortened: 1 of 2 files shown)


peterzhuamazon:
I think rpm validation at least need some wait right?
Test is one thing but rpm validation is literally catch early if the rpm package is valid or not.
It wont take very long either.

gaiksaya:
> I think rpm validation at least need some wait right? Test is one thing but rpm validation is literally catch early if the rpm package is valid or not. It wont take very long either.

You mean you want to hard fail if validations fail? Wondering if this can be a development blocker.

prudhvigodithi (approved):

prudhvigodithi:
> I think rpm validation at least need some wait right?
> Test is one thing but rpm validation is literally catch early if the rpm package is valid or not.
> It wont take very long either.

@peterzhuamazon I feel we should isolate everything that is not part of build and assemble from this distribution job, so that it can be as resilient as possible.

peterzhuamazon (approved):

peterzhuamazon:
> > I think rpm validation at least need some wait right?
> > Test is one thing but rpm validation is literally catch early if the rpm package is valid or not.
> > It wont take very long either.
> 
> @peterzhuamazon I feel we should isolate everything that is not part of build and assemble from this distribution job, so that it can be as resilient as possible. We should track all those other validation/test/scan jobs outside of this distribution job, considering each job has its own use case and one should not depend on each other.

True, thanks!a

Issue #2144, opened 2022-05-24
[Enhancement] Run Dashboards integTest in parallel on Jenkins


peterzhuamazon:
Hi,

This is a build repo follow up to https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/110.

1. Add scripts / tweaks in FT repo script finder to allow components separation.
2. Python code in build repo able to pass --component properly or add something like --test to specify specific test.
3. Tweak jenkinsfile to consume this changes and run on Jenkins.

- [x] Add support in python code base
- [x] https://github.com/opensearch-project/opensearch-build/issues/3412

peterzhuamazon:
* 20220524: https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/234

gaiksaya:
In order to run integration tests for dashboard plugins, the plugins need to be part of the test manifest https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.5.0/opensearch-dashboards-2.5.0-test.yml

Right now only functional-test repo is added as one of the component.

One of the approach to go forward is have functional test repo as default and then parse through the manifest to collect all the components and pass each component as `-t <component>` to functional test repo.

gaiksaya:
### Approach:

Clone functional test repo as default and add all plugins as a single component to test manifest

In this approach, functional test repo will be cloned by default based on input manifest (branch). We will add all other components to test manifest as below. The test workflow will iterate over all the components and pass the component name as parameter to functional test repo.
This makes sure we have all the components in test manifest for accountability and single point of truth what tests are run for which component and with what test configuration.

```yaml
---
schema-version: '1.0'
name: OpenSearch Dashboards
ci:
  image:
    name: opensearchstaging/ci-runner:ci-runner-rockylinux8-opensearch-dashboards-integtest-v2
components:
  - name: OpenSearch-Dashboards
    bwc-test:
      test-configs:
        - with-security
        - without-security
  - name: observabilityDashboards
    integ-test:
      test-configs:
        - with-security
        - without-security
  - name: securityDashboards
    integ-test:
      test-configs:
        - with-security
        - without-security
  - name: anomalyDetectionDashboards
     integ-test:
      test-configs:
        - with-security
        - without-security

Divyaasm:
Closed via https://github.com/opensearch-project/opensearch-build/pull/3316 and https://github.com/opensearch-project/opensearch-build/pull/3465

PR #3316, opened 2023-03-16
Updated integ-tests implementation to run all the OSD plugins. 

Divyaasm:
Signed-by: divyaasm@amazon.com

Fix to issue https://github.com/opensearch-project/opensearch-build/issues/2144
### Description
-Made changes to run OSD plugins altogether or individually using --component arg.
-OSD integtests when triggered from buildrepo are now executed via opensearch-dashboards-functional-test repo.
-Updated test-cases accordingly.

### DIFF
--- a/src/test_workflow/integ_test/integ_test_suite.py
+++ b/src/test_workflow/integ_test/integ_test_suite.py
@@ -6,22 +6,15 @@
 # compatible open source license.
 
 import abc
-import json
 import logging
-import os
 from pathlib import Path
 from typing import Any, Dict
 
-from git.git_repository import GitRepository
 from manifests.build_manifest import BuildManifest
 from manifests.bundle_manifest import BundleManifest
-from paths.script_finder import ScriptFinder
-from system.execute import execute
 from test_workflow.dependency_installer import DependencyInstaller
-from test_workflow.integ_test.topology import ClusterEndpoint, NodeEndpoint
 from test_workflow.test_recorder.log_recorder import LogRecorder
 from test_workflow.test_recorder.test_recorder import TestRecorder
-from test_workflow.test_recorder.test_result_data import TestResultData
 from test_workflow.test_result.test_component_results import TestComponentResults
 
 
@@ -33,7 +26,6 @@ class IntegTestSuite(abc.ABC):
     dependency_installer: DependencyInstaller
     bundle_manifest: BundleManifest
     build_manifest: BuildManifest
-    repo: GitRepository
     save_logs: LogRecorder
     additional_cluster_config: dict
 
@@ -61,13 +53,6 @@ class IntegTestSuite(abc.ABC):
         self.bundle_manifest = bundle_manifest
         self.build_manifest = build_manifest
 
-        self.repo = GitRepository(
-            self.component.repository,
-            self.component.commit_id,
-            os.path.join(self.work_dir, self.component.name),
-            test_config.working_directory
-        )
-
         self.save_logs = test_recorder.test_results_logs
         self.additional_cluster_config = None
 
@@ -75,46 +60,9 @@ class IntegTestSuite(abc.ABC):
     def execute_tests(self) -> TestComponentResults:
         pass
 
+    @abc.abstractmethod
     def multi_execute_integtest_sh(self, cluster_endpoints: list, security: bool, test_config: str) -> int:
-        script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
-
-        def custom_node_endpoint_encoder(node_endpoint: NodeEndpoint) -> dict:
-            return {"endpoint": node_endpoint.endpoint, "port": node_endpoint.port, "transport": node_endpoint.transport}
-        if os.path.exists(script):
-            if len(cluster_endpoints) == 1:
-                single_data_node = cluster_endpoints[0].data_nodes[0]
-                cmd = f"bash {script} -b {single_data_node.endpoint} -p {single_data_node.port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
-            else:
-                endpoints_list = []
-                for cluster_details in cluster_endpoints:
-                    endpoints_list.append(cluster_details.__dict__)
-                endpoints_string = json.dumps(endpoints_list, indent=0, default=custom_node_endpoint_encoder)
-                cmd = f"bash {script} -e '"
-                cmd = cmd + endpoints_string + "'"
-                cmd = cmd + f" -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
-            self.repo_work_dir = os.path.join(
-                self.repo.dir, self.test_config.working_directory) if self.test_config.working_directory is not None else self.repo.dir
-            (status, stdout, stderr) = execute(cmd, self.repo_work_dir, True, False)
-            test_result_data = TestResultData(
-                self.component.name,
-                test_config,
-                status,
-                stdout,
-                stderr,
-                self.test_artifact_files
-            )
-            self.save_logs.save_test_result_data(test_result_data)
-            if stderr:
-                logging.info("Stderr reported for component: " + self.component.name)
-                logging.info(stderr)
-            return status
-        else:
-            logging.info(f"{script} does not exist. Skipping integ tests for {self.component.name}")
-            return 0
-
-    def execute_integtest_sh(self, endpoint: str, port: int, security: bool, test_config: str) -> int:
-        cluster_endpoint_port = [ClusterEndpoint("cluster1", [NodeEndpoint(endpoint, port, 9300)], [])]
-        return self.multi_execute_integtest_sh(cluster_endpoint_port, security, test_config)
+        pass
 
     def is_security_enabled(self, config: str) -> bool:
         if config in ["with-security", "without-security"]:
--- a/src/test_workflow/integ_test/integ_test_suite_opensearch.py
+++ b/src/test_workflow/integ_test/integ_test_suite_opensearch.py
@@ -6,23 +6,29 @@
 # compatible open source license.
 
 import glob
+import json
 import logging
 import os
 from pathlib import Path
 from typing import Any
 
+from git.git_repository import GitRepository
 from manifests.build_manifest import BuildManifest
 from manifests.bundle_manifest import BundleManifest
+from paths.script_finder import ScriptFinder
+from system.execute import execute
 from test_workflow.dependency_installer_opensearch import DependencyInstallerOpenSearch
 from test_workflow.integ_test.integ_test_suite import IntegTestSuite, InvalidTestConfigError
-from test_workflow.integ_test.topology import Topology
+from test_workflow.integ_test.topology import NodeEndpoint, Topology
 from test_workflow.test_recorder.test_recorder import TestRecorder
+from test_workflow.test_recorder.test_result_data import TestResultData
 from test_workflow.test_result.test_component_results import TestComponentResults
 from test_workflow.test_result.test_result import TestResult
 
 
 class IntegTestSuiteOpenSearch(IntegTestSuite):
     dependency_installer: DependencyInstallerOpenSearch
+    repo: GitRepository
 
     def __init__(
         self,
@@ -42,6 +48,12 @@ class IntegTestSuiteOpenSearch(IntegTestSuite):
             dependency_installer_opensearch,
             bundle_manifest_opensearch,
             build_manifest_opensearch
+        )
+        self.repo = GitRepository(
+            self.component.repository,
+            self.component.commit_id,
+            os.path.join(self.work_dir, self.component.name),
+            test_config.working_directory
         )
 
     def execute_tests(self) -> TestComponentResults:
@@ -88,6 +100,43 @@ class IntegTestSuiteOpenSearch(IntegTestSuite):
             os.chdir(self.work_dir)
             self.pretty_print_message("Running integration tests for " + self.component.name + " " + config)
             return self.multi_execute_integtest_sh(endpoints, security, config)
+
+    def multi_execute_integtest_sh(self, cluster_endpoints: list, security: bool, test_config: str) -> int:
+        script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
+
+        def custom_node_endpoint_encoder(node_endpoint: NodeEndpoint) -> dict:
+            return {"endpoint": node_endpoint.endpoint, "port": node_endpoint.port, "transport": node_endpoint.transport}
+        if os.path.exists(script):
+            if len(cluster_endpoints) == 1:
+                single_data_node = cluster_endpoints[0].data_nodes[0]
+                cmd = f"bash {script} -b {single_data_node.endpoint} -p {single_data_node.port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
+            else:
+                endpoints_list = []
+                for cluster_details in cluster_endpoints:
+                    endpoints_list.append(cluster_details.__dict__)
+                endpoints_string = json.dumps(endpoints_list, indent=0, default=custom_node_endpoint_encoder)
+                cmd = f"bash {script} -e '"
+                cmd = cmd + endpoints_string + "'"
+                cmd = cmd + f" -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
+            self.repo_work_dir = os.path.join(
+                self.repo.dir, self.test_config.working_directory) if self.test_config.working_directory is not None else self.repo.dir
+            (status, stdout, stderr) = execute(cmd, self.repo_work_dir, True, False)
+            test_result_data = TestResultData(
+                self.component.name,
+                test_config,
+                status,
+                stdout,
+                stderr,
+                self.test_artifact_files
+            )
+            self.save_logs.save_test_result_data(test_result_data)
+            if stderr:
+                logging.info("Stderr reported for component: " + self.component.name)
+                logging.info(stderr)
+            return status
+        else:
+            logging.info(f"{script} does not exist. Skipping integ tests for {self.component.name}")
+            return 0
 
     @property
     def test_artifact_files(self) -> dict:
--- a/src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
+++ b/src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
@@ -5,17 +5,23 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+import logging
 import os
 from pathlib import Path
 
+from git.git_repository import GitRepository
 from manifests.build_manifest import BuildManifest
 from manifests.bundle_manifest import BundleManifest
 from manifests.test_manifest import TestComponent
+from paths.script_finder import ScriptFinder
+from system.execute import execute
 from test_workflow.dependency_installer_opensearch import DependencyInstallerOpenSearch
 from test_workflow.dependency_installer_opensearch_dashboards import DependencyInstallerOpenSearchDashboards
 from test_workflow.integ_test.integ_test_suite import IntegTestSuite
 from test_workflow.integ_test.local_test_cluster_opensearch_dashboards import LocalTestClusterOpenSearchDashboards
+from test_workflow.integ_test.topology import ClusterEndpoint, NodeEndpoint
 from test_workflow.test_recorder.test_recorder import TestRecorder
+from test_workflow.test_recorder.test_result_data import TestResultData
 from test_workflow.test_result.test_component_results import TestComponentResults
 from test_workflow.test_result.test_result import TestResult
 
@@ -24,6 +30,7 @@ class IntegTestSuiteOpenSearchDashboards(IntegTestSuite):
     dependency_installer_opensearch_dashboards: DependencyInstallerOpenSearchDashboards
     bundle_manifest_opensearch_dashboards: BundleManifest
     build_manifest_opensearch_dashboards: BuildManifest
+    repo: GitRepository
 
     def __init__(
         self,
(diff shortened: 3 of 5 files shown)


src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -114,7 +116,8 @@ def custom_node_endpoint_encoder(node_endpoint: NodeEndpoint) -> dict:
 
     def execute_integtest_sh(self, endpoint: str, port: int, security: bool, test_config: str) -> int:
         cluster_endpoint_port = [ClusterEndpoint("cluster1", [NodeEndpoint(endpoint, port, 9300)], [])]
-        return self.multi_execute_integtest_sh(cluster_endpoint_port, security, test_config)
+        script = ScriptFinder.find_integ_test_script("functionalTestDashboards", "https://github.com/opensearch-project/opensearch-dashboards-functional-test.git")
```
gaiksaya, line 119: This will break for opensearch plugins. functional test only applies to dashboards. Also the I don't see `ref` anywhere, this will by default clone main. You need to get the repo link and ref from input manifest, https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.6.0/opensearch-dashboards-2.6.0.yml#L13-L15

src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -75,15 +75,16 @@ def __init__(
...
         def custom_node_endpoint_encoder(node_endpoint: NodeEndpoint) -> dict:
-            return {"endpoint": node_endpoint.endpoint, "port": node_endpoint.port, "transport": node_endpoint.transport}
+            return {"endpoint": node_endpoint.endpoint, "port": node_endpoint.port,
+                    "transport": node_endpoint.transport}
+
         if os.path.exists(script):
             if len(cluster_endpoints) == 1:
                 single_data_node = cluster_endpoints[0].data_nodes[0]
-                cmd = f"bash {script} -b {single_data_node.endpoint} -p {single_data_node.port} -s {str(security).lower()} -v {self.bundle_manifest.build.version}"
+                cmd = f"bash {script} -b {single_data_node.endpoint} -p {single_data_node.port} -s {str(security).lower()} -t {self.component.name} -v {self.bundle_manifest.build.version}"
```
gaiksaya, line 87: This is again not true for opensearch components,

gaiksaya (changes requested):
I believe the methods need to be separated into child classes due to `-t` option. It only applies to dashboards and not opensearch.

src/test_workflow/integ_test/integ_test_suite.py
Divyaasm: I see the function execute_integtest_sh in used only for OSD which calls multi_execute_integtest_sh, whereas OS directly uses multi_execute_integtest_sh.

gaiksaya: Yeah we need to change that behavior. Can you fix that as well? Adding @peterzhuamazon @prudhvigodithi for more input

src/test_workflow/integ_test/integ_test_suite.py
Divyaasm: I can make multi_execute_integtest_sh abstract and implement it separately for OS and OSD.

peterzhuamazon: Just use a filter to remove -t if not needed.

src/test_workflow/integ_test/integ_test_suite.py
peterzhuamazon: Use the git lib in build repo to checkout ftrepo, do not need to use a find script call like this, because if cwd already have ftrepo content, you can just call the find script for integtest.sh and it will find it in root.

src/test_workflow/integ_test/integ_test_suite_opensearch.py
```diff
@@ -87,7 +88,8 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
         ) as (endpoints):
             os.chdir(self.work_dir)
             self.pretty_print_message("Running integration tests for " + self.component.name + " " + config)
-            return self.multi_execute_integtest_sh(endpoints, security, config)
+            script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
```
peterzhuamazon, line 91: Why do we need to call the scriptfinder here?

src/test_workflow/integ_test/integ_test_suite.py
peterzhuamazon: Another problem is what if user does not enter any component, then it should default run everything in the yml. Therefore, you method might not work as it will include ftrepo as a component to run.

peterzhuamazon (commented):
Hi @Divyaasm, some comments here thanks.

src/test_workflow/integ_test/integ_test_suite.py
Divyaasm: Sure Peter, started working on this change. Trying to fetch the integtest.sh from ftrepo based on ref branch from input manifest.

src/test_workflow/integ_test/integ_test_suite.py
Divyaasm: Hey @peterzhuamazon , the plan is to implement the multi_execute_integtest_sh in the child classes for OS and OSD. If that's the case I think the implementaion works well for OS too, let me know if I just need to add a filter to remove -t instead of having different implementations OS and OSD.

Currently I can see the tests are running sequentially for all OSD and OS plugins in yml if we do not provide --component names.

src/test_workflow/integ_test/integ_test_suite_opensearch.py
Divyaasm: To make sure I separate the ftrepo clone for dashboards, Will change it back in the child class implementations.

peterzhuamazon:
Have offline discussion with @Divyaasm on this and there are several issues:
1. The ftrepo is not checked out at all as the initial implementation assuming only components within the manifest get checked out, so removing it from the manifest would cause it not to be checked out and actually trying to use the default gradle integtest to run cypress OSD test.
2. @Divyaasm will do a filter `none` check before sending the component to `-t`. So that we can make sure the component has value before using -t, else remove.
https://github.com/dblock/opensearch-build/blob/8babdde32e2262b9e2de3e3672daa198bef74c4f/src/build_workflow/builder_from_source.py#L48
3. Refine how the script finder is called once ftrepo is checked out.


```
Service is available
2023-03-17 18:43:05 INFO     ===============================================
2023-03-17 18:43:05 INFO     Running integration tests for customImportMapDashboards
2023-03-17 18:43:05 INFO     ===============================================
2023-03-17 18:43:05 INFO     Executing "bash /local/home/divyaasm/divyaasm/opensearch-build/scripts/default/integtest.sh -b localhost -p 5601 -s true -t customImportMapDashboards -v 2.6.0" in /tmp/tmpv4kasw4_/customImportMapDashboards
2023-03-17 18:43:05 INFO     Recording component test results for customImportMapDashboards at /local/home/divyaasm/divyaasm/opensearch-build/test-results/2e5262448e9e454cb9790bc9892f6415/integ-test/customImportMapDashboards/with-security/test-results
2023-03-17 18:43:05 INFO     Sending SIGKILL to PID 27123
2023-03-17 18:43:05 INFO     Process killed with exit code None
2023-03-17 18:43:05 INFO     Cleanup /tmp/tmpv4kasw4_/local-test-cluster/* content after the test
2023-03-17 18:43:06 INFO     Sending SIGKILL to PID 27241
2023-03-17 18:43:06 INFO     Process killed with exit code None
2023-03-17 18:43:06 INFO     Cleanup /tmp/tmpv4kasw4_/local-test-cluster/* content after the test
2023-03-17 18:43:06 INFO     Recording local cluster logs for customImportMapDashboards with test configuration as with-security at /local/home/divyaasm/divyaasm/opensearch-build/test-results/2e5262448e9e454cb9790bc9892f6415/integ-test/customImportMapDashboards/with-security/local-cluster-logs
2023-03-17 18:43:06 INFO     Recording local cluster logs for customImportMapDashboards with test configuration as with-security at /local/home/divyaasm/divyaasm/opensearch-build/test-results/2e5262448e9e454cb9790bc9892f6415/integ-test/customImportMapDashboards/with-security/local-cluster-logs
2023-03-17 18:43:06 INFO     opensearch distribution: tar
2023-03-17 18:43:06 INFO     opensearch-dashboards distribution: tar
2023-03-17 18:43:06 INFO     Downloading bundle artifact
2023-03-17 18:43:06 INFO     Downloading https://ci.opensearch.org/ci/dbc/distribution-build-opensearch-dashboards/2.6.0/5520/linux/x64/tar/dist/opensearch-dashboards/opensearch-dashboards-2.6.0-linux-x64.tar.gz into /tmp/tmpv4kasw4_/local-test-cluster/opensearch-dashboards-2.6.0-linux-x64.tar.gz ...
2023-03-17 18:43:07 INFO     Downloaded bundle to /tmp/tmpv4kasw4_/local-test-cluster/opensearch-dashboards-2.6.0-linux-x64.tar.gz
2023-03-17 18:43:07 INFO     Installing /tmp/tmpv4kasw4_/local-test-cluster/opensearch-dashboards-2.6.0-linux-x64.tar.gz in /tmp/tmpv4kasw4_/local-test-cluster/opensearch-dashboards-2.6.0
Removing securityDashboards...
Plugin removal complete
2023-03-17 18:43:30 INFO   
```

src/test_workflow/integ_test/integ_test_suite.py
```diff
@@ -76,45 +61,7 @@ def execute_tests(self) -> TestComponentResults:
         pass
 
     def multi_execute_integtest_sh(self, cluster_endpoints: list, security: bool, test_config: str) -> int:
```
gaiksaya, line 63: Make this method abstract. So that is its implemented whoever extends this suite.

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -81,6 +94,38 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
             os.chdir(self.work_dir)
             return self.execute_integtest_sh(endpoint, port, security, config)
 
+    def multi_execute_integtest_sh(self, cluster_endpoints: list, security: bool, test_config: str) -> int:
+        script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
```
gaiksaya, line 98: Add tests for this asserting that `self.repo` is always `functionalTestDashboards` for opensearch dashboards

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -48,6 +55,12 @@ def __init__(
             bundle_manifest_opensearch,
             build_manifest_opensearch
         )
+        self.repo = GitRepository(
+            build_manifest_opensearch_dashboards.components["functionalTestDashboards"].repository,
+            build_manifest_opensearch_dashboards.components["functionalTestDashboards"].commit_id,
```
gaiksaya, line 60: nit: Use single quotes

gaiksaya (changes requested):
Looks good, Some minor improvements.

src/test_workflow/integ_test/integ_test_suite.py
Divyaasm: Sure Sayali

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
Divyaasm: Sure Sayali

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -48,6 +55,12 @@ def __init__(
             bundle_manifest_opensearch,
             build_manifest_opensearch
         )
+        self.repo = GitRepository(
+            build_manifest_opensearch_dashboards.components["functionalTestDashboards"].repository,
```
peterzhuamazon, line 59: Nit: Adding some comments letting people know why we hardcode functionalTestDashboards.

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -81,6 +94,38 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
             os.chdir(self.work_dir)
             return self.execute_integtest_sh(endpoint, port, security, config)
 
+    def multi_execute_integtest_sh(self, cluster_endpoints: list, security: bool, test_config: str) -> int:
+        script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
+
+        def custom_node_endpoint_encoder(node_endpoint: NodeEndpoint) -> dict:
+            return {"endpoint": node_endpoint.endpoint, "port": node_endpoint.port, "transport": node_endpoint.transport}
+        if os.path.exists(script):
+            single_data_node = cluster_endpoints[0].data_nodes[0]
```
peterzhuamazon, line 103: Naming can be better since there is no so-called "data-node" in OSD setups.

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -81,6 +94,38 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
...
+                self.test_artifact_files
+            )
+            self.save_logs.save_test_result_data(test_result_data)
+            if stderr:
+                logging.info("Stderr reported for component: " + self.component.name)
+                logging.info(stderr)
+            return status
+        else:
+            logging.info(f"{script} does not exist. Skipping integ tests for {self.component.name}")
+            return 0
```
peterzhuamazon, line 123: Nit: I kinda feel like this entire part can be put into another wrapper to be called.

peterzhuamazon (changes requested):
Hi @Divyaasm ,

Some good improvements, just some nits.

Thanks.

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
Divyaasm: Do you think it's better to have a different function to implement this part @peterzhuamazon ?
If that's the case we can implement this in the base class both for OS and OSD.

gaiksaya: I think this is okay for now (since it's a nit). We can revisit this later.

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
Divyaasm: I was expecting this is form the cluster_endpoints properties https://github.com/opensearch-project/opensearch-build/blob/59ab4340d1e788d84c6058c04e4ea75f86a29fa8/src/test_workflow/integ_test/topology.py#L186

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
peterzhuamazon: > Do you think it's better to have a different function to implement this part @peterzhuamazon ? If that's the case we can implement this in the base class both for OS and OSD.

That is what I mean as a improvement later.

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
Divyaasm: Done Peter

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
Divyaasm: Done

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
peterzhuamazon: Thanks.

peterzhuamazon (approved):

peterzhuamazon:
Thanks @Divyaasm this is a great addition to OSD test so that it can run parallel like OS test.

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -81,6 +97,38 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
...
             return self.execute_integtest_sh(endpoint, port, security, config)
 
+    def multi_execute_integtest_sh(self, cluster_endpoints: list, security: bool, test_config: str) -> int:
+        script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
+
+        def custom_node_endpoint_encoder(node_endpoint: NodeEndpoint) -> dict:
+            return {"endpoint": node_endpoint.endpoint, "port": node_endpoint.port, "transport": node_endpoint.transport}
+        if os.path.exists(script):
+            single_node = cluster_endpoints[0].data_nodes[0]
+            cmd = f"bash {script} -b {single_node.endpoint} -p {single_node.port} -s {str(security).lower()} -t {self.component.name} -v {self.bundle_manifest.build.version}"
```
Divyaasm, line 107: Hi @tianleh , seems there is an issue in running integtests when -t is passed from python code base @line 107.

tianleh: Checked with @peterzhuamazon . Please raise the PR to FTRep once ready.

rishabh6788:
How about we introduce a new parameter, say `sub-component` that is only acknowledged when the calling class is `IntegTestSuiteOpenSearchDashboards`. 
This way we don't have to have multiple implementations of `custom_node_endpoint_encoder` method [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/integ_test_suite.py#L78). 

All we need to do would be add a condition to test which sub-class is calling this method and if it comes from `IntegTestSuiteOpenSearchDashboards` and if a subcomponent has been passed in the argument then add that to the command. We wouldn't have to change anything in the manifests as well. 
Since I was debugging the issue with OSD integ test running for all components even after passing the specific component, to test I tried this slightly different approach to achieve this. [Here](https://github.com/opensearch-project/opensearch-build/pull/3446) is a draft PR just for the demo purpose. 
I am okay with going ahead with the current implementation as well. 
Thoughts? @peterzhuamazon @Divyaasm @gaiksaya @dblock

gaiksaya:
>Thoughts? @peterzhuamazon @Divyaasm @gaiksaya @dblock

The idea is good but the sub-component is actually the component of dashboards product. That's why in order to avoid confusion we went with refactoring the current set up. It is also true that they are sub-components of functional-test repo which is quite confusing but that is just from test perspective. For build workflow, we have similar thing where we build specific components. Wanted to keep the same for test as well.

rishabh6788:
> 

The logic is such that if `sub-component` is passed when running OS integ-test it will be ignored, it will only be acknowledged when the test suite is executed from OSD class.

gaiksaya:
> > 
> 
> The logic is such that if `sub-component` is passed when running OS integ-test it will be ignored, it will only be acknowledged when the test suite is executed from OSD class.

Yeah got that but again will be confusing for component as well as other teams between components and sub-components for OS and OSD. I am happy to revisit this logic in future. The reason we turned`multi_execute_integtest_sh` into abstract method was because eventually the way integration tests are run for OS and OSD may vary in future as these two are different products. This is the start of diverging path.

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -81,6 +97,38 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
...
             return self.execute_integtest_sh(endpoint, port, security, config)
 
+    def multi_execute_integtest_sh(self, cluster_endpoints: list, security: bool, test_config: str) -> int:
+        script = ScriptFinder.find_integ_test_script(self.component.name, self.repo.working_directory)
+
+        def custom_node_endpoint_encoder(node_endpoint: NodeEndpoint) -> dict:
+            return {"endpoint": node_endpoint.endpoint, "port": node_endpoint.port, "transport": node_endpoint.transport}
+        if os.path.exists(script):
+            single_node = cluster_endpoints[0].data_nodes[0]
+            cmd = f"bash {script} -b {single_node.endpoint} -p {single_node.port} -s {str(security).lower()} -t {self.component.name} -v {self.bundle_manifest.build.version} -o default"
```
gaiksaya, line 107: if `self.component.name` is empty will the tests still run for all components by default?

rishabh6788 (approved):
LGTM!

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
Divyaasm: The list of components is fetched [here](https://github.com/Divyaasm/opensearch-build/blob/305a8abada0d470f6d0d695b37b44c3c830c0279/src/test_workflow/integ_test/integ_test_runner.py#L42) based on input argument --component if that's not passed it runs for all the components in the test manifest. self.component.name has the plugin name and has it's value from the manifest  and is not empty in any case.  

We alter the plugin names passed to self.component.name based on --component argument to the workflow but run them individually each time with -t component.name as per test manifest.

gaiksaya: Thanks! This makes it clear.

gaiksaya (approved):
I believe https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/644 needs to go in first before this PR can be merged?

Divyaasm:
> I believe [opensearch-project/opensearch-dashboards-functional-test#644](https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/644) needs to go in first before this PR can be merged?

Yes, and Peter seem to have PR related to integ-test OSD in line

peterzhuamazon (approved):

PR #3320, opened 2023-03-16
Add tini to allow termination signal handling

### LINKED ISSUE #3229: [BUG] OpenSearch container does not allow for graceful termination
**Describe the bug**

OpenSearch container does not do a graceful shutdown when a termination signal (SIGTERM/SIGINT) is sent.
This issue was discovered when running the official docker image within Kubernetes, where a rollout would cause the container to be terminated forcefully after `terminationGracePeriodSeconds ` instead of catching the `SIGTERM` signal sent and shutting down cleanly.

This could be fixed by using a proper init process (like `tini`) as PID 1 instead of a shell, which would register signal handlers for these termination signals and allow for graceful termination. An example for this change is in the gist below.

Resources with more details on this issue:
* https://hynek.me/articles/docker-signals/
* https://engineeringblog.yelp.com/2016/01/dumb-init-an-init-for-docker.html

**To Reproduce**
1. Run an OpenSearch container using the official docker image (tested on both 1.3.7 and 2.5.0).
2. Send a `SIGTERM` or `SIGINT` to the container and it will not terminate.

Gist here with more detailed reproduction/remediation steps: [OpenSearch container termination signal handling](https://gist.github.com/shikharbhardwaj/9ba1f31fe9f5083b9f64a8976414c37e)

**Expected behavior**
The container should listen for these termination signals and shutdown gracefully.

**Plugins**
N/A

**Host/Environment (please complete the following information):**
 - OS: Ubuntu 22.04
 - Docker: 20.10.16
 - OpenSearch versions: 1.3.7, 2.5.0

**Additional context**

shikharbhardwaj:
### Description

- Adds tini as the PID1 process in opensearch and opensearch-dashboards container to allow for termination signal handling.

### Issues Resolved
- Closes #3229


docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile
```diff
@@ -52,12 +52,17 @@ FROM amazonlinux:2
...
 ARG OPENSEARCH_DASHBOARDS_HOME=/usr/share/opensearch-dashboards
+ENV TINI_VERSION=v0.19.0
 
 # Update packages
 # Install the tools we need: tar and gzip to unpack the OpenSearch tarball, and shadow-utils to give us `groupadd` and `useradd`.
 # Install which to allow running of securityadmin.sh
 RUN yum update -y && yum install -y tar gzip shadow-utils which && yum clean all
 
+# Add tini to use as init (PID1) process.
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /bin/tini
```
peterzhuamazon, line 63: This wont work as this download is only the x64 version, will crash on the arm64 here.
```
tini: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=a3f773dca5734fef07e2ea80c0450b3c60bbaf74, stripped
```

docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile
```diff
@@ -93,5 +98,5 @@ LABEL org.label-schema.schema-version="1.0" \
   org.label-schema.build-date="$BUILD_DATE"
 
 # CMD to run
-ENTRYPOINT ["./opensearch-dashboards-docker-entrypoint.sh"]
+ENTRYPOINT ["tini", "--", "./opensearch-dashboards-docker-entrypoint.sh"]
```
peterzhuamazon, line 101: Is it possible to use ENTRYPOINT + CMD together?
```

ENTRYPOINT ["/tini", "--"]

# Run your program under Tini
CMD ["/your/program", "-and", "-its", "arguments"]
```

docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile
```diff
@@ -52,12 +52,17 @@ FROM amazonlinux:2
...
+ENV TINI_VERSION=v0.19.0
 
 # Update packages
 # Install the tools we need: tar and gzip to unpack the OpenSearch tarball, and shadow-utils to give us `groupadd` and `useradd`.
 # Install which to allow running of securityadmin.sh
 RUN yum update -y && yum install -y tar gzip shadow-utils which && yum clean all
 
+# Add tini to use as init (PID1) process.
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /bin/tini
+RUN chmod +x /bin/tini
```
peterzhuamazon, line 64: I kinda prefer to not make it globally available, and add full path in the entrypoint.
Tho I am not strongly against it.

peterzhuamazon: probably use 755 specifically instead of +x so it is precise?

peterzhuamazon (commented):
Hi @shikharbhardwaj ,

Thanks very much for this PR.
Would you mind send some test results/logs if possible?
Showcasing the before/after of the change?

Also, not sure if this can be considered as a major change.
Since this is the first time we are changing how the process is being started.
So probably a good timing to merge this fix is when 3.x launch?

cc: @bbarani 

Thanks.

peterzhuamazon (changes requested):
Add a block to this PR now just in case it get accidentally merged.
Thanks.

docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile
peterzhuamazon: Need code to handle download for x64 vs arm64.

shikharbhardwaj: Updated the download source according to the TARGETARCH supplied by docker

docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile
shikharbhardwaj: Not sure about this since this may change behavior existing invocations, but AFAICS non-`opensearch` command invocations will just exec the supplied command with a couple of env var `exports`

docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile
shikharbhardwaj: Sounds good, updated the chmod to 755. Open to path suggestions where we can keep this binary

shikharbhardwaj:
@peterzhuamazon sorry for the delay on turning this around, made some updates here along with the steps I did to test this. Let me know if there's a better set of steps to test this PR.

Not exactly sure about the major version requirement, since this is more of a fix than a new feature, but it is possible something downstream depends on this behavior.

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
```diff
@@ -58,7 +58,8 @@ function setupPerformanceAnalyzerPlugin {
             echo "Disabling execution of $OPENSEARCH_HOME/bin/$PERFORMANCE_ANALYZER_PLUGIN/performance-analyzer-agent-cli for OpenSearch Performance Analyzer Plugin"
         else
             echo "Enabling execution of OPENSEARCH_HOME/bin/$PERFORMANCE_ANALYZER_PLUGIN/performance-analyzer-agent-cli for OpenSearch Performance Analyzer Plugin"
-            $OPENSEARCH_HOME/bin/opensearch-performance-analyzer/performance-analyzer-agent-cli > $OPENSEARCH_HOME/logs/PerformanceAnalyzer.log 2>&1 &
+            # Launch performance plugin and disown to PID1 (tini).
+            $OPENSEARCH_HOME/bin/opensearch-performance-analyzer/performance-analyzer-agent-cli > $OPENSEARCH_HOME/logs/PerformanceAnalyzer.log 2>&1 & disown
```
peterzhuamazon, line 62: Good catch, tho I never try to use disown after a command tho.
I think it does the same thing as nohup in front.

peterzhuamazon: I will add this change into another PR so this PR of yours only focused on tini.
If we need to revert later on we dont have to revert the entire tini implementation.
Thanks.

peterzhuamazon: * https://github.com/opensearch-project/opensearch-build/pull/3479

docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile
peterzhuamazon: TARGETARCH seems only support multi-arch building?
Tried on the single arch build script and it could not figure out arch.
https://docs.docker.com/desktop/extensions-sdk/extensions/multi-arch/#adding-multi-arch-binaries

```

Step 20/39 : ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-${TARGETARCH} /bin/tini
ADD failed: failed to GET https://github.com/krallin/tini/releases/download/v0.19.0/tini- with status 404 Not Found: <!DOCTYPE html>
<html lang="en" data-color-mode="auto" data-light-theme="light" data-dark-theme="dark" data-a11y-animated-images="system">
```

Although we release for multi-arch we also want to support single-arch if people want to build in their local.
Probably add this in the build-image-single-arch.sh?
```
--build-arg TARGETARCH=${arch_uname}
```

docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile
peterzhuamazon: Seems like you add this now?

docker/release/dockerfiles/opensearch-dashboards.al2.dockerfile
peterzhuamazon: Thanks!

peterzhuamazon: As of path, just use `WORKDIR`? It is the home dir of os or osd on docker.
So instead of using `/bin/tini` just use `./tini`?

peterzhuamazon:
Thanks @shikharbhardwaj I have reviewed for another pass.
Please take a look at the comments.

Thanks!

peterzhuamazon:
You might also want to rebase after #3479 merged to consume all the CVE fixes, so that security scan can pass.
Thanks.

jordarlu:
Hello, @shikharbhardwaj , would you be able to rebase and resolve the conflct to have the WhiteSource Security Check to validate again? thanks !

peterzhuamazon:
Hi @shikharbhardwaj welcome back and would you mind check the follow up comments I added in previous callouts?
Thanks.

jordarlu:
Greetings, @shikharbhardwaj , since it has been a while, wondering if you got time to comment on the feedback from @peterzhuamazon yet?  thanks ..

bbarani:
Hi @shikharbhardwaj, can you please take a look at the comments and provide your feedback?

chawinphat:
@peterzhuamazon 

I was able to gracefully terminate using Tini as described above only with the path being '/bin/tini". Using the relative path "./tini" gives the error: 

`teddyt@Teddys-MacBook-Pro docker % docker run -it -p 9200:9200 -p 9600:9600 -e "discovery.type=single-node" chawinphat/opensearch:1.3.9
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "./tini": stat ./tini: no such file or directory: unknown.`

dblock:
@peterzhuamazon Let's close if this is not going to be merged?

[Catch All Triage w/ [1](https://github.com/dblock), [2](https://github.com/krisfreedain), [3](https://github.com/andrross)]

Issue #3205, opened 2023-02-16
Enhancement on validation_docker script


jordarlu:
### Is your feature request related to a problem? Please describe

We want to make a few of enhancemens on the docker validation script including the following : 

- [x] Adding a flag/argument to do Only the digest validaiton without moving to run cluster, nor the API test
- [x] Adding a flag/argument to use Only staging images ( in which, we will not check digest )

> By doing this, we can validate the OpenSearch API prior of publishing to the product account at DockerHub
- [x] Adding a max retry for testing if the OpenSearch cluster is ready for accepting HTTP request
- [x] Using Curl to replace 'requests' library in Python
> it was found that the requests lib did not response quick enough to catch the return status code
- [x] Return 0 instead of True to the caller when the script ends successfully 
> it was satisfy Jenkins CI process
### Describe the solution you'd like

We will update the script accordingly

jordarlu:
PR has been merged .. closing the issue

PR #3264, opened 2023-03-02
docker validation improvement

jordarlu:
Description
This PR is to address https://github.com/opensearch-project/opensearch-build/issues/3205

Issues Resolved
Adding a flag/argument to do Only the digest validaiton without moving to run cluster, nor the API test
Adding a flag/argument to use Only staging images ( in which, we will not check digest )
Adding a max retry for testing if the OpenSearch cluster is ready for accepting HTTP request
Using Curl to replace 'requests' library in Python
Return 0 instead of True to the caller when the script ends successfully

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3325, opened 2023-03-17
docker validation improvement

jordarlu:
Description
This PR is to address https://github.com/opensearch-project/opensearch-build/issues/3205

Issues Resolved

- Adding a flag/argument to do Only the digest validaiton without moving to run cluster, nor the API test
- Adding a flag/argument to use Only staging images ( in which, we will not check digest )
- Adding a max retry for testing if the OpenSearch cluster is ready for accepting HTTP request
- Return 0 instead of True to the caller when the script ends successfully

### DIFF
--- a/src/validation_workflow/api_request.py
+++ b/src/validation_workflow/api_request.py
@@ -6,9 +6,8 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+import subprocess
 from typing import Any
-
-import requests
 
 """
 This class is to run API test againt on local OpenSearch API URL with default port 9200.
@@ -20,15 +19,11 @@ class ApiTest:
 
     def __init__(self, request_url: str) -> None:
         self.request_url = request_url
-        self.apiHeaders_auth = {"Authorization": "Basic YWRtaW46YWRtaW4="}  # default user/pass "admin/admin" in Base64 format
-        self.apiHeaders_accept = {"Accept": "*/*"}
-        self.apiHeaders_content_type = {"Content-Type": "application/json"}
-        self.apiHeaders = {}
-        self.apiHeaders.update(self.apiHeaders_auth)
-        self.apiHeaders.update(self.apiHeaders_accept)
-        self.apiHeaders.update(self.apiHeaders_content_type)
 
     def api_get(self) -> Any:
-
-        response = requests.get(self.request_url, headers=self.apiHeaders, verify=False)
-        return response.status_code, response.text
+        self.command = ['curl', self.request_url, '-u', 'admin:admin', '--insecure', '-i']
+        result = subprocess.run(self.command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        response = result.stdout.decode().strip()
+        response_code = int(response.split()[1])
+        response_content = response.split('\r\n\r\n', 1)[1]
+        return response_code, response_content
--- a/src/validation_workflow/api_request.py
+++ b/src/validation_workflow/api_request.py
@@ -6,8 +6,9 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
-import subprocess
 from typing import Any
+
+import requests
 
 """
 This class is to run API test againt on local OpenSearch API URL with default port 9200.
@@ -19,11 +20,15 @@ class ApiTest:
 
     def __init__(self, request_url: str) -> None:
         self.request_url = request_url
+        self.apiHeaders_auth = {"Authorization": "Basic YWRtaW46YWRtaW4="}  # default user/pass "admin/admin" in Base64 format
+        self.apiHeaders_accept = {"Accept": "*/*"}
+        self.apiHeaders_content_type = {"Content-Type": "application/json"}
+        self.apiHeaders = {}
+        self.apiHeaders.update(self.apiHeaders_auth)
+        self.apiHeaders.update(self.apiHeaders_accept)
+        self.apiHeaders.update(self.apiHeaders_content_type)
 
     def api_get(self) -> Any:
-        self.command = ['curl', self.request_url, '-u', 'admin:admin', '--insecure', '-i']
-        result = subprocess.run(self.command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        response = result.stdout.decode().strip()
-        response_code = int(response.split()[1])
-        response_content = response.split('\r\n\r\n', 1)[1]
-        return response_code, response_content
+
+        response = requests.get(self.request_url, headers=self.apiHeaders, verify=False)
+        return response.status_code, response.text
--- a/src/validation_workflow/api_test_cases.py
+++ b/src/validation_workflow/api_test_cases.py
@@ -39,9 +39,10 @@ class ApiTestCases:
             success_status_code = test_case.__getitem__(1)
             validate_string = test_case.__getitem__(2)
 
-            status_code, response_text = ApiTest(str(request_url)).api_get()
-            logging.info(f"\nStatus_code ->{status_code} \nresponse_text ->{response_text}")
             time.sleep(3)
+            status_code, response_text = ApiTest(str(request_url)).api_get()
+            logging.info(f"Test Request -> {str(request_url)}")
+            logging.info(f"\nStatus_code ->{status_code} \nresponse_text ->{response_text}")
 
             if status_code == success_status_code and (not validate_string or validate_string in response_text):
                 pass_counter += 1
--- a/src/validation_workflow/api_test_cases.py
+++ b/src/validation_workflow/api_test_cases.py
@@ -6,7 +6,6 @@
 # compatible open source license.
 
 import logging
-import time
 from typing import Any
 
 from validation_workflow.api_request import ApiTest
@@ -39,9 +38,8 @@ class ApiTestCases:
             success_status_code = test_case.__getitem__(1)
             validate_string = test_case.__getitem__(2)
 
-            time.sleep(3)
             status_code, response_text = ApiTest(str(request_url)).api_get()
-            logging.info(f"Test Request -> {str(request_url)}")
+            logging.info(f"\nRequest_url ->{str(request_url)} \n")
             logging.info(f"\nStatus_code ->{status_code} \nresponse_text ->{response_text}")
 
             if status_code == success_status_code and (not validate_string or validate_string in response_text):
--- a/src/validation_workflow/docker/validation_docker.py
+++ b/src/validation_workflow/docker/validation_docker.py
@@ -62,7 +62,7 @@ class ValidateDocker(Validation):
         # STEP 2 . inspect image digest between opensearchproject(downloaded/local) and opensearchstaging(dockerHub)
         if not self.args.using_staging_artifact_only:
             self.image_names_list = [self.args.OS_image, self.args.OSD_image]
-            self.image_digests = list(map(lambda x: self.inspect_docker_image(x[0], x[1]), zip(self.image_ids, self.image_names_list)))
+            self.image_digests = list(map(lambda x: self.inspect_docker_image(x[0], x[1]), zip(self.image_ids, self.image_names_list)))  # type: ignore
 
             if all(self.image_digests):
                 logging.info('Image digest is validated.\n\n')
@@ -210,7 +210,7 @@ class ValidateDocker(Validation):
             s = s.replace(old_string, new_string)
             f.write(s)
 
-    def inspect_docker_image(self, image_id: str, image_name: str) -> str:
+    def inspect_docker_image(self, image_id: str, image_name: str) -> Any:
         return InspectDockerImage(image_id, image_name, self.args.version).inspect_digest()
 
     def is_container_daemon_running(self) -> Any:
--- a/src/validation_workflow/docker/validation_docker.py
+++ b/src/validation_workflow/docker/validation_docker.py
@@ -13,7 +13,10 @@ import time
 from subprocess import PIPE
 from typing import Any
 
+import requests
+
 from system.temporary_directory import TemporaryDirectory
+from validation_workflow.api_request import ApiTest
 from validation_workflow.api_test_cases import ApiTestCases
 from validation_workflow.docker.inspect_docker_image import InspectDockerImage
 from validation_workflow.validation import Validation
@@ -27,48 +30,17 @@ class ValidateDocker(Validation):
 
     def download_artifacts(self) -> bool:
         try:
-            assert self.is_container_daemon_running(), 'Docker Daemon is not running. Exiting the docker validation.'
+            assert self.is_container_daemon_running(), 'Docker daemon is not running. Exiting the docker validation.'
 
-            image_names = {
-                'dockerhub': {
-                    'opensearch': {
-                        'staging': 'opensearchstaging/opensearch',
-                        'production': 'opensearchproject/opensearch'
-                    },
-                    'opensearchdashboards': {
-                        'staging': 'opensearchstaging/opensearch-dashboards',
-                        'production': 'opensearchproject/opensearch-dashboards'
-                    }
-                },
-                'ecr': {
-                    'opensearch': {
-                        'staging': 'public.ecr.aws/opensearchstaging/opensearch',
-                        'production': 'public.ecr.aws/opensearchproject/opensearch'
-                    },
-                    'opensearchdashboards': {
-                        'staging': 'public.ecr.aws/opensearchstaging/opensearch-dashboards',
-                        'production': 'public.ecr.aws/opensearchproject/opensearch-dashboards'
-                    }
-                }
-            }
-
-            # Choose image names based on input arguments
-            self.using_staging_artifact_only = 'staging' if self.args.using_staging_artifact_only else 'production'
-            self._OS_image_name = image_names[self.args.docker_source]['opensearch'][self.using_staging_artifact_only]
-            self._OSD_image_name = image_names[self.args.docker_source]['opensearchdashboards'][self.using_staging_artifact_only]
-
-            self.local_image_OS_id = (
-                self.get_image_id(self._OS_image_name, ValidationArgs().stg_tag('opensearch').replace(" ", ""))
-                if self.args.using_staging_artifact_only
-                else self.get_image_id(self._OS_image_name, self.args.version)
-            )
-            self.local_image_OSD_id = (
-                self.get_image_id(self._OSD_image_name, ValidationArgs().stg_tag('opensearch_dashboards').replace(" ", ""))
-                if self.args.using_staging_artifact_only
-                else self.get_image_id(self._OSD_image_name, self.args.version)
-            )
-            logging.info(f'the OpenSearch image ID is : {self.local_image_OS_id}')
-            logging.info(f'the OpenSearch Dasboards image ID is : {self.local_image_OSD_id} \n\n')
+            # STEP 1 . pull the images for OS and OSD
+            product_names = ["opensearch", "opensearch_dashboards"]
+            using_staging_artifact_only = 'staging' if self.args.using_staging_artifact_only else 'production'
+            get_image_id = lambda product: self.get_image_id(  # noqa: E731
+                self.get_artifact_image_name(product, using_staging_artifact_only),
+                self.args.version if not self.args.using_staging_artifact_only else ValidationArgs().stg_tag(product).replace(" ", ""))
+            self.image_ids = list(map(get_image_id, product_names))
+            logging.info(f'the opensearch image ID is : {self.image_ids[0]}')
+            logging.info(f'the opensearch-dashboards image ID is : {self.image_ids[1]} \n\n')
             return True
 
         except AssertionError as e:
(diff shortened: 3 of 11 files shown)


jordarlu:
Hi, the update on api_test will not impact others who use this method because the input and output format staying the same as before.
In src/run_validation.py, the test_result will be either True or False which taken from the actual job runner, so, I transform it to 0 for True, and 1 for False.
This PR also has two additional arguments added to commit with feedback from dry-run test a while ago, pls see the output for those two additional optional argument below.
The inplace_change makes the changes on copied docker-compose.yml at temp directory. The source file is not touched.

jordarlu:
My apologies on making the review diffcult by replacing a new PR with the previous one ..if it helps, here is the link to the previosu PR https://github.com/opensearch-project/opensearch-build/pull/3264.  Thank you for speding extra effort on reviewing it ...

src/validation_workflow/docker/validation_docker.py
```diff
@@ -29,23 +26,56 @@ def __init__(self, args: ValidationArgs) -> None:
...
+                    }
+                },
+                'ecr': {
+                    'os': {
+                        True: 'public.ecr.aws/opensearchstaging/opensearch',
+                        False: 'public.ecr.aws/opensearchproject/opensearch'
+                    },
+                    'osd': {
+                        True: 'public.ecr.aws/opensearchstaging/opensearch-dashboards',
+                        False: 'public.ecr.aws/opensearchproject/opensearch-dashboards'
```
gaiksaya, line 50: I did not get the True/False mapping. Can you explain? AFAI understand, its whether its staging or prod. Would ask to name the key of the mapping accordingly. True/False is confusing or add comments.
Adding @prudhvigodithi / @peterzhuamazon if they can suggest something better.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -29,23 +26,56 @@ def __init__(self, args: ValidationArgs) -> None:
...
             return True
 
-        else:
-            raise Exception('Docker Daemon is not running. Exiting the docker validation.')
+        except AssertionError as e:
+            logging.error(str(e))
+            return False
+
+        finally:
+            pass
```
gaiksaya, line 78: Log something here instead of `pass` like what process was completed.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -56,15 +86,21 @@ def start_cluster(self) -> bool:
...
 
     def validation(self) -> bool:
-        # STEP 2 . inspect image digest betwwen opensearchproject(downloaed/local) and opensearchstaging(dockerHub)
-
-        self._OS_inspect_digest = InspectDockerImage(self.local_image_OS_id, self.args.OS_image, self.args.version)
-        self._OSD_inspect_digest = InspectDockerImage(self.local_image_OSD_id, self.args.OSD_image, self.args.version)
-
-        if self._OS_inspect_digest.inspect_digest() and self._OSD_inspect_digest.inspect_digest():
-            logging.info('Image digest is validated\n\n')
+        # STEP 2 . inspect image digest between opensearchproject(downloaed/local) and opensearchstaging(dockerHub)
```
gaiksaya, line 89: nit: typo in downloaed. Should be downloaded/local

src/validation_workflow/docker/validation_docker.py
```diff
@@ -73,12 +109,19 @@ def validation(self) -> bool:
...
+                    time.sleep(10)
                     if self.check_http_request():
                         logging.info('\n\ncluster is now ready for API test\n\n')
                         break
-                    logging.info('sleeping 10sec')
-                    time.sleep(10)
+                    self.retry_count += 1
+                else:
+                    logging.warning(f"Maximum number of retries ({self.max_retry}) reached. Cluster is not ready for API test.")
+                    raise Exception(f"Maximum number of retries ({self.max_retry}) reached. Cluster is not ready for API test.")
```
gaiksaya, line 124: Do we need to warn and raise exception both?

src/validation_workflow/docker/validation_docker.py
```diff
@@ -87,43 +130,50 @@ def validation(self) -> bool:
...
     def check_http_request(self) -> bool:
         try:
-            status_code, response_text = ApiTest('https://localhost:9200/').api_get()
-            if status_code == 200:
-                return True
-            else:
-                return False
-        except (requests.exceptions.ConnectionError, requests.exceptions.ConnectTimeout) as e:
-            logging.error(f'Error connecting to https://localhost:9200/: {e}')
+            subprocess.check_output(['curl', 'https://localhost:9200', '-u', 'admin:admin', '--insecure', '-s'])
```
gaiksaya, line 143: Wondering if we want the `curl` to be silent "-s". Let it be loud. Even if it fails exact error can be viewed then.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -87,43 +130,50 @@ def validation(self) -> bool:
...
-            logging.info('cleanup is Not completed')
+        try:
+            # clean up
+            if self.args.validate_digest_only:
+                return True
+            if self.cleanup_process():
+                logging.info('cleanup is completed')
+                return True
+            else:
+                logging.info('cleanup is Not completed')
```
gaiksaya, line 159:
```suggestion
                logging.info('cleanup is not complete')
```

src/validation_workflow/docker/validation_docker.py
```diff
@@ -87,43 +130,50 @@ def validation(self) -> bool:
...
-            logging.info('cleanup is completed')
-            return True
-        else:
-            logging.info('cleanup is Not completed')
+        try:
+            # clean up
+            if self.args.validate_digest_only:
+                return True
+            if self.cleanup_process():
+                logging.info('cleanup is completed')
```
gaiksaya, line 156:
```suggestion
                logging.info('cleanup is complete')
```

src/validation_workflow/docker/validation_docker.py
```diff
@@ -87,43 +130,50 @@ def validation(self) -> bool:
...
+                return True
+            else:
+                logging.info('cleanup is Not completed')
+                return False
+        except Exception as e:
+            logging.error(f'An error occurred during cleanup: {e}')
             return False
+        finally:
+            pass
```
gaiksaya, line 165:
```suggestion
            logging.info('Docker validation cleanup is complete')
```

src/validation_workflow/docker/validation_docker.py
```diff
@@ -87,43 +130,50 @@ def validation(self) -> bool:
...
-        result = subprocess.run(docker_command, shell=True, stdout=PIPE, stderr=PIPE, universal_newlines=True)
+        self.docker_compose_down = f'docker-compose -f {self._target_yml_file} down'
+        result = subprocess.run(self.docker_compose_down, shell=True, stdout=PIPE, stderr=PIPE, universal_newlines=True)
 
         # remove docker-compose.yml from the tmp folder
         try:
             os.remove(self._target_yml_file)
         except OSError as e:
-            print("Error: %s - %s." % (e.filename, e.strerror))
+            logging.info("Error: %s - %s." % (e.filename, e.strerror))
```
gaiksaya, line 176:
```suggestion
            logging.error("Error: %s - %s." % (e.filename, e.strerror))
```
Also raise an exception that just logging.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -147,17 +196,41 @@ def run_container(self, OpenSearch_image: str, OpenSearchDashboard_image: str, v
...
+                if self.args.using_staging_artifact_only
+                else f'{OpenSearchDashboard_image}:{version}'
+            )
+            # spin up containers
+            self.docker_compose_up = f'docker-compose -f {self.target_yml_file} up -d'
+            result = subprocess.run(self.docker_compose_up, shell=True, stdout=PIPE, stderr=PIPE, universal_newlines=True)
+            return ('returncode=0' in (str(result)), self.target_yml_file)
+
+        except Exception as e:
+            print(f"An error occurred: {e}")
```
gaiksaya, line 231: Raise an exception than just printing it.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -147,17 +196,41 @@ def run_container(self, OpenSearch_image: str, OpenSearchDashboard_image: str, v
...
+        finally:
+            pass
```
gaiksaya, line 233: Log what process completed rather than pass

gaiksaya (changes requested):
Since the validation depends heavily on docker-compose, please add tests ensuring that those URLs are reachable. If someone raises a PR to delete those files, docker validation test cases should fail.

src/validation_workflow/docker/validation_docker.py
jordarlu: thanks, @gaiksaya.. because we only have Staging and Production, and I want to determine which one to apply by one of the input argument, 'self.args.using_staging_artifact_only', which will be either True or False; hence, I wanted to map Staging = True, and Production = False. 
Therefore, at L56~57, I can apply to the correct URL based on the input argument 'self.args.using_staging_artifact_only'.

src/validation_workflow/docker/validation_docker.py
jordarlu: got it ..it makes sense ..let me update it in next commit

src/validation_workflow/docker/validation_docker.py
jordarlu: oops ./ thanks for the correction .. let me get that right in my next commit

src/validation_workflow/docker/validation_docker.py
jordarlu: sure, @gaiksaya .. I wanted to slience it because there was too much noises, informations and caused the log/output difficult to read, but I can certainly remove the "-s" ...

src/validation_workflow/docker/validation_docker.py
jordarlu: :) got it!!

src/validation_workflow/docker/validation_docker.py
jordarlu: got it !! makes sense.

src/validation_workflow/docker/validation_docker.py
jordarlu: mm... I wanted to also show a simple string on the output as well ....but yes,, I can certainly remove it ... let me do that .!! thanks!!

jordarlu:
> Since the validation depends heavily on docker-compose, please add tests ensuring that those URLs are reachable. If someone raises a PR to delete those files, docker validation test cases should fail.

HI, @gaiksaya ..just want to make sure I got it right, you meant to have a test as validation on file existing of those two docker-compose-1/2.x.yml in build repo.... correct? thanks

gaiksaya:
> > Since the validation depends heavily on docker-compose, please add tests ensuring that those URLs are reachable. If someone raises a PR to delete those files, docker validation test cases should fail.
> 
> HI, @gaiksaya ..just want to make sure I got it right, you meant to have a test as validation on file existing of those two docker-compose-1/2.x.yml in build repo.... correct? thanks

Yes

src/validation_workflow/docker/validation_docker.py
jordarlu: Myabe we can change to something like this?

jordarlu:
Hi, @gaiksaya , if you can have another round of review on updated portion, thanks !!

src/validation_workflow/api_request.py
```diff
@@ -20,15 +19,11 @@ class ApiTest:
...
-        self.apiHeaders = {}
-        self.apiHeaders.update(self.apiHeaders_auth)
-        self.apiHeaders.update(self.apiHeaders_accept)
-        self.apiHeaders.update(self.apiHeaders_content_type)
 
     def api_get(self) -> Any:
-
-        response = requests.get(self.request_url, headers=self.apiHeaders, verify=False)
-        return response.status_code, response.text
+        self.command = ['curl', self.request_url, '-u', 'admin:admin', '--insecure', '-i']
```
dblock, line 24: Why are we replacing a native HTTP implementation with `curl`?

src/validation_workflow/docker/validation_docker.py
```diff
@@ -147,17 +195,41 @@ def run_container(self, OpenSearch_image: str, OpenSearchDashboard_image: str, v
...
-            shutil.copy2(self.source_file, self.target_yml_file)
+        try:
+            source_file = None
+            if self.version_number in self.docker_compose_files:
+                source_file = os.path.join('docker', 'release', 'dockercomposefiles', self.docker_compose_files[self.version_number])
+                shutil.copy2(source_file, self.target_yml_file)
 
-        self.inplace_change(self.target_yml_file, f'opensearchproject/opensearch:{version[0]}', f'{OpenSearch_image}:{version}')
-        self.inplace_change(self.target_yml_file, f'opensearchproject/opensearch-dashboards:{version[0]}', f'{OpenSearchDashboard_image}:{version}')
+            if (self.args.using_staging_artifact_only):
```
dblock, line 204: Use inheritance. This should be refactored into different validation implementations, one for staging and another for otherwise.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -147,17 +195,41 @@ def run_container(self, OpenSearch_image: str, OpenSearchDashboard_image: str, v
...
+                if self.args.using_staging_artifact_only
+                else f'{OpenSearchDashboard_image}:{version}'
+            )
+            # spin up containers
+            self.docker_compose_up = f'docker-compose -f {self.target_yml_file} up -d'
+            result = subprocess.run(self.docker_compose_up, shell=True, stdout=PIPE, stderr=PIPE, universal_newlines=True)
+            return ('returncode=0' in (str(result)), self.target_yml_file)
+
+        except Exception as e:
+            raise Exception(f"An error occurred: {e}")
```
dblock, line 230: There's no need to catch and re-raise, just let it raise an exception.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -29,23 +26,57 @@ def __init__(self, args: ValidationArgs) -> None:
...
-
-            self.local_image_OS_id = self.get_image_id(self._OS_image_name, self.args.version)
-            self.local_image_OSD_id = self.get_image_id(self._OSD_image_name, self.args.version)
-            logging.info(f'the OS image ID is : {self.local_image_OS_id}')
-            logging.info(f'the OSD image ID is : {self.local_image_OSD_id} \n\n')
-
+        try:
+            assert self.is_container_daemon_running(), 'Docker Daemon is not running. Exiting the docker validation.'
+
+            image_names = {
```
dblock, line 32: move to a constant

src/validation_workflow/api_request.py
jordarlu: Thanks for the comment, @dblock .. a few weeks back, when I was doing testing with @peterzhuamazon , we found out that the Curl is more stable on privding the result than using the request lib for some reasons.

src/validation_workflow/docker/validation_docker.py
jordarlu: thanks, sure, let me update it in my next commit..

src/validation_workflow/docker/validation_docker.py
```diff
@@ -29,23 +26,57 @@ def __init__(self, args: ValidationArgs) -> None:
         super().__init__(args)
 
     def download_artifacts(self) -> bool:
-
-        # Check if the docker daemon is running; or else, bring it up
-        if (self.is_container_daemon_running()):
-
-            # STEP 1 . pull the images for OS and OSD
-            self._OS_image_name = 'opensearchproject/opensearch' if self.args.docker_source == 'dockerhub' else 'public.ecr.aws/opensearchproject/opensearch'
-            self._OSD_image_name = 'opensearchproject/opensearch-dashboards' if self.args.docker_source == 'dockerhub' else 'public.ecr.aws/opensearchproject/opensearch-dashboards'
```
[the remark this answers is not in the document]
jordarlu: Thanks for the comment, @dblock , the new code hopefully explains more on selecting the image source between dockerhub and ecr, and plus supporting staging and production...
the code will choose either dockerhub or ecr ; and either using staging-only or comparing between staging and production before briing up a container, and it will be based on the arguemnts taken from the user input

sorry that I am lost when you comment on 'pass the image name dynamically' ... I guess you don't mean to have user to input the full image name as an argument if I am not mistaken .. thanks,

dblock (commented):
I made some directional suggestions below. Break up the code into functions and use inheritance so that we don't have too many different things going on in the same function x2 with opensearch and opensearch dashboards

src/validation_workflow/api_request.py
jordarlu: Thanks, dB, I will keep using the requests libarary. so, no change on this script.

src/validation_workflow/api_test_cases.py
```diff
@@ -39,9 +39,10 @@ def test_cases(self) -> Any:
             success_status_code = test_case.__getitem__(1)
             validate_string = test_case.__getitem__(2)
 
+            time.sleep(3)
```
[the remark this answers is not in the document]
jordarlu: It was originally putting a delay between each test, not required anymore . .let me remove it .. thanks for the comment

src/validation_workflow/docker/validation_docker.py
```diff
@@ -56,29 +87,40 @@ def start_cluster(self) -> bool:
...
-
-        self._OS_inspect_digest = InspectDockerImage(self.local_image_OS_id, self.args.OS_image, self.args.version)
-        self._OSD_inspect_digest = InspectDockerImage(self.local_image_OSD_id, self.args.OSD_image, self.args.version)
-
-        if self._OS_inspect_digest.inspect_digest() and self._OSD_inspect_digest.inspect_digest():
-            logging.info('Image digest is validated\n\n')
+        # STEP 2 . inspect image digest between opensearchproject(downloaded/local) and opensearchstaging(dockerHub)
+        if not self.args.using_staging_artifact_only:
+            self._OS_inspect_digest = InspectDockerImage(self.local_image_OS_id, self.args.OS_image, self.args.version).inspect_digest()
+            self._OSD_inspect_digest = InspectDockerImage(self.local_image_OSD_id, self.args.OSD_image, self.args.version).inspect_digest()
```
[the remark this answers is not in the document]
jordarlu: Thanks for the meeting, and comments, @dblock 
I made some updates.. would you mind have a look again? Thanks!!

src/validation_workflow/docker/validation_docker.py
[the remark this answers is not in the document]
jordarlu: Thanks, @dblock ..I made an update just now .. hopefully it will work betteer this time..

src/validation_workflow/docker/validation_docker.py
```diff
@@ -56,121 +59,131 @@ def start_cluster(self) -> bool:
...
+                    'staging': 'public.ecr.aws/opensearchstaging/opensearch',
+                    'production': 'public.ecr.aws/opensearchproject/opensearch'
+                },
+                'opensearch_dashboards': {
+                    'staging': 'public.ecr.aws/opensearchstaging/opensearch-dashboards',
+                    'production': 'public.ecr.aws/opensearchproject/opensearch-dashboards'
+                }
+            }
+        }
+        return self.image_names[self.args.docker_source][artifact][using_staging_artifact_only]
```
dblock, line 186: You should make this OO. Create a class called `DockerImage` that takes source, artifact and staging as input and generates the name as a method and returns `public.ecr.aws/opensearchstaging/opensearch-dashboards` and such as a property. Each one of the parts of the result has meaning. For example a method called `host` would return `null` if the registry is dockerhub, public.ecr.aws if it's ecr, etc.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -29,23 +29,26 @@ def __init__(self, args: ValidationArgs) -> None:
...
-            logging.info(f'the OSD image ID is : {self.local_image_OSD_id} \n\n')
-
+            product_names = ["opensearch", "opensearch_dashboards"]
+            using_staging_artifact_only = 'staging' if self.args.using_staging_artifact_only else 'production'
+            get_image_id = lambda product: self.get_image_id(  # noqa: E731
+                self.get_artifact_image_name(product, using_staging_artifact_only),
+                self.args.version if not self.args.using_staging_artifact_only else ValidationArgs().stg_tag(product).replace(" ", ""))
+            self.image_ids = list(map(get_image_id, product_names))
+            logging.info(f'the opensearch image ID is : {self.image_ids[0]}')
+            logging.info(f'the opensearch-dashboards image ID is : {self.image_ids[1]} \n\n')
```
dblock, line 43: This code should be

```python
self.images = DockerImages(["OpenSearch, "OpenSearch Dashboards"])
self.images.download()
```

You're missing a collection class `DockerImages` and an individual class `DockerImage`.

dblock (approved):
This is still very _not_ object oriented with a lot of `if`'s and exception catching. 
I think you should fix the parameter `_` into `-` and let's merge this, then see if you can take some time to refactor it further after that?

jordarlu:
> This is still very _not_ object oriented with a lot of `if`'s and exception catching. I think you should fix the parameter `_` into `-` and let's merge this, then see if you can take some time to refactor it further after that?

Thanks for the approval, @dblock .. yes .. I will definitely optimit it later on with using aoother PR .. and thanks for the chatting as well

jordarlu:
> Since the validation depends heavily on docker-compose, please add tests ensuring that those URLs are reachable. If someone raises a PR to delete those files, docker validation test cases should fail.

Thank you, @gaiksaya , if you can give another review..

gaiksaya (approved):
Hope to see a follow up PR with changes suggested by dB!

src/validation_workflow/validation_args.py
```diff
@@ -84,6 +84,17 @@ def __init__(self) -> None:
             choices=["x64", "arm64"],
             default="x64"
         )
+        group = parser.add_mutually_exclusive_group()
+        group.add_argument(
+            "--validate_digest_only",
```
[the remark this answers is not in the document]
gaiksaya: @jordarlu Do you want to think small thing in before merge?

jordarlu:
> Hope to see a follow up PR with changes suggested by dB!

Certainly , let me create an issue for following up on the comments from dB  :)    thanks again!!

src/validation_workflow/validation_args.py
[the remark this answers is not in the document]
jordarlu: sorry,, I missed it ... thanks for notifying me this .. and yes... let me update it really quick before merging ...and will create another issue to follow up on comments for further improvements .. thanks !!

jordarlu:
https://github.com/opensearch-project/opensearch-build/issues/3475 is opened to track the enhancement

PR #3356, opened 2023-03-30
[OSD] include Node 14.21.3

kavilla:
### Description
Including Node 14.21.3 to the env image since this is the version we call out in the nvm file.

We do a relaxed version check on main branch.

### Issues Resolved
Potentially related to:
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3282

I see the failures on fiber for arm deb which could be the case if the system but will need to verify it's not the diff in Node first.

### DIFF
--- a/manifests/1.4.0/opensearch-dashboards-1.4.0.yml
+++ b/manifests/1.4.0/opensearch-dashboards-1.4.0.yml
@@ -5,7 +5,7 @@ build:
   version: 1.4.0
 ci:
   image:
-    name: opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028
+    name: opensearchstaging/ci-runner:ci-runner-centos7-opensearch-dashboards-build-v2
 components:
   - name: OpenSearch-Dashboards
     repository: https://github.com/opensearch-project/OpenSearch-Dashboards.git
--- a/docker/ci/dockerfiles/current/build.centos7.opensearch-dashboards.x64.arm64.dockerfile
+++ b/docker/ci/dockerfiles/current/build.centos7.opensearch-dashboards.x64.arm64.dockerfile
@@ -95,7 +95,7 @@ ENV PATH=/usr/share/opensearch/.gem/gems/fpm-1.14.2/bin:$PATH
 # nvm environment variables
 ENV NVM_DIR /usr/share/opensearch/.nvm
 ENV NODE_VERSION 10.24.1
-ARG NODE_VERSION_LIST="10.24.1 14.19.1 14.20.0 14.20.1"
+ARG NODE_VERSION_LIST="10.24.1 14.19.1 14.20.0 14.20.1 14.21.3"
 COPY --chown=1000:1000 config/build-opensearch-dashboards-entrypoint.sh /usr/share/opensearch
 # install nvm
 # https://github.com/creationix/nvm#install-script
--- a/docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
+++ b/docker/ci/dockerfiles/current/release.centos.clients.x64.arm64.dockerfile
@@ -135,7 +135,7 @@ ENV PATH=$RUBY_HOME:$RVM_HOME:$PATH
 # nvm environment variables
 ENV NVM_DIR /usr/share/opensearch/.nvm
 ENV NODE_VERSION 14.19.1
-ARG NODE_VERSION_LIST="10.24.1 14.19.1 14.20.0 14.20.1"
+ARG NODE_VERSION_LIST="10.24.1 14.19.1 14.20.0 14.20.1 14.21.3"
 
 # Installing nvm
 # https://github.com/creationix/nvm#install-script


manifests/1.4.0/opensearch-dashboards-1.4.0.yml
```diff
@@ -5,7 +5,7 @@ build:
   version: 1.4.0
 ci:
   image:
-    name: opensearchstaging/ci-runner:centos7-x64-arm64-jdkmulti-node10.24.1-cypress6.9.1-20211028
+    name: opensearchstaging/ci-runner:ci-runner-centos7-opensearch-dashboards-build-v2
```
kavilla, line 8: should resolve: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2685

peterzhuamazon (changes requested):
Hi @kavilla 

1. Please update all the existing dockerfile in `current` dir to this new version, if they install node as part of the building process.
2. Need to update the code here to reflect Windows.
https://github.com/opensearch-project/opensearch-ci/blob/main/packer/scripts/windows/scoop-install-commons.ps1#L130
3. All images needs to be rebuild before consuming.

And could you please confirm what is this change for? Which version?
This needs to be coordinate with our team.

Thanks.

manifests/1.4.0/opensearch-dashboards-1.4.0.yml
peterzhuamazon: Hi @kavilla isnt 1.x will lock to node 10.24.1?

kavilla: This image hasn't been updated for 1.4 but all the other builds on 1.x use this image. For example, https://github.com/opensearch-project/opensearch-build/blob/main/manifests/1.3.9/opensearch-dashboards-1.3.9.yml#LL8C59-L8C87

kavilla:
Hey @peterzhuamazon,

>     1. Please update all the existing dockerfile in `current` dir to this new version, if they install node as part of the building process.

I think I've updated the hardcoded refs of the node version to just have it in the path. Just as long as it's related to the build.

>     2. Need to update the code here to reflect Windows.
>        https://github.com/opensearch-project/opensearch-ci/blob/main/packer/scripts/windows/scoop-install-commons.ps1#L130

Sorry about that. Here it is: https://github.com/opensearch-project/opensearch-ci/pull/264

>     3. All images needs to be rebuild before consuming.

Yeah, not a pressing priority, I wouldn't drop any other priorities. I would track this only for 3.0.0 and to help get the builds successful for OSD 3.0.0. Since we don't have a current successful release of 3.0.0 for a little I wanted to get the OSD project building on the .nvm version in our repo https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/.nvmrc. Which is causing some issues with our CI in the FTRepo.

The current build failure with DEB ARM seems to be related to the fibers not being available for the node version and platform:
```
Please make sure you are using a supported platform and node version. If you
would like to compile fibers on this machine please make sure you have setup your
build environment--
Windows + OS X instructions here: https://github.com/nodejs/node-gyp
Ubuntu users please run: `sudo apt-get install g++ build-essential`
RHEL users please run: `yum install gcc-c++` and `yum groupinstall 'Development Tools'` 
Alpine users please run: `sudo apk add python make g++`
/bin/sh: nodejs: command not found
```

Just wanted to get this one to get the repos aligned just in case it does help and should probably be done anyways. Right now the node version check is relaxed so we can build on other machines successfully even if the node version available is not the matching version.

peterzhuamazon:
Thanks @kavilla for the responses I mistakenly identified another image having nodejs reference.
If this is ongoing for the 3.x I will just approve these and we can sync up next week on the builds.

Thanks.

peterzhuamazon:
Also please fix the check error, thanks.

manifests/2.4.0/opensearch-2.4.0.yml
```diff
@@ -112,8 +112,8 @@ components:
     checks:
       - gradle:properties:version
       - gradle:dependencies:opensearch.version
-  - name: dashboards-reports
-    repository: https://github.com/opensearch-project/dashboards-reports.git
+  - name: opensearch-reports
+    repository: https://github.com/opensearch-project/reporting.git
```
peterzhuamazon, line 116: Hi,
There is a repo split on the 2.5.0 version.
2.4.0 is untouched.
The tag exists for the dashboards-reports.
This should not be changed here.
Thanks.

peterzhuamazon (changes requested):

peterzhuamazon:
Async is having issues on 3.x still.
```

* What went wrong:
A problem occurred evaluating root project 'asynchronous-search'.
> Can't get https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.7.0/latest/linux/x64/tar/builds/opensearch/plugins/opensearch-asynchronous-search-2.7.0.0.zip to /tmp/tmpp1cctb6b/asynchronous-search/src/test/resources/org/opensearch/search/asynchronous/bwc/2.7.0.0/opensearch-asynchronous-search-2.7.0.0.zip

```

Not related, merged.

peterzhuamazon (approved):

PR #3357, opened 2023-03-30
Disable tests with security for knn in 2.7.0

martin-gaievski:
### Description
We need to disable tests with security for knn in 2.7.0 release.
Plugin tests need special adjustments for checks that security plugin start enforcing. Team has been working on solution from 2.6 (https://github.com/opensearch-project/security/issues/2265), but this requires more effort/time. This is to unblock pipeline for other plugins/components. 

### DIFF
--- a/manifests/2.7.0/opensearch-2.7.0-test.yml
+++ b/manifests/2.7.0/opensearch-2.7.0-test.yml
@@ -71,7 +71,6 @@ components:
   - name: k-NN
     integ-test:
       test-configs:
-        - with-security
         - without-security
 
   - name: neural-search


gaiksaya (changes requested):
Hi @martin-gaievski ,

This tests are not blocker for any other components. each component runs independently and on separate server. Its fine if they fail. We want to keep track of what is failing and what is fixed.

bbarani:
> Hi @martin-gaievski ,
> 
> These tests are not blocker for any other components. each component runs independently and on separate server. Its fine if they fail. We want to keep track of what is failing and what is fixed. Adding @bbarani!

Yes, build failures are blocker but integration test failures are specific to your component and doesn't affect other components.

gaiksaya:
Closing the PR as we want to track the tests for k-NN at distribution level

PR #3404, opened 2023-04-14
Fix missing logs of local test cluster of test workflow

### LINKED ISSUE #3382: [Bug]: Fix missing local test cluster logs for integration tests
### Describe the bug

Currently, the local test cluster only records stdout and stderr for local cluster that is deployed by the test workflow. 
Missing logs files:
```
- gc.log.00
- opensearchcluster1.log
- opensearchcluster1_deprecation.json
- opensearchcluster1_deprecation.log
- opensearchcluster1_index_indexing_slowlog.json
- opensearchcluster1_index_indexing_slowlog.log
- opensearchcluster1_index_search_slowlog.json
- opensearchcluster1_index_search_slowlog.log
- opensearchcluster1_server.json
- opensearchcluster1_task_detailslog.json
- opensearchcluster1_task_detailslog.log
```

### To reproduce

Run below command locally and see that test-results/../local-test-cluster folder only includes stdout and stderr files
```
./test.sh integ-test manifests/2.7.0/opensearch-2.7.0-test.yml --component geospatial -p opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.7.0/7325/linux/x64/tar --keep
``` 

### Expected behavior

Should record all log files 

### Host / Environment

_No response_

### Additional context

Estimate: 3 points

### Acceptance Criteria

- [ ]  Local cluster logs should be recorded in test-results folder 
- [ ] The temp directory (where local opensearch and dashboards is downloaded) should be cleaned up after each test
- [ ] Add tests cases verifying the behavior

zelinh:
### Description
Currently, we force to delete everything from the local cluster when uninstalling. This process would also remove the local cluster logs so that the test_recorder is not able to record it after tests. 
This PR is to restore missing logs of local test cluster but remove everything else when uninstalling.

### Issues Resolved
#3382 

### DIFF
--- a/src/test_workflow/integ_test/distribution_tar.py
+++ b/src/test_workflow/integ_test/distribution_tar.py
@@ -39,5 +39,9 @@ class DistributionTar(Distribution):
         return start_cmd_map[self.filename]
 
     def uninstall(self) -> None:
-        logging.info(f"Cleanup {self.work_dir}/* content after the test")
-        subprocess.check_call(f"rm -rf {self.work_dir}/*", shell=True)
+        logging.info(f"Cleanup {self.work_dir}/* content after the test and keep {self.install_dir}/logs for recording.")
+        subprocess.check_call(f"mv {self.install_dir}/logs {self.work_dir}", shell=True)
+        subprocess.check_call(f"rm -rf {self.install_dir}", shell=True)
+        subprocess.check_call(f"mkdir {self.install_dir}", shell=True)
+        subprocess.check_call(f"mv {self.work_dir}/logs {self.install_dir}", shell=True)
+        logging.info(f"{self.work_dir}/* would be deleted or kept subject to TemporaryDirectory rules.")
--- a/src/test_workflow/integ_test/distribution_zip.py
+++ b/src/test_workflow/integ_test/distribution_zip.py
@@ -39,5 +39,9 @@ class DistributionZip(Distribution):
         return start_cmd_map[self.filename]
 
     def uninstall(self) -> None:
-        logging.info(f"Cleanup {self.work_dir}/* content after the test")
-        subprocess.check_call(f"rm -rf {self.work_dir}/*", shell=True)
+        logging.info(f"Cleanup {self.work_dir}/* content after the test and keep {self.install_dir}/logs for recording.")
+        subprocess.check_call(f"mv {self.install_dir}/logs {self.work_dir}", shell=True)
+        subprocess.check_call(f"rm -rf {self.install_dir}", shell=True)
+        subprocess.check_call(f"mkdir {self.install_dir}", shell=True)
+        subprocess.check_call(f"mv {self.work_dir}/logs {self.install_dir}", shell=True)
+        logging.info(f"{self.work_dir}/* would be deleted or kept subject to TemporaryDirectory rules.")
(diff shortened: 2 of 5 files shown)


zelinh:
I'm able to see `opensearch-service-logs` inside `test-results` includes 
```
[opensearch@5f1ec38a6deb opensearch-service-logs]$ ls
gc.log                               opensearchcluster1_index_indexing_slowlog.json  opensearchcluster1.log
gc.log.00                            opensearchcluster1_index_indexing_slowlog.log   opensearchcluster1_server.json
opensearchcluster1_deprecation.json  opensearchcluster1_index_search_slowlog.json    opensearchcluster1_task_detailslog.json
opensearchcluster1_deprecation.log   opensearchcluster1_index_search_slowlog.log     opensearchcluster1_task_detailslog.log
```
with command below to reproduce
```
./test.sh integ-test manifests/2.6.0/opensearch-2.6.0-test.yml --component sql -p opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.6.0/7160/linux/x64/tar
```

src/test_workflow/integ_test/distribution_tar.py
```diff
@@ -39,5 +39,9 @@ def start_cmd(self) -> str:
         return start_cmd_map[self.filename]
 
     def uninstall(self) -> None:
-        logging.info(f"Cleanup {self.work_dir}/* content after the test")
-        subprocess.check_call(f"rm -rf {self.work_dir}/*", shell=True)
+        logging.info(f"Cleanup {self.work_dir}/* content after the test and keep {self.install_dir}/logs for recording.")
+        subprocess.check_call(f"mv {self.install_dir}/logs {self.work_dir}", shell=True)
+        subprocess.check_call(f"rm -rf {self.install_dir}", shell=True)
```
dblock, line 44: This is super dangerous. There's no guarantee that `self.install_dir` does not have spaces, or is blank. you'll be deleting all kinds of things. Replace all these system commands with Python code.

dblock (commented):
I won't block the PR for the problem below because existing code already does `rm -rf`, but I really think you should rewrite it. It also doesn't work on Windows.

src/test_workflow/integ_test/distribution_tar.py
```diff
@@ -39,5 +39,9 @@ def start_cmd(self) -> str:
...
+        subprocess.check_call(f"mkdir {self.install_dir}", shell=True)
+        subprocess.check_call(f"mv {self.work_dir}/logs {self.install_dir}", shell=True)
```
gaiksaya, line 46: +1 to db's comment. Also, We are moving the logs around instead of logging it with test-recorder. Can you check if test recorder can record these before they are wiped out? Looks like hacky way to just keep logs. We might need more information in the future from these folder like config, etc.

src/test_workflow/integ_test/distribution_tar.py
zelinh: Thanks for comments. Yeah I think current logics for uninstalling the cluster for tarball and windows is to force remove the entire `opensearch-{version}` folder. What do you think about we leave them alone and we only delete them along with the tmp workspace?

src/test_workflow/integ_test/distribution_tar.py
zelinh: It is using test-recorder to record the test-results. This part is to keep the original files undeleted and then the test-recorder could find it. I agree the original logic is hacky as well; just simply remove the entire folder to uninstall.

src/test_workflow/integ_test/distribution_tar.py
dblock: I am good with any solution that doesn't call `rm -rf` in shell :)

zelinh:
Close this PR with an alternative PR #3453 to resolve this issue.

PR #3415, opened 2023-04-18
Workflow to run performance tests using opensearch-benchmark

rishabh6788:
### Description
This PR adds support to run performance tests against OpenSearch cluster using opensearch-benchmark. The workflow is very similar to `perf-tests` with below mentioned updates and improvements:
* Uses public cdk infra package [opensearch-cluster-cdk](https://github.com/opensearch-project/opensearch-cluster-cdk) instead of private opensearch-infra repo. 
* Uses opensearch-benchmark for running and analyzing performance run tests instead of custom internal tool. 
* Supports single and as well as multi-node clusters. 
* Supports adding additional OS config from the command line. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-benchmark/issues/102


src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -0,0 +1,158 @@
...
+    ) -> None:
+        self.manifest = bundle_manifest
+        self.current_workspace = current_workspace
+        self.args = args
+        self.output_file = "output.json"
+        role = config["Constants"]["Role"]
+        params_dict = self.setup_cdk_params(config)
+        params_list = []
+        for key, value in params_dict.items():
+            print(key, value)
```
dblock, line 55: Leftover print?

src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -0,0 +1,111 @@
...
+        parser.add_argument("--suffix", dest="suffix", help="Suffix to be added to stack name for performance test")
+        parser.add_argument("--component", dest="component", default="OpenSearch",
+                            help="Component name that needs to be performance tested")
+        parser.add_argument("--config", type=argparse.FileType("r"), help="Config file.", required=True)
+        parser.add_argument(
+            "--without-security", dest="insecure", action="store_true",
+            help="Force the security of the cluster to be disabled.", default=False)
+        parser.add_argument("--keep", dest="keep", action="store_true",
+                            help="Do not delete the working temporary directory.")
+        parser.add_argument("--singleNode", dest="singleNode", action="store_true",
```
dblock, line 55: Are arguments elsewhere `camelCase`? I think not.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -0,0 +1,158 @@
...
+            if value:
+                if key == 'additionalConfig':
```
dblock, line 57: Why is this needed?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -0,0 +1,158 @@
...
+                    params_list.append(f" -c {key}={value}")
+        role_params = (
+            f" --require-approval=never --plugin cdk-assume-role-credential-plugin"
+            f" -c assume-role-credentials:writeIamRoleName={role} -c assume-role-credentials:readIamRoleName={role} "
+        )
+        self.params = "".join(params_list) + role_params
+        self.is_endpoint_public = False
+        self.cluster_endpoint = None
+        self.cluster_endpoint_with_port = None
+        self.stack_name = 'opensearch-infra-stack-' + self.args.stack_suffix + '-' + self.manifest.build.id + '-' + self.manifest.build.architecture
```
dblock, line 69: Use `f'...`

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def start(self) -> None:
+        # os.chdir(self.work_dir)
+        command = f"npm install && cdk deploy \"*\" {self.params} --outputs-file {self.output_file}"
+
+        logging.info(f'Executing "{command}" in {os.getcwd()}')
+        subprocess.check_call(command, cwd=os.getcwd(), shell=True)
+        with open(self.output_file, "r") as read_file:
+            load_output = json.load(read_file)
+            print(load_output[self.stack_name])
```
dblock, line 79: Use `logging` if you need to print something.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -0,0 +1,158 @@
...
+            "region": config["Constants"]["Region"],
+            "suffix": suffix,
+            "securityDisabled": "true" if self.args.insecure else "false",
+            "cpuArch": self.manifest.build.architecture,
+            "singleNodeCluster": "true" if self.args.singleNode else "false",
+            "distVersion": self.manifest.build.version,
+            "minDistribution": "true" if self.args.minDistribution else "false",
+            "serverAccessType": config["Constants"]["serverAccessType"],
+            "restrictServerAccessTo": config["Constants"]["restrictServerAccessTo"],
+            "additionalConfig": self.args.additionalConfig,
```
dblock, line 134: There's a lot of pass through params. I wonder whether we're better off not exposing all these command line options and only accepting a config file and passing it through here?

dblock (commented):
Looks pretty clean! Some nits.

Please document this in .md as part of this PR?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: Removed.

src/test_workflow/benchmark_test/benchmark_args.py
rishabh6788: this was a miss, updated to use uniform formatting.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: It has more to do how opensearch-cluster-cdk accepts the json type parameter, the input has to be enclosed in single-quote in order to be properly parsed in the CDK Code [here](https://github.com/opensearch-project/opensearch-cluster-cdk/blob/main/lib/os-cluster-entrypoint.ts#L139-L147). 
So in the above code I am just making sure that if the parameter is of type `additional-config` make sure the value is enclosed in single-quotes. Hope that makes sense. 
@dblock

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: Updated

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: Not required, removed.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: Could you please give an example here, not sure if I understood the comment properly. 
@dblock

src/test_workflow/benchmark_test/benchmark_test_cluster.py
dblock: It seems weird. Is there a more generic version of this? For example if the value has quotes it should be single quoted?

src/run_benchmark_test.py
```diff
@@ -0,0 +1,29 @@
...
+def main() -> int:
+    """
+        Entry point for Benchmark Test with bundle manifest, config file containing the required arguments for running
+        benchmarking test. Will call out in test.sh with benchmark as argument
+    """
+    benchmark_args = BenchmarkArgs()
+    console.configure(level=benchmark_args.logging_level)
+    manifest = BundleManifest.from_file(benchmark_args.bundle_manifest)
+    BenchmarkTestRunners.from_args(benchmark_args, manifest).run()
+    return 0
```
dblock, line 25: Should a failure return a non-zero error code?

src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -0,0 +1,115 @@
...
+        parser.add_argument("--manager-node-count", dest="managerNodeCount", help="Number of cluster manager nodes, "
+                                                                                  "default is 3")
+        parser.add_argument("--data-node-count", dest="dataNodeCount", help="Number of data nodes, default is 2")
+        parser.add_argument("--client-node-count", dest="clientNodeCount", help="Number of dedicated client nodes, "
+                                                                                "default is 0")
+        parser.add_argument("--ingest-node-count", dest="ingestNodeCount", help="Number of dedicated ingest nodes, "
+                                                                                "default is 0")
+        parser.add_argument("--ml-node-count", dest="mlNodeCount", help="Number of dedicated machine learning nodes, "
+                                                                        "default is 0")
+        parser.add_argument("--jvmSysProps", dest="jvmSysProps", help="A comma-separated list of key=value pairs that "
```
dblock, line 71: Another camelCase

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -0,0 +1,156 @@
...
+            suffix = self.args.stack_suffix + '-' + self.manifest.build.id + '-' + self.manifest.build.architecture
+        else:
+            suffix = self.manifest.build.id + '-' + self.manifest.build.architecture
+        return {
+            "distributionUrl": self.manifest.build.location,
+            "vpcId": config["Constants"]["VpcId"],
+            "account": config["Constants"]["AccountId"],
+            "region": config["Constants"]["Region"],
+            "suffix": suffix,
+            "securityDisabled": "true" if self.args.insecure else "false",
```
dblock, line 125: Should be able to do str(self.args.insecure) and always get the right value of "true" or "false" - default value for insecure should be `false`

src/test_workflow/benchmark_test/benchmark_test_cluster.py
dblock: I think I'm saying that almost all `Count` arguments come from args. I don't know whether that's the right answer, but a config file that contained all these options could be passed as 1 single argument `--config foobar.yml`. Then this code doesn't need to know about `dataNodeCount` for example, that would be contained inside `foobar.yml`.

rishabh6788: The idea to keep them as args is that there may be cases where a user wants to try different cluster configurations for running benchmark in parallel with different node counts for each cluster. Another is the ML use case where a user wants to have ML node enabled. 
@dblock

src/run_benchmark_test.py
rishabh6788: makes sense, will remove it.

src/test_workflow/benchmark_test/benchmark_args.py
rishabh6788: fixed it!

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: I guess I resorted to this because the boolean in python is `True` or `False` but in typescript and many other languages it is `true` or `false`. I think better way to handle this would be:
`str(self.args.insecure).lower()`.

dblock (changes requested):
You still have the `camelCase` variables all over the place, that's not the convention in the rest of the project. Example: https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/perf_test/perf_args.py#L21

src/test_workflow/benchmark_test/benchmark_test_cluster.py
dblock: Does quoting everything work, ie. doing `params_list.append(f" -c {key}=\'{value}\'")` all the time?

If not, add a TODO: fixme here or something like that with an explanation that this is a hack

Next, if you want to clean this up further,  you can write the loop with a `map` and a lambda that will make this a lot shorter. Finally, add role params that are `-c` into `params_dict` and let them be transformed with this code to avoid hard-coding strings. You'll still need the `--require-approval` and the `--plugin` params I guess. I'm ok to leave it as is, but consider doing it.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -0,0 +1,156 @@
...
+            f" -c assume-role-credentials:writeIamRoleName={role} -c assume-role-credentials:readIamRoleName={role} "
+        )
+        self.params = "".join(params_list) + role_params
+        self.is_endpoint_public = False
+        self.cluster_endpoint = None
+        self.cluster_endpoint_with_port = None
+        self.stack_name = f"opensearch-infra-stack-{self.args.stack_suffix}-{self.manifest.build.id}-{self.manifest.build.architecture}"
+
+    def start(self) -> None:
+        # os.chdir(self.work_dir)
```
dblock, line 71: remove commented code

src/test_workflow/benchmark_test/benchmark_test_cluster.py
dblock: That's fine. I think we're just arguing command line params vs. a config file.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: Ack! Will add a TODO and take this up in the next iteration.

rishabh6788:
> You still have the `camelCase` variables all over the place, that's not the convention in the rest of the project. Example: https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/perf_test/perf_args.py#L21

Got it! let me refactor everything and update accordingly. Too much coding in TS got me into this habit :-|

rishabh6788:
> You still have the `camelCase` variables all over the place, that's not the convention in the rest of the project. Example: https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/perf_test/perf_args.py#L21

Fixed it. Also added the TODO to get rid of hack I put to quote the json input argument. 
@dblock

src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -0,0 +1,118 @@
...
+                            help="Component name that needs to be performance tested")
+        parser.add_argument("--config", type=argparse.FileType("r"), help="Config file.", required=True)
+        parser.add_argument(
+            "--without-security", dest="insecure", action="store_true",
+            help="Force the security of the cluster to be disabled.", default=False)
+        parser.add_argument("--keep", dest="keep", action="store_true",
+                            help="Do not delete the working temporary directory.")
+        parser.add_argument("--single-node", dest="single_node", action="store_true",
+                            help="Is this a single node cluster")
+        parser.add_argument("--min-distribution", dest="min_distribution", action="store_true", help="Is it the "
```
dblock, line 57: Put each argument such as `help` on its own line and it won't need to be wrapped like this.

src/test_workflow/test_jsonargs.py
```diff
@@ -0,0 +1,24 @@
...
+class JsonArgs(argparse.Action):
+    def __call__(self, parser: Any, namespace: argparse.Namespace, values: Union[str, Sequence[Any], None], option_string: str = None) -> None:
+        setattr(namespace, self.dest, dict())
+        print(f"values are {values}")
+        for value in values:
+            key, value = value.split(':')
+            getattr(namespace, self.dest)[key] = value
+
+
+JsonArgs.__test__ = False  # type:ignore
```
dblock, line 24: Call this file `json_args.py` and this won't be needed. It's also the convention.

src/test_workflow/test_jsonargs.py
```diff
@@ -0,0 +1,24 @@
...
+# GitHub history for details.
+
+import argparse
+from typing import Any, Sequence, Union
+
+
+class JsonArgs(argparse.Action):
+    def __call__(self, parser: Any, namespace: argparse.Namespace, values: Union[str, Sequence[Any], None], option_string: str = None) -> None:
+        setattr(namespace, self.dest, dict())
+        print(f"values are {values}")
```
dblock, line 18: Stray `printf`, remove.

src/test_workflow/benchmark_test/benchmark_test_suite.py
```diff
@@ -0,0 +1,54 @@
...
+        # Pass the cluster endpoints with -t for multi-cluster use cases(e.g. cross-cluster-replication)
+        if args.benchmark_config:
+            base_command = f"docker run -v {args.benchmark_config}:/opensearch-benchmark/.benchmark/benchmark.ini"
+        else:
+            base_command = "docker run"
+
+        self.command = (
+            f"{base_command} opensearchproject/opensearch-benchmark:latest execute_test "
+            f"--workload={self.args.workload} --test-mode --pipeline=benchmark-only --target-hosts={endpoint}"
+        )
```
dblock, line 44: Looks messy, and uses an unnecessary local `base_command` variable.

```python
self.command = "docker run"
if args.benchmark_config:
   self.command += " -v ..."
self.command += "opensearchproject/opensearch-benchmark:latest execute_test:
...
```

dblock (changes requested):
Needs documentation in .md files for these new tests.

Code: I'm down to nits. Address those and it's ready to merge.

rishabh6788:
> Needs documentation in .md files for these new tests.
> 
> Code: I'm down to nits. Address those and it's ready to merge.

Added `README` for benchmarking test. Have a TODO for adding jenkins job details once it is ready. It is still WIP. 
Addressed the comments. 
@dblock

src/test_workflow/benchmark_test/benchmark_test_suite.py
```diff
@@ -0,0 +1,54 @@
...
+
+        if args.user_tag:
+            user_tag = f"--user-tag=\"{args.user_tag}\""
+            self.command = f"{self.command} {user_tag}"
```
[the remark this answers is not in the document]
rishabh6788: Done!

src/test_workflow/benchmark_test/benchmark_test_suite.py
rishabh6788: Thanks, done!

src/test_workflow/test_jsonargs.py
rishabh6788: Done!

src/test_workflow/test_jsonargs.py
rishabh6788: Makes sense, done!

src/test_workflow/benchmark_test/benchmark_args.py
rishabh6788: Done!

src/test_workflow/README.md
```diff
@@ -174,6 +174,17 @@ Internal tools provide dashboards for monitoring cluster behavior during these t
 |Indexing Latency|Consistent during each test iteration|upward trends|
 |Query Latency|Varies based on the query being issued|upward trends|
 
+### Benchmarking Tests
```
dblock, line 177: Update TOC

src/test_workflow/README.md
```diff
@@ -174,6 +174,17 @@ Internal tools provide dashboards for monitoring cluster behavior during these t
...
+Runs benchmarking test on a remote opensource OpenSearch cluster. Uses [OpenSearch Benchmark](https://github.com/opensearch-project/OpenSearch-Benchmark) to run benchmark tests.
+At a high-level the benchmarking test workflow uses [opensearch-cluster-cdk](https://github.com/opensearch-project/opensearch-cluster-cdk.git) to first set-up an OpenSearch
```
dblock, line 179: Don't line wrap, https://code.dblock.org/2021/06/07/to-wrap-or-not-to-wrap-in-markdown.html

src/test_workflow/README.md
```diff
@@ -174,6 +174,17 @@ Internal tools provide dashboards for monitoring cluster behavior during these t
 |Indexing Latency|Consistent during each test iteration|upward trends|
 |Query Latency|Varies based on the query being issued|upward trends|
 
+### Benchmarking Tests
+Runs benchmarking test on a remote opensource OpenSearch cluster. Uses [OpenSearch Benchmark](https://github.com/opensearch-project/OpenSearch-Benchmark) to run benchmark tests.
```
dblock, line 178: test or tests?

you can remove "to run benchmark tests", that's what it does

src/test_workflow/README.md
```diff
@@ -174,6 +174,17 @@ Internal tools provide dashboards for monitoring cluster behavior during these t
...
+At a high-level the benchmarking test workflow uses [opensearch-cluster-cdk](https://github.com/opensearch-project/opensearch-cluster-cdk.git) to first set-up an OpenSearch 
+cluster (single/multi-node) and then executes `opensearch-benchmark` to run benchmark test against that cluster. The performance metric that opensearch-benchmark generates 
+during the run are ingested into another OS cluster for further analysis and dashboarding purpose.
+
+The benchmarking tests will be run nightly and if you have a feature in any released/un-released OS version that you want to benchmark periodically please create an issue
```
dblock, line 183: I believe we don't abbreviate OS, spell OpenSearch.

src/test_workflow/README.md
```diff
@@ -237,6 +248,9 @@ After the performance test completes, it will report back the test results as we
 
 The development is tracked by meta issue [#126](https://github.com/opensearch-project/opensearch-build/issues/126)
 
+#### benchmarkTest job
```
dblock, line 251: Capitalize Like Other Titles

You can remove this for now since we don't have content.

dblock (changes requested):
Minor stuff for the markdown please.

src/test_workflow/README.md
rishabh6788: Sorry about that, updated!

src/test_workflow/README.md
rishabh6788: Ack! fixed!

src/test_workflow/README.md
rishabh6788: Done!

src/test_workflow/README.md
rishabh6788: Updated!

src/test_workflow/README.md
rishabh6788: Removed for now. Will add once we have the job ready to be merged.

dblock (approved):

Issue #2504, opened 2022-08-24
[Bug]: The cdk clusters used for performance testings are not destroyed if build is aborted


gaiksaya:
### Describe the bug

If the build is aborted in https://build.ci.opensearch.org/job/perf-test/ the underlying performance test clusters brought up using cdk are not destroyed causing them to be abandoned until manually cleaned

### To reproduce

Run the above job and abort it.
Got to the related aws account and see that cloudformation stack for that job is not destroyed

### Expected behavior

Stacks should be cleaned irrespective of success or failure

### Acceptance criteria 

- [ ] When the Jenkins/CI job is abortted/interrupted/terminated, CDK cluster that was created by code shall be deleted  
- [x] When the Jenkins/CI job is completed successfully, CDK cluster that was created by code shall be deleted
- [x] When the Jenkins/CI job is fail, CDK cluster that was created by code shall be deleted

jordarlu:
https://stackoverflow.com/questions/25047537/running-a-post-build-script-when-a-jenkins-job-is-aborted
https://www.jenkins.io/doc/book/pipeline/syntax/#post

gaiksaya:
When a workflow is aborted it is usually due to issue with the jenkins example: timeouts, connectivity error, etc.
One of the approach can be:
Export the name of the stack form python code (AFAIK they are static so not really required) and use the aborted block to destroy the stack using aws cli command.

What it would look like:
```
post {
    aborted {
        node(AGENT_LABEL) {
            script {
                withCredentials([string(credentialsId: "${args.roleAccountNumberCred}", variable: 'AWS_ACCOUNT_NUMBER')]) {
                    withAWS(role: args.assumedRoleName, roleAccount: "${AWS_ACCOUNT_NUMBER}", duration: 900, roleSessionName: 'jenkins-session', region: "${region}") {
                        cfnDelete(stack:'my-stack', pollInterval:1000, roleArn: 'args.assumedRoleNam')
                    }
                }
            }
        }
    }
}

jordarlu:
this issue is resolved. Test ref is here https://build.ci.opensearch.org/job/perf-test/3962/console  , thanks

PR #3438, opened 2023-04-25
Correction on CDK cluster destroyed issue when the perf-test job is aborted

jordarlu:
### Description
After the testing from the previous fix, we found the CDK cluster was still not destroyed due to assuming role issue; hence, we want to update the role to fix it.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2504

### DIFF
--- a/jenkins/opensearch/perf-test.jenkinsfile
+++ b/jenkins/opensearch/perf-test.jenkinsfile
@@ -272,7 +272,7 @@ pipeline {
                     "test-single-security-${BUILD_ID}-${ARCHITECTURE}-${BUILD_NUMBER}"
                     ]
                     withCredentials([string(credentialsId: 'jenkins-aws-account-public', variable: 'AWS_ACCOUNT_PUBLIC')]) {
-                        withAWS(role: 'opensearch-test', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session', region: 'eu-west-1') {
+                        withAWS(role: 'cfn-set-up', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session', region: 'eu-west-1') {
                             try {
                                 for (String stackName : stackNames) {
                                     def stack = null


jenkins/opensearch/perf-test.jenkinsfile
```diff
@@ -272,7 +272,7 @@ pipeline {
                     "test-single-security-${BUILD_ID}-${ARCHITECTURE}-${BUILD_NUMBER}"
                     ]
                     withCredentials([string(credentialsId: 'jenkins-aws-account-public', variable: 'AWS_ACCOUNT_PUBLIC')]) {
-                        withAWS(role: 'opensearch-test', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session', region: 'eu-west-1') {
+                        withAWS(role: 'cfn-set-up', roleAccount: "${AWS_ACCOUNT_PUBLIC}", duration: 900, roleSessionName: 'jenkins-session', region: 'eu-west-1') {
```
peterzhuamazon, line 275: I am not getting why the role needs to be changed.

I checked and the new policies are already in opensearch-test role, and the cfn-set-up role not even exist.

You added a new policy to an existing role.

peterzhuamazon (changes requested):

jenkins/opensearch/perf-test.jenkinsfile
gaiksaya: cfn-set-up is the role actually used to deploy the clusters internally. Lot of hacky ways due to opensearch-infra being private repo. The Jenkins main node and agent node can assume `cfn-set-up` so we are good.

peterzhuamazon: Hi,

We need to add these items:
1. Add a new secret to host the proper test cluster account id (named `odfe-aws-account-test` possible?).
2. change the withCredentials section to retrieve the account id from `odfe-aws-account-test`.
3. revert previous change that add the permissions on wrong role.

cc: @jordarlu @gaiksaya 

Thanks.

jordarlu: Thanks for the suggestions , @peterzhuamazon @gaiksaya .. I made the update accordingly

jordarlu: After our dicussion, we decided the name of the new ID would be   perf-test-account-id

PR #3472, opened 2023-05-04
Enable knn in 2.8

jmazanec15:
### Description
Enable k-NN in 2.8.

Upgraded 2.x to 2.8 in https://github.com/opensearch-project/k-NN/pull/886

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3473, opened 2023-05-04
Move old manifests to legacy

gaiksaya:
### Description
Moves some old version manifests to legacy manifest folder

### Issues Resolved


gaiksaya:
> So the only thing that is not legacy is the latest version?

Kept few more for reference in manifests folder along with latest version. Here is the list https://github.com/gaiksaya/opensearch-build/tree/ml/manifests

peterzhuamazon (approved):

gaiksaya:
Looks like 3.0.0 manifest check is failing for multiple components such as ml-commons, neural-search, k-NN. Should we CI checks for 3.0.0 OS manifest?
@bbarani @dblock @prudhvigodithi @peterzhuamazon

dblock:
> Looks like 3.0.0 manifest check is failing for multiple components such as ml-commons, neural-search, k-NN. Should we disable CI checks for 3.0.0 OS manifest? @bbarani @dblock @prudhvigodithi @peterzhuamazon

No, we should be fixing the issues!

jmazanec15:
For k-NN, we are in process of fixing our CI. Problem is there were a couple breaks that have caused a bit of a tangle up in the k-NN CI.

The jackson/gradle upgrade (https://github.com/opensearch-project/OpenSearch/pull/7300) and immutable maps removal (https://github.com/opensearch-project/OpenSearch/pull/7309) has broken a few things. We raised this PR: https://github.com/opensearch-project/k-NN/pull/885. However, CI is failing for secure test because the docker distribution has yet to be updated due to the jackson/gradle issue (circular). Additionally, BWC tests are failing because k-NN 2.8 has not been added to distribution yet (https://github.com/opensearch-project/opensearch-build/pull/3472). neural-search is stuck because it depends on k-NN.

In the long term, I believe solution is for plugins to not take dependencies on distribution for secure tests and bwc tests - instead, depend on opensearch-core min artifact and individual dependent plugin artifacts. I think this will break the circular CI dependency chain mentioned above.

In the meantime, for k-NN, I would like to merge this: https://github.com/opensearch-project/opensearch-build/pull/3472 so that BWC tests can pass. Then merge https://github.com/opensearch-project/k-NN/pull/885 to unblock distro build and then fix neural search.

gaiksaya:
In that case just moving the old manifests to legacy for now. 3.0.0 will remain broken until plugin teams fix the issue.
Thanks!

PR #3479, opened 2023-05-05
Resolve PA proc not disown and ml_dashboards settings missing in OSD docker entrypoint

peterzhuamazon:
### Description
Resolve PA proc not disown and ml_dashboards settings missing in OSD docker entrypoint.
The former issue was discovered by @shikharbhardwaj , thanks.

### Issues Resolved

### DIFF
--- a/docker/release/config/opensearch/opensearch-docker-entrypoint.sh
+++ b/docker/release/config/opensearch/opensearch-docker-entrypoint.sh
@@ -58,7 +58,7 @@ function setupPerformanceAnalyzerPlugin {
             echo "Disabling execution of $OPENSEARCH_HOME/bin/$PERFORMANCE_ANALYZER_PLUGIN/performance-analyzer-agent-cli for OpenSearch Performance Analyzer Plugin"
         else
             echo "Enabling execution of OPENSEARCH_HOME/bin/$PERFORMANCE_ANALYZER_PLUGIN/performance-analyzer-agent-cli for OpenSearch Performance Analyzer Plugin"
-            $OPENSEARCH_HOME/bin/opensearch-performance-analyzer/performance-analyzer-agent-cli > $OPENSEARCH_HOME/logs/PerformanceAnalyzer.log 2>&1 &
+            $OPENSEARCH_HOME/bin/opensearch-performance-analyzer/performance-analyzer-agent-cli > $OPENSEARCH_HOME/logs/PerformanceAnalyzer.log 2>&1 & disown
         fi
     else
         echo "OpenSearch Performance Analyzer Plugin does not exist, disable by default"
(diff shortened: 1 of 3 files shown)


docker/release/config/opensearch/opensearch-docker-entrypoint.sh
```diff
@@ -58,7 +58,7 @@ function setupPerformanceAnalyzerPlugin {
             echo "Disabling execution of $OPENSEARCH_HOME/bin/$PERFORMANCE_ANALYZER_PLUGIN/performance-analyzer-agent-cli for OpenSearch Performance Analyzer Plugin"
         else
             echo "Enabling execution of OPENSEARCH_HOME/bin/$PERFORMANCE_ANALYZER_PLUGIN/performance-analyzer-agent-cli for OpenSearch Performance Analyzer Plugin"
-            $OPENSEARCH_HOME/bin/opensearch-performance-analyzer/performance-analyzer-agent-cli > $OPENSEARCH_HOME/logs/PerformanceAnalyzer.log 2>&1 &
+            $OPENSEARCH_HOME/bin/opensearch-performance-analyzer/performance-analyzer-agent-cli > $OPENSEARCH_HOME/logs/PerformanceAnalyzer.log 2>&1 & disown
```
prudhvigodithi, line 61: thanks @peterzhuamazon does `disown` remove the job? the idea is to run the `performance-analyzer-agent-cli` outputting the logs to `$OPENSEARCH_HOME/logs/PerformanceAnalyzer.log`?

prudhvigodithi: should it be `disown -h`?

peterzhuamazon: Nope, `& disown` is very similar to `nohup` in the sense of avoid getting interrupted by `SIGHUP`.
`nohup` will still have the process in job list while `disown` will remove it.

In this case since it is a non-interactive shell anyway we wont really need to bring it from background to foreground, both `nohup` and `& disown` satisfy the use case right here.

I can change it to nohup if you want to, doesnt hurt.

peterzhuamazon: I think one advantage of this is, nohup will have a new pid while disown make sure the pid given is the actual process pid.

peterzhuamazon: `-h` is not needed as it requires a jobid.
Plus, we are ok if someone specifically send a `SIGHUP` signal to PA, but not when the shell is detached.
The current implementation satisfied our need. Thanks.

prudhvigodithi (approved):

prudhvigodithi (approved):

docker/release/config/opensearch/opensearch-docker-entrypoint.sh
prudhvigodithi: Thanks, approved peter, `nohup` also adds a `nohup.out` file, just in case if we need that file.

peterzhuamazon: > Thanks, approved peter, `nohup` also adds a `nohup.out` file, just in case if we need that file.

Since we are already redirect the log to another file, the `nohup.out` will just not create and redirect there.

PR #3529, opened 2023-05-17
Add Neural-search to 3.0.0 manifest

martin-gaievski:
### Description
Add neural-search plugin to 3.0.0 manifest

### DIFF
--- a/manifests/3.0.0/opensearch-3.0.0.yml
+++ b/manifests/3.0.0/opensearch-3.0.0.yml
@@ -23,6 +23,15 @@ components:
     checks:
       - gradle:publish
       - gradle:properties:version
+  - name: ml-commons
+    repository: https://github.com/opensearch-project/ml-commons.git
+    ref: main
+    platforms:
+      - linux
+      - windows
+    checks:
+      - gradle:properties:version
+      - gradle:dependencies:opensearch.version: opensearch-ml-plugin
   - name: job-scheduler
     repository: https://github.com/opensearch-project/job-scheduler.git
     ref: main
--- a/manifests/3.0.0/opensearch-3.0.0.yml
+++ b/manifests/3.0.0/opensearch-3.0.0.yml
@@ -73,6 +73,15 @@ components:
     checks:
       - gradle:properties:version
       - gradle:dependencies:opensearch.version
+  - name: neural-search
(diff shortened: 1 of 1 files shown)


gaiksaya:
Looks like the CI passed. I assume neural does not need ml-commons as a dependency during the build since this manifest lacks it?

martin-gaievski:
> Looks like the CI passed. I assume neural does not need ml-commons as a dependency during the build since this manifest lacks it?

We use ml-commons jars from public repo, as to my knowledge ml-commons is doing GHA to publish artifacts to maven, it's independent of the distribution build.

Currently ml-commons is unblocked, I can add it with this PR if that's ok

gaiksaya:
> > Looks like the CI passed. I assume neural does not need ml-commons as a dependency during the build since this manifest lacks it?
> 
> We use ml-commons jars from public repo, as to my knowledge ml-commons is doing GHA to publish artifacts to maven, it's independent of the distribution build.
> 
> Currently ml-commons is unblocked, I can add it with this PR if that's ok

Yes please do. Maven only has snapshots version of ml-commons. However while building distribution, the snapshot is set to false. I believe it would fail then since it is unable to find it in  maven local.

martin-gaievski:
Added ml-commons as requested, fyi @b4sjoo

manifests/3.0.0/opensearch-3.0.0.yml
```diff
@@ -23,6 +23,15 @@ components:
...
       - gradle:properties:version
+  - name: ml-commons
+    repository: https://github.com/opensearch-project/ml-commons.git
+    ref: main
+    platforms:
+      - linux
+      - windows
+    checks:
+      - gradle:properties:version
+      - gradle:dependencies:opensearch.version: opensearch-ml-plugin
```
gaiksaya, line 34: Not sure if JS is one of the ml-commons's dependency? Care to make it below job-scheduler? Just to be safe?
Thanks!

martin-gaievski: order ml-commons - js was in the manifest before it got removed last time, basically I restored manifest to the form it was before that change 5 days ago - https://github.com/opensearch-project/opensearch-build/commit/7cc32169662445fbd1d177153c2b459573cd7cb7

gaiksaya (approved):

Issue #3010, opened 2022-12-07
Docker Healthcheck


markusa380:
**Is your feature request related to a problem? Please describe.**
It's harder than necessary to detect when OpenSearch Docker containers are ready to accept requests, for example when using [testcontainers](https://www.testcontainers.org).

**Describe the solution you'd like**
I'd like to see the official Docker images use the `HEALTHCHECK` instruction (which should probably request `/_cat/health?v=status`) so other systems can depend on it.

**Describe alternatives you've considered**
Alternative solutions revolve around the systems using the image, e.g. by implementing custom health check loops or, like testcontainers with ElasticSearch, by using a regex on STDOUT.

**Additional context**
n/a

bbarani:
@markusa380 Thanks for opening the issue. We are looking in to this issue. In the meantime, feel free to contribute this enhancement.

Tokesh:
Hello! I'm newbie at Open Source Community. If it still relevant issue, can you provide information about the dockerfiles in which the heartbeat status needs to be added?
I can try to solve this issue. I'll do my best!

dblock:
Is https://github.com/opensearch-project/opensearch-go/pull/315 relevant?

bbarani:
@prudhvigodithi @peterzhuamazon @rishabh6788 @gaiksaya Do you have some inputs for @Tokesh ?

gaiksaya:
Hi @Tokesh 

These are the relevant docker files which would need health check:
https://github.com/opensearch-project/opensearch-build/tree/main/docker/release/dockerfiles

Associated docker images:
https://hub.docker.com/r/opensearchproject/opensearch
https://hub.docker.com/r/opensearchproject/opensearch-dashboards

prudhvigodithi:
Hey @Tokesh welcome to the OpenSearch community. Here is some useful [link](https://dev.to/aws-builders/implementing-a-healthcheck-for-a-docker-container-4m9h), we can add a health check with OpenSearch API that shows the health of the container. For helm charts this [how](https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/values.yaml#L349-L371) its done you can use this as reference.

Tokesh:
Thank you all for the details! I am currently taking my college exams. I will try to send the PR for consideration within a week. I don't seem to be blocking anyone.

prudhvigodithi:
All the best for your exam @Tokesh.

peterzhuamazon:
[...]

PR #3557, opened 2023-05-24
adding healthcheck to docker-compose

Tokesh:
### Description
Here few ways how we can add healthcheck, but i add some settings to docker-compose files based on their ports.
I test it in docker ps. Statuses added healthy or unhealthy (problems with SSL certification)
I'm ready to listen all your suggestions and rework this.

### Issues Resolved
[#3010]

### DIFF
--- a/docker/release/dockercomposefiles/docker-compose-1.x.yml
+++ b/docker/release/dockercomposefiles/docker-compose-1.x.yml
@@ -25,6 +25,11 @@ services:
       - 9600:9600  # required for Performance Analyzer
     networks:
       - opensearch-net
+    healthcheck:
+      test: curl --fail -s http://localhost:9200/ || exit 1
+      interval: 30s
+      timeout: 5s
+      retries: 3
   opensearch-node2:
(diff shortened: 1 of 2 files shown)


peterzhuamazon:
Thanks @Tokesh on the contribution here.

@prudhvigodithi I am not sure if using curl on docker-compose file is a good way to do healthcheck?
Since this is only applying to docker-compose file.

Tokesh:
> Thanks @Tokesh on the contribution here.
> 
> @prudhvigodithi I am not sure if using curl on docker-compose file is a good way to do healthcheck? Since this is only applying to docker-compose file.

Thank you for reviewing of pr! 

But really, i'm not sure about docker-compose, because i can add it everywhere, like in dockerfiles. 
This I wanted to ask you, maybe there is a more efficient option where you can add it? I just thought that all the images that are needed will be raised through the docker compose and you can add them there and that will be enough. But I'll still be happy to hear what I can improve in my code. I am so glad that I can learn from you!

Issue #3589, opened 2023-06-01
[BUG] OSD integTest lack additional_config entrypoint for both OS and OSD clusters


peterzhuamazon:
Hi,

This is related to OSD specific integTest.

In OS, the additional config entry can be grab from the opensearch-<>-test.json and added to the opensearch.yml.

In OSD, this functional is half-complete in service_opensearch_dashboards.py and only allow hardcoded entries, without interpreting the entries from opensearch-dashboards-<>-test.json.

Moreover, in OSD integTest you need to start both OS and OSD, means the additional config in OS also need a new param to hold, so it wont get mixed up with OSD additional config. Right now both are under `additional-cluster-configs` key and will both be pushed to opensearch_dashboards.yml.

Thank.

peterzhuamazon:
Only partially resolved this issue in PR:
* 3590

gaiksaya:
[Triage] The issue is partially resolved.

peterzhuamazon:
We have implemented the OSD side of configs already.
As for OS configs during OSD test there does not seems to have a specific need from plugins.
Will close this issue for now until further notices. Thanks.

peterzhuamazon:
Seems like there is a use case here: 
* https://github.com/opensearch-project/opensearch-build/pull/4402

kavilla:
Low priority: for OpenSearch Dashboards we set `cluster.routing.allocation.disk.threshold_enabled=false` in our GitHub workflows. It would be a nice-to-have feature to be configure OpenSearch when we needed

Priority is low because we should figure out why that threshold is being hit.

ashwin-pc:
I wouldn't categorize this as a low priority since this is blocking running the tests with the same environment today locally and in the infra pipeline. Having identical test environments is key to reproducibility. Figuring out why we have to increase the memory limits is a different issue and I don't think we should tie the two together since that may be a bigger problem to solve and we shouldn't hold off on this feature on the hopes that that will be resolved.

PR #3590, opened 2023-06-01
Add additional cluster config support in OSD integTest

peterzhuamazon:
### Description
Add additional cluster config support in OSD integTest

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3589 partially
#3434

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3593, opened 2023-06-02
Initial empty dictionary for None additional cluster configs

peterzhuamazon:
### Description
Initial empty dictionary for None additional cluster configs

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3589 partially
#3434 


bbarani (approved):

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -84,6 +84,9 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
             self.additional_cluster_config = self.test_config.integ_test.get("additional-cluster-configs")
             logging.info(f"Additional config found: {self.additional_cluster_config}")
 
+        if self.additional_cluster_config is None:
```
rishabh6788, line 87: how about just initialize `self.additional_cluster_config` as empty directory and check the `len` or `keys()` method to check if it has any values? @peterzhuamazon

peterzhuamazon:
This is also to follow up #3590.

src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
peterzhuamazon: That can be done but right now I am just matching the OS side implementation.
We can improve this in later PRs.

dblock: I'm late to this game, but assigning of a default should happen in the constructor, not inside the execution.

Otherwise do `self.additional_cluster_config or {}` anywhere it's used.

PR #3684, opened 2023-06-27
Add support to run benchmark against min distribution

### LINKED ISSUE #3516: Add option to run benchmark test againt OpenSearch min distribution
The benchmark test in its current form runs only against OpenSearch bundle distribution.
We need to be able to run benchmarking test agains `min` distribution as well.

rishabh6788:
### Description
This PR adds support to run benchmark tests against min bundle. 

### Issues Resolved
#3516 

### DIFF
--- a/src/run_benchmark_test.py
+++ b/src/run_benchmark_test.py
@@ -6,7 +6,9 @@
 # compatible open source license.
 
 import sys
+from typing import Union
 
+from manifests.build_manifest import BuildManifest
 from manifests.bundle_manifest import BundleManifest
 from system import console
 from test_workflow.benchmark_test.benchmark_args import BenchmarkArgs
@@ -20,7 +22,8 @@ def main() -> int:
     """
     benchmark_args = BenchmarkArgs()
     console.configure(level=benchmark_args.logging_level)
-    manifest = BundleManifest.from_file(benchmark_args.bundle_manifest)
+    manifest: Union[BundleManifest, BuildManifest] = BundleManifest.from_file(benchmark_args.bundle_manifest) if not benchmark_args.min_distribution else \
+        BuildManifest.from_file(benchmark_args.bundle_manifest)
     BenchmarkTestRunners.from_args(benchmark_args, manifest).run()
     return 0
 
--- a/src/test_workflow/benchmark_test/benchmark_test_cluster.py
+++ b/src/test_workflow/benchmark_test/benchmark_test_cluster.py
@@ -11,18 +11,19 @@ import logging
 import os
 import subprocess
 from contextlib import contextmanager
-from typing import Any, Generator
+from typing import Any, Generator, Union
 
 import requests
 from requests.auth import HTTPBasicAuth
 from retry.api import retry_call  # type: ignore
 
+from manifests.build_manifest import BuildManifest
 from manifests.bundle_manifest import BundleManifest
 from test_workflow.benchmark_test.benchmark_args import BenchmarkArgs
 
 
 class BenchmarkTestCluster:
-    manifest: BundleManifest
+    manifest: Union[BundleManifest, BuildManifest]
     work_dir: str
     current_workspace: str
     args: BenchmarkArgs
@@ -39,7 +40,7 @@ class BenchmarkTestCluster:
 
     def __init__(
             self,
-            bundle_manifest: BundleManifest,
+            bundle_manifest: Union[BundleManifest, BuildManifest],
             config: dict,
             args: BenchmarkArgs,
             current_workspace: str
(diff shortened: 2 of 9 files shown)


src/run_benchmark_test.py
```diff
@@ -18,9 +20,11 @@ def main() -> int:
         Entry point for Benchmark Test with bundle manifest, config file containing the required arguments for running
         benchmarking test. Will call out in test.sh with benchmark as argument
     """
+    manifest: Union[BundleManifest, BuildManifest]
     benchmark_args = BenchmarkArgs()
     console.configure(level=benchmark_args.logging_level)
-    manifest = BundleManifest.from_file(benchmark_args.bundle_manifest)
+    manifest = BundleManifest.from_file(benchmark_args.bundle_manifest) if not benchmark_args.min_distribution else \
+        BuildManifest.from_file(benchmark_args.bundle_manifest)
```
peterzhuamazon, line 27: I think you can just combine these two?

manifest: Union[BundleManifest, BuildManifest] = <> ?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -121,7 +122,9 @@ def setup_cdk_params(self, config: dict) -> dict:
         else:
             suffix = self.manifest.build.id + '-' + self.manifest.build.architecture
         return {
-            "distributionUrl": self.manifest.build.location,
+            "distributionUrl": self.manifest.build.location if isinstance(self.manifest, BundleManifest) else
+            f"https://artifacts.opensearch.org/snapshots/core/opensearch/{self.manifest.build.version}/opensearch-min-"
+            f"{self.manifest.build.version}-linux-{self.manifest.build.architecture}-latest.tar.gz",
```
peterzhuamazon, line 127: Not really liking this but does build manifest not providing any of this information?
If so I would seriously thinking about Prudhvi suggestion to just assemble raw min without plugin before pushing to artifact.opensearch.org.

peterzhuamazon (commented):
Hi @rishabh6788 add some comments here.

Thanks!

src/test_workflow/benchmark_test/benchmark_test_cluster.py
peterzhuamazon: Probably create an issue related to this as a later improvements?

gaiksaya: I think we are already doing that https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L156-L168

Just need to upload the manifest too?

rishabh6788: As of now the BuildManifest doesn't have `location` field in the manifest therefore had to format the url to download the min distribution. 
I will create an issue to track this and update once the field becomes available.

gaiksaya: I believe what we are saying is today snapshots `min` is created same way as we build distribution but just with 1 component `OpenSearch`. So it has all the things that distribution does. Regarding commit id that you want, we can get that if we also upload the manifest.yml to artifacts.opensearch.org in the same way as we upload min tarball.

gaiksaya: See code here:
https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L156-L168
and 
uploading part here:
https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/uploadMinSnapshotsToS3.groovy#L43-L59

rishabh6788: @gaiksaya Peter's change uploads the min-snapshot manifest too, it is of type BuildManifest which doesn't have `location` field in it as per schema.  
min-snapshot manifest https://artifacts.opensearch.org/snapshots/core/opensearch/2.9.0-SNAPSHOT/opensearch-min-2.9.0-SNAPSHOT-linux-arm64-latest.tar.gz.build-manifest.yml
@peterzhuamazon

peterzhuamazon: @gaiksaya we are creating with build, not assemble after build, this means only build manifest is present which does not have the location information.

By running another round of assemble without plugin would just repackage min into a tar with a startup script and bundle manifest. The reason I am hesitate because that startup script will break due to no SEC and PA plugins.

Therefore, I would say put this as a later improvement in an issue.

rishabh6788: Bundling min with startup script is not the right way forward as it will fail due to lack of security plugin.

gaiksaya: I see what you mean! Yeah we need to add check if plugin exists in start up scripts.

rishabh6788: Created issue https://github.com/opensearch-project/opensearch-build/issues/3686 to track above mentioned requirements. @gaiksaya @peterzhuamazon

gaiksaya (approved):

peterzhuamazon (approved):

PR #3710, opened 2023-07-07
Allow dylib to be signed by macos signing and fix verification command

gaiksaya:
### Description
Allow dylib to be signed by macos signing and fix verification command. pkgutil should only be used when the type of pkg signed is a package (.pkg) else use codesign by default.

### Issues Resolved

### DIFF
--- a/src/sign_workflow/signer_mac.py
+++ b/src/sign_workflow/signer_mac.py
@@ -18,7 +18,7 @@ This class is responsible for signing macos artifacts using the OpenSearch-signe
 
 
 class SignerMac(Signer):
-    ACCEPTED_FILE_TYPES = [".pkg", ".dmg"]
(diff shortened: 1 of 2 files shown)


src/sign_workflow/signer_mac.py
```diff
@@ -18,7 +18,7 @@
 
 
 class SignerMac(Signer):
-    ACCEPTED_FILE_TYPES = [".pkg", ".dmg"]
+    ACCEPTED_FILE_TYPES = [".pkg", ".dmg", ".dylib"]
```
zelinh, line 21: Does our signer sign this file extension? I haven't signed any of this type of files, want to make sure we tested it before.

gaiksaya: Yes we do! It is a part of sql ODBC signing part. We tested and released the product too.

zelinh (approved):
LGTM! Thanks.

PR #3728, opened 2023-07-12
Add check if security plugin exists in startup script

rishabh6788:
### Description
There are certain cases when a distribution build is built without security plugin. In that case when `opensearch-tar-install.sh` script is run to start the OS process it fails at this [line](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/startup/tar/linux/opensearch-tar-install.sh#L12) as there is no `opensearch-security` plugin present in the bundle. 
To overcome this adding a simple condition to check whether `opensearch-security` plugin exists in the bundle. 

### Issues Resolved


peterzhuamazon (approved):
Offline talk with @rishabh6788 confirms linux work and also test on Windows runs on my machine.

Thanks.

gaiksaya:
Thanks for fixing this @rishabh6788 Great step for making all RC candidates executable. 🙌

Issue #3196, opened 2023-02-13
[Bug]: Fix OS integration tests for Windows platform


rishabh6788:
### Describe the bug

In the current scenario the OS integration tests have to be run manually for Windows platform as they don't run as expected. The major problem is of failure during clean-up process after the tests have run but the process fails as it is not able to delete the temp files generated during test run. 

Sample Logs: WIP

### To reproduce

On Windows 2019 machine checkout `opensearch-build` repo and run `./test.sh integ-test manifests/1.3.8/opensearch-1.3.8-test.yml --component index-management` 

### Expected behavior

The integ test should run without failing to delete the temp file generated and should display the final result of `PASS` or `FAILURE` once the test has completed.

zelinh:
I'm able to reproduce this issue on my Windows instance. 
Here are the logs:
```
Traceback (most recent call last):
    status = self.__setup_cluster_and_execute_test_config(config)
  File "C:\Users\Administrator\workspace\opensearch-build\src\test_workflow\integ_test\integ_test_suite_opensearch_dashbo
ards.py", line 104, in __setup_cluster_and_execute_test_config
    return self.execute_integtest_sh(endpoint, port, security, config)
  File "C:\Users\Administrator\scoop\apps\python39\3.9.13\lib\contextlib.py", line 126, in __exit__
    next(self.gen)
  File "C:\Users\Administrator\workspace\opensearch-build\src\test_workflow\test_cluster.py", line 64, in create
    cluster.terminate()
  File "C:\Users\Administrator\workspace\opensearch-build\src\test_workflow\test_cluster.py", line 85, in terminate
    self.termination_result = self.service.terminate()
  File "C:\Users\Administrator\workspace\opensearch-build\src\test_workflow\integ_test\service.py", line 65, in terminate

    self.return_code = self.process_handler.terminate()
  File "C:\Users\Administrator\workspace\opensearch-build\src\system\process.py", line 63, in terminate
    os.remove(self.stdout.name)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users
\\ADMINI~1\\AppData\\Local\\Temp\\2\\tmpix2x74us'
```

zelinh:
According to my research, the `NamedTemporaryFile` is being accessed by `python.exe`. 
In our implementation, we are creating `self.stdout` in the `process.py` class and we are opening the file again with `open()` to read the data. 
[...]

PR #3811, opened 2023-07-25
Fix integ tests for windows platform

zelinh:
### Description
Fixing the current error of unable to remove temp files when terminating the process and also adding `UTF-8` encoding when executing the process for results record purpose.

According to the [module document](https://docs.python.org/3/library/tempfile.html) for `NamedTemporaryFile`, the temp file is open when we creating it and it may be supported to re-open it again by its name on Unix but not especially on windows platform. 

Also during my testing, the file was still open by python process even after we closed it by the file name. 
`item is popenfile(path='C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\2\\tmp2mlahvrq', fd=-1) and proc is psutil.Process(pid=6168, name='python.exe', status='running', started='23:49:13')`. Adding the for loop to check if it's closed before we remove it from the system will avoid `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\2\\tmpix2x74us'` which caused the issue before.

When the process completed, our current `test-recorder` is not able to record the `stdout.txt` for the dashboards on windows because it uses `cp1252` for encoding by default and we would get `UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 1357: character maps to <undefined>`. I am adding `utf-8` encoding to the subprocess execution to resolve this.

Sample running tests:
Running on dashboards: `./test.sh integ-test ./manifests/2.8.0/opensearch-dashboards-2.8.0-test.yml -p opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.8.0/7935/windows/x64/zip opensearch-dashboards=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch-dashboards/2.8.0/6182/windows/x64/zip --test-run-id 1234 --component alertingDashboards`
```
2023-07-24 23:22:58 INFO     | alertingDashboards   | with-security        | PASS  |
2023-07-24 23:22:58 INFO     | alertingDashboards   | without-security     | PASS  |
```

Running for OpenSearch: `./test.sh integ-test ./manifests/2.8.0/opensearch-2.8.0-test.yml -p opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.8.0/7935/windows/x64/zip  --component alerting --test-run-id 123`
```
2023-07-25 00:17:21 INFO     | alerting             | with-security        | PASS  |
2023-07-25 00:17:21 INFO     | alerting             | without-security     | PASS  |
```

I have also run integ tests on our linux docker and the back compatibility of the changes is passing.
`[opensearch@a49844d86db9 opensearch-build]$ ./test.sh integ-test ./manifests/2.8.0/opensearch-dashboards-2.8.0-test.yml -p opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.8.0/7935/linux/x64/tar opensearch-dashboards=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch-dashboards/2.8.0/6182/linux/x64/tar --test-run-id 1236 --component mlCommonsDashboards`
```
2023-07-25 00:26:58 INFO     | mlCommonsDashboards  | with-security        | PASS  |
2023-07-25 00:26:58 INFO     | mlCommonsDashboards  | without-security     | PASS  |
```

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3196

### DIFF
--- a/src/system/execute.py
+++ b/src/system/execute.py
@@ -20,7 +20,7 @@ def execute(command: str, dir: str, capture: bool = True, raise_on_failure: bool
     :returns a tuple containing the exit code, stdout, and stderr.
     """
     logging.info(f'Executing "{command}" in {dir}')
-    result = subprocess.run(command, cwd=dir, shell=True, capture_output=capture, text=True)
+    result = subprocess.run(command, cwd=dir, shell=True, capture_output=capture, text=True, encoding='utf-8')
     if raise_on_failure:
         result.check_returncode()
     return result.returncode, result.stdout, result.stderr
--- a/src/system/process.py
+++ b/src/system/process.py
@@ -26,8 +26,8 @@ class Process:
         if self.started:
             raise ProcessStartedError(self.pid)
 
-        self.stdout = tempfile.NamedTemporaryFile(mode="r+", delete=False)
-        self.stderr = tempfile.NamedTemporaryFile(mode="r+", delete=False)
+        self.stdout = tempfile.NamedTemporaryFile(mode="r+", delete=False, encoding='utf-8')
+        self.stderr = tempfile.NamedTemporaryFile(mode="r+", delete=False, encoding='utf-8')
 
         self.require_sudo = require_sudo
 
@@ -58,15 +58,33 @@ class Process:
         logging.info(f"Process killed with exit code {self.process.returncode}")
 
         if self.stdout:
-            self.__stdout_data__ = open(self.stdout.name, 'r').read()
+            self.stdout.seek(0)
+            self.__stdout_data__ = self.stdout.read()
+            self.stdout.flush()
             self.stdout.close()
-            os.remove(self.stdout.name)
+            for proc in psutil.process_iter():
+                try:
+                    for item in proc.open_files():
+                        if self.stdout.name == item.path:
+                            logging.warning(f"stdout {item} is being used by process with this detail {proc}")
+                except Exception:
+                    pass
+            os.unlink(self.stdout.name)
             self.stdout = None
 
         if self.stderr:
-            self.__stderr_data__ = open(self.stderr.name, 'r').read()
+            self.stderr.seek(0)
+            self.__stderr_data__ = self.stderr.read()
+            self.stderr.flush()
             self.stderr.close()
-            os.remove(self.stderr.name)
+            for proc in psutil.process_iter():
+                try:
+                    for item in proc.open_files():
+                        if self.stderr.name == item.path:
+                            logging.warning(f"stderr {item} is being used by process with this detail {proc}")
+                except Exception:
+                    pass
+            os.unlink(self.stderr.name)
             self.stderr = None
 
         self.return_code = self.process.returncode
--- a/src/system/process.py
+++ b/src/system/process.py
@@ -66,9 +66,9 @@ class Process:
                 try:
                     for item in proc.open_files():
                         if self.stdout.name == item.path:
-                            logging.warning(f"stdout {item} is being used by process with this detail {proc}")
-                except Exception:
-                    pass
+                            raise Exception(f"stdout {item} is being used by process {proc}")
+                except Exception as err:
+                    logging.warning(f"{err.args}")
             os.unlink(self.stdout.name)
             self.stdout = None
 
@@ -81,9 +81,9 @@ class Process:
                 try:
                     for item in proc.open_files():
                         if self.stderr.name == item.path:
-                            logging.warning(f"stderr {item} is being used by process with this detail {proc}")
-                except Exception:
-                    pass
+                            raise Exception(f"stderr {item} is being used by process {proc}")
+                except Exception as err:
+                    logging.warning(f"{err.args}")
             os.unlink(self.stderr.name)
             self.stderr = None
 
--- a/src/system/process.py
+++ b/src/system/process.py
@@ -68,7 +68,7 @@ class Process:
                         if self.stdout.name == item.path:
                             raise Exception(f"stdout {item} is being used by process {proc}")
                 except Exception as err:
-                    logging.warning(f"{err.args}")
+                    logging.error(f"{err.args}")
             os.unlink(self.stdout.name)
             self.stdout = None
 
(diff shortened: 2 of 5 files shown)


src/system/process.py
```diff
@@ -58,15 +58,33 @@ def terminate(self) -> int:
         logging.info(f"Process killed with exit code {self.process.returncode}")
 
         if self.stdout:
-            self.__stdout_data__ = open(self.stdout.name, 'r').read()
+            self.stdout.seek(0)
+            self.__stdout_data__ = self.stdout.read()
+            self.stdout.flush()
             self.stdout.close()
-            os.remove(self.stdout.name)
+            for proc in psutil.process_iter():
```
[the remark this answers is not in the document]
zelinh: this would list out all running process in local machine to check if the file is successfully closed before removal.

src/system/process.py
```diff
@@ -58,15 +58,33 @@ def terminate(self) -> int:
...
+                try:
+                    for item in proc.open_files():
+                        if self.stdout.name == item.path:
+                            logging.warning(f"stdout {item} is being used by process with this detail {proc}")
+                except Exception:
+                    pass
```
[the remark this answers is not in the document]
zelinh: Yeah make sense. Let me add some more for exception handle.

src/system/process.py
```diff
@@ -58,15 +58,33 @@ def terminate(self) -> int:
         logging.info(f"Process killed with exit code {self.process.returncode}")
 
         if self.stdout:
-            self.__stdout_data__ = open(self.stdout.name, 'r').read()
+            self.stdout.seek(0)
+            self.__stdout_data__ = self.stdout.read()
+            self.stdout.flush()
```
peterzhuamazon, line 63: I can understand the flush but why do we need seek(0) here?

src/system/process.py
```diff
@@ -58,15 +58,33 @@ def terminate(self) -> int:
...
             self.stdout.close()
-            os.remove(self.stdout.name)
+            for proc in psutil.process_iter():
+                try:
+                    for item in proc.open_files():
+                        if self.stdout.name == item.path:
+                            raise Exception(f"stdout {item} is being used by process {proc}")
+                except Exception as err:
+                    logging.warning(f"{err.args}")
```
peterzhuamazon, line 71: I would treat this as info level or error level not warning.

src/system/process.py
```diff
@@ -58,15 +58,33 @@ def terminate(self) -> int:
...
+            self.stderr.flush()
             self.stderr.close()
-            os.remove(self.stderr.name)
+            for proc in psutil.process_iter():
+                try:
+                    for item in proc.open_files():
+                        if self.stderr.name == item.path:
+                            raise Exception(f"stderr {item} is being used by process {proc}")
+                except Exception as err:
+                    logging.warning(f"{err.args}")
```
peterzhuamazon, line 86: Same as above

src/system/execute.py
```diff
@@ -20,7 +20,7 @@ def execute(command: str, dir: str, capture: bool = True, raise_on_failure: bool
     :returns a tuple containing the exit code, stdout, and stderr.
     """
     logging.info(f'Executing "{command}" in {dir}')
-    result = subprocess.run(command, cwd=dir, shell=True, capture_output=capture, text=True)
+    result = subprocess.run(command, cwd=dir, shell=True, capture_output=capture, text=True, encoding='utf-8')
```
peterzhuamazon, line 23: Are we able to set the encoding globally instead of manually adding to each subprocess run?

peterzhuamazon (changes requested):
Hi @zelinh adding some comments here. 
Thanks.

src/system/process.py
zelinh: From my understanding, after we `write()` to the tempfile, basically the file handle would be at the end of the file. Then when we `read` from it, we would get nothing from the end. Therefore moving the file handle to the beginning position would enable us to read the complete file. 
Check this example as well. https://docs.python.org/3/library/tempfile.html#examples

src/system/process.py
zelinh: Make sense. Updated.

src/system/execute.py
zelinh: I think according to my research, the encoding would be default depending on the platform. Not sure whether we could set it globally somewhere.

peterzhuamazon: So does it affect the behavior on Linux?

zelinh: I don't think so. The default encoding for deb/linux should be `UTF-8` but not for windows.

peterzhuamazon: Thanks.

peterzhuamazon (approved):

src/system/process.py
```diff
@@ -58,15 +58,33 @@ def terminate(self) -> int:
...
+            self.stdout.flush()
             self.stdout.close()
-            os.remove(self.stdout.name)
+            for proc in psutil.process_iter():
+                try:
+                    for item in proc.open_files():
+                        if self.stdout.name == item.path:
+                            raise Exception(f"stdout {item} is being used by process {proc}")
+                except Exception as err:
+                    logging.error(f"{err.args}")
```
gaiksaya, line 71: Do we want to retry closing here if the proc is still using the file?

src/system/process.py
```diff
@@ -58,15 +58,33 @@ def terminate(self) -> int:
...
+            self.stderr.flush()
             self.stderr.close()
-            os.remove(self.stderr.name)
+            for proc in psutil.process_iter():
+                try:
+                    for item in proc.open_files():
+                        if self.stderr.name == item.path:
+                            raise Exception(f"stderr {item} is being used by process {proc}")
+                except Exception as err:
+                    logging.error(f"{err.args}")
```
gaiksaya, line 86: Same as above

src/system/process.py
zelinh: Not really. That's unexpected and we are supposed to get error I think.

gaiksaya (approved):

Issue #3524, opened 2023-05-16
[BUG] OpenSearch-Dashboards integTest does not uninstall OpenSearch after completion


peterzhuamazon:
Hi,

When running the OSD integTest through Python, both OS and OSD will be installed.
But when completing the test or failed the test, only OSD will be uninstalled.

The logic here should be finetuned, this is only for DEB/RPM, not for TAR due to how `uninstall` function is being written:
https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/distribution_rpm.py#L56
https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/distribution_deb.py#L54

Thanks.

peterzhuamazon:
We might able to retrieve the above service list and run the uninstall based on that.
In src/test_workflow/test_cluster.py

zelinh:
[Triage] We have enough information here. @Divyaasm Please help take a look here. Thanks.

PR #3823, opened 2023-07-27
 Uninstall OpenSearch after testing on dashboards and modify the exception handle to skip irrelevant exception

zelinh:
### Description
Uninstall OpenSearch after testing on dashboards and modify the exception handle to skip irrelevant exception.

I noticed our previous fix on the integ tests would be logging multiple empty errors just like this. 
```
2023-07-26 22:49:42 ERROR    ()
2023-07-26 22:49:42 ERROR    ()
2023-07-26 22:49:42 INFO     Sending SIGKILL to PID 6410
2023-07-26 22:49:42 INFO     Process killed with exit code None
2023-07-26 22:49:42 ERROR    ()
2023-07-26 22:49:42 ERROR    ()
2023-07-26 22:49:42 ERROR    ()
2023-07-26 22:49:42 ERROR    ()
```

This is because when we terminate the process. We are iterating all the running process and see if the file is successfully closed before we remove it. 
However, we are catching multiple irrelevant exception and print out them with no more context. Some of this exception come from the race condition(e.g. the process ends before we go through its opened files) or some process we have no access on. We don't need to worry about these. Therefore according to my research, it's no harm to skip them regarding of our testing, and we would only need to log what we need under if statement `self.stderr.name == item.path`.

one of a use case on using `Process.open_files`
https://stackoverflow.com/questions/64599502/how-can-i-check-if-a-file-is-actively-open-in-another-process-outside-of-my-pyth

### Issues Resolved
closes https://github.com/opensearch-project/opensearch-build/issues/3524

### DIFF
--- a/src/system/process.py
+++ b/src/system/process.py
@@ -66,9 +66,9 @@ class Process:
                 try:
                     for item in proc.open_files():
                         if self.stdout.name == item.path:
-                            raise Exception(f"stdout {item} is being used by process {proc}")
-                except Exception as err:
-                    logging.error(f"{err.args}")
+                            logging.error(f"stdout {item} is being used by process {proc}")
+                except Exception:
+                    pass
             os.unlink(self.stdout.name)
             self.stdout = None
 
@@ -81,9 +81,9 @@ class Process:
                 try:
                     for item in proc.open_files():
                         if self.stderr.name == item.path:
-                            raise Exception(f"stderr {item} is being used by process {proc}")
-                except Exception as err:
-                    logging.error(f"{err.args}")
+                            logging.error(f"stderr {item} is being used by process {proc}")
+                except Exception:
+                    pass
             os.unlink(self.stderr.name)
             self.stderr = None
 
--- a/src/test_workflow/test_cluster.py
+++ b/src/test_workflow/test_cluster.py
@@ -87,6 +87,7 @@ class TestCluster(abc.ABC):
         for service in self.dependencies:
             termination_result = service.terminate()
             self.__save_test_result_data(termination_result)
+            service.uninstall()
 
         if not self.termination_result:
             raise ClusterServiceNotInitializedException()
--- a/src/test_workflow/test_cluster.py
+++ b/src/test_workflow/test_cluster.py
@@ -84,6 +84,7 @@ class TestCluster(abc.ABC):
         if self.service:
             self.termination_result = self.service.terminate()
             self.__save_test_result_data(self.termination_result)
+            self.service.uninstall()
 
         for service in self.dependencies:
             termination_result = service.terminate()
@@ -92,8 +93,6 @@ class TestCluster(abc.ABC):
 
         if not self.termination_result:
             raise ClusterServiceNotInitializedException()
-
-        self.service.uninstall()
 
     def __save_test_result_data(self, termination_result: ServiceTerminationResult) -> None:
         test_result_data = TestResultData(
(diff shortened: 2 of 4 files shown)


src/test_workflow/test_cluster.py
```diff
@@ -87,6 +87,7 @@ def terminate(self) -> None:
         for service in self.dependencies:
             termination_result = service.terminate()
             self.__save_test_result_data(termination_result)
+            service.uninstall()
```
peterzhuamazon, line 90: Could you make sure the uninstallation of dependencies wont affect `__save_test_result_data`?

src/system/process.py
```diff
@@ -66,9 +66,9 @@ def terminate(self) -> int:
                 try:
                     for item in proc.open_files():
```
gaiksaya, line 67: I believe this happened because we are checking all the process that is running on the host. The host has a number of processes running that may or may not be related to testing. Can you narrow down what process and open files we are checking?

src/system/process.py
```diff
@@ -66,9 +66,9 @@ def terminate(self) -> int:
...
                         if self.stdout.name == item.path:
-                            raise Exception(f"stdout {item} is being used by process {proc}")
-                except Exception as err:
-                    logging.error(f"{err.args}")
+                            logging.error(f"stdout {item} is being used by process {proc}")
+                except Exception:
+                    pass
```
gaiksaya, line 71: Catching and exception and not doing anything is not making sense. Can you elaborate what you are trying to do here?

src/system/process.py
zelinh: I think that's the properties of `psutil` to iterate all the running process though some of them are not related to our testing at all. I don't think there is a way to narrow down to specific process across platforms. Also some of the exception may be caused by race condition and we think we could pass on them since they are already outdated.

src/system/process.py
zelinh: These exceptions caught are mainly some process that were not related to our testing or they were already closed during the iteration so we may not have permission to go through it. I have seen quite few examples of using this regarding of whether certain file is closed/not used by any process. e.g. https://stackoverflow.com/questions/64599502/how-can-i-check-if-a-file-is-actively-open-in-another-process-outside-of-my-pyth. I think it's safe to do this.

src/system/process.py
gaiksaya: You can get the process id from line 48. See if the process is subprocess of it? Rather than iterating over all the existing processes?

zelinh: @gaiksaya Basically in windows, according to my research, the `python.exe` was the process that was holding the file from removing which is not the `OpenSearch/Dashboards` process so that process id is not the one we want to check.

peterzhuamazon: The pid thing needs another research and bug fix as it was initially only designed for tar, not any other dist that can be handled by external means.

As of now the pid is not accurate unless is tar.

gaiksaya (approved):

src/test_workflow/test_cluster.py
```diff
@@ -83,16 +83,16 @@ def start(self) -> None:
...
 
         if not self.termination_result:
             raise ClusterServiceNotInitializedException()
 
-        self.__save_test_result_data(self.termination_result)
-
         self.service.uninstall()
```
peterzhuamazon, line 96: Could you move this to the `if self.service` as well so uninstall is handled in the correct order?
Thanks.

peterzhuamazon (approved):

PR #3834, opened 2023-08-02
Do not fail the bundle build on plugin component failure

rishabh6788:
### Description
In the current scenario in case of any plugin build failure the complete distribution build is discarded and it keeps failing till the component owner fixes it. This can take from a few days to weeks blocking everyone else to continue with their testing and development using bundle build with their latest changes.
This PR adds support to continue the distribution build in case of any plugin failure so that a usable bundle artifact is generated with successful plugins installed.

### Issues Resolved

### DIFF
--- a/src/build_workflow/build_args.py
+++ b/src/build_workflow/build_args.py
@@ -26,6 +26,7 @@ class BuildArgs:
     platform: str
     architecture: str
     distribution: str
+    continue_on_error: bool
 
     def __init__(self) -> None:
         parser = argparse.ArgumentParser(description="Build an OpenSearch Distribution")
@@ -95,6 +96,13 @@ class BuildArgs:
             default="tar",
             dest="distribution"
         )
+        parser.add_argument(
+            "--continue-on-error",
+            dest="continue_on_error",
+            default=False,
+            action="store_true",
+            help="Do not fail the distribution build on any plugin component failure.",
+        )
 
         args = parser.parse_args()
         self.logging_level = args.logging_level
@@ -107,6 +115,7 @@ class BuildArgs:
         self.architecture = args.architecture
         self.distribution = args.distribution
         self.script_path = sys.argv[0].replace("/src/run_build.py", "/build.sh")
+        self.continue_on_error = args.continue_on_error
 
     def component_command(self, name: str) -> str:
         return " ".join(
--- a/src/run_build.py
+++ b/src/run_build.py
@@ -24,6 +24,7 @@ def main() -> int:
     args = BuildArgs()
     console.configure(level=args.logging_level)
     manifest = InputManifest.from_file(args.manifest)
+    failed_plugins = []
 
     if args.ref_manifest:
         manifest = manifest.stable()
@@ -70,10 +71,14 @@ def main() -> int:
                 logging.info(f"Successfully built {component.name}")
             except:
                 logging.error(f"Error building {component.name}, retry with: {args.component_command(component.name)}")
-                raise
+                if args.continue_on_error and component.name not in ['OpenSearch', 'OpenSearch Dashboards']:
+                    failed_plugins.append(component.name)
+                    continue
+                else:
+                    raise
 
         build_recorder.write_manifest()
-
+    logging.info(f"Failed plugins are {failed_plugins}")
     logging.info("Done.")
     return 0
 
--- a/src/run_build.py
+++ b/src/run_build.py
@@ -71,14 +71,15 @@ def main() -> int:
                 logging.info(f"Successfully built {component.name}")
             except:
                 logging.error(f"Error building {component.name}, retry with: {args.component_command(component.name)}")
-                if args.continue_on_error and component.name not in ['OpenSearch', 'OpenSearch Dashboards']:
+                if args.continue_on_error and component.name not in ['OpenSearch', 'OpenSearch-Dashboards']:
                     failed_plugins.append(component.name)
                     continue
                 else:
                     raise
 
         build_recorder.write_manifest()
-    logging.info(f"Failed plugins are {failed_plugins}")
+    if len(failed_plugins) > 0:
+        logging.error(f"Failed plugins are {failed_plugins}")
     logging.info("Done.")
     return 0
 
--- a/tests/test_run_build.py
+++ b/tests/test_run_build.py
@@ -38,6 +38,7 @@ class TestRunBuild(unittest.TestCase):
 
     OPENSEARCH_MANIFEST = os.path.realpath(os.path.join(MANIFESTS, "templates", "opensearch", "1.x", "os-template-1.1.0.yml"))
     OPENSEARCH_MANIFEST_1_2 = os.path.realpath(os.path.join(MANIFESTS, "templates", "opensearch", "1.x", "os-template-1.2.0.yml"))
+    NON_OPENSEARCH_MANIFEST = os.path.realpath(os.path.join(MANIFESTS, "templates", "opensearch", "1.x", "non-os-template-1.1.0.yml"))
 
     @patch("argparse._sys.argv", ["run_build.py", OPENSEARCH_MANIFEST, "-p", "linux"])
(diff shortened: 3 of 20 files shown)


rishabh6788:
Logs and results from local run for OpenSearch-3.0.0 manifest:

Failed plugins:
```
2023-08-02 01:38:52 INFO     Failed plugins are ['opensearch-reports', 'performance-analyzer', 'opensearch-observability', 'k-NN', 'neural-search', 'geospatial', 'anomaly-detection', 'cross-cluster-replication', 'sql']
2023-08-02 01:38:52 INFO     Done.
```

Successfully built plugins in `tar/builds/opensearch/plugins` directory:
```
[opensearch@e1e34ce2625f plugins]$ ls -l
total 346052
-rw-rw-r-- 1 opensearch opensearch   8205015 Aug  2 01:38 opensearch-alerting-3.0.0.0.zip
-rw-rw-r-- 1 opensearch opensearch   3276536 Aug  2 01:34 opensearch-job-scheduler-3.0.0.0.zip
-rw-rw-r-- 1 opensearch opensearch 287440521 Aug  2 01:33 opensearch-ml-3.0.0.0.zip
-rw-rw-r-- 1 opensearch opensearch   2581320 Aug  2 01:37 opensearch-notifications-3.0.0.0.zip
-rw-rw-r-- 1 opensearch opensearch  12180463 Aug  2 01:37 opensearch-notifications-core-3.0.0.0.zip
-rw-rw-r-- 1 opensearch opensearch  40659900 Aug  2 01:35 opensearch-security-3.0.0.0.zip
```

Attached is the generated build manifest. 
[local-test-manifest.yml.zip](https://github.com/opensearch-project/opensearch-build/files/12235212/local-test-manifest.yml.zip)
@prudhvigodithi @peterzhuamazon @dblock @gaiksaya

src/build_workflow/build_args.py
```diff
@@ -95,6 +96,13 @@ def __init__(self) -> None:
             default="tar",
             dest="distribution"
         )
+        parser.add_argument(
+            "--continue-on-error",
+            dest="continue_on_error",
+            default=False,
+            action="store_true",
+            help="Do not fail the distribution build on any plugin component failure.",
```
gaiksaya, line 104:
```suggestion
            help="Do not fail the distribution build on any component failure.",
```

src/run_build.py
```diff
@@ -70,10 +71,15 @@ def main() -> int:
...
+                if args.continue_on_error and component.name not in ['OpenSearch', 'OpenSearch Dashboards']:
+                    failed_plugins.append(component.name)
+                    continue
+                else:
+                    raise
 
         build_recorder.write_manifest()
-
+    if len(failed_plugins) > 0:
+        logging.error(f"Failed plugins are {failed_plugins}")
```
gaiksaya, line 82: If we error this out the exit code will be a non-zero value right? We still want the build to pass (with exit code as 0). I believe this should be `logging.info` ?

gaiksaya (changes requested):
Can you try to add test case asserting the failed components are logged and exit code is 0 if `--continue-on-error`  is true?

Thanks!

src/run_build.py
```diff
@@ -70,10 +71,15 @@ def main() -> int:
                 logging.info(f"Successfully built {component.name}")
             except:
                 logging.error(f"Error building {component.name}, retry with: {args.component_command(component.name)}")
-                raise
+                if args.continue_on_error and component.name not in ['OpenSearch', 'OpenSearch Dashboards']:
```
peterzhuamazon, line 74: Is the component name `OpenSearch Dashboards` or OpenSearch-Dashboards`?

src/run_build.py
peterzhuamazon: I would let the Python code to fail but use the Jenkins workflow to decide whether it should fail the run.

The python code define as `continue on error` means it is `error`, just keep going, but we should not treat it as success.

peterzhuamazon (commented):
Thanks @rishabh6788 add some comments here.

src/run_build.py
gaiksaya: Its difficult to mark something error-ed out as not an error on CI level. The user wanted to build everything in manifest, it did with some failing components. I understand its a partial failure but if we want hard failure we can always omit the `--continue-on-error` option.
WDYT @prudhvigodithi @dblock?

src/build_workflow/build_args.py
rishabh6788: any component would include OpenSearch and OSD as well which is not the case. We need to fail the build the if core component fails.

src/run_build.py
rishabh6788: It is just logging level and doesn't exit with code 1 but 0. 
Agree with @gaiksaya , if you want to fail the build on component failure then skip the optional flag.

src/run_build.py
rishabh6788: fixed it to `OpenSearch-Dashboards`.

rishabh6788:
@peterzhuamazon I am of the opinion is if we are using `--continue-on-error` then we want to have a successful build at the end. 
We can handle the logic of failing the jenkins job based on the presence of `2023-08-02 18:22:36 ERROR    Failed plugins are []` string or can set an env variable in the python code for jenkins job to make decision. 
Open to other suggestions. 
@peterzhuamazon

rishabh6788:
> Can you try to add test case asserting the failed components are logged and exit code is 0 if `--continue-on-error` is true?
> 
> Thanks!

Will take a look at this, given it is working as expected in local testing would not want this to be blocker.

src/run_build.py
rishabh6788: Even with `continue-on-error` we will be logging the exception message which will make sure our github auto-create issue workflow would still cut issues for failing components.
The failing components will be logged.  @bbarani

rishabh6788:
> Can you try to add test case asserting the failed components are logged and exit code is 0 if `--continue-on-error` is true?
> 
> Thanks!

Added new tests. @gaiksaya

tests/test_run_build.py
```diff
@@ -160,3 +161,45 @@ def test_main_manifest_new_lock_with_overrides(self, mock_logging: Mock, mock_to
...
+    @patch("run_build.TemporaryDirectory")
+    @patch("run_build.logging.error")
+    def test_fail_plugins_continue_on_error(self, mock_logging_error: Mock, mock_temp: Mock, mock_recorder: Mock, mock_builder_from: Mock, *mocks: Any) -> None:
+        mock_temp.return_value.__enter__.return_value.name = tempfile.gettempdir()
+        mock_builder = Mock()
+        mock_builder.build.side_effect = Exception("Error during build")
+        mock_builder_from.return_value = mock_builder
+
+        main()
+        mock_logging_error.assert_called_with("Failed plugins are ['common-utils', 'job-scheduler', 'sql', 'alerting']")
```
gaiksaya, line 191: So this is failing but OpenSearch is not a part of the manifest?

rishabh6788: Yes, the test logic picks the first component from the manifest and mocks an exception, had to remove OS to test behavior is the component is a plugin.

peterzhuamazon (approved):

tests/test_run_build.py
gaiksaya: But in case of core failure we want the entire build to fail right?

rishabh6788: Yes `def test_fail_core_component_continue_on_error` tests exception is raised, not just caught.

gaiksaya (approved):

src/run_build.py
```diff
@@ -24,6 +24,7 @@ def main() -> int:
     args = BuildArgs()
     console.configure(level=args.logging_level)
     manifest = InputManifest.from_file(args.manifest)
+    failed_plugins = []
```
dblock, line 27: Should have been called `failed_components` to be consistent.

dblock (commented):
Have a naming nit, better late than never ;)

Issue #1274, opened 2021-12-08
[META] After tests are complete there should be a test-run manifest that allows for finding all related test artifacts


peternied:
### Problem
After integration tests are run there is alot of information on the standard out and a final summary message that looks like:
```
2021-11-22 22:39:29 INFO     | alerting             | with-security        | PASS  |    0 |
2021-11-22 22:39:29 INFO     | alerting             | without-security     | PASS  |    0 |
2021-11-22 22:39:29 INFO     | anomaly-detection    | with-security        | PASS  |    0 |
2021-11-22 22:39:29 INFO     | anomaly-detection    | without-security     | PASS  |    0 |
2021-11-22 22:39:29 INFO     | asynchronous-search  | with-security        | PASS  |    0 |
2021-11-22 22:39:29 INFO     | asynchronous-search  | without-security     | PASS  |    0 |
2021-11-22 22:39:29 INFO     | dashboards-reports   | without-security     | PASS  |    0 |
2021-11-22 22:39:29 INFO     | index-management     | with-security        | PASS  |    0 |
2021-11-22 22:39:29 INFO     | index-management     | without-security     | PASS  |    0 |
2021-11-22 22:39:29 INFO     | k-NN                 | with-security        | PASS  |    0 |
2021-11-22 22:39:29 INFO     | k-NN                 | without-security     | PASS  |    0 |
2021-11-22 22:39:29 INFO     | sql                  | with-security        | PASS  |    0 |
2021-11-22 22:39:29 INFO     | sql                  | without-security     | PASS  |    0 |
```

This is useful for humans, but harder for our automation systems to integrate with.  If we are going to automatically file bugs or analyze test results there is critical information missing such as how to look up service logs, or test result logs.

### Proposal 1: Create a test-run manifest / update component manifests

Rather than modify on this table based approach which is very useful for the happy path, I suggest we create a new test-run yml manifest.  The existing component manifest files would be augmented to include the hierarchy of all files under them for deep diving.

test-run manifest contents:
  - Exact command used to execute, for reproducabillity
  - Extracted details about the test circumstances
  - List of components (Same list as from [test-manifest](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/1.3.0/opensearch-1.3.0-test.yml#L2))
    - Name of the component
    -  List configurations ran against that component
      - Name of the configuration
      - Arguments to the configuration
      - Test component manifest with additional and all associated artifacts
[...]

PR #3864, opened 2023-08-10
Update the name to test-report and add README for report workflow

zelinh:
### Description
Update the name to test-report and add README for report workflow

### Issues Resolved
Part of #1274

### DIFF
--- a/src/report_workflow/README.md
+++ b/src/report_workflow/README.md
@@ -1,0 +1,24 @@
+#### Generate test-report manifest for each test.
+The input requires a mandatory `test-manifest-path`, `--artifact-paths`, `--base-path`, `--test-type` and `--test-run-id`, optional `--output-path` and `--component` to automatically generate the test-report manifest after testing.
+
+*Usage*
+```
+./report.sh <test-manifest-path> --artifact-paths opensearch=<...> opensearch-dashboards=<...> --test-run-id <...> --test-type integ-test --base-path <...>
+```
+e.g.
+```
+./report.sh manifests/2.9.0/opensearch-2.9.0-test.yml -p opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.9.0/8172/linux/x64/tar --test-run-id 5328 --test-type integ-test --base-path https://ci.opensearch.org/ci/dbc/integ-test/2.9.0/8172/linux/x64/tar
+```
+The following options are available.
+
+| name                 | description                                                                    |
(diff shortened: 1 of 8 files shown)


src/report_workflow/README.md
```diff
@@ -0,0 +1,24 @@
+#### Generate test-report manifest for each test.
+The input requires a mandatory `test-manifest-path`, `--artifact-paths`, `--base-path`, `--test-type` and `--test-run-id`, optional `--output-path` and `--component` to automatically generate the test-report manifest after testing.
```
gaiksaya, line 2: nit: Add `<required>` and `<optional>` in the below table so that all information related to params is in one place

src/report_workflow/README.md
```diff
@@ -0,0 +1,24 @@
...
+The following options are available.
+
+| name                 | description                                                                    |
+|----------------------|--------------------------------------------------------------------------------|
+| test-manifest-path   | Specify a test manifest path.                                                  |
+| -p, --artifact-paths | Artifact paths of distributions used for testing.                              |
+| --base-path          | Base paths of testing logs.                                                    |
+| --test-type          | Type of tests report generates on.                                             |
+| --output-path        | Specify the path location for the test-report manifest.                        |
+| --test-run-id        | Specify the unique execution id for the test.                                  |
```
gaiksaya, line 21: This test-run-id should be equal to the the test runs build number right?

src/report_workflow/README.md
zelinh: Added.

src/report_workflow/README.md
zelinh: Yes, this id will be used to retrieve the component yml.

src/report_workflow/README.md
gaiksaya: You can remove this from here and instead add more description what this workflow actually does. 
```suggestion
As the name specifies, the test report workflow helps to automatically generate a consolidated report of the tests run at distribution level along with commands to reproduce the error and associated failures.
These tests include integration tests. BWC tests.
```

src/report_workflow/README.md
gaiksaya: Can you specify that it should match the test-run-id from already run tests? Seems like you can enter anything which is not right

zelinh: Thanks. Just updated.

gaiksaya (approved):

PR #3876, opened 2023-08-11
Add support to select ec2 instance type for data nodes

rishabh6788:
### Description
Add support to provide EC2 instance type for data nodes while running nightly benchmark. 

### Issues Resolved

### DIFF
--- a/src/test_workflow/benchmark_test/benchmark_args.py
+++ b/src/test_workflow/benchmark_test/benchmark_args.py
@@ -33,6 +33,7 @@ class BenchmarkArgs:
     ml_node_storage: int
     jvm_sys_props: str
     additional_config: str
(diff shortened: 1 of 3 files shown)


jordarlu (approved):

peterzhuamazon (changes requested):
Hi @rishabh6788 ,

Just want to know if this is something we can change in jenkinsfile instead of python code?

The instance type is very specific to AWS, and I think it is more suitable for pipeline specific settings, instead of in the Python code. Thanks.

rishabh6788:
@peterzhuamazon This change needs to go into python code as it finally gets passed to the CDK as one of the context parameter. 
The code is anyway tightly coupled to AWS since we are already using AWS CDK to spin up the infra.

PR #3885, opened 2023-08-15
Added support for data-instance-type and capture-segment-replication-stat additional parameters

rishabh6788:
This PR:
1. Adds support to provide data node instance type and
2. Option to capture segment replication stats in nightly benchmark runs. 

### Description
_Describe what this change achieves._

### Issues Resolved


gaiksaya:
Hi @rishabh6788 ,

Wanted to know what are the side effects of enabling this parameter if underlying segRep feature is disabled. 
If both are tightly coupled, would recommend to add a check in the jenkinsFile to check if both are true.

rishabh6788:
Not too tightly coupled, in worst case there would not segrep related metric. 
We can take this optimization in future but at the moment this will not cause any job failure if enabled on non-segrep clusters. 
@gaiksaya

gaiksaya (approved):

gaiksaya:
> Not too tightly coupled, in worst case there would not segrep related metric. We can take this optimization in future but at the moment this will not cause any job failure if enabled on non-segrep clusters. @gaiksaya

Sounds good. Let's keep watch and add this functionality if required. Thanks!

PR #3886, opened 2023-08-15
Enable continue on error for distribution build jobs

gaiksaya:
### Description
This change adds below features:
- Enable continue on error for distribution build jobs for OS and OSD
- Mark build as `UNSTABLE` if there are failed plugins in the build
- Always search for build failures and create issues.

### Issues Resolved

### DIFF
--- a/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
+++ b/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
@@ -1,4 +1,4 @@
-lib = library(identifier: 'jenkins@5.4.1', retriever: modernSCM([
+lib = library(identifier: 'jenkins@5.5.0', retriever: modernSCM([
     $class: 'GitSCMSource',
     remote: 'https://github.com/opensearch-project/opensearch-build-libraries.git',
 ]))
@@ -73,6 +73,11 @@ pipeline {
         booleanParam(
             name: 'CREATE_GITHUB_ISSUE',
             description: 'To create a github issue for failing component or not.',
+            defaultValue: true
+        )
+        booleanParam(
+            name: 'CONTINUE_ON_ERROR',
+            description: 'Continue building the distribution even if a single or few component fails',
             defaultValue: true
         )
     }
@@ -161,7 +166,8 @@ pipeline {
                                 inputManifest: "manifests/${INPUT_MANIFEST}",
                                 platform: 'linux',
                                 architecture: 'x64',
-                                distribution: 'tar'
+                                distribution: 'tar',
+                                continueOnError: ${params.CONTINUE_ON_ERROR}
                             )
                             String buildManifestUrl = buildManifestObj.getUrl(JOB_NAME, BUILD_NUMBER)
                             String buildManifestUrlOpenSearch = [buildManifestObj.getArtifactRootUrl(JOB_NAME_OPENSEARCH, "latest"), "builds", "opensearch", "manifest.yml"].join("/")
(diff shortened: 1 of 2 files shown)


jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -75,6 +75,11 @@ pipeline {
             description: 'To create a github issue for failing component or not.',
             defaultValue: true
         )
+        booleanParam(
+            name: 'CONTINUE_ON_ERROR',
+            description: 'Continue building the distribution even if a single or few component fails',
```
peterzhuamazon, line 80: `a single or few` to `one or more`?

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -161,7 +166,8 @@ pipeline {
                                 inputManifest: "manifests/${INPUT_MANIFEST}",
                                 platform: 'linux',
                                 architecture: 'x64',
-                                distribution: 'tar'
+                                distribution: 'tar',
+                                continueOnError: ${params.CONTINUE_ON_ERROR}
```
peterzhuamazon, line 170: I think you either just use params.<> or wrap double quotes around it?
I could be mistaken here tho.

peterzhuamazon: Assuming test both true and false on staging jenkins already?

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -826,11 +833,21 @@ pipeline {
             node(AGENT_X64) {
                 checkout scm
                 script {
+                    def findFailedPlugins = buildMessage(search: 'Failed plugins')
```
peterzhuamazon, line 836: Is there a better way than search for this text?
e.g. a flag set by the lib or similar?
Or just return code of the `build.sh` run in the python code through jenkins lib?

peterzhuamazon (commented):
Hi @gaiksaya add some comments here.

Thanks.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
gaiksaya: Yes tested on both. Works as expected!

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
gaiksaya: Not a better way! Python code returns 0 as exit code and we want to flag is as unstable.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
gaiksaya: Fixed it! Thanks!

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
peterzhuamazon: I am not very convinced by this method.
But giving the circumstances I can approve it for now.
We need to improve this to some degree later.

Thanks.

peterzhuamazon (approved):

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
gaiksaya: You were right! It needs to be either of you mentioned. Fixed here https://github.com/opensearch-project/opensearch-build/pull/3888

PR #3898, opened 2023-08-17
Improve ownership and permissions of files in deb and rpm packages

### LINKED ISSUE #3815: Defaut config permission too relaxed in deb and rpm packages
After installing the Debian package, all users can read all configuration files:

```sh-session

$ ls -dl /etc/opensearch                    
drwxr-xr-x 9 opensearch opensearch 4096 25 juil. 15:57 /etc/opensearch/
$ ls -l /etc/opensearch/opensearch.yml 
-rw-r--r-- 1 opensearch opensearch 6503 25 juil. 15:57 /etc/opensearch/opensearch.yml
$ ls -dl /etc/opensearch/opensearch-security
drwxr-xr-x 2 opensearch opensearch 4096 25 juil. 15:57 /etc/opensearch/opensearch-security/
$ ls -l /etc/opensearch/opensearch-security 
total 76
-rw-r--r-- 1 opensearch opensearch    50 14 oct.   2022 action_groups.yml
-rw-r--r-- 1 opensearch opensearch  1973 14 oct.   2022 allowlist.yml
-rw-r--r-- 1 opensearch opensearch  2541 14 oct.   2022 audit.yml
-rw-r--r-- 1 opensearch opensearch 10063 14 oct.   2022 config.yml
-rw-r--r-- 1 opensearch opensearch  1689 14 oct.   2022 internal_users.yml
-rw-r--r-- 1 opensearch opensearch   154 14 oct.   2022 nodes_dn.yml
-rw-r--r-- 1 opensearch opensearch 12381 14 oct.   2022 opensearch.yml.example
-rw-r--r-- 1 opensearch opensearch   844 14 oct.   2022 roles_mapping.yml
-rw-r--r-- 1 opensearch opensearch 12649 14 oct.   2022 roles.yml
-rw-r--r-- 1 opensearch opensearch   170 14 oct.   2022 tenants.yml
-rw-r--r-- 1 opensearch opensearch  1973 14 oct.   2022 whitelist.yml
```

Some of these files contain or can contain sensitive data, for example:
* `internal_users.yml` contain hashed password;
* `config.yml` can contain clear-text credential for authentication against an LDAP directory.

Upon initial investigation, this is the result of this `chmod` on the whole tree of files installed by the package:

https://github.com/opensearch-project/opensearch-build/blob/b422ae0bfddd590cea5041dbe4e44e642c0b8d64/scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh#L44

It looks like the tarball of OpenSearch has different and more restrictive permissions by default
| Filename | Debian package permission | Tarball permission |
|---|---:|---:|
| `config` (read `/etc/opensearch` with the Debian package) | 755 | 755 |
| `config/opensearch.yml` | 644 | 640 |
| `config/opensearch-security` | 755 | 750 |
| `config/opensearch-security/*.yml` | 644 | 640 |

I am opening this issue in order to discuss this packaging issue and fix it.  IMHO, the tarball default permissions are better than the Debian ones, but there is still room for improvement such as changing files ownership to root:opensearch in order to prevent a compromised service from rewriting it's configuration (and yes, currently all the files installed by the package are owned by the opensearch user so he can overwrite all the application code, but let's tackle this in another PR).

What do you think?

smortex:
### Description

When installing OpenSearch from .deb or .rpm packages, the files permissions are unexpectedly relaxed allowing all local users to access all files from the OpenSearch, and allowing OpenSearch itself to update any file installed in the package.

This PR adjust the permissions so that the configuration, logs and data cannot be accessed by random users, and the application code cannot be updated by the opensearch user.

### Issues Resolved

Fixes #3815

### DIFF
--- a/scripts/pkg/build_templates/opensearch/deb/debian/postinst
+++ b/scripts/pkg/build_templates/opensearch/deb/debian/postinst
@@ -38,7 +38,6 @@ if ! grep -q '## OpenSearch Performance Analyzer' ${config_dir}/jvm.options; the
 fi
 
 # Set owner
-chown -R opensearch.opensearch ${product_dir}
 chown -R opensearch.opensearch ${config_dir}
 chown -R opensearch.opensearch ${log_dir}
 chown -R opensearch.opensearch ${data_dir}
--- a/scripts/pkg/build_templates/opensearch/deb/debian/postinst
+++ b/scripts/pkg/build_templates/opensearch/deb/debian/postinst
@@ -40,7 +40,7 @@ fi
 # Set ownership and permissions
 # FIXME: the opensearch service should not have w permission in the config directory
 chown -R opensearch.opensearch ${config_dir}
-chmod -R o-rx ${config_dir}
+chmod -R u=rwX,g=rX,o= ${config_dir}
 
 chown -R opensearch.adm ${log_dir}
 chmod 750 ${log_dir}
--- a/scripts/pkg/build_templates/opensearch/deb/debian/postinst
+++ b/scripts/pkg/build_templates/opensearch/deb/debian/postinst
@@ -46,6 +46,8 @@ chown -R opensearch.adm ${log_dir}
 chmod 750 ${log_dir}
 
 chown -R opensearch.opensearch ${data_dir}
+chmod 750 ${data_dir}
+
 chown -R opensearch.opensearch ${pid_dir}
 
 # Reload systemctl daemon
--- a/scripts/pkg/build_templates/opensearch/deb/debian/postinst
+++ b/scripts/pkg/build_templates/opensearch/deb/debian/postinst
@@ -42,7 +42,9 @@ fi
 chown -R opensearch.opensearch ${config_dir}
 chmod -R o-rx ${config_dir}
 
-chown -R opensearch.opensearch ${log_dir}
+chown -R opensearch.adm ${log_dir}
+chmod 750 ${log_dir}
+
 chown -R opensearch.opensearch ${data_dir}
 chown -R opensearch.opensearch ${pid_dir}
 
--- a/scripts/pkg/build_templates/opensearch/deb/debian/postinst
+++ b/scripts/pkg/build_templates/opensearch/deb/debian/postinst
@@ -37,8 +37,11 @@ if ! grep -q '## OpenSearch Performance Analyzer' ${config_dir}/jvm.options; the
    echo "--add-opens=jdk.attach/sun.tools.attach=ALL-UNNAMED" >> ${config_dir}/jvm.options
 fi
 
-# Set owner
+# Set ownership and permissions
+# FIXME: the opensearch service should not have w permission in the config directory
 chown -R opensearch.opensearch ${config_dir}
+chmod -R o-rx ${config_dir}
+
 chown -R opensearch.opensearch ${log_dir}
 chown -R opensearch.opensearch ${data_dir}
 chown -R opensearch.opensearch ${pid_dir}
--- a/scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh
+++ b/scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh
@@ -41,6 +41,6 @@ ln -s ${data_dir} ${buildroot}${product_dir}/data
 ln -s ${log_dir}  ${buildroot}${product_dir}/logs
 
 # Change Permissions
-chmod -Rf a+rX,u+w,g-w,o-w ${buildroot}/*
+chmod -Rf u=rwX,g=rX,o=rX ${buildroot}/*
 
 exit 0
--- a/scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh
+++ b/scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh
@@ -41,6 +41,7 @@ ln -s ${data_dir} ${buildroot}${product_dir}/data
 ln -s ${log_dir}  ${buildroot}${product_dir}/logs
 
 # Change Permissions
+chmod -Rf g-s ${buildroot}/*
 chmod -Rf u=rwX,g=rX,o=rX ${buildroot}/*
 
 exit 0
--- a/scripts/pkg/build_templates/opensearch/rpm/opensearch.rpm.spec
+++ b/scripts/pkg/build_templates/opensearch/rpm/opensearch.rpm.spec
@@ -68,7 +68,8 @@ if [ ! -f %{buildroot}%{data_dir}/performance_analyzer_enabled.conf ]; then
     echo 'true' > %{buildroot}%{data_dir}/performance_analyzer_enabled.conf
 fi
 # Change Permissions
-chmod -Rf a+rX,u+w,g-w,o-w %{buildroot}/*
+chmod -Rf g-s %{buildroot}/*
+chmod -Rf u=rwX,g=rX,o= %{buildroot}/etc
 exit 0
 
 %pre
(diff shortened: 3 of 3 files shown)


peterzhuamazon:
In deb we also set the initial permissions here:
https://github.com/opensearch-project/opensearch-build/blob/main/scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh#L44

scripts/pkg/build_templates/opensearch/deb/debian/postinst
```diff
@@ -37,11 +37,17 @@ if ! grep -q '## OpenSearch Performance Analyzer' ${config_dir}/jvm.options; the
    echo "--add-opens=jdk.attach/sun.tools.attach=ALL-UNNAMED" >> ${config_dir}/jvm.options
 fi
 
-# Set owner
-chown -R opensearch.opensearch ${product_dir}
+# Set ownership and permissions
+# FIXME: the opensearch service should not have w permission in the config directory
```
peterzhuamazon, line 41: Write permission has to be there for the security demo script to run and make changes.

peterzhuamazon: I see you keep the write for others now.

scripts/pkg/build_templates/opensearch/rpm/opensearch.rpm.spec
```diff
@@ -175,16 +169,27 @@ exit 0
...
 
+%dir %attr(750, %{name}, %{name}) %{data_dir}
+%dir %attr(750, %{name}, %{name}) %{log_dir}
+%dir %attr(750, %{name}, %{name}) %{pid_dir}
+
+# Permissions
+%defattr(-, root, root)
+
+# Root dirs/docs/licenses
+%dir %{product_dir}
```
peterzhuamazon, line 180: So we dont want opensearch:opensearch to own product_dir?

scripts/pkg/build_templates/opensearch/rpm/opensearch.rpm.spec
```diff
@@ -175,16 +169,27 @@ exit 0
 %attr(0644, root, root) %config(noreplace) %{_prefix}/lib/sysctl.d/%{name}.conf
 %attr(0644, root, root) %config(noreplace) %{_prefix}/lib/tmpfiles.d/%{name}.conf
 
+%dir %attr(750, %{name}, %{name}) %{data_dir}
+%dir %attr(750, %{name}, %{name}) %{log_dir}
+%dir %attr(750, %{name}, %{name}) %{pid_dir}
```
peterzhuamazon, line 174: I think I initially put log and pid dir without %dir is to prevent them to be cleaned up by rpm:
```
Identifies that the path is a directory that should be owned by this RPM. This is important so that the RPM file manifest accurately knows what directories to clean up on uninstall. 
```

I dont remember why I dont do the same for data.

peterzhuamazon (commented):
Hi @smortex thanks for the contribution here.

I have some comments and once we clear these, I would need to run a full-ledge build on ci images with this code, to ensure it is running without issues.

Thanks.

scripts/pkg/build_templates/opensearch/deb/debian/postinst
```diff
@@ -37,11 +37,17 @@ if ! grep -q '## OpenSearch Performance Analyzer' ${config_dir}/jvm.options; the
    echo "--add-opens=jdk.attach/sun.tools.attach=ALL-UNNAMED" >> ${config_dir}/jvm.options
 fi
 
-# Set owner
-chown -R opensearch.opensearch ${product_dir}
+# Set ownership and permissions
```
peterzhuamazon, line 40: Need to make some changes here as well:
 https://github.com/opensearch-project/opensearch-build/blob/main/scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh#L44

scripts/pkg/build_templates/opensearch/deb/debian/postinst
smortex: I recursively remove the r and x bit for others, so the onwer has rwx permission and the group has rx permission.  Ideally, the files should be owner by root so that the service can not rewrite them, but on first start the service want to create `opensearch.keystore` so for now I left the owner to `opensearch:opensearch`.

scripts/pkg/build_templates/opensearch/rpm/opensearch.rpm.spec
smortex: We need to dig into this: I am not used to rpm and with or without `%dir` I see the same behavior: directory is not removed if new files where added into it.

We indeed do not want to remove the data dir when uninstalling.

scripts/pkg/build_templates/opensearch/rpm/opensearch.rpm.spec
smortex: Correct.  Just like the rest of the system, the user running the service should not need to (be able to) overwrite its files.  The default ownership `root:root` make sense.

scripts/pkg/build_templates/opensearch/deb/debian/postinst
smortex: This is supposed to be noop, but for consistency I added a commit to express the permissions explicitly rather than adding and removing permissions from a default that is implied.

peterzhuamazon: Got it, thanks.

peterzhuamazon:
Thanks @smortex, once you address my last comment / open conversation, we can queue this up for merging.

Bare in mind we might have a patch release before 2.10.0, so we might keep this PR open a bit until we enter the 2.10.0 release cycle, since there are a lot of essential changes on both deb/rpm.

I will start the test on my side once all the existing conversation addressed.

Thanks.

scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh
```diff
@@ -41,6 +41,6 @@ ln -s ${data_dir} ${buildroot}${product_dir}/data
 ln -s ${log_dir}  ${buildroot}${product_dir}/logs
 
 # Change Permissions
-chmod -Rf a+rX,u+w,g-w,o-w ${buildroot}/*
+chmod -Rf u=rwX,g=rX,o=rX ${buildroot}/*
```
peterzhuamazon, line 44: Do we want to have this line here as well, just for the sake of consistency?
```
chmod -Rf g-s %{buildroot}/*
```

scripts/pkg/build_templates/opensearch/rpm/opensearch.rpm.spec
smortex: Done!

scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh
smortex: From what I see, the packaging does not copy the SGID bit when installing, so I think we can skip it.

smortex: Ah, maybe when installing from rpm it is needed.  I cannot test ATM, but will be able next week.

peterzhuamazon:
Hi @smortex ,

Add a comment and want to know if https://github.com/opensearch-project/OpenSearch/pull/9447 needs to go in before this?

Thanks.

scripts/pkg/build_templates/opensearch/deb/debmake_opensearch_install.sh
smortex: Yes, I am more used to Debian and the .deb does not set the SGID bit.  Either it is not copied in the package, or ignored when installing.  I'll add the explicit removal so that the .rpm package is in sync with the Debian one and with what I would expect for a package.

smortex:
@peterzhuamazon I added a commit to have the .deb and .rpm packaging doing the same stripping of the SGID bit.  I am good with this PR as it is now, no need to wait for https://github.com/opensearch-project/OpenSearch/pull/9447 to be merged IMHO as we have all required workarounds here.

When https://github.com/opensearch-project/OpenSearch/pull/9447 is merged, we can do the cleanup in this repo in a new PR.

peterzhuamazon (approved):
Thanks @smortex for the help and commit to fix this issue.

PR #3912, opened 2023-08-21
Add support to accept telemetry parameters

### LINKED ISSUE #3911: Add support to pass additional telemetry parameters for node-stats telemetry
For tracking remote store nightly runs, we need the changes to be done in opensearch-build repo to enable collection of node-stats/segments stats. The below is the change summarised -
`--telemetry="node-stats" --telemetry-params='{"node-stats-include-indices":"true"}'` in OSB run command.  

{"node-stats-include-indices":"true","node-stats-include-indices-metrics":"segments"} - parameters required.  

Also, we are seeing discrepancy in the Remote Store nightly runs where the throughput is showing almost 1/3rd of the expected throughput. [Link](https://benchmarks.opensearch.org/app/dashboards#/view/f4278450-15d2-11ee-a68c-1330e5a77b4c?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-7d,to:now))&_a=(description:'',filters:!(),fullScreenMode:!f,options:(hidePanelTitles:!f,useMargins:!t),query:(language:kuery,query:''),timeRestore:!f,title:'Remote%20store',viewMode:view))
Please help to get the above 2 points addressed (introduction of remote store metrics and fixing the remote store dashboard) as we need to track the performance characteristics and couple of remote store specific stats at the earliest to avoid any unintentional regression (by catching it from the nightly remote store dashboards).

rishabh6788:
### Description
This PR adds support to accept additional telemetry parameters to enable non-default settings when using `node-stats` or any other type of telemetry. 
The settings are only applied when any type of telemetry, e.g. `node-stats` is enabled, else it will be ignored. 

### Issues Resolved
#3911 

### DIFF
--- a/src/test_workflow/benchmark_test/benchmark_test_suite.py
+++ b/src/test_workflow/benchmark_test/benchmark_test_suite.py
@@ -35,23 +35,25 @@ class BenchmarkTestSuite:
         self.args = args
         # Pass the cluster endpoints with -t for multi-cluster use cases(e.g. cross-cluster-replication)
         self.command = 'docker run --rm'
-        if args.benchmark_config:
+        if self.args.benchmark_config:
             self.command += f" -v {args.benchmark_config}:/opensearch-benchmark/.benchmark/benchmark.ini"
         self.command += f" opensearchproject/opensearch-benchmark:latest execute-test --workload={self.args.workload} " \
                         f"--pipeline=benchmark-only --target-hosts={endpoint}"
 
-        if args.workload_params:
+        if self.args.workload_params:
             logging.info(f"Workload Params are {args.workload_params}")
             self.command += f" --workload-params '{args.workload_params}'"
 
-        if args.user_tag:
+        if self.args.user_tag:
             user_tag = f"--user-tag=\"{args.user_tag}\""
             self.command += f" {user_tag}"
 
-        if args.telemetry:
+        if self.args.telemetry:
             self.command += " --telemetry "
-            for value in args.telemetry:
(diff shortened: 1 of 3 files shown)


src/test_workflow/benchmark_test/benchmark_test_suite.py
```diff
@@ -35,23 +35,25 @@ def __init__(
...
             self.command += f" {user_tag}"
 
-        if args.telemetry:
+        if self.args.telemetry:
             self.command += " --telemetry "
-            for value in args.telemetry:
+            for value in self.args.telemetry:
                 self.command += f"{value},"
+            if self.args.telemetry_params:
+                self.command += f" --telemetry-params '{self.args.telemetry_params}'"
```
peterzhuamazon, line 56: A bit confused.

If I have value for `telemetry_params`, but didnt set `telementry` does that mean it wont take effect?

peterzhuamazon: Just realize `telementry` is actually coming from param `--capture-segment-replication-stat`, which is a bit confusing to start with.

peterzhuamazon (approved):

src/test_workflow/benchmark_test/benchmark_test_suite.py
rishabh6788: yeah, since there are different types of telemetry devices, the idea is to give an option to choose from them and internally we pass the selected items as a list.

PR #3918, opened 2023-08-23
Add documentation to generate the token

gaiksaya:
### Description
Add documentation to generate the token for 1-click release process. 

### Issues Resolved


peterzhuamazon (approved):
Note that `-30` can be change to other number as well, but it must be a number so limited lines will be retrieved as base.

Also, try to use `urandom` all the time instead of the `random` one as the former one is more robust, efficient, and random.

Thanks.

PR #3937, opened 2023-08-30
Add Jenkins job for Docker Patch jenkins shared library

### LINKED ISSUE #3930: Docker re-release Jenkins Job for the Docker Patch library
Create a new docker job to utilize the patchDockerImage() jenkins shared library. To re-release the latest docker images.

#3303

Divyaasm:
### Description
Add a Jenkins job in build library to call the docker patch functionality 
### Issues Resolved
#3930


jenkins/docker/docker-re-release.jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+        )
+        string(
+            name: 'TAG',
+            description: "Enter tag of the Product",
+            trim: true
+        )
+        booleanParam(
+            name: 'RE_RELEASE',
+            defaultValue: true,
+            description: "Forces to create a Docker-Copy if the Docker Build is successful"
```
gaiksaya, line 25: This description is not clear, can you please elaborate or be more clear? 
Thanks!

tests/jenkins/TestDockerReRelease.groovy
```diff
@@ -0,0 +1,93 @@
...
+                'tests/jenkins/jenkinsjob-regression-files/docker/docker-re-release.jenkinsfile')
+    }
+
+    @Test
+    void checkForTriggeredJobs(){
+        runScript('jenkins/docker/docker-re-release.jenkinsfile')
+        assertThat(getCommandExecutions('build', ''), hasItem('{job=docker-build, propagate=true, wait=true, parameters=[null, null, null]}'))
+        assertThat(getCommandExecutions('build', ''), hasItem('{job=docker-copy, propagate=true, wait=true, parameters=[null, null, null, null]}'))
+        assertThat(getCommandExecutions('build', ''), hasItem('{job=docker-scan, propagate=true, wait=true, parameters=[null]}'))
+        assertThat(getCommandExecutions('build', ''), hasItem('{job=docker-promotion, propagate=true, wait=true, parameters=[null, null, null]}'))
```
gaiksaya, line 73: Why are the parameters going `null` here? Shouldn't they be actual parameters?

jenkins/docker/docker-re-release.jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+    agent none
+    parameters {
+        string(
+            name: 'PRODUCT',
+            description: "OpenSearch / OpenSearch-Dashboards",
+            trim: true
+        )
+        string(
+            name: 'TAG',
+            description: "Enter tag of the Product",
```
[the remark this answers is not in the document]
Divyaasm: It's just a tag Sayali, The input would be 1 or 2 for example

tests/jenkins/TestDockerReRelease.groovy
Divyaasm: Not sure why the parameters are taken as null though the above lines shows the exact parameter values that are passed. This is the same case when the jenkins job is triggered inside the library checkout previous implementation [here](https://github.com/opensearch-project/opensearch-build-libraries/blob/560f671d486af6e78af1bb21bc7591c5388184d1/tests/jenkins/jobs/BuildDockerImage_Jenkinsfile_builds_both.txt#L15)

jenkins/docker/docker-re-release.jenkinsfile
```diff
@@ -0,0 +1,59 @@
+lib = library(identifier: 'jenkins@5.7.1', retriever: modernSCM([
+    $class: 'GitSCMSource',
+    remote: 'https://github.com/opensearch-project/opensearch-build-libraries.git',
+]))
+
+pipeline {
+    options {
+        timeout(time: 7, unit: 'HOURS')
```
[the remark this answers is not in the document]
Divyaasm: Had some confusion on this. Should 5 hrs be sufficient.

jenkins/docker/docker-re-release.jenkinsfile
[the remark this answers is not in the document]
gaiksaya: Can you add an example in the description to be more clear? Also Does this take care to update latest and stuff? Also how many tags can you pass?

tests/jenkins/TestDockerReRelease.groovy
gaiksaya: Can you try `addParam` instead of binding variable? Binding variables are used for mocking env variables mostly. https://github.com/jenkinsci/JenkinsPipelineUnit#mocking-jenkins-variables

jenkins/docker/docker-re-release.jenkinsfile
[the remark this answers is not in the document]
Divyaasm: Yes @gaiksaya, the latest is updated only when the corresponding version is latest, currently it's for 2.x. And it's completely taken care by the shared library. Currently tag and product are strings. I can change them to list. So we can pass multiple values at a time.

jenkins/docker/docker-re-release.jenkinsfile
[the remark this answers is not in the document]
peterzhuamazon: Why does it needs to be 5 hours?
Max this activity should be 2 hours isnt it?

The most time you take on this is rebuild, and the rebuild does not need more than 1 hour, if not 30min.

I am confused of how this 5hr number coming from.

jenkins/docker/docker-re-release.jenkinsfile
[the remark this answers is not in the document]
peterzhuamazon: @Divyaasm I would suggest use a choice param so people can choose the major version.

You did not give any example in the description and it will naturally confuse people.

jenkins/docker/docker-re-release.jenkinsfile
peterzhuamazon: Agree ^^

jenkins/docker/docker-re-release.jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+    }
+    agent none
+    parameters {
+        string(
+            name: 'PRODUCT',
```
peterzhuamazon, line 13: Also change this to choice param.

jenkins/docker/docker-re-release.jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+            description: "OpenSearch / OpenSearch-Dashboards",
+            trim: true
+        )
+        string(
+            name: 'TAG',
+            description: "Enter tag of the Product",
+            trim: true
+        )
+        booleanParam(
+            name: 'RE_RELEASE',
```
peterzhuamazon, line 23: Not sure the point of this param if the workflow itself is aimed for re-release already?
What is the point of rebuild but not re-release?
What does it force in this case?

peterzhuamazon (changes requested):
A lot of issues observed in this design.
Add some comments.

Thanks.

jenkins/docker/docker-re-release.jenkinsfile
Divyaasm: This parameter acts as a buildOption to create a docker-copy after the successful build inside the [buildDockerImage()](https://github.com/opensearch-project/opensearch-build-libraries/blob/560f671d486af6e78af1bb21bc7591c5388184d1/vars/buildDockerImage.groovy#L71) 
I will update the description.

jenkins/docker/docker-re-release.jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+    agent none
+    parameters {
+        choice(
+            name: 'PRODUCT',
+            choices: ['opensearch', 'opensearch-dashboards'],
+            description: "Choose the product type among OpenSearch / OpenSearch-Dashboards."
+        )
+        choice(
+            name: 'TAG',
+            choices: ['1', 'latest'],
```
peterzhuamazon, line 19: This should be 1 and 2, not latest.

jenkins/docker/docker-re-release.jenkinsfile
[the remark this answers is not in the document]
peterzhuamazon: Yes @Divyaasm that is particularly set for the release ci image, as that one takes at least 4 hours.

jenkins/docker/docker-re-release.jenkinsfile
peterzhuamazon: So in which case you do not want to docker-copy?
If not for docker-copy after we can just use docker-build without this workflow tho.
I am still not really sure why we need this at all.

peterzhuamazon (changes requested):

jenkins/docker/docker-re-release.jenkinsfile
gaiksaya: @peterzhuamazon Can't this value be 2.9.0, 1.3.11?

Divyaasm: I think this workflow is intended to concentrate only on the latest versions. So tag should do the work and update the major tag and latest images. Or providing the latest version number in each branch be it 1.x or 2.x should do the same some job as per library implementation.

peterzhuamazon: You dont need to be 2.9.0/1.3.11 as we only patch the latest supported version.
And major 1/2 tag already tag to the latest version.

jenkins/docker/docker-re-release.jenkinsfile
```diff
@@ -16,14 +16,9 @@ pipeline {
         )
         choice(
             name: 'TAG',
-            choices: ['1', 'latest'],
+            choices: ['1', '2'],
```
peterzhuamazon, line 19: This should be going from latest to earliest, as if `2 1` instead of `1 2`.

peterzhuamazon (approved):
Thanks @Divyaasm ,

One more nitpik and we are good to go.

Thanks.

gaiksaya (approved):

PR #3952, opened 2023-09-01
Fix files ownership and permissions for Dashboards

smortex:
Similar to the issue fixed in #3898, OpenSearch-Dashboards package has
unexpected files owner and permissions.

This ensure the installed files are not owner by the
opensearch-dashboards user (preventing the program to overwrite itself
with malicious code if the service has some kind of vulnerability), and
make sure logs and data cannot be accessed by random users.

Signed-off-by: Romain Tartière <romain@blogreen.org>


smortex:
Cc @peterzhuamazon who worked on the OpenSearch part of this PR.

Endalaust (approved):
Approved. Well done.

peterzhuamazon:
Taking a look on it soon.

scripts/pkg/build_templates/opensearch-dashboards/rpm/opensearch-dashboards.rpm.spec
```diff
@@ -56,7 +56,8 @@ chmod 0755 %{buildroot}%{product_dir}/bin/*
 ln -s %{data_dir} %{buildroot}%{product_dir}/data
 ln -s %{log_dir}  %{buildroot}%{product_dir}/logs
 # Change Permissions
-chmod -Rf a+rX,u+w,g-w,o-w %{buildroot}/*
+chmod -Rf g-s %{buildroot}/*
+chmod -Rf u=rwX,g=rX,o= %{buildroot}/*
```
peterzhuamazon, line 60: Is this supposed to be on `%{buildroot}/*` or `%{buildroot}/etc`?
Because in #3898 it is the latter one only for rpm, but not for deb.

And in this PR it is matching between rpm and deb.


Also deb is o=rX while rpm is o=

Just want to make sure on here.

Thanks.

peterzhuamazon (changes requested):

scripts/pkg/build_templates/opensearch-dashboards/rpm/opensearch-dashboards.rpm.spec
smortex: > Is this supposed to be on `%{buildroot}/*` or `%{buildroot}/etc`?

Good catch!  I amended my commit to fix it. :+1: 

> Also deb is o=rX while rpm is o=

I think this is okay: when packaging the .deb, permissions are currently set in two places:

1. `hmod -Rf u=rwX,g=rX,o=rX ${buildroot}/*` in `scripts/pkg/build_templates/opensearch-dashboards/deb/debmake_opensearch_dashboards_install.sh` set the whole permissions to 755 / 644
2. `chmod -R u=rwX,g=rX,o= ${config_dir}` in `~/Projects/opensearch/opensearch-build/scripts/pkg/build_templates/opensearch-dashboards/deb/debian/postinst` set the config to 750 / 640 post-installation.

Arguably there is place for improvement but I will have a look at it later.

peterzhuamazon (approved):
Thanks @smortex, approved.

scripts/pkg/build_templates/opensearch-dashboards/rpm/opensearch-dashboards.rpm.spec
peterzhuamazon: Got it!
Thanks for fixing this one quick.

Issue #1563, opened 2022-01-27
[Proposal] Support AL2023/NodeJS18 for OpenSearch/Dashboards Releases


peterzhuamazon:
With the announcement of Amazon Linux 2023 Preview available for testing, we want to setup some notes here for the support of AL2023 once it is fully released to public.
https://aws.amazon.com/about-aws/whats-new/2021/11/preview-amazon-linux-2022/
https://aws.amazon.com/about-aws/whats-new/2023/03/amazon-linux-2023/

* Evaluate the compatibility of AL2023 when running OpenSearch/Dashboards
  * OpenSearch/Dashboards latest binary can run on AL2023 with the same expected behaviors on AL2
  * Working alongside plugin teams to ensure all the plugins are running as expected
  * Performance testing to ensure there is no performance regression
  * Evaluate the package/dependencies availability on AL2023 default repository
  * Evaluate the size of the image vs contents integrity
* Evaluate AL2023 availability
  * AL2023 available on docker for both x64 and arm64
  * AL2023 available on EC2 for both x64 and arm64 for testing
* Evaluate Build Process
  * What level of changes do we need to support AL2023 with current build process and scripts
  * Can we support both AL2 and AL2023 in the same system
  * Build tool related libs/packages/dependencies availability on AL2023
* Evaluate deprecation plan for older OS
  * When would we deprecate AL2 while supporting AL2023
  * Will this introduce breaking changes to customer
  * Would we support older version of OpenSearch/Dashboards with new OS?

We welcome more discussion on this.

Thanks.

peterzhuamazon:
Hi @nknize @dblock what do you think about this change?

From our teams perspective it makes sense to upgrade the main docker image from AL2 to AL2023 starting 3.0.0. It also make sense that we use this opportunities to upgrade other OS such as Ubuntu and Rockylinux to their latest versions.

Thanks.

dblock:
I really don't know enough about this @peterzhuamazon. What are pros/cons?

peterzhuamazon:
Now with this issue on the horizon, we need to also think about how to proceed with the changes:
* opensearch-project/opensearch-dashboards#3601 


This means:
* We have to deprecate ci image that uses centos7 or al2.
* Our ci and release docker images have to move from al2 to al2023
* #3351 
  * Python upgrade from 3.7 to 3.9 must follow as al2023 is default to newer version of python
[...]

PR #3967, opened 2023-09-06
[NodeJS18 Upgrade] Switch to AL2023 Runner and Rockylinux8 for NodeJS18 Upgrade

peterzhuamazon:
### Description
[NodeJS18 Upgrade] Switch to AL2023 Runner and Rockylinux8 for NodeJS18 Upgrade

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1563
https://github.com/opensearch-project/opensearch-build/issues/3743


peterzhuamazon:
The failure is not related, the issue is that 2.x is not 2.10.0 anymore but 2.11.0, as 2.10 just cut the branches now.

bbarani (approved):

rishabh6788 (approved):

jenkins/data-prepper/release-data-prepper-all-artifacts.jenkinsfile
```diff
@@ -25,7 +25,7 @@ pipeline {
         stage('Promote Archives') {
             agent {
                 docker {
-                    label 'Jenkins-Agent-AL2-X64-C54xlarge-Docker-Host'
+                    label 'Jenkins-Agent-AL2023-X64-C54xlarge-Docker-Host'
```
gaiksaya, line 28: Let's not change for any non-distribution workflows yet! It would be difficult to test as these are not exercised regularly.

peterzhuamazon: Hi,

I think it is ok to proceed with this change as it is simply running a docker container on the host.
Plus it has been there since weeks ago and these servers are stable.

This is a two-way door and I would suggest we go with the change than not so we can actively test these changes from now on. If it failed on that workflow we can always change back.

Thanks.

PR #4027, opened 2023-09-15
[Windows Container] Add Windows IntegTest support to Jenkins with Windows Docker Container

peterzhuamazon:
### Description
[Windows Container] Add Windows IntegTest support to Jenkins with Windows Docker Container

### Issues Resolved
https://github.com/opensearch-project/opensearch-ci/issues/281
https://github.com/opensearch-project/opensearch-build/issues/3743


src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -99,7 +99,7 @@ def __setup_cluster_and_execute_test_config(self, config: str) -> int:
             config,
             self.test_recorder,
         ) as (endpoint, port):
-            self.pretty_print_message("Running integration tests for " + self.component.name)
+            self.pretty_print_message("Running integration tests for " + self.component.name + " " + config)
```
peterzhuamazon, line 102: This is missed in the osd part but already in os part.

jenkins/opensearch-dashboards/integ-test.jenkinsfile
```diff
@@ -195,6 +207,35 @@ pipeline {
...
+                                                            localPath: "${WORKSPACE}/artifacts",
+                                                            force: true
+                                                          )
+                                                        sh("cp -a $WORKSPACE/artifacts/${artifactPath} $WORKSPACE")
+                                                        sh("rm -rf $WORKSPACE/artifacts")
+                                                    }
+                                                    else {
+                                                        echo "Not on Windows, unstash repository+artifacts"
+                                                    }
+
```
gaiksaya, line 238: Can you elobrate on what you are trying to do if its only windows? 
I believe because its windows you want to download again and not unstash?

peterzhuamazon: The stash/unstash action internally achieve the file in an slow manner.
Running such action is slow but still relatively reasonable on linux.

However on windows, we face multiple issues with this approach, this is on both the EC2 directly and on Container:
1. Takes forever to stash the same 3GB+ artifacts, 3x more time than on linux in general.
2. Will crash when try to unstash the archive, or just error out without output.
3. The download from S3 itself is slower on windows than on linux.

The best solution I am able to achieve right now, is to let non-windows keeps the original behavior, download once, stash, distribute, unstash so you only download once.

On Windows, you need to initialize download on every container individually, which is much smoother and quicker than stash/unstash, and avoid errors.

You might ask, why dont I download on each container instead of stash/unstash? That is because with that approach, it used to throttle the linux runner due to the amount of data is coming in and it is unnecessary. On Windows, this is a must have to keep the show going, so I intentionally increase the delay of each container from 10seconds to 20seconds. It is still have room to improve, but at least it is running well now.

I also removed the end step where you unstash again just to create reports. You do not need that as you can just create reports with build repo and build manifests.

Thanks.

gaiksaya: We do need data for generating reports along with build repo and build manifest. @zelinh Can add more.

Also I would prefer all to download in that case? What do you think? Switch linux behavior to download as well rather than unstash?

zelinh: @gaiksaya Yes. I had an offline discussion with @peterzhuamazon, for generating the reports, we would only need the build repo and build manifest (and dashboards build manifests if for dashboards) to run the library for creating report.

peterzhuamazon: As @zelinh clarified in above, we dont need data but the yml on s3 and the build manifest.
That would be all we needed here.

Thanks.

gaiksaya (approved):

PR #4038, opened 2023-09-19
Allow opensearch-dashboard user to read its config

smortex:
In  #3952, the permissions where changed to fix some inconsistencies in
the .deb and .rpm packaging.

This change restricted access to the configuration files (which where
previously readable by all users) but failed to adjust the files
ownership so that the service can access these files.

Ensure the configuration directory and files belong to the root user and
the opensearch-dashboards group

Signed-off-by: Romain Tartière <romain@blogreen.org>

### DIFF
--- a/scripts/pkg/build_templates/opensearch-dashboards/deb/debian/postinst
+++ b/scripts/pkg/build_templates/opensearch-dashboards/deb/debian/postinst
@@ -37,6 +37,7 @@ echo "### You can start opensearch-dashboards service by executing"
 echo " sudo systemctl start opensearch-dashboards.service"
 
 # Set ownership and permissions
+chown -R root.opensearch-dashboards ${config_dir}
 chmod -R u=rwX,g=rX,o= ${config_dir}
 
 chown -R opensearch-dashboards.adm ${log_dir}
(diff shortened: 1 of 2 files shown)


smortex:
@peterzhuamazon I saw this trying RC .deb packages (first RC I try).  It looks like it slept-through when I was working on #3952…  I think it needs special labeling so that it is included in the 2.10.0 release.  Thank you!

peterzhuamazon:
Hi @smortex ,

We do see OpenSearch rpm/deb test failed to start cluster as well.

Do you need to do the same fix on OpenSearch?

Thanks.

smortex:
@peterzhuamazon the OpenSearch package is fine according to my testing, it already had some config that changed the owner and we kept it unchanged (just changed the permissions)

peterzhuamazon (approved):

peterzhuamazon:
> @peterzhuamazon the OpenSearch package is fine according to my testing, it already had some config that changed the owner and we kept it unchanged (just changed the permissions)

Thanks @smortex I think it might be a plugin that caused some hiccups in test.
I checked in my local again with latest pkg and it runs.

I will go ahead and merge this. Thanks.

PR #4064, opened 2023-09-22
Automatically merge manifest changes after checks pass

peternied:
### Description
Automatically merge manifest changes after checks pass if the contributor is one of the maintainers or a bot.  This list could be expanded to release owners as time goes by.

- Restricted to certain authors
- Only allowed for manifest files changes

This action can be expanded to allow more authors or files for consideration.

### Additional Context
- _Shameless copy+paste from https://github.com/opensearch-project/OpenSearch/issues/10211_

#### Is your feature request related to a problem? Please describe.
There are an increasing number of automated tools and small changes that don't need as much attention from maintainers, such as version bumps where CI is sufficient to ensure the quality of the build. These pull requests are often distracting and easy to ignore and they tend to pile up, it would be great if changes of this type could be automated.

#### Describe the solution you'd like
It would be great if there was a tool that checked after all CI checks had passed and would then automatically merge changes.


.github/workflows/merge-common-changes.yml
peternied, line 1: I've got a pull request in my fork that I am testing this change with https://github.com/peternied/opensearch-build/pull/43

I'll update this thread when I've confirm its working as expected

peternied:
I haven't been able to get the `check_run` action to trigger on my repo, I suspect this is because the inciting action was triggered by a pull request and this is a measure [1] to protect against recursive actions triggers.

The clear alternative is `pull_request_target`, however, those check will take a non-trivial amount of time.  Seems like going to a polling system of some kind is the only way to be sure.  Anyone else have thoughts?

- [1] https://github.com/orgs/community/discussions/33804

peternied:
I think I found a way to do it, I could add a new workflow triggered by [workflow_run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) and then find the event that triggered that workflow.  If the trigger wasn't a pull request, I can ignore it, if it was a pull request I can use the checks API to see if all checks passed, if not ignore it.  While it will run 'extra' times for all of the workflows, it will ensure whenever the last workflow completes that all the checks will have passed or failed.

prudhvigodithi:
Hey @peternied thanks for the contribution, can you please elaborate the problem statement what are we trying to solve with this GH action ?
Thanks

peternied:
@prudhvigodithi I've added more details in the description.  TDLR; it would be really nice if there were some files that were automatically merged immediately after checks pass, such as adding a component to a fresh manifest

peternied:
@prudhvigodithi What do you think about this change?

prudhvigodithi:
Hey @peternied just an elevator pitch.

When PR (updated/added) created  by `allowed-authors` for any of the manifest files `manifests/*/*.yml` and if the workflows in the `workflows`  list are all passed it would auto-merge the PR. Tomorrow if a new workflow is added it should be part of the `workflows` list and like wise if a new maintainer is added (to ensure his/her PR's are auto merged) should be part of the `allowed-authors`. 

Please correct me if I'm wrong. If this is the case I'm ok to move forward.
@peterzhuamazon @gaiksaya

peternied:
@prudhvigodithi - yup you've got it, this is the case.

prudhvigodithi (approved):

.github/workflows/automatic-merges.yml
```diff
@@ -0,0 +1,47 @@
...
+            bbarani
+            gaiksaya
+            rishabh6788
+            zelinh
+            jordarlu
+            prudhvigodithi
+            Divyaasm
+            dblock
+            tianleh
+            peternied
```
gaiksaya, line 45: How about dynamically get this list from https://github.com/opensearch-project/opensearch-build/blob/main/.github/CODEOWNERS similar to https://github.com/opensearch-project/opensearch-py/blob/main/.github/workflows/release-drafter.yml#L14-L16

peternied: Hmm, that works well for repo maintainers, but I'd think this list might be larger since release owners are from all over the OpenSearch-Project.

Maybe this could be a follow up item?

gaiksaya: Sounds good!

Issue #3448, opened 2023-04-26
Support for multi node clusters and logging


monusingh-1:
### Is your feature request related to a problem? Please describe

During integration tests of cross cluster replication, clusters with following Topology are created
Leader - Data node -2 
Follower- Data node -2

The data nodes join to form a  cluster in these distribution builds
2.7.0/opensearch-2.7.0-test.yml, 2.7.0, arm64, 7759, tar,
2.7.0/opensearch-2.7.0-test.yml, 2.7.0, x64, 7759, tar
https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test/detail/integ-test/4653/pipeline

The nodes fail to join to form common cluster in these distributions
2.7.0/opensearch-2.7.0-test.yml, 2.7.0, x64, 7764, deb
2.7.0/opensearch-2.7.0-test.yml, 2.7.0, arm64, 7764, rpm

https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test/detail/integ-test/4658/pipeline

https://build.ci.opensearch.org/job/integ-test/4660/execution/node/858/log/?consoleFull 
Search for 
Node 1  at `2023-04-26 08:51:30` 
Node 2 at `2023-04-26 08:50:57`
We can see
```
"number_of_nodes":1,"number_of_data_nodes":1
```

After investigation we were able to check that when the data nodes fail to join, the clusters are not reachable and the integ tests starts throwing

```
java.net.ConnectException: Connection refused
        at org.opensearch.client.RestClient.extractAndWrapCause(RestClient.java:953)
        at org.opensearch.client.RestClient.performRequest(RestClient.java:332)
        at org.opensearch.client.RestClient.performRequest(RestClient.java:320)
        at org.opensearch.replication.MultiClusterRestTestCase.stopAllReplicationJobs(MultiClusterRestTestCase.kt:424)
        at org.opensearch.replication.MultiClusterRestTestCase.wipeIndicesFromCluster(MultiClusterRestTestCase.kt:448)
        at org.opensearch.replication.MultiClusterRestTestCase.wipeCluster(MultiClusterRestTestCase.kt:374)
        at org.opensearch.replication.MultiClusterRestTestCase.wipeClusters(MultiClusterRestTestCase.kt:369)
        at jdk.internal.reflect.GeneratedMethodAccessor13.invoke(Unknown Source)
[...]

PR #4097, opened 2023-10-03
Support multinode cluster logging

monusingh-1:
### Description
Allow logging of multiple nodes if a component uses multi node configuration to run integtest.

Fixes following when there are multiple clusters being created for integ test run:
FileExistsError: [Errno 17] File exists: '/var/jenkins/workspace/integ-test@4/test-results/5327/integ-test/cross-cluster-replication/with-security/local-cluster-logs/opensearch-service-logs'

### Issues Resolved

https://github.com/opensearch-project/opensearch-build/issues/3448


monusingh-1:
Will work on fixing integration tests

peterzhuamazon:
Hi @monusingh-1 could you add some quick log or list of dir to showcase this function work in action?

Thanks.

src/test_workflow/test_recorder/test_recorder.py
```diff
@@ -95,10 +95,12 @@ class LocalClusterLogs(LogRecorder):
 
     def __init__(self, parent_class: TestRecorder) -> None:
         self.parent_class = parent_class
+        self.number_of_nodes = 0
 
     def save_test_result_data(self, test_result_data: TestResultData) -> None:
         base = self.parent_class._create_base_folder_structure(test_result_data.component_name, test_result_data.component_test_config)
-        dest_directory = os.path.join(base, "local-cluster-logs")
+        dest_directory = os.path.join(base, "local-cluster-logs/id-" + str(self.number_of_nodes))
+        self.number_of_nodes += 1
```
peterzhuamazon, line 103: The problem with this design as per your log, is that once with-security test complete, the id will not reset and result in id-2/id-3 in without-security dir.

Would suggest reset the number after each run. So it will results in this:
```
with-security/id-0 id-1
without-security/id-0 id-1
```

instead of:
```
with-security/id-0 id-1
without-security/id-2 id-3
```

bbarani, line 103: This seems to be a generic change. Can you make sure that this doesn't impact other plugins? CC: @peterzhuamazon @gaiksaya

peterzhuamazon: I think it might impact the resports workflows that @zelinh was working on.

Need confirmation on how he retrieve the logs.

Thanks.

gaiksaya: Tests are triggered separately for with and without security so we should be good I believe.

gaiksaya (approved):
@rishabh6788 @bbarani Need to know if it is okay to merge this PR now or if we should hold on until 2.11 is done.
Thanks!

src/test_workflow/test_recorder/test_recorder.py
monusingh-1: The current behavior continues to increase the id throughout different components.

bbarani:
> @rishabh6788 @bbarani Need to know if it is okay to merge this PR now or if we should hold on until 2.11 is done. Thanks!

I think the tests have been disabled for CCR now. Lets merge it and enable tests for CCR once 2.11.0 is out?

src/test_workflow/test_recorder/test_recorder.py
zelinh: Yeah, I was worried this might impact the logs path for component yml file we generated during test record; but i just checked and confirmed that we are walking through the test folder [here](https://github.com/opensearch-project/opensearch-build/blob/318450ec8f4c8a3596321b8ae12adacb81e804c2/src/test_workflow/test_recorder/test_recorder.py#L79-L83). The inside file/directory structure should also be recorded properly. 
I also checked that we aren't hardcorded anywhere `local-cluster-logs` so I think this should be fine.

zelinh (approved):

Issue #118, opened 2021-07-29
Open-source signing infrastructure


dblock:
**Is your feature request related to a problem? Please describe.**
We've developed a service that does signing. Open-source it.

peterzhuamazon:
We will not open source the signer client for now.

dblock:
Is this final or is it a matter of time? If it's the former, I'd like to understand here what the plan for publicly reproducible signing infrastructure is, and if it's the latter, I'd like to reopen the issue.

peterzhuamazon:
> Is this final or is it a matter of time? If it's the former, I'd like to understand here what the plan for publicly reproducible signing infrastructure is, and if it's the latter, I'd like to reopen the issue.

@peternied could you explain more in details to @dblock I think our decision is this is reproducible with gpg and it is not necessarily that we would want to open source the internal code to run the signing process.

dblock:
> > Is this final or is it a matter of time? If it's the former, I'd like to understand here what the plan for publicly reproducible signing infrastructure is, and if it's the latter, I'd like to reopen the issue.
> 
> @peternied could you explain more in details to @dblock I think our decision is this is reproducible with gpg and it is not necessarily that we would want to open source the internal code to run the signing process.

That's fine, I am looking for an issue for replacing the existing signing infrastructure with gpg, then.

peternied:
For our maven artifacts this is relatively straight forward using GPG.  We can make this process more transparent - do you have an example of what you would to align towards?

Our other platforms like Mac/Windows signing processes are Amazon internal and would require considerable work to make accessible.

PR #4098, opened 2023-10-03
Add support to run benchmarks using OpenSearch distribution url

### LINKED ISSUE #4073: Add option to run benchmark against OS artifact url
The nightly benchmarks currently depend on the distribution manifest file to get the location of the OpenSearch distribution tarball and then set up the cluster and run benchmark. 
There have been a few requests to run ad-hoc benchmark tests using nightly benchmark framework against the distribution tarball provided by requester. Since the tarball was created locally and had no manifest associated with it we could not accept the request and guided the user to run it locally. 
Add a feature in the workflow to be able to run the benchmark workflow using either a distribution manifest url or the actual artifact url.

rishabh6788:
### Description
In the current phase the nightly benchmark runs are tightly coupled with distribution manifest file generated as part of our OpenSearch distribution build process. 
There have been a request from developers to be able to use our benchmark platform to run ad-hoc benchmarks to test their local changes. For this the most basic requirement is to directly pass the OpenSearch distribution url instead of code reading it from `manifest.yml` file. 
The existing functionality would still work as the user can either use distribution manifest or distribution url as inputs. 

We are defaulting to `x64` arch if distribution url is provided as input and will add support for `arm64` in subsequent PRs. 

### Issues Resolved
#4073 
https://github.com/opensearch-project/OpenSearch/issues/9418


src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -131,3 +137,8 @@ def __init__(self) -> None:
         self.telemetry = args.telemetry
         self.telemetry_params = args.telemetry_params if args.telemetry_params else None
         self.logging_level = args.logging_level
+
+        if self.bundle_manifest is None and self.distribution_url is None:
+            raise Exception('Please provide either --bundle-manifest or --distribution-url to run the performance test.')
```
peterzhuamazon, line 142: What if both are not NONE, then which one will be taken?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -112,27 +114,37 @@ def wait_for_processing(self, tries: int = 3, delay: int = 15, backoff: int = 2)
         logging.info(f"Waiting for domain at {self.endpoint} to be up")
         protocol = "http://" if self.args.insecure else "https://"
         url = "".join([protocol, self.endpoint, "/_cluster/health"])
-        request_args = {"url": url} if self.args.insecure else {"url": url, "auth": HTTPBasicAuth("admin", "admin"), "verify": False}  # type: ignore
+        request_args = {"url": url} if self.args.insecure else {"url": url, "auth": HTTPBasicAuth("admin", "admin"),  # type: ignore
```
peterzhuamazon, line 117: Verify:False not needed?

src/test_workflow/benchmark_test/benchmark_args.py
rishabh6788: If you are talking about `--bundle-manifest` and `--distribution-version` being provided at the same time then workflow will default to `--bundle-manifest` and use that to run.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: It is there on line #118

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -112,27 +114,37 @@ def wait_for_processing(self, tries: int = 3, delay: int = 15, backoff: int = 2)
...
-        else:
+        elif self.manifest:
             suffix = self.manifest.build.id + '-' + self.manifest.build.architecture
+        elif self.args.stack_suffix:
+            suffix = self.args.stack_suffix
+
+        if self.manifest:
+            artifact_url = self.manifest.build.location if isinstance(self.manifest, BundleManifest) else \
+                f"https://artifacts.opensearch.org/snapshots/core/opensearch/{self.manifest.build.version}/opensearch-min-" \
+                f"{self.manifest.build.version}-linux-{self.manifest.build.architecture}-latest.tar.gz"
```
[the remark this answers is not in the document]
rishabh6788: Unfortunately the build manifest doesn't have `location` attribute like bundle-manifest, so had to fall back to using formatted url.

peterzhuamazon (approved):

Issue #4094, opened 2023-10-02
[Security] Demo configuration script requires admin password 


RyanL1997:
### Is your feature request related to a problem? Please describe

According to the changes in security plugin of security demo configuration script: https://github.com/opensearch-project/security/pull/3329. This change introduces an alternative to the default credentials admin:admin the security plugin currently uses.
In this implementation, ~a new setting is added to the opensearch.yml file. This setting, `plugins.security.authcz.admin.password` is parsed on plugin start up and then propagated into the internal user store of each node when they launch~ During the set up of the security demo config, it will ask user to provide an initial admin password. User can either provide that by defining an environment variable 'initialAdminPassword' with a password string, or create a file 'initialAdminPassword.txt' with a single line that contains the password string and place it under the config folder. The absent of this 'initialAdminPassword' will lead to the failure of the demo config script.

For adapting this change, we need to change the build script to handle the setup for admin password.

### Describe the solutions you'd like

We have a talk with @RyanL1997 @DarshitChanpura @peterzhuamazon on this for a bit:

**Solution A: Adding a parameter for running the demo config script to set the default admin password as 'admin' ([sample implementation in security](https://github.com/opensearch-project/security/pull/3441)):**

We can add a parameter to security demo configuration setup script, which if not provide this parameter, user will be prompted to provide a secure admin password, and if user provides this parameters which means user decides to skip the setup of admin password, then we’ll proceed with "admin" as default admin password and tell user that they acknowledge to accept an insecure password for this non-prod env.

We believe this change is justifiable given that the primary intent of the demo configuration script is for non-production environments. Our documentation clearly underscores our recommendation against using this demo config for actual production setups. In line with this, introducing this parameter remains consistent with the script's intended use, emphasizing that it is not suitable for production deployments.

**Solution B: Users have to be prompted to enter an initial admin password during the set up of security demo configuration script ([sample implementation in security](https://github.com/opensearch-project/security/pull/3449)):**

[...]

PR #4105, opened 2023-10-04
Add a warning info and hard stop for security demo setup

RyanL1997:
### Description
Add a warning info and hard stop for security demo setup

### Issues Resolved
* Relate https://github.com/opensearch-project/opensearch-build/issues/4094

### DIFF
--- a/scripts/startup/tar/linux/opensearch-tar-install.sh
+++ b/scripts/startup/tar/linux/opensearch-tar-install.sh
@@ -10,7 +10,10 @@ cd $OPENSEARCH_HOME
 KNN_LIB_DIR=$OPENSEARCH_HOME/plugins/opensearch-knn/lib
 ##Security Plugin
 if [ -d "$OPENSEARCH_HOME/plugins/opensearch-security" ]; then
-        bash $OPENSEARCH_HOME/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s
+        echo "OpenSearch 2.11.0 onwards, the security plugin introduces a change that requires an initial password for 'admin' user."
+        echo "Please define an environment variable 'initialAdminPassword' with a password string."
+        echo “Or create a file 'initialAdminPassword.txt' with a single line that contains the password string and place it under $OPENSEARCH_PATH_CONF folder.”
+        bash $OPENSEARCH_HOME/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s || exit 1
         echo "done security"
 fi
 


gaiksaya (changes requested):
Hi @RyanL1997 

Is this change being backported to 1.x too? 

If not @bbarani we need to branch this repository to accommodate these changes.
Also @peterzhuamazon I believe you have better knowledge what to do in terms of rpm, debian, windows zip and docker. 

Thanks!

DarshitChanpura:
This change is meant for 2.11 and above.
There will be subsequent PRs for RPM, DEB, windows and docker.

This PR intentionally only targets TAR to test and verify the change.

scripts/startup/tar/linux/opensearch-tar-install.sh
```diff
@@ -10,7 +10,10 @@ cd $OPENSEARCH_HOME
 KNN_LIB_DIR=$OPENSEARCH_HOME/plugins/opensearch-knn/lib
 ##Security Plugin
 if [ -d "$OPENSEARCH_HOME/plugins/opensearch-security" ]; then
-        bash $OPENSEARCH_HOME/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s
+        echo "OpenSearch 2.11.0 onwards, the security plugin introduces a change that requires an initial password for 'admin' user."
+        echo "Please define an environment variable 'initialAdminPassword' with a password string."
+        echo “Or create a file 'initialAdminPassword.txt' with a single line that contains the password string and place it under $OPENSEARCH_PATH_CONF folder.”
+        bash $OPENSEARCH_HOME/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s || exit 1
```
rishabh6788, line 16: What happens if `plugins.security.authcz.admin.password` setting is set? @RyanL1997 
Does it set the `initialAdminPassword` env variable so the check in `tools/install_demo_configuration.sh` doesn't fail?

RyanL1997: Hi @rishabh6788, thanks for the feedback. Yes, so if that setting is set, the security plugin will use that password, and yes it will run the install demo config script.

RyanL1997:
Hi @gaiksaya, thanks for double checking. @DarshitChanpura 's reply is right, we do not need this change in 1.x, since the original change in security is only backported into 2.x.

bbarani:
> Hi @gaiksaya, thanks for double checking. @DarshitChanpura 's reply is right, we do not need this change in 1.x, since the original change in security is only backported into 2.x.

This will force us to branch the build workflows as well since its not setup that way at this point in time.

gaiksaya:
> > Hi @gaiksaya, thanks for double checking. @DarshitChanpura 's reply is right, we do not need this change in 1.x, since the original change in security is only backported into 2.x.
> 
> This will force us to branch the build workflows as well since its not setup that way at this point in time.

+1 

Today distribution workflows only runs on main as the code is generic. For config files we have 1.x and 2.x folders specifying the related versions. 
With this change, security plugin being a crucial one for running tests, etc, our source code for running integrations tests, bwc tests, etc would change as well. For example, we need to make sure this is taken care before triggering any tests in our test jobs. Also since this only applies to 2.x and not 1.x need to handle that scenario as well. Either via CI or via source code. 
```
Please define an environment variable 'initialAdminPassword' with a password string or create a file 'initialAdminPassword.txt' with a single line that contains the password string and place it under $OPENSEARCH_PATH_CONF 
```

scripts/startup/tar/linux/opensearch-tar-install.sh
rishabh6788: So would that plugin setting set the `initialAdminPassword` env variable?

Because as per [this](https://github.com/opensearch-project/security/blob/main/tools/install_demo_configuration.sh#L395-L401), if the code doesn't find the `initialAdminPassword` env set or `ADMIN_PASSWORD_FILE="$OPENSEARCH_CONF_DIR/initialAdminPassword.txt"` file then it is going to error out.

rishabh6788: Or is the plugin setting adding password to `$OPENSEARCH_CONF_DIR/initialAdminPassword.txt` file at runtime?

DarshitChanpura:
> > > Hi @gaiksaya, thanks for double checking. @DarshitChanpura 's reply is right, we do not need this change in 1.x, since the original change in security is only backported into 2.x.
> > 
> > 
> > This will force us to branch the build workflows as well since its not setup that way at this point in time.
> 
> +1
> 
> Today distribution workflows only runs on main as the code is generic. For config files we have 1.x and 2.x folders specifying the related versions. With this change, security plugin being a crucial one for running tests, etc, our source code for running integrations tests, bwc tests, etc would change as well. For example, we need to make sure this is taken care before triggering any tests in our test jobs.
> 
> ```
> Please define an environment variable 'initialAdminPassword' with a password string or create a file 'initialAdminPassword.txt' with a single line that contains the password string and place it under $OPENSEARCH_PATH_CONF 
> ```
> 
> Also since this only applies to 2.x and not 1.x need to handle that scenario as well. Either via CI or via source code.

Could you give us some pointers on how to test this?

scripts/startup/tar/linux/opensearch-tar-install.sh
RyanL1997: > Because as per [this](https://github.com/opensearch-project/security/blob/main/tools/install_demo_configuration.sh#L395-L401), if the code doesn't find the initialAdminPassword env set or ADMIN_PASSWORD_FILE="$OPENSEARCH_CONF_DIR/initialAdminPassword.txt" file then it is going to error out.

I think this is what we are expecting, since we would like to force user to setup the initial admin password instead of using the `admin:admin` So to make the long story short, no more admin:admin, and we want to fail both demo config setup and initiation of the cluster if there is an absent of this initial admin password provided by user.

gaiksaya:
> Could you give us some pointers on how to handle and test this?

I believe we need to plan it properly and handle it from infra side. Branching being the key element here (whether to do it or not OR handle the scenario efficiently based on versions.)
This is the code for all tests https://github.com/opensearch-project/opensearch-build/tree/main/src/test_workflow
You can try to run integration tests on any component and test it out. Instructions should be in the readme. 
Basically, setting up the testing cluster for any tests would need some changes.

P.S: Did a rough search on admin keyword. Yielded these results: https://github.com/search?q=repo%3Aopensearch-project%2Fopensearch-build+admin+language%3APython&type=code&l=Python

rishabh6788:
We can use the new setting to keep the password `admin` to avoid updates to the python code. Right? @gaiksaya

scripts/startup/tar/linux/opensearch-tar-install.sh
rishabh6788: FYI, I am not able to find `plugins.security.authcz.admin.password` setting in Security plugin code. Was this implemented? @RyanL1997

gaiksaya:
> We can use the new setting to set the password `admin` to avoid updates to the python code. Right? @gaiksaya

That's what I was thinking as well. However, what password you use is not the issue. We need to make code compatible like if version is `1.x` run install.sh normally. If 2.x/3.x use env variable. 
Also, if not branching another way to go around is separate all specfiles/config scripts into 1.x and 2.x like you said. That should include everything from tarball, rpm,deb,windows and docker. 
There might be duplicates which is okay rather than risking wrong version config.

scripts/startup/tar/linux/opensearch-tar-install.sh
RyanL1997: @rishabh6788 Sorry for the confusions. No this is not implemented. I will re-edit the description of that issue and original PR, and this was some stale design. I think the current change we are having right now should be only in the install of demo configuration script.

DarshitChanpura: There is no such setting in the opensearch.yml. The only way to set admin password on boot-up with the new change is to set it via demo installation script.

DarshitChanpura:
@gaiksaya @bbarani @peterzhuamazon @rishabh6788 What is the path forward to accommodate this change?

bbarani:
> Is it a possibility that we get this change in now for 2.11, and then fix it later when we release 1.x tarball. This will address the immediate need which is to get it in for 2.x, and we can create a follow-up to fix it for 1.x before next 1.x (which should give enough time to fix it before release i.e. 1.3.14 is scheduled to be release on December 5th)

It would basically break anyone using this code to build 1.x on their own as well. Also, this would hinder our ability to release patch release in timely manner for 1.x version (if we need to address any critical CVE's).

RyanL1997:
Hi @bbarani, thank you and the other folks for all the inputs on this. Yes, we can see the problem now. It seems like we need to address the `1.x`'s build breaking before we implement the script changes like this since we currently don't have branching strategy for the build flow. (cc: @DarshitChanpura @peternied @davidlago )

DarshitChanpura:
> It would basically break our automation (nightly builds, benchmarking etc..) as well as for anyone using this code to build 1.x on their own. Also, this would hinder our ability to release patch release in timely manner for 1.x version (if we need to address any critical CVE's) before planned patch release timeline.

Makes sense. How do we move forward with this change?

bbarani:
> separate all specfiles/config scripts into 1.x and 2.x l

I think these are the possible 2 option I can infer from the discussion here,

1.) Create branch for 1.x and 2.x
2.) Separate all specfiles/config scripts into 1.x and 2.x.

Both the above changes would also involve CI, workflow changes as well to consume the code / configs from the right location for corresponding versions. 

@gaiksaya @peterzhuamazon @rishabh6788 what do you think is the level of effort for either of the approaches?

gaiksaya:
> These are the possible options I can infer from the discussion here,
> 
> 1.) Create branch for 1.x and 2.x to isolate the build workflow changes to 2.x versions 2.) Separate all specfiles / config scripts into 1.x and 2.x.
> 
> Both the above changes would also involve CI, workflow changes as well to consume the code / configs from the right location for corresponding versions.
> 
> @gaiksaya @peterzhuamazon @rishabh6788 what do you think is the level of effort for either of the approaches?

I like option 1. In this way whatever test, build code we have for 1.x remains intact. Also whatever new changes come up (today its 2.11, in future 3.0.0) the way we set up test clusters and run test can be defined and modified per major version. 
Level of effort compared to both is also less for option 1. We however need sometime to set up CI and make sure its working as expected. 

Option 2 is good in terms of config files. However this does not solve our code compatibility issues (integration, perf tests as well as validation workflows that were newly added recently). Whatever deploys clusters with security would need to be compatible with both 1.x and 2.x.

scripts/startup/tar/linux/opensearch-tar-install.sh
peternied, line 16: I don't think we need the version specific text - so long as if the script returns a non-zero code the docker install fails, the console output should include the information needed to troubleshoot the issue.

I think this would avoid the git branches or version detection in the near term, what do you think of this notion @gaiksaya?

RyanL1997: I was checking with @peterzhuamazon, and it seems like we need to add these to warn the user before we run the flow from build script too, even we had our own lines inside our own script. 

> I think this would avoid the git branches or version detection in the near term, what do you think of this notion @gaiksaya?
For this, if I was understanding this correct yesterday, we need the branching strategy from the build team no matter we have these specific lines or not because the build team is using their workflows on `main` to build our `3.x`,`2.x` and `1.x`. In another word, everything is relying on main for build repo. Since we are not having our password change in our 1.x line, if we make the change like this, their `1.x` will be broken because the changes we are make changes the behavior of this script. (Please correct me if I'm wrong. cc: @gaiksaya @rishabh6788 )

peternied:
> We need the branching strategy from the build team no matter we have these specific lines or not because the build team is using their workflows on main to build our 3.x,2.x and 1.x. In another word, everything is relying on main for build repo. Since we are not having our password change in our 1.x line, if we make the change like this, their 1.x build will be broken because the build flow should not be failed/exit on 1.x

@RyanL1997 Based on this comment it seems like we should close out this pull request and create an issue, then add proposals for how to address these concerns.  Then before implementing, make sure to review the plan with folks from opensearch-build.  How does that sound?

DarshitChanpura:
Let's reconvene on how to bring the demo configuration changes in. Until then let's close this PR. We can re-open this once we resume work.

Issue #3441, opened 2023-04-25
Integrate Validation mechanism with OS and OSD builds


Divyaasm:
### Is your feature request related to a problem? Please describe

The idea is to integrate a Validation mechanism to each successful build in OS and OSD, and update the latest build only when the validation is successful. 

### Describe the solution you'd like

Utilize the [Validation](https://github.com/Divyaasm/opensearch-build/tree/main/src/validation_workflow) framework, the idea is to validate the artifacts by downloading them, start a cluster and run the api commands to test the cluster similar to https://opensearch.org/docs/latest/install-and-configure/install-opensearch/tar/

Sub issues under the meta issue:

- [x] [Update the current Validation mechanism to download OS and OSD artifacts from CI bucket](https://github.com/opensearch-project/opensearch-build/issues/3690)
- [x] [Update existing Validation Framework to run individually for OS and OSD](https://github.com/opensearch-project/opensearch-build/issues/3691)
- [ ] The approach to call the validation library will be done in integtest workflow.
       Related issue : #4424 

Acceptance Criteria:

1. Separate the Validation Framework using -- type OS or OSD currently it does for OS and OSD together.
2. Add a new functionality to fetch the artifacts from CI bucket
3. The idea is to start with Tar distribution seperately for OS and OSD
4. ~~Call the Validation Groovy library in the post method before the[ /latest](https://github.com/opensearch-project/opensearch-build/blob/68a2fe7a67f30171545ceff689a3493dcb6d7558/jenkins/opensearch/distribution-build.jenkinsfile#L844) is updated.~~

### Additional context

A possible solution to address https://github.com/opensearch-project/opensearch-build/issues/3390

bbarani:
Tagging the [PR](https://github.com/opensearch-project/opensearch-build/pull/4377)

Divyaasm:
Validation workflow is now integrated with Integ-test workflow for now instead of Build Workflow.

PR #4116, opened 2023-10-10
Jenkins job updation for validation workflow

Divyaasm:
### Description
Changes made to jenkins job related to validation workflow to accomodate the recent changes to python codebase
### Issues Resolved
Addresses https://github.com/opensearch-project/opensearch-build/issues/3441


jenkins/validate-artifacts/validate-artifacts.jenkinsfile
```diff
@@ -65,6 +76,19 @@ pipeline {
...
         )
+        string(
+            name: 'PROJECTS',
+            description: 'Enter opensearch or opensearch-dashboards',
+            trim: true,
+            defaultValue: "opensearch"
+        )
+        string(
+            name: 'ARTIFACT_TYPE',
+            description: 'Determine the artifact type among staging or production ',
```
gaiksaya, line 87: Is local an option too? Should this be a `CHOICE` instead of string parameter?

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
```diff
@@ -65,6 +76,19 @@ pipeline {
             trim: true,
             defaultValue: "using-staging-artifact-only"
         )
+        string(
+            name: 'PROJECTS',
+            description: 'Enter opensearch or opensearch-dashboards',
+            trim: true,
+            defaultValue: "opensearch"
+        )
```
gaiksaya, line 84: Should this be a `CHOICE`?

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
```diff
@@ -73,12 +97,29 @@ pipeline {
...
+                    OPENSEARCH_DASHBOARDS_ARTIFACT_URL ? url_paths['opensearch-dashboards'] = OPENSEARCH_DASHBOARDS_ARTIFACT_URL : null
+
+                    if (OPENSEARCH_DASHBOARDS_ARTIFACT_URL && OPENSEARCH_ARTIFACT_URL == "") {
                         currentBuild.result = 'ABORTED'
-                        error('VERSION cannot be empty. Please provide version of the artifacts to validate')
+                        error("Provide OPENSEARCH_ARTIFACT_URL to validate")
+                    }
+
+                    if(OPENSEARCH_ARTIFACT_URL) {
+                        ARCHITECTURE = url_paths.values().find { it.contains("x64") } ? "x64" : "arm64"
```
gaiksaya, line 111: This is not true I believe? Windows does not have arm64.

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
```diff
@@ -73,12 +97,29 @@ pipeline {
...
+                        DISTRIBUTION = url_paths.values().find { it.contains("tar.gz") } ? "tar" : (url_paths.values().find { it.contains("rpm") } ? "rpm" : "yum")
```
gaiksaya, line 112: Did we miss debian here?

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
```diff
@@ -73,12 +97,29 @@ pipeline {
...
                     }
 
-                    if (params.DISTRIBUTION.contains('docker') && (OS_BUILD_NUMBER == '' || OSD_BUILD_NUMBER == '')) {
+                    env.FILE_PATH = url_paths
+
+                    if(VERSION == '' && OPENSEARCH_ARTIFACT_URL == '') {
+                        currentBuild.result = 'ABORTED'
+                        error("Provide either version or URLs to validate")
```
gaiksaya, line 119:
```suggestion
                        error("Both VERSION and OPENSEARCH_ARTIFACT_URL cannot be empty. Please provide either value")
```

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
```diff
@@ -175,5 +225,47 @@ pipeline {
                 }
             }
         }
+        stage('Validate artifacts from urls'){
+            when {
+                expression {
+                    VERSION == '' && OPENSEARCH_ARTIFACT_URL != ''
```
gaiksaya, line 231: I believe this check is already done is verify-parameters stage. So can be changed to just expression `{
VERSION}` right?

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
Divyaasm: Local is not an option, it's automatically detected in python code based on the url format. User can just provide OS local path and OSD local path in this case. In this case we just copy the local artifacts from their respective path to temporary folder instead of downloading them.

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
Divyaasm: Not sure if that's required. It takes values like opensearch or opensearch opensearch-dashboards and validate accordingly.

gaiksaya: Does it take both or either? I believe if it want to validate both , there should be an option for that as well? 
I think right option here is `OPENSEARCH` OR `ALL` since we need both os and osd for osd to be validated? What do you think? OSD cannot be validated on its own right?

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
Divyaasm: I see @gaiksaya currently we do not support windows. But that is based on user input. When windows arm64 does not exist we do not provide the artifacts details at all while calling the job . It should always be x64 when it comes to windows.

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
Divyaasm: The validation workflow currently supports only tar, rpm, yum and docker. So I didn't include windows and debian.

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
Divyaasm: Yeah I'll check this out and update. Thanks

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
Divyaasm: both or either(only opensearch). Yes that case is covered here. The input could be `opensearch` or `opensearch opensearch-dashboards`. Both the scenarios are validated here @gaiksaya

gaiksaya: Sorry if that was not clear. When options are provided to the user I believe it should be `OPENSEARCH` OR `ALL` instead of `opensearch or opensearch-dashboards`

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
```diff
@@ -45,25 +56,34 @@ pipeline {
         )
         string(
             name: 'ARCHITECTURE',
-            description: 'Enter the architecture type among x64 and arm64 separated by space. Defaults "x64,arm64"',
+            description: 'Enter the architecture type among x64 and arm64 separated by space. Defaults "x64 arm64"',
```
gaiksaya, line 59:
```suggestion
            description: 'Enter the architecture type among x64 and arm64 separated by space. Defaults to "x64 arm64"',
```

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
gaiksaya: Do we need to add check for that? I beleive we need to keep this in mind when we add validation for windows?

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
```diff
@@ -175,5 +226,48 @@ pipeline {
                 }
             }
         }
+        stage('Validate artifacts from urls'){
+            when {
+                expression {
+                    VERSION == ''
+                }
+            }
```
gaiksaya, line 234: Why do we need this check? Sometimes if a user adds both URLS and version I believe URL takes precedence? In that case add an warning statement stating `VERSION is not empty, ignoring VERSION since URLs were specified`

gaiksaya (commented):
Some nites but looks good overall.

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
Divyaasm: Yeah Sayali. I will add a check when we add windows.

jenkins/validate-artifacts/validate-artifacts.jenkinsfile
Divyaasm: I found multiple stages clumsy for version and url's validation. Just updated the Job to validate all the conditions in a single stage. Now it should be straightforward.

Issue #3966, opened 2023-09-06
[Meta] Ensure Github workflow runs on docker image used by Production Distribution Build


peterzhuamazon:
[Meta] Ensure Github workflow pull docker image used by Infra Jenkins build.

This is to ensure every plugin repo will use the exact docker images we used in Jenkins build, to check their PRs and run tests before merging the code, so that issues can be detected earlier, and environment can be identical across teams.

This is to make sure, integration tests are run on same environment as that of infra build and help catch issues early. 

### Also this requires the user of the run being changed to `root` then switch to `1000` so both OS and OSD can start running without issues on linux: https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user

* https://github.com/opensearch-project/k-NN/issues/1042 (Trial run)
  * https://github.com/opensearch-project/k-NN/pull/1099
  * https://github.com/opensearch-project/k-NN/pull/1204
  * https://github.com/opensearch-project/k-NN/pull/1249
  * https://github.com/opensearch-project/k-NN/pull/1252
* https://github.com/opensearch-project/alerting/pull/1250
  * https://github.com/opensearch-project/alerting/pull/1264
  * https://github.com/opensearch-project/alerting/pull/1271
* https://github.com/opensearch-project/cross-cluster-replication/pull/1249
* https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/620
* https://github.com/opensearch-project/ml-commons-dashboards/pull/280
  * https://github.com/opensearch-project/ml-commons-dashboards/pull/281
* https://github.com/opensearch-project/ml-commons/pull/1538
* https://github.com/opensearch-project/job-scheduler/pull/534
* https://github.com/opensearch-project/alerting-dashboards-plugin/pull/789
* https://github.com/opensearch-project/anomaly-detection/pull/1092
  * https://github.com/opensearch-project/anomaly-detection/pull/1093
  * https://github.com/opensearch-project/anomaly-detection/pull/1094
* https://github.com/opensearch-project/asynchronous-search/pull/507
* https://github.com/opensearch-project/common-utils/pull/557
* https://github.com/opensearch-project/reporting/pull/924
* https://github.com/opensearch-project/dashboards-visualizations/pull/287
* https://github.com/opensearch-project/index-management/pull/1025
* https://github.com/opensearch-project/index-management-dashboards-plugin/pull/907
* https://github.com/opensearch-project/notifications/pull/809
[...]

PR #4124, opened 2023-10-12
Add build_ref in preparation for build repo branching

peterzhuamazon:
### Description
Add build_ref in preparation for build repo branching

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3966

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -9,7 +9,7 @@ jobs:
     with:
       platform: centos7
       usage: opensearch
-      build_ref: main
+      build_ref: ${GITHUB_REF}
 
   python-tests:
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/python-tests.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     with:
       platform: centos7
       usage: opensearch
+      build_ref: ${github.ref}
```
gaiksaya, line 12: I dont think we need this. The workflow runs on respective branches. So we commit or backport to 2.x, the workflows will run on 2.x not on main.

peterzhuamazon: It is designed specifically in this way so other repos can reuse this yaml:
https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/get-ci-image-tag.yml#L31-L36

peterzhuamazon:
Anyway, thinking people can just use the params to get the results, this can be closed for now.
Thanks.

Issue #4157, opened 2023-10-19
Add dependent component logic in input manifest


zelinh:
Similar to what we suggested [here](https://github.com/opensearch-project/opensearch-build/issues/441#issuecomment-1275192880), we would need to add a new entry `depends_on` in the input manifests for those components that has dependency plugin from Maven. 
We will also need to figure the exact dependent component for each of plugin. 

| Components | Dependent components |
|--------|--------|
| k-NN | N/A |
| geospatial | job-scheduler |
| **cross-cluster-replication** | common-utils |
| **ml-commons** | common-utils |
| neural-search | k-NN, ml-commons | 
| **notification/notifications-core** | common-utils |
| **observability** | common-utils |
| **opensearch-reporting** | job-scheduler, common-utils |
| **sql** | ml-commons |
| **asynchronous-search** | common-utils | 
| **anomaly-detection** | job-scheduler, common-utils |
| **alerting** | common-utils |
| security-analytics | common-utils |
| **index-management** | job-scheduler, common-utils |
| performance-analyzer | N/A | 
| custom-codecs | N/A |

zelinh:
I will be testing to build each of the components on 2.x branch to see their direct dependency components.

zelinh:
When building `geospatials` directly `./build.sh manifests/2.12.0/opensearch-2.12.0.yml -p linux -a x64 -d tar --component geospatial`, seeing
```
* What went wrong:
Execution failed for task ':compileJava'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not find org.opensearch:opensearch-job-scheduler-spi:2.12.0.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/opensearch/opensearch-job-scheduler-spi/2.12.0.0/opensearch-job-scheduler-spi-2.12.0.0.pom
       - file:/usr/share/opensearch/.m2/repository/org/opensearch/opensearch-job-scheduler-spi/2.12.0.0/opensearch-job-scheduler-spi-2.12.0.0.pom
       - https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/opensearch-job-scheduler-spi/2.12.0.0/opensearch-job-scheduler-spi-2.12.0.0.pom
       - https://plugins.gradle.org/m2/org/opensearch/opensearch-job-scheduler-spi/2.12.0.0/opensearch-job-scheduler-spi-2.12.0.0.pom
     Required by:
         project :
```

[...]

PR #4179, opened 2023-10-26
Add depends_on to the input manifest and increment its schema to 1.1

zelinh:
### Description
Add depends_on to the input manifest and increment its schema to 1.1.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4157


src/build_workflow/builders.py
```diff
@@ -11,15 +11,15 @@
...
 from build_workflow.builder_from_source import BuilderFromSource
-from manifests.input_manifest import InputComponent, InputComponentFromDist, InputComponentFromSource
+from manifests.input_manifest import InputComponent
 
 
 class Builders(ABC):
     @classmethod
     def builder_from(self, component: InputComponent, target: BuildTarget) -> Builder:
-        if type(component) is InputComponentFromDist:
+        if hasattr(component, "dist"):
```
dblock, line 20: This causes the need to do a `type: ignore` below, why is this needed?

zelinh: @dblock Good point! Those below will give us `error: "InputComponent" has no attribute "repository"  [attr-defined]`.
I think the reason is that it's iterating through `manifest.components.select` and that method is set to return an iterator of `InputComponent` type even though we have the check to make sure it has `repository` attribute.

rishabh6788: Little confused too, why it can't be handled with existing implementation?

zelinh: I think the usages of `component` are different in these two scenarios. In this `builders.py` file, we don't iterate through components and don't access any of the attribute. 

In below example from `run_checkout.py`
```
        for component in manifest.components.select():
            logging.info(f"Checking out {component.name}")
            if hasattr(component, "repository"):
                with GitRepository(
                    component.repository,  # type: ignore[attr-defined]
                    component.ref,  # type: ignore[attr-defined]
                    os.path.join(work_dir.name, component.name),
                    component.working_directory,  # type: ignore[attr-defined]
                ) as repo:
                    logging.debug(f"Checked out {component.name} into {repo.dir}")
```
We are getting an iterator of `InputComponent` type. Even though we have a conditional statement 
`if hasattr(component, "repository")`, python type check seems not able to tell it's a `InputComponentFromSource` and when we try to access its attribute such as `component.repository`, it says `error: "InputComponent" has no attribute "repository"  [attr-defined]`.

The workflow can be run correctly with no error when I was testing on them and the python tests are passing as well.

rishabh6788: As per my understanding you are only adding a new `depends_on` field in the input manifest schema. This change should not break the existing code flow, ideally the component object type would always be either `InputComponentFromSource` or `InputComponentFromDist`. 
What is the error you are getting with your changes, is it related to type incompatibility?
Please add the error you are facing with type check implementation, we can work from there to figure out what is going on. 
@dblock Do you think this is the right way forward?

dblock: 100%, this change is unexpected

zelinh: Yeah this was unexpected for me at first as well. When looking into the reason, this change is actually to support backward compatibility. 
Because when we incrementing the schema version to `1.1`, we keep the previous input_manifest class as `input_manifest_1_0` to support any previous 1.0 manifest. 
If in this case `builders` class, we are building `1.0` version input manifest, the condition statement would never fulfilled because the component object types are either `InputComponentFromSource_1_0` or `InputComponentFromDist_1_0` imported from `Input_Manifest_1_0`. 
These changes are making the conditional check more generic and sustainable for future schema version change for `input_manifest` as well. 
Let me know what you think about. Thanks! @rishabh6788 @dblock

src/manifests/input/input_manifest_1_0.py
```diff
@@ -0,0 +1,289 @@
...
+
+        :param List[str] focus: Choose some components.
+        :param str platform: Only components targeting a given platform.
+        :return: Collection of components.
+        :raises ValueError: Invalid platform or component name specified.
+        """
+        if focus and len(focus) > 0:
+            invalid = [item for item in focus if item not in self]
+            if len(invalid) > 0:
+                raise ValueError(f"Unknown component{'s'[:len(invalid) != 1]}={','.join(invalid)}.")
```
rishabh6788, line 191: please add missing tests.

src/manifests/input/input_manifest_1_0.py
```diff
@@ -0,0 +1,289 @@
...
+        self.checks = list(map(lambda entry: Check_1_0(entry), data.get("checks", [])))
+
+    @classmethod
+    def _from(self, data: dict) -> 'InputComponent_1_0':
+        if "repository" in data:
+            return InputComponentFromSource_1_0(data)
+        elif "dist" in data:
+            return InputComponentFromDist_1_0(data)
+        else:
+            raise ValueError(f"Invalid component data: {data}")
```
rishabh6788, line 215: please add missing tests.

src/manifests/input/input_manifest_1_0.py
zelinh: These functions are basically the same as the main `input_manifest` class, just the previous `1.0` schema version. However I did add some tests for different schema versions in `tests/tests_manifests/test_input_manifest.py`. 
I don't see we have separate tests for old schema version `build_manifest` or `bundle_manifest`. Do we want to create duplicate tests for this class especially?

src/manifests/input/input_manifest_1_0.py
zelinh: Replied above. ^^

src/build_workflow/builders.py
rishabh6788: Also do you think adding an optional field warrant a new minor version upgrade in the schema? 
What is the disadvantage of editing the existing schema and adding the optional `depends_on` field in the existing input manifest schema?
@zelinh @dblock

zelinh: > @zelinh Can you try making it `type(component) in ['InputComponentFromSource_1_0', 'InputComponentFromSource']` and try? Same for `Dist`

I don't think it's sustainable if we do this. We would need to add it each time if we have a new schema version of the component type.

zelinh: > Also do you think adding an optional field warrant a new minor version upgrade in the schema? What is the disadvantage of editing the existing schema and adding the optional `depends_on` field in the existing input manifest schema? @zelinh @dblock

Basically we are adding a new field in our input manifest so typically we would upgrade to a new schema version and keep the previous version in a sub-class. For example our `build_manifest` : https://github.com/opensearch-project/opensearch-build/tree/main/src/manifests/build.
If we are modifying the existing schema without incrementing a version, I'm worried that it may cause backward incompatibility in some unexpected cases on other workflows (e.g. build/ci workflow)

rishabh6788: > > @zelinh Can you try making it `type(component) in ['InputComponentFromSource_1_0', 'InputComponentFromSource']` and try? Same for `Dist`
> 
> I don't think it's sustainable if we do this. We would need to add it each time if we have a new schema version of the component type.

I am okay to approve this if there is no other way to fix apart from type-ignore.
@dblock LMK your thoughts.

zelinh: Tagging @gaiksaya for the concerns here.

dblock: I think either all `InputComponentFromDist*` classes need a common parent class (`InputComponentFromDist`), or you can explicitly check for `InputComponentFromDist` and `InputComponentFromDist_1_0` here?

dblock:
One thought. The `depends_on` feature is a net, optional, addition. Why did you decide to increment the schema version for it - can it be done as such on the existing schema?

src/build_workflow/builders.py
zelinh: @dblock I also thought about having a common parent class such as `InputComponentFromDist`. While the current parent class of `InputComponentFromDist/InputComponentFromDist_1_0` are `InputComponent` and `InputComponent_1_0` which inherit from generic `Component` so it's kinda the same situation for them. 
We could maybe create a new general class as the parent class but I think it may create more chaos. 

I think importing both `InputComponentFromDist` and `InputComponentFromDist_1_0` could work but in long term, we may have to add more and more class for type checking and it's not easy to maintain.

dblock (approved):
I tried to come up with a different approach, but didn't do better. This LGTM!

PR #4181, opened 2023-10-30
Add opensearch-cluster-cdk 1.x git branch support

rishabh6788:
### Description
We have a requirement to run benchmarks for OpenSearch 1.x versions. 
Since all the code for 1.x support in opensearch-cluster-cdk resides in `1.x` branch in opensearch-cluster-cdk repo, need to add support to select which branch to check-out during git repo checkout step in the workflow. It was hard-coded to `main` repo. 

### Issues Resolved

### DIFF
--- a/src/test_workflow/benchmark_test/benchmark_test_runner.py
+++ b/src/test_workflow/benchmark_test/benchmark_test_runner.py
@@ -29,7 +29,8 @@ class BenchmarkTestRunner(abc.ABC):
         else:
             self.security = False
 
-        self.tests_dir = os.path.join(os.getcwd(), "test-results", "benchmark-test", f"{'with' if self.security else 'without'}-security")
+        self.tests_dir = os.path.join(os.getcwd(), "test-results", "benchmark-test",
+                                      f"{'with' if self.security else 'without'}-security")
         os.makedirs(self.tests_dir, exist_ok=True)
 
     @abc.abstractmethod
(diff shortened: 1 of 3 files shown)


src/test_workflow/benchmark_test/benchmark_test_runner.py
```diff
@@ -38,3 +39,17 @@ def run_tests(self) -> None:
 
     def run(self) -> None:
         self.run_tests()
+
+    def get_git_ref(self) -> str:
+        if self.test_manifest:
+            os_major_version = self.test_manifest.version.split(".")[0]
+            if os_major_version in ['2', '3']:
+                return 'main'
```
peterzhuamazon, line 47: Is there a better way of handling this?
Hardcoding 2/3 as main branch feels hard to maintain over time.
Would appreciate a version property file or manifest to retrieve latest version vs others.
Or any other way of not hardcoding.

rishabh6788: The cluster cdk branch is tightly coupled to OpenSearch version, even if we move it to a property file or as such, this check would still be required to figure out for which version of OpenSearch we are running the benchmark against. 

Also, I do not see this changing in near future as we only have 3 major versions to maintain out of which 3.x is yet to be released next year and hopefully will be in development for next couple of years till 4.x comes along.

rishabh6788: Once we end support for 1.x, this whole logic can be removed, assuming that there will not be breaking change with respect to cluster bootstrapping.

rishabh6788: This could have been ` if os_major_version not in ['1']:` but since we have been asked to run becnhmark against 7.10.2 it becomes ` if os_major_version not in ['1', '7']:` which then doesn't make it better.

peterzhuamazon: Can you open an issue to track this removal of 1.x?

peterzhuamazon (approved):

Issue #4201, opened 2023-11-06
[Incremental Build] Differentiate the commits changes and finalize rebuild components


zelinh:
This is a sub-issue from https://github.com/opensearch-project/opensearch-build/issues/3922.

When enable incremental build, the build workflow should be able to:
  - Lock the references from input manifest to commits format.
  - Compare the components from input manifest to the local build manifest from previous build, and identify new addition components.
  - Find components with commit changes.
  - Go through the input manifest with `depends_on` field to locate any affecting dependency components when rebuilding.

Acceptance criteria:
  - Get a list of components for rebuilding. 
  - The components in the return list should include all modified plugins with commit changes, components depending on these modified component, and all plugin component if core(OpenSearch/OpenSearch-Dashboards) changes.

Exception:
For OpenSearch Dashboards, since we haven't enabled building components. https://github.com/opensearch-project/opensearch-build/issues/606. We would need to also include `OpenSearch-Dashboards` repo for rebuilding if any of the dashboard plugin changes.

zelinh:
There will be essentially three parts to achieve the goal here:

1. Implement a logic to get all components depending on a given plugin in the input manifests. 
```
   # Return a list of plugin that will depend on the given plugin
   def plugins_depend_on(self, plugin: str) -> List[str]:
```
2. Compare the input manifest and build manifest from previous build to find out new additional components and all components with commit changes.

3. Finalize components for rebuilding. This should include: new additional components; components with commit changes; all direct or indirect components that depend on previous two groups of components.

zelinh:
For more exception I'm thinking about:
- If the build manifest is not found from local folder with path e.g. `tar/builds/opensearch/manifest.yml`, we would build core with all other plugins from the input manifest.
- The build manifest within local build folder can come from previous version. We also need a check to verify that the build version of current input manifest matching the version of the build manifest from previous build. Otherwise, we would build core with all other plugins.

zelinh:
Closing this issue as it's completed.

PR #4207, opened 2023-11-10
Add commits differentiation logic for incremental build

zelinh:
### Description
This is a milestone for the incremental build. Two methods in the `run_incremental.py` class are used to find whatever components have updated commits from previous build and find whatever plugins depend on them according to the input manifest.

Those additions in `run_build.py` is a placeholder to further implementation on incremental build.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4201


gaiksaya (commented):
Taking a first pass at this PR. 

Thanks @zelinh . Is this PR a 2 part? Wondering what are we doing with plugins that needs to be rebuild? The code for incremental build in `run_build.py` is only listing the diffs as of now. Confused as what difference is it making from build point of view.

src/run_build.py
```diff
@@ -41,6 +42,12 @@ def main() -> int:
 
     output_dir = BuildOutputDir(manifest.build.filename, args.distribution).dir
 
+    if args.incremental:
+        buildIncremental = BuildIncremental(manifest, args.distribution)
+        list_of_updated_plugins = buildIncremental.commits_diff(manifest)
+        logging.info(f"Plugins for incremental build: {buildIncremental.rebuild_plugins(list_of_updated_plugins, manifest)}")
+        return 0
```
gaiksaya, line 49: Pointing the exact question from the overview here: Wondering what difference is this making in overall build.

zelinh: Yeah as I stated in the description of this issue, these additions in `run_build.py` is a placeholder to further implementation on incremental build. For this PR, it's intended for getting a list of components for rebuild.

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,66 @@
...
+class BuildIncremental:
+    def __init__(self, input_manifest: InputManifest, distribution: str):
+        self.distribution = distribution
+        self.input_manifest = input_manifest
+
+    # Given input manifest and return a list of what components changed and added.
+    def commits_diff(self, input_manifest: InputManifest) -> List[str]:
+        build_manifest_path = os.path.join(self.distribution, "builds", input_manifest.build.filename, "manifest.yml")
+        if not os.path.exists(build_manifest_path):
+            logging.info("Previous build manifest is not exists. Rebuilding Core.")
```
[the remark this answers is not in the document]
zelinh: It would rebuilding all components since core changed according to the logic in below.

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,66 @@
...
+                continue
+        return plugins
+
+    # Given updated plugins and look into the depends_on of all components to finalize a list of rebuilding components.
+    def rebuild_plugins(self, changed_plugins: List, input_manifest: InputManifest) -> List[str]:
+        if not changed_plugins:
+            return []
+
+        if any(core in changed_plugins for core in ("OpenSearch", "OpenSearch-Dashboards")):
+            logging.info("Core changes, rebuilding all.")
```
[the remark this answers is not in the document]
zelinh: Updated thanks.

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,66 @@
...
+            return [input_manifest.build.name.replace(" ", "-")]
+        plugins = []
+        for component in stable_input_manifest.components.select():
+            if component.name not in previous_build_manifest.components:
+                plugins.append(component.name)
+                logging.info(f"Adding {component.name} since it's not in previous build manifest")
+                continue
+            if component.ref != previous_build_manifest.components[component.name].commit_id:  # type: ignore[attr-defined]
+                plugins.append(component.name)
+                logging.info(f"Adding {component.name} because it has different commit ID and needs to be incremental built.")
```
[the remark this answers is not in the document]
zelinh: Updated in the latest commit.

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,66 @@
...
+        previous_build_manifest = BuildManifest.from_path(build_manifest_path)
+        stable_input_manifest = input_manifest.stable()
+        if previous_build_manifest.build.version != stable_input_manifest.build.version:
+            logging.info("The version of previous build manifest doesn't match the current input manifest. Rebuilding Core.")
+            return [input_manifest.build.name.replace(" ", "-")]
+        plugins = []
+        for component in stable_input_manifest.components.select():
+            if component.name not in previous_build_manifest.components:
+                plugins.append(component.name)
+                logging.info(f"Adding {component.name} since it's not in previous build manifest")
```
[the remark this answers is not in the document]
zelinh: Updated.

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class BuildIncremental:
+    def __init__(self, input_manifest: InputManifest, distribution: str):
+        self.distribution = distribution
+        self.input_manifest = input_manifest
+
+    # Given input manifest and return a list of what components changed and added.
+    def commits_diff(self, input_manifest: InputManifest) -> List[str]:
+        build_manifest_path = os.path.join(self.distribution, "builds", input_manifest.build.filename, "manifest.yml")
```
rishabh6788, line 23: I am assuming we will be downloading the build-manifest and storing it in the path mentioned before this logic executes?

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,64 @@
...
+        if not os.path.exists(build_manifest_path):
+            logging.info("Previous build manifest does not exist. Rebuilding Core.")
+            return [input_manifest.build.name.replace(" ", "-")]
+        previous_build_manifest = BuildManifest.from_path(build_manifest_path)
+        stable_input_manifest = input_manifest.stable()
+        if previous_build_manifest.build.version != stable_input_manifest.build.version:
```
rishabh6788, line 29: Why this check is here?
Could there be a case where the previous build manifest for an OS version differ from that in input manifest?
If yes, can you please give an example.

src/manifests/input_manifest.py
```diff
@@ -130,6 +130,13 @@ def stable(self) -> 'InputManifest':
         manifest.components.__stabilize__()
         return manifest
 
+    def plugins_depend_on(self, plugin: str) -> List[str]:
```
rishabh6788, line 133: Where is this being used?

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,64 @@
...
+                continue
+            if component.ref != previous_build_manifest.components[component.name].commit_id:  # type: ignore[attr-defined]
+                plugins.append(component.name)
+                logging.info(f"Adding {component.name} because it has different commit ID and needs to be rebuilt.")
+                continue
+        return plugins
+
+    # Given updated plugins and look into the depends_on of all components to finalize a list of rebuilding components.
+    def rebuild_plugins(self, changed_plugins: List, input_manifest: InputManifest) -> List[str]:
+        if not changed_plugins:
```
rishabh6788, line 46: May be add this check in the `run_build.py` and call this method only if you have plugins to rebuild as the method name suggests.

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,64 @@
...
+        for component in stable_input_manifest.components.select():
+            if component.name not in previous_build_manifest.components:
+                plugins.append(component.name)
+                logging.info(f"Adding {component.name} since it is missing from previous build manifest")
+                continue
+            if component.ref != previous_build_manifest.components[component.name].commit_id:  # type: ignore[attr-defined]
+                plugins.append(component.name)
+                logging.info(f"Adding {component.name} because it has different commit ID and needs to be rebuilt.")
+                continue
+        return plugins
```
rishabh6788, line 42: nit: rename it to `components` as it could be OpenSearch or Dashboards itself.

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,64 @@
...
+            return []
+
+        if any(core in changed_plugins for core in ("OpenSearch", "OpenSearch-Dashboards")):
+            logging.info("Core engine has new changes, rebuilding all components.")
+            return [component.name for component in input_manifest.components.select()]
+
+        queue = changed_plugins
+        rebuild_list = []
+
+        rebuild_list.append("OpenSearch-Dashboards") if input_manifest.build.filename == "opensearch-dashboards" else rebuild_list
```
rishabh6788, line 56: Did not understand this. 
Aren't we supposed to rebuild everything if core is present in the `changed_plugins` list at line 49?

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,64 @@
...
+        while queue:
+            plugin = queue.pop(0)
+            rebuild_list.append(plugin) if plugin not in rebuild_list else rebuild_list
```
rishabh6788, line 59: Why this?
Is there a possibility to have duplicates in the passed `changed_plugins` list?
If yes, shouldn't that be handled before it is being passed here.

src/build_workflow/build_incremental.py
```diff
@@ -0,0 +1,64 @@
...
+    # Given updated plugins and look into the depends_on of all components to finalize a list of rebuilding components.
+    def rebuild_plugins(self, changed_plugins: List, input_manifest: InputManifest) -> List[str]:
+        if not changed_plugins:
+            return []
+
+        if any(core in changed_plugins for core in ("OpenSearch", "OpenSearch-Dashboards")):
+            logging.info("Core engine has new changes, rebuilding all components.")
+            return [component.name for component in input_manifest.components.select()]
+
+        queue = changed_plugins
```
rishabh6788, line 53: seems redundant as you can use `changed_plugins` itself.

src/manifests/input_manifest.py
```diff
@@ -130,6 +130,13 @@ def stable(self) -> 'InputManifest':
...
+        plugins = []
+        for component in self.components.select():
+            if component.depends_on is not None and plugin in component.depends_on:
```
rishabh6788, line 136: This assumes that each component in the input_manifest has `depends_on` key. Is the the case?
If not, it should ideally check if the component has `depends_on` key and then process further. 
Also, please add test for this.

src/build_workflow/build_incremental.py
rishabh6788: Or for OSD plugins to rebuild, we need OSD to be rebuilt again?

src/run_build.py
```diff
@@ -41,6 +42,12 @@ def main() -> int:
 
     output_dir = BuildOutputDir(manifest.build.filename, args.distribution).dir
 
+    if args.incremental:
```
rishabh6788, line 45: please add test.

src/build_workflow/build_incremental.py
zelinh: This came out of my mind when I was testing on local. When we are building artifacts in local, we don't have any sub-folder for different version. All built artifacts stored in the same directory following this pattern: `<architecture>/builds/<opensearch/opensearch-dashboards>/<plugins/... or manifest.yml>`. Even we built input manifests of different versions. Latter build-manifest will overwrite the previous one but all artifacts would all be in place without any of removal. 

So I believe at least we need to check we are incrementally building the same version artifacts.

src/build_workflow/build_incremental.py
zelinh: Yes sure. This would be considered in the current implementation on the building process for incremental build.

src/build_workflow/build_incremental.py
zelinh: Yes. OSD doesn't have any those built-tool available just like OS. Every time we rebuild any of dashboards plugin, we have to build OSD core.

src/build_workflow/build_incremental.py
zelinh: I'm popping this queue so I was thinking to have a new list; while we can still have the original `changed_plugins` list unchanged.

src/build_workflow/build_incremental.py
zelinh: This is to make sure there is no duplicate in the `rebuild_list` which we would return.

src/manifests/input_manifest.py
zelinh: A little confused. It firstly check `depends_on` key is present and then check if the current component is depending on the given plugin.

src/run_build.py
zelinh: This is a placeholder for the next PR and it's just logging a list of plugins for rebuilding. I have complete tests for all those functions within `buildIncremental` class.

src/manifests/input_manifest.py
rishabh6788: Okay, I thought `component.depends_on is not None` is checking if the value of `depends_on` key. 
Please add the test.

src/run_build.py
rishabh6788: It still is part of the code flow and should be tested, if not in this PR then in subsequent PRs.

src/build_workflow/build_incremental.py
rishabh6788: Okay, I didn't see `changed_plugins` being used anywhere else.

src/run_build.py
zelinh: Will absolutely add tests in my following PR.

src/build_workflow/build_incremental.py
zelinh: Updated in the latest commit.

src/manifests/input_manifest.py
zelinh: Test is added in the latest commit.

rishabh6788 (approved):

PR #4224, opened 2023-11-18
Mark build stage as unstable if plugins failed to build

### LINKED ISSUE #4018: Enable different visualization color for staging if unstable

Currently we enable a new `-continue-on-error` into our distribution build; in this way, the build jenkins job won't stop if it fails on some plugins, but gives `unstable` at the end of the job results. 

However, looking into each stages inside the build jenkins job, all of the stages are green.
e.g.
Screenshot 2023-09-14 at 1 23 38 PM
It's hard for us to locate which stage actually fails unless we search through the entire console log which is extremely time-consuming.

Jenkins should have the option to change the stage status to `unstable` so it will display different color as yellow in Blue Ocean as well. 
https://www.jenkins.io/blog/2019/07/05/jenkins-pipeline-stage-result-visualization-improvements/

I'm suggesting if we are running distribution build job with `continue-on-error`, we should not only update the final job status to unstable but also the specific stage that has error. This will be more informative and easier for maintenance.

_No response_

### Additional context

_No response_

### Acceptance Criteria:
1. If a non-mandatory component failure happens at a stage level, that stage should be marked as `unstable`. 
2. The overall job status should be `unstable`.

gaiksaya:
### Description
The `continue-on-error` option does not throw an error or an non-zero exit code. Hence it comes difficult for component owners to parse jenkins build log to find the exactly failure. This PR marks the stage as unstable if plugins failed to build.

### Issues Resolved
resolves #4018 

### DIFF
--- a/jenkins/opensearch/distribution-build.jenkinsfile
+++ b/jenkins/opensearch/distribution-build.jenkinsfile
@@ -213,13 +213,19 @@ pipeline {
                         }
                     }
                     post {
+                        success {
+                            script {
+                                if (params.CONTINUE_ON_ERROR) {
+                                    markStageUnstableIfPluginsFailedToBuild()
+                                }
+                            }
+                        }
                         always {
                             script {
                                 lib.jenkins.Messages.new(this).add(
                                     "${STAGE_NAME}",
                                     lib.jenkins.Messages.new(this).get(["${STAGE_NAME}"])
                                 )
-                                markStageUnstableIfPluginsFailedToBuild()
                                 postCleanup()
                             }
                         }
@@ -258,11 +264,15 @@ pipeline {
                                 }
                             }
                             post {
-                                always {
+                                success {
                                     script {
-                                        markStageUnstableIfPluginsFailedToBuild()
-                                        postCleanup()
+                                        if (params.CONTINUE_ON_ERROR) {
+                                            markStageUnstableIfPluginsFailedToBuild()
+                                        }
                                     }
+                                }
+                                always {
+                                    postCleanup()
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


jenkins/opensearch/distribution-build.jenkinsfile
```diff
@@ -824,3 +836,10 @@ pipeline {
         }
     }
 }
+
+def markStageUnstableIfPluginsFailedToBuild() {
+    def stageLogs = getLogsForStage(stageName: "${STAGE_NAME}")
+    if (stageLogs.any{e -> e.contains('Failed plugins are')}) {
```
zelinh, line 842: Seems like we are reading through all of the logs from Jenkins stage even if there are no failed plugins right? 
I feel this might be really inefficient and delay the building process. Can we maybe modify the `--continue-on-error` from the build workflow so that it could send signal to whenever it skips one plugin?
It could be having a new variable storing a list of failed plugins or just a boolean saying we continued on error.

gaiksaya: The only time that string is thrown as an error is when continue on error is enabled. Otherwise it is not. I think it will be too many if/else statements to call this function if continue on error is enabled. 
Another work around is to only call this upon success as that when this error string is thrown. Will make that change. Making this PR as draft until dependent PRs are merged.

gaiksaya: Made the changes! The function is only called when `CONTINUE_ON_ERROR = true` and only for success stage. 
Thanks!

gaiksaya:
Hi @zelinh,
The related PRs have been merged. Can you please take another look?
Thanks!

zelinh (approved):
LGTM!

PR #4242, opened 2023-11-28
debian package: fix installation issues on systems having systemd installed but not running it as init.

automaticserver:
### Description
Instead of checking for the existence of systemd binaries check for actually running systemd, as seen by the reference result when using dh_installsystemd:

```bash
# Automatically added by dh_installsystemd/13.11.4
if [ -z "${DPKG_ROOT:-}" ] && [ "$1" = upgrade ] && [ -d /run/systemd/system ] ; then
	deb-systemd-invoke stop 'opensearch.service' >/dev/null || true
fi
# End automatically added section
```

This is only a fix based on the current implementation. I'm not changing the way systemctl is invoked.

### Issues Resolved
- This use case has not been reported yet exactly this way.

**This PR can be rejected if my upcoming proposal issue is considered** (will post link below as comment once created)

/dm

### DIFF
--- a/scripts/pkg/build_templates/opensearch-dashboards/deb/debian/postinst
+++ b/scripts/pkg/build_templates/opensearch-dashboards/deb/debian/postinst
@@ -20,12 +20,12 @@ log_dir=/var/log/opensearch-dashboards
 pid_dir=/var/run/opensearch-dashboards
 
 # Reload systemctl daemon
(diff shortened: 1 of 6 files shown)


automaticserver:
As announced see also #4243 to decide whether to keep this PR open or not

prudhvigodithi (approved):

PR #4250, opened 2023-11-29
Updates tar distribution to conform to changes in install demo configuration script in security plugin

DarshitChanpura:
### Description
Following up on the changes to demo configuration script in security plugin, https://github.com/opensearch-project/security/pull/3669, the OS supported distributions need to adapt to changes in the behavior of demo configuration script.

#### What has changed?

The `./opensearch-tar-install.sh` will now have these execution paths:
1. `OPENSEARCH_INITIAL_ADMIN_PASSWORD` value is set to a strong password: Demo configuration script execution succeeds
2. If weak password is provided in either of the cases, `tar-install.sh` execution will fail.
3. If no password is provided, the script will exit and installation will fail.

For detailed testing output, refer to [this comment](https://github.com/opensearch-project/security/issues/3634#issuecomment-1832676518) on the issue mentioned below.


scripts/startup/tar/linux/opensearch-tar-install.sh
```diff
@@ -10,7 +10,11 @@ cd $OPENSEARCH_HOME
 KNN_LIB_DIR=$OPENSEARCH_HOME/plugins/opensearch-knn/lib
 ##Security Plugin
 if [ -d "$OPENSEARCH_HOME/plugins/opensearch-security" ]; then
-        bash $OPENSEARCH_HOME/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s
+        echo "OpenSearch 2.12.0 onwards, the security plugin introduces a change that requires an initial password for 'admin' user."
+        echo "Please define an environment variable 'initialAdminPassword' with a password string."
+        echo "Or create a file 'initialAdminPassword.txt' with a single line that contains the password string and place it under $OPENSEARCH_PATH_CONF folder."
+        echo "If none of these are provided, a password will be generated and printed out."
+        bash $OPENSEARCH_HOME/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s || exit 1
```
peterzhuamazon, line 17: How do we handle the docker changes tho? As it uses the same demo scripts.
Especially that it is a bad idea to save a text file to have admin:admin as default.

peterzhuamazon: DEB/RPM as well.
This change here is simply exit out the startup if there is no password given, a breaking change.

rishabh6788:
Just to be clean, the user is still `admin` and only the password have to be provided, right?

scripts/startup/tar/linux/opensearch-tar-install.sh
DarshitChanpura: This PR focuses on TAR only. There will be subsequents PRs for different distributions.

prudhvigodithi:
Adding @derek-ho, looking this is having a problem to deal multi node cluster? Derek can you please add more details. Thank you

dblock:
Is the variable really `initialAdminPassword`?! All OpenSearch variables are OPENSEARCH_SOMETHING_SOMETHING, where did we deviate from this convention?

DarshitChanpura:
@prudhvigodithi @rishabh6788 Would you please review this?

rishabh6788:
I am okay with approving this change but we cannot merge this until security plugin changes have been back-ported to 2.x as it will start breaking 2.12.0 tests and benchmark runs.

DarshitChanpura:
> I am okay with approving this change but we cannot merge this until security plugin changes have been back-ported to 2.x as it will start breaking 2.12.0 tests and benchmark runs.

The backport PR is currently in draft mode, waiting couple of PRs to be merged into main: https://github.com/opensearch-project/security/pull/3845

DarshitChanpura:
https://github.com/opensearch-project/security/pull/3845 is merged and this PR is unblocked.

scripts/startup/tar/linux/opensearch-tar-install.sh
```diff
@@ -10,7 +10,10 @@ cd $OPENSEARCH_HOME
 KNN_LIB_DIR=$OPENSEARCH_HOME/plugins/opensearch-knn/lib
 ##Security Plugin
 if [ -d "$OPENSEARCH_HOME/plugins/opensearch-security" ]; then
-        bash $OPENSEARCH_HOME/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s
+        echo "OpenSearch 2.12.0 onwards, the OpenSearch Security Plugin introduces a change that requires an initial password for 'admin' user."
+        echo "Please define an environment variable 'OPENSEARCH_INITIAL_ADMIN_PASSWORD' with a strong password string."
+        echo "If a password is not provided, the setup will quit."
+        bash $OPENSEARCH_HOME/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s || exit 1
```
peterzhuamazon, line 16: After talking to @prudhvigodithi we should remove the exit 1, and add `set -e` to the install script, as the original design removed that due to `chmod 777 /dev/shm`, which has been removed some time ago.

peterzhuamazon (changes requested):

scripts/startup/tar/linux/opensearch-tar-install.sh
peterzhuamazon: More conversation with @DarshitChanpura and seems like `set -e` might have more impact on docker side. He will get more info on that and we will need some more discussion on it with @prudhvigodithi.

I am ok with either approach, just think it would be better if we are consistent across all distributions.

DarshitChanpura: set -e affects the entire file setup and analyzing that change is out of scope, IMO. @prudhvigodithi @peterzhuamazon please lmk if we should indeed use `set -e`. TAR is much simple to analyze and can be changed to use set -e, however, docker setup is bit more involved in terms of understanding all possible code paths. Windows however doesn't have `set -e` from what I understand, and hence will require some form of `|| exit /b 1`.

So, for consistency purposes I'd vote on keeping it as is with `|| exit 1`. This has been extensively tested on TAR, docker and Windows, and the setup works as expected with this change.

@prudhvigodithi @peterzhuamazon @rishabh6788 thoughts?

DarshitChanpura:
@peterzhuamazon @prudhvigodithi @rishabh6788 Could I please get reviews on this?

peterzhuamazon:
Need to change these as well if you plan to change docker-compose:
https://opensearch.org/samples/docker-compose.yml
https://opensearch.org/docs/latest/install-and-configure/install-opensearch/docker/#sample-docker-composeyml

DarshitChanpura:
> Also need @prudhvigodithi to confirm on @DarshitChanpura comment above regarding keeping the exit 1 for docker and tar, but remove for deb and rpm.

This PR only address TAR ball, no docker/windows/rpm/deb changes are present

DarshitChanpura:
@peterzhuamazon Could I get some re-reviews on this PR?

prudhvigodithi:
I'm ok with `exit 1`, even though having `set -e` is better as the execution codes are directly dictated by `install_demo_configuration.sh` but needs proper testing on how the script would behave (using `set -e`) with docker etc.
@peterzhuamazon add your final thought.

prudhvigodithi:
@DarshitChanpura can you please use as echo -e "This is a line with a new line.\nThis is the second line." to print statements in a new line. We dont need 3 echo lines.

DarshitChanpura:
echo -e maybe slightly unstable. I've used printf instead which is more stable for combining 3 lines with \n

DarshitChanpura:
@peterzhuamazon @prudhvigodithi @rishabh6788 Could I get some more reviews?

prudhvigodithi (approved):

scripts/startup/tar/linux/opensearch-tar-install.sh
```diff
@@ -10,7 +10,8 @@ cd $OPENSEARCH_HOME
 KNN_LIB_DIR=$OPENSEARCH_HOME/plugins/opensearch-knn/lib
 ##Security Plugin
 if [ -d "$OPENSEARCH_HOME/plugins/opensearch-security" ]; then
-        bash $OPENSEARCH_HOME/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s
+        printf "OpenSearch 2.12.0 onwards, the OpenSearch Security Plugin introduces a change that requires an initial password for 'admin' user. \nPlease define an environment variable 'OPENSEARCH_INITIAL_ADMIN_PASSWORD' with a strong password string. \nIf a password is not provided, the setup will quit."
```
peterzhuamazon, line 13: `echo -e` should be more posix compliant than `printf`.

DarshitChanpura:
@peterzhuamazon is there anything blocking this PR still?

peterzhuamazon (approved):

PR #4271, opened 2023-12-06
Allow manifest templates to include all the contents while changing version and ref based on core branch properties

### LINKED ISSUE #3707: Add placeholder plugins when auto-generating next version's manifests

Coming from https://github.com/opensearch-project/opensearch-plugins/pull/219#issuecomment-1621215855 the problem we'd like to solve is to know which plugins are going to be in some future release of OpenSearch. This is useful to, for example, run tests for the plugins on every PR into OpenSearch core (https://github.com/opensearch-project/OpenSearch/pull/8486).

1. Allow manifests to include placeholder component references. These would not be packaged, ie. simply skipped in all the packaging/testing/etc. workflows.

```
components:
  - name: OpenSearch
    ref: 2.x
  - name: common-utils
    repository: https://github.com/opensearch-project/common-utils.git
    ref: 2.x
    placeholder: true # this is not a real component!
```

2. When automatically creating a new manifest, copy all components from the previous manifest as placeholders.
3. Releasing should check that there are no placeholder components.

_No response_

### Additional context

_No response_

peterzhuamazon:
### Description
Allow manifest templates to include all the contents while changing version and ref based on core branch properties.

All changes:
1. Remove very old code which assumes 1.x is the only line with componentsfromsource.
2. Add all version new templates with proper list of components.
3. Remove component alternation in favor of directly copy contents from templates, then replace versions and refs before writing to disk.

Thanks.

### Issues Resolved
Closes #3707 


tests/jenkins/data/zip/builds/opensearch/manifest.yml
zelinh, line 1: Why do we need this build manifest?

src/manifests_workflow/input_manifests.py
```diff
@@ -95,76 +94,44 @@ def update(
 
             logging.info(f"Checking {self.name} {branches} branches")
             for branch in branches:
-                c = min_klass.checkout(
-                    path=os.path.join(work_dir.name, self.name.replace(" ", ""), branch),
+                min_component = min_klass.checkout(
```
rishabh6788, line 97: is it just for better readability?
Did not understand the motive behind this change? 
If I understand correctly, instead `OpenSearchDashboards`, this will be `OpenSearch-Dashboard`, right?

tests/jenkins/data/zip/builds/opensearch/manifest.yml
peterzhuamazon: Seems like it is created by the test.
Need removal.

src/manifests_workflow/input_manifests.py
peterzhuamazon: The `self.name.replace(" ", "")` field is just the folder name after the clone, really does not matter right here.
I just change it to something that is not containing spaces.

Thanks.

src/manifests_workflow/input_manifests.py
```diff
@@ -95,76 +94,44 @@ def update(
...
                 self.add_to_cron(release_version)
                 self.add_to_versionincrement_workflow(release_version)
 
-    def create_manifest(self, version: str, components: List = []) -> InputManifest:
+    def create_manifest(self, version: str, branch: str) -> InputManifest:
         templates_base_path = os.path.join(self.manifests_path(), "templates")
         template_version_folder = version.split(".")[0] + ".x"
         template_full_path = os.path.join(templates_base_path, self.prefix, template_version_folder, "manifest.yml")
         if not os.path.exists(template_full_path):
+            logging.info(f"Not found: {template_full_path}")
```
gaiksaya, line 118:
```suggestion
            logging.info(f"Manifest template not found: {template_full_path}")
```

src/manifests_workflow/input_manifests.py
```diff
@@ -95,76 +94,44 @@ def update(
...
-                        component_version = component.version
-                        if component_version:
-                            release_version = ".".join(component_version.split(".")[:3])
-                            if release_version not in main_versions.keys():
-                                main_versions[release_version] = []
-                            main_versions[release_version].append(component)
-                            logging.info(f"{component.name}#main is version {release_version} (from {component_version})")
-
-            # summarize
-            logging.info("Found versions on main:")
```
gaiksaya, line 135: Won't we be checking the versions on main now?

src/manifests_workflow/input_manifests.py
```diff
@@ -95,76 +94,44 @@ def update(
...
                 )
 
-                version = c.version
+                version = min_component.version
                 logging.info(f"{self.name}#{branch} is version {version}")
                 if version not in main_versions.keys():
-                    main_versions[version] = [c]
-
-            if component_klass is not None:
-                # components can increment their own version first without incrementing min
```
gaiksaya, line 109: Looks like this logic was not right and not useful that's why it was removed? since components won't be able to increment without min

src/manifests_workflow/input_manifests.py
rishabh6788: nit: `min_component_klass`?

src/manifests_workflow/input_manifests.py
rishabh6788: This was working but only limited to OpenSearch and OpenSearch-Dashboards, please correct me if I am wrong? @peterzhuamazon

rishabh6788:
Looks good but I have one query. 
Now that we are updating templates with all the components for the workflow to read, update the refs to 2.x branch and write it back for the upcoming release version. 
To make it even more simple, why can't we just read last released version input manifest, update the refs to 2.x and write to the new release version manifest? @peterzhuamazon 

Also, I did not see where we are confirming whether each component has updated the next release version on 2.x/1.x branch before we update the refs and check in the new input manifest? 
If this workflow runs just after 2.12 release, and if I understand the code flow correctly, this workflow will basically add 2.13 input manifest and will update the refs to 2.x (assuming OS/OSD have already updated the versions on their 2.x branch), but it may not be necessary that all the other components have updated the version on their 2.x?

peterzhuamazon:
Hi,

The original setups is assuming there is only 1.x version thus always use the latest one (aka latest patch) as the new template to add components, which is not suitable for now.

The original componentfromsource and componentfromdist is useless now as we are not using python to directly call http pull anymore since it is already getting replace with Zelin's incremental build. So that logic is uselss at this point and just add confusing.

The new logic is very simple. If called, get the major version to find the template, copy the entire template and replace version with version, ref with branch. If there is a new entry in the new release, then release manager just add it into the template.

If we dont like the above logic then I would suggest completely remove the template system written by Zelin, and just always pick the latest patch version from the same line.

Thanks.

rishabh6788:
> Hi,
> 
> The original setups is assuming there is only 1.x version thus always use the latest one (aka latest patch) as the new template to add components, which is not suitable for now because you cannot use 3.0.0 on 1.x or 2.x.
> 
> The original componentfromsource and componentfromdist is useless now as we are not using python to directly call http pull anymore since it is already getting replace with Zelin's incremental build. So that logic is uselss at this point and just add confusing.
> 
> The new logic is very simple. If called, get the major version to find the template, copy the entire template and replace version with version, ref with branch. If there is a new entry in the new release, then release manager just add it into the template.
> 
> If we dont like the above logic then I would suggest completely remove the template system written by Zelin, and just always pick the latest patch version from the same line.
> 
> Thanks.

What happens when all the other components have not yet updated release version on their respective major version branch, e.g., 2x, do we let the builds for next release version fail for those components?

It is fine for non-critical ones but in case CU or JS don't update their versions on their major branches then even though OS is up-to-date the overall build will keep failing till the critical components also update the release version on their major branches. 

If that is the case then we should make sure all the critical components, CU and JS along with Core, also update their versions immediately after release, the rest can follow later.

gaiksaya:
> It is fine for non-critical ones but in case CU or JS don't update their versions on their major branches then even though OS is up-to-date the overall build will keep failing till the critical components also update the release version on their major branches.
> 
> If that is the case then we should make sure all the critical components, CU and JS along with Core, also update their versions immediately after release, the rest can follow later.

+1 I believe even JS CU won't be able to upgrade as almost all plugin's CI dependent on min snapshot artifact and that is only generated once manifest is added to this repo. The flow in current situation looks like this:

1. OpenSearch repo bumps the version
2. [manifest-update.jenkinsFile](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/manifests-update.jenkinsFile) detects the change and creates a manifest in build
3. [check-for-build.jenkinsFile](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/check-for-build.jenkinsfile) detects the change and triggers [distribution-build-opensearch](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile) job which is responsible for triggering [publish-min-snapshot](https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/publish-min-snapshots.jenkinsfile) responsible for creating min snapshots that are used in plugin CIs.

Also important note that manifest checks will fail since the code will not be compiled as version increment PRs are not merged.

peterzhuamazon:
We have a different approach on this:
* https://github.com/opensearch-project/opensearch-build/issues/3707#issuecomment-1915482325

Issue #4210, opened 2023-11-13
[Incremental Build] Build incremental components through build workflow


zelinh:
This is a sub-issue from https://github.com/opensearch-project/opensearch-build/issues/3922

Once we have a list of rebuilding plugins obtained from [this issue](https://github.com/opensearch-project/opensearch-build/issues/4201), we should be able to run the build workflow with these components. 

Acceptance Critieria:
- The build workflow will run through only those designated components.
- The build manifest will be updated with rebuilding artifacts and keep unchanged artifacts from previous build.
- The build itself should run with no errors.

Implementation plan.
We would run our current build_workflow with these components until recording/generating the build_manifest.

Since we don’t want to erase other previously build components from the build manifest, we would take input of the build manifest with relative path `builds/opensearch/manifest.yml`. After reading through the previous manifest, we would be updating the manifest with whatever components that we just incremental build. 

We have `BuildManifestBuilder` class in `build_recorder.py`. Normally it starts from scratch; if we only build few components, it would only record those built plugins. 

However, we can add a functionality to the builder itself. When `incremental` argument is enable, it would firstly read through `builds/opensearch/manifest.yml` and append/update the currently building components to the manifest.

gaiksaya:
Hi @zelinh ,

Looks like associated PR is merged. Can we close this issue?
Thanks!

zelinh:
This issue is resolved.

PR #4289, opened 2023-12-14
Build incremental components through build workflow with loading previous build manifest

zelinh:
### Description
This is the last PR to enable incremental build on the Python workflow. 

The idea of this PR is to build those components that were detected by `rebuild_component` from `build_incremental` class. 

Firstly we add a new parameter to the `build_recorder` class so that it could load existing build manifest if provided; otherwise generate a brand new manifest.

Then similar to how we build in `run_build.py`, here for incremental build, we are importing the previous build manifest and only build the list of components from `buildIncremental.rebuild_plugins()`.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4210


zelinh:
I'm looking to add additional tests to fix the cov.

zelinh:
PR is ready for review. Seems like some issues with Maven so that the groovy test failed.

src/build_workflow/build_incremental.py
```diff
@@ -62,3 +68,56 @@ def rebuild_plugins(self, changed_plugins: List, input_manifest: InputManifest)
...
+                        failed_plugins.append(component.name)
+                        continue
+                    else:
+                        raise
+
+            build_recorder_incremental.write_manifest()
+
+        if len(failed_plugins) > 0:
+            logging.error(f"Failed plugins are {failed_plugins}")
+        logging.info("Done.")
```
gaiksaya, line 123: Confused as this entire block is duplicate of https://github.com/opensearch-project/opensearch-build/blob/main/src/run_build.py ? 
Can you explain why are you trying to achieve? 

Also, the build recorder object will always be either build or incremental_build. Accordingly you can initialize the object.

Adding @prudhvigodithi @dblock to take a look too.

gaiksaya (changes requested):

src/build_workflow/build_incremental.py
zelinh: Just had an offline discussion with @gaiksaya. I will try to squash the build logic here and those from `run_build.py` together so that it could save us from a lot of chaos.

src/run_build.py
```diff
@@ -45,8 +48,16 @@ def main() -> int:
...
+        components = buildIncremental.rebuild_plugins(list_of_updated_plugins, manifest)
+        logging.info(f"Plugins for incremental build: {components}")
+
+        build_manifest_path = os.path.join(args.distribution, "builds", manifest.build.filename, "manifest.yml")
+        if not os.path.exists(build_manifest_path):
+            logging.error(f"Previous build manifest missing at path: {build_manifest_path}")
+
+        logging.info(f"Build {components} incrementally.")
+
+        build_manifest_data = BuildManifest.from_path(build_manifest_path).__to_dict__()
```
dblock, line 60: Rename this to `build_manifest` and pass around as `BuildManifest`, rather than a dict.

dblock (commented):
This is pretty clean. See below, rename/simplify. Document this feature?

src/run_build.py
zelinh: Thanks. Working on the changes.

src/build_workflow/build_recorder.py
```diff
@@ -53,18 +53,25 @@ def write_manifest(self) -> None:
...
-            self.data["build"]["version"] = target.opensearch_version
-            self.data["build"]["platform"] = target.platform
-            self.data["build"]["architecture"] = target.architecture
-            self.data["build"]["distribution"] = target.distribution if target.distribution else "tar"
-            self.data["schema-version"] = "1.2"
             self.components_hash: Dict[str, Dict[str, Any]] = {}
 
+            if build_manifest:
+                build_manifest_data = build_manifest.__to_dict__()
+                self.data = build_manifest_data
```
dblock, line 62: I believe you should be able to iterate through components without turning the build manifest into a dictionary first, which would remove the need for the `build_manifest_data` variable.

zelinh: I think that would require to change the logics in other sections for example `to_manifest`  from `build_recorder` since it generates the build manifest based on dictionary. Do you think it's necessary or it may be overcomplicated.

zelinh:
> This is pretty clean. See below, rename/simplify. Document this feature?

@dblock Just added some documentations. Please help review it as well. Thanks!

src/build_workflow/build_recorder.py
dblock: Whatever you need to add to `build_manifest` is a good idea. It should look like this here:

```python
self.data = dict(build_manifest)
for component in build_manifest.components:
    self.components_hash[component.name] = dict(component)
```

zelinh: I see what you mean! Updated in the latest commit. Thanks for suggestion!

src/build_workflow/README.md
```diff
@@ -101,3 +102,15 @@ fi
...
+
+This functionality augments the existing build process by introducing the `--incremental` binary parameter.
+
+Sample command: `./build.sh manifests/2.12.0/opensearch-2.12.0.yml --incremental`.
+
+The build workflow will examine the build manifest from the previous build using the example path `tar/builds/opensearch/manifest.yml` when this command is executed.
+The build workflow will be executed in accordance with the comparison between the commits for each component in the preceding build manifest and the current input manifest.
+It will contain every modified component, and every component that relies on these revised components based on the `depends_on` entry in the input manifest.
+
+Once build is finished, a new build manifest will be generated using the previous build manifest as a reference, ensuring that all non-modified components remain unchanged.
```
gaiksaya, line 116: I think the the build manifest path is constant? `{distribution}/builds/opensearch/manifest.yml` Better to mention it that it will look for build manifest in that location.

Also worth mention that new artifacts will overwrite the previously built artifacts as well as manifest.

src/build_workflow/build_recorder.py
```diff
@@ -75,6 +81,7 @@ def append_component(self, name: str, version: str, repository_url: str, ref: st
                 "version": version,
             }
             self.components_hash[name] = component
+            logging.info(f"Appended {name} component in input manifest.")
```
gaiksaya, line 84: Should this be build manifest?

src/build_workflow/README.md
zelinh: Updated. Thanks.

src/build_workflow/build_recorder.py
zelinh: Good catch. Updated in the latest commit. Thanks.

src/run_build.py
```diff
@@ -63,11 +74,11 @@ def main() -> int:
             architecture=args.architecture or manifest.build.architecture,
         )
 
-        build_recorder = BuildRecorder(target)
+        build_recorder = BuildRecorder(target, build_manifest) if args.incremental else BuildRecorder(target)
 
         logging.info(f"Building {manifest.build.name} ({target.architecture}) into {target.output_dir}")
 
-        for component in manifest.components.select(focus=args.components, platform=target.platform):
+        for component in manifest.components.select(focus=components, platform=target.platform):
```
gaiksaya, line 81: Won't this disable building the individual components provided by `--component`?

zelinh: Yes, it would. We could have some enhancements discussed later to accommodate both `--component` and `--incremental`

gaiksaya: I don't think we should break the existing code compatibilty. All the GH issues created for build failures of any component recommend to build using `--component` args. Example: https://github.com/opensearch-project/alerting/issues/1326#issue-2036863756

In order to unblock and move forward we can do below:
1. Make `--component` and `--incremental` args mutually exclusive
2. Create an issue to accommodate the feature where if `--component <name> --incremental` is passed, it should compare the component from the previous manifest and using the diff make a decision to either rebuilt or skip.

peterzhuamazon:
Just reviewed with @zelinh on this PR.

I have no issues with the implementation and is clear on the details.
Regarding the conflict between `--component` and `--incremental`, is it possible to go these two routes:
1. If both parameters present, error out the build and provide information in console, that user can only use either parameter, not both at the same time.
2. If both parameters present, output a warning msg in the console, notify user that `--component` will not take effect when `--incremental` is presented at the same time.

The 1st route needs more implementation, while the 2nd route requires one more logging.
Please let me know what you think.


Thanks.

peterzhuamazon:
After discussing again among @gaiksaya @zelinh and I, we have decided to make both `--component` and `--incremental` mutually exclusive by following the solution here:
https://github.com/opensearch-project/opensearch-build/blob/main/src/validation_workflow/validation_args.py#L114

Later, we can make both parameters work together, in cases where user want to use `--incremental` function to verify if there is a change within the component(s) defined by `--component` parameter.

src/run_build.py
peterzhuamazon, line 81: If `component` is empty then `select()` will return all components here. Need to take care of it here.

zelinh: Just realize if components list is empty. This will rebuild all components. Looking into this.

peterzhuamazon: Need to exit if there is no component changed.

zelinh:
Updated based on what @peterzhuamazon @gaiksaya and I discussed earlier.
I make `--components` and `--incremental` mutually exclusive for now. We can discuss about the future enhancement later. 
If the rebuild list is empty, the whole build process will stop and skip building from everything.

Please help review it again. Thanks! @peterzhuamazon @gaiksaya @dblock

src/run_build.py
zelinh: Resolved in the latest commit. Please take a look thanks.

src/run_build.py
```diff
@@ -45,8 +48,20 @@ def main() -> int:
     if args.incremental:
         buildIncremental = BuildIncremental(manifest, args.distribution)
         list_of_updated_plugins = buildIncremental.commits_diff(manifest)
-        logging.info(f"Plugins for incremental build: {buildIncremental.rebuild_plugins(list_of_updated_plugins, manifest)}")
-        return 0
+        components = buildIncremental.rebuild_plugins(list_of_updated_plugins, manifest)
+        if not components:
+            logging.info("Nothing changes. Skip all.")
```
gaiksaya, line 53:
```suggestion
            logging.info("No commit difference found between any components. Skipping the build")
```

src/run_build.py
```diff
@@ -45,8 +48,20 @@ def main() -> int:
...
+            return 0
```
gaiksaya, line 54: Just confirming, this return will exit the entire build?

gaiksaya (commented):
Almost there!

Please add test for skipping the entire build and ensure the graceful exit.
Thanks!

src/run_build.py
zelinh: Yes. I tried locally. 
```
[opensearch@c81bdd48ecdb opensearch-build]$ ./build.sh manifests/2.12.0/opensearch-2.12.0-inc.yml --incremental
Installing dependencies from Pipfile.lock (fcf5b2)...
Running ./src/run_build.py manifests/2.12.0/opensearch-2.12.0-inc.yml --incremental ...
2024-01-04 00:26:44 INFO     Loading tar/builds/opensearch/manifest.yml
2024-01-04 00:26:44 INFO     Updating ref for https://github.com/opensearch-project/OpenSearch.git from 923aa25367006b830a0b66aca82eb9c47dafba22 to 923aa25367006b830a0b66aca82eb9c47dafba22 (923aa25367006b830a0b66aca82eb9c47dafba22)
2024-01-04 00:26:45 INFO     Updating ref for https://github.com/opensearch-project/common-utils.git from 2.x to 8e2537b6c1cde03449f0818242a1a36089a69522 (refs/heads/2.x)
2024-01-04 00:26:45 INFO     Updating ref for https://github.com/opensearch-project/job-scheduler.git from 2.x to 482c90fccf59a3e02bc5cb60124580c164a9d7c3 (refs/heads/2.x)
2024-01-04 00:26:45 INFO     Updating ref for https://github.com/opensearch-project/geospatial.git from 2.x to 273fa8fe051d469b5a79ee37a1ce203738239ddb (refs/heads/2.x)
2024-01-04 00:26:46 INFO     Updating ref for https://github.com/opensearch-project/security.git from 2.x to 2d7d6702c8aab1fbb25ab660f3432062fa564e10 (refs/heads/2.x)
2024-01-04 00:26:46 INFO     Nothing changes. Skip all.
```

zelinh:
> Almost there!
> 
> Please add test for skipping the entire build and ensure the graceful exit. Thanks!

Added new tests. I tried locally and it works.
```
[opensearch@c81bdd48ecdb opensearch-build]$ ./build.sh manifests/2.12.0/opensearch-2.12.0-inc.yml --incremental
Installing dependencies from Pipfile.lock (fcf5b2)...
Running ./src/run_build.py manifests/2.12.0/opensearch-2.12.0-inc.yml --incremental ...
2024-01-04 00:47:28 INFO     Loading tar/builds/opensearch/manifest.yml
2024-01-04 00:47:29 INFO     Updating ref for https://github.com/opensearch-project/OpenSearch.git from 923aa25367006b830a0b66aca82eb9c47dafba22 to 923aa25367006b830a0b66aca82eb9c47dafba22 (923aa25367006b830a0b66aca82eb9c47dafba22)
2024-01-04 00:47:29 INFO     Updating ref for https://github.com/opensearch-project/common-utils.git from 2.x to 8e2537b6c1cde03449f0818242a1a36089a69522 (refs/heads/2.x)
2024-01-04 00:47:29 INFO     Updating ref for https://github.com/opensearch-project/job-scheduler.git from 2.x to 482c90fccf59a3e02bc5cb60124580c164a9d7c3 (refs/heads/2.x)
2024-01-04 00:47:30 INFO     Updating ref for https://github.com/opensearch-project/geospatial.git from 2.x to 273fa8fe051d469b5a79ee37a1ce203738239ddb (refs/heads/2.x)
2024-01-04 00:47:30 INFO     Updating ref for https://github.com/opensearch-project/security.git from 2.x to 2d7d6702c8aab1fbb25ab660f3432062fa564e10 (refs/heads/2.x)
2024-01-04 00:47:30 INFO     No commit difference found between any components. Skipping the build
```

gaiksaya (approved):
LGTM! Great enhancement @zelinh 👏

src/run_build.py
gaiksaya: Hold on! 
`Running ./src/run_build.py manifests/2.12.0/opensearch-2.12.0-inc.yml --incremental ...`

Is the build manifest name renamed? Was this was just manual test?

zelinh: Yeah sorry for confusion. This is the manifest that I was testing on my local. Basically the same as our input manifest but with fewer components. Let me know if you have more questions. Thanks!

gaiksaya: Thanks! I think we are good.

peterzhuamazon (approved):
Thanks @zelinh , please open the improvement issue related to both parameters when you have time.

Thanks.

zelinh:
Issue created here as we discussed. https://github.com/opensearch-project/opensearch-build/issues/4320

PR #4274, opened 2023-12-08
Updates Docker and windows distribution build files to reflect to changes in admin user setup

DarshitChanpura:
### Description
As a follow-up of #4250, this PR updates docker and windows distribution install script to adapt to changes in demo configuration setup.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4302, opened 2023-12-21
Pass in strong password instead to fix integtest workflow

derek-ho:
### Description
With recent changes in security plugin requiring a strong initial admin password to be set, I've seen a few autocuts with integtest failures. This PR attempts to resolve those by passing in a strong password and using that to run the tests

### Issues Resolved

### DIFF
--- a/scripts/default/integtest.sh
+++ b/scripts/default/integtest.sh
@@ -89,7 +89,7 @@ fi
 
 if [ -z "$CREDENTIAL" ]
 then
-  CREDENTIAL="admin:admin"
+  CREDENTIAL="admin:myStrongPassword123!"
 fi
 
 USERNAME=`echo $CREDENTIAL | awk -F ':' '{print $1}'`
--- a/scripts/default/integtest.sh
+++ b/scripts/default/integtest.sh
@@ -89,7 +89,12 @@ fi
 
 if [ -z "$CREDENTIAL" ]
 then
-  CREDENTIAL="admin:myStrongPassword123!"
+    # OpenSearch 2.12 onwards security plugins requires a password to be set to setup admin user
+  if [ "$(echo "${VERSION} 2.12" | awk '{print ($1 >= $2)}')" -eq 1 ]; then
+    CREDENTIAL="admin:myStrongPassword123!"
+  else
+    CREDENTIAL="admin:admin"
+  fi
 fi
 
 USERNAME=`echo $CREDENTIAL | awk -F ':' '{print $1}'`
--- a/scripts/default/integtest.sh
+++ b/scripts/default/integtest.sh
@@ -87,10 +87,13 @@ then
   SNAPSHOT="false"
 fi
 
+ IFS='.' read -ra version_array <<< "$OPENSEARCH_VERSION"
+
+
 if [ -z "$CREDENTIAL" ]
 then
-    # OpenSearch 2.12 onwards security plugins requires a password to be set to setup admin user
-  if [ "$(echo "${VERSION} 2.12" | awk '{print ($1 >= $2)}')" -eq 1 ]; then
+  # Starting in 2.12.0, security demo configuration script requires an initial admin password
+  if (( ${version_array[0]} > 2 || (${version_array[0]} == 2 && ${version_array[1]} >= 12) )); then
     CREDENTIAL="admin:myStrongPassword123!"
   else
     CREDENTIAL="admin:admin"
--- a/scripts/default/integtest.sh
+++ b/scripts/default/integtest.sh
@@ -87,16 +87,15 @@ then
   SNAPSHOT="false"
 fi
 
- IFS='.' read -ra version_array <<< "$OPENSEARCH_VERSION"
-
-
+OPENSEARCH_REQUIRED_VERSION="2.12.0"
 if [ -z "$CREDENTIAL" ]
 then
   # Starting in 2.12.0, security demo configuration script requires an initial admin password
-  if (( ${version_array[0]} > 2 || (${version_array[0]} == 2 && ${version_array[1]} >= 12) )); then
-    CREDENTIAL="admin:myStrongPassword123!"
-  else
+  COMPARE_VERSION=`echo $OPENSEARCH_REQUIRED_VERSION $OPENSEARCH_VERSION | tr ' ' '\n' | sort -V | uniq | head -n 1`
+  if [ "$COMPARE_VERSION" != "$OPENSEARCH_REQUIRED_VERSION" ]; then
     CREDENTIAL="admin:admin"
+  else
+    CREDENTIAL="admin:myStrongPassword123!"
   fi
 fi
 
--- a/src/test_workflow/integ_test/distribution_deb.py
+++ b/src/test_workflow/integ_test/distribution_deb.py
@@ -30,6 +30,10 @@ class DistributionDeb(Distribution):
         logging.info("deb installation requires sudo, script will exit if current user does not have sudo access")
         deb_install_cmd = " ".join(
             [
+                'sudo'
+                'env'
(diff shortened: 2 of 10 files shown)


derek-ho:
Am I correct in assuming that we also need to change any repo which has its own integtest.sh file to use default admin:myStrongPassword123! with this change? How do we handle cases in which integtest pass in its own credentials, or we don't need to worry about those?

src/test_workflow/integ_test/distribution_deb.py
```diff
@@ -30,6 +30,10 @@ def install(self, bundle_name: str) -> None:
         logging.info("deb installation requires sudo, script will exit if current user does not have sudo access")
         deb_install_cmd = " ".join(
             [
+                'sudo'
+                'env'
+                'OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123!'
```
rishabh6788, line 35: This will only set the env variable for the duration of the `sudo env` command execution and will not persist. Recommend to use `export` to set the environment variable.

src/test_workflow/integ_test/distribution_rpm.py
```diff
@@ -30,6 +30,10 @@ def install(self, bundle_name: str) -> None:
         logging.info("rpm installation requires sudo, script will exit if current user does not have sudo access")
         rpm_install_cmd = " ".join(
             [
+                'sudo'
+                'env'
+                'OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123!'
```
rishabh6788, line 35: same as above.

src/test_workflow/integ_test/distribution_tar.py
```diff
@@ -33,7 +33,7 @@ def install(self, bundle_name: str) -> None:
     @property
     def start_cmd(self) -> str:
         start_cmd_map = {
-            "opensearch": "./opensearch-tar-install.sh",
+            "opensearch": "OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123! ./opensearch-tar-install.sh",
```
rishabh6788, line 36: I believe this should be `export OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123! && ./opensearch-tar-install.sh`

rishabh6788:
Added a few comments, please address them. 
Also please fix the failing tests.

scripts/default/integtest.sh
```diff
@@ -89,7 +89,7 @@ fi
 
 if [ -z "$CREDENTIAL" ]
 then
-  CREDENTIAL="admin:admin"
+  CREDENTIAL="admin:myStrongPassword123!"
```
rishabh6788, line 92: This wouldn't work for OS version lower than 2.12. 
It may not make much sense to run integration tests for released 2.x versions but it will definitely break integ tests for 1.x line.

gaiksaya: Right! The code is generic for all versions as of today. Until we solve the branching problem.

prudhvigodithi: To unblock this we can try adding if/else condition to differentiate with 1.x and above until we have this branching/versioning for the build repo solved. The script already has an argument `OPENSEARCH_VERSION` so adding if/else should be straight forward.

derek-ho: I think issue with that is what if patch version 2.0 < patch version < 2.12 come along? I was looking into bash script to compare semantic version, but it seems complex, not sure how do you folks feel about adding something like this: https://stackoverflow.com/questions/4023830/how-to-compare-two-strings-in-dot-separated-version-format-in-bash?

derek-ho: @rishabh6788 @gaiksaya @prudhvigodithi I changed the logic in the script itself to set a strong password if the version is after 2.12.0. Let me know if that makes sense to keep. I also included this logic in the integtest.sh files of repo's which have their own versions of integtest.sh.

scripts/default/integtest.sh
```diff
@@ -87,9 +87,17 @@ then
...
+
 if [ -z "$CREDENTIAL" ]
 then
-  CREDENTIAL="admin:admin"
+  # Starting in 2.12.0, security demo configuration script requires an initial admin password
+  if (( ${version_array[0]} > 2 || (${version_array[0]} == 2 && ${version_array[1]} >= 12) )); then
+    CREDENTIAL="admin:myStrongPassword123!"
+  else
+    CREDENTIAL="admin:admin"
+  fi
```
peterzhuamazon, line 100: 1. You are changing IFS without resetting it back to original
2. Why not use awk or cut with separator like on line 95/103?

peterzhuamazon (changes requested):
This is just the default integTest.sh, there are more in each plugin's repo.

derek-ho:
> This is just the default integTest.sh, there are more in each plugin's repo.

I believe most plugins have removed their custom integTest.sh in favor of this default. There are two repos I found which need their own custom ones, we have open PRs in those, which should be merged prior to 2.12.0 release, correct me if my understanding is incorrect:
https://github.com/opensearch-project/cross-cluster-replication/pull/1298
https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/981
If you can take a look at these, that would also be awesome! I do not know if their repo's branching strategy is honored. @kavilla suggested we don't need logic in these repos to determine version since those repo's integtest.sh is run against their own branch.

scripts/default/integtest.sh
derek-ho: Fixed as discussed with this commit: [7480dde](https://github.com/opensearch-project/opensearch-build/pull/4302/commits/7480dde2a05f95ac25de7c9644b7197239e1875f)

DarshitChanpura (approved):

peterzhuamazon (commented):
Waiting for final decision based on this issue before approval:
https://github.com/opensearch-project/security/issues/3916

prudhvigodithi (approved):

src/test_workflow/integ_test/distribution_tar.py
```diff
@@ -33,7 +33,7 @@ def install(self, bundle_name: str) -> None:
     @property
     def start_cmd(self) -> str:
         start_cmd_map = {
-            "opensearch": "./opensearch-tar-install.sh",
+            "opensearch": "export OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123! && ./opensearch-tar-install.sh",
```
prudhvigodithi, line 36: FYI, the `export OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123!` would be redundant for versions below 2.12.0, but should not block the installation.

peterzhuamazon (approved):

peterzhuamazon:
You can either add docker or open a new PR for docker:
https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/config/opensearch/opensearch-docker-entrypoint.sh#L33-L39

derek-ho:
> You can either add docker or open a new PR for docker: https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/config/opensearch/opensearch-docker-entrypoint.sh#L33-L39

Is this guy being used in docker integtest? Can you point me to where docker integtest is run, if at all? I would rather merge in this PR as is, since the main purpose is to sniff out if this will solve the autocuts that have been cut against a lot of repos, and close any false alarm that is being raised.

prudhvigodithi:
@derek-ho Merging the PR to make sure 2.12.0 `with-security` integ tests pass, please keep an eye on failures if there is anything breaking with this change.

DarshitChanpura:
> You can either add docker or open a new PR for docker:
https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/config/opensearch/opensearch-docker-entrypoint.sh#L33-L39

This is addressed in this PR: https://github.com/opensearch-project/opensearch-build/pull/4274

PR #4304, opened 2023-12-22
Support security demo script changes

rishabh6788:
### Description
There were recent changes to security demo configuration setup which now requires a custom admin Password to be setup for OpenSearch version 2.12.0 and above. 
This PR adds appropriate code changes to support the nightly benchmark runs. 

### Issues Resolved
https://github.com/opensearch-project/security/issues/3624

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -39,6 +39,7 @@ zipp = "~=3.8.1"
 importlib-metadata = "~=4.12.0"
 ruamel-yaml = "~=0.17.21"
 mistune = "~=3.0.1"
+semver = ">=3,<4"
 
 [dev-packages]
 
--- a/src/test_workflow/benchmark_test/benchmark_test_cluster.py
+++ b/src/test_workflow/benchmark_test/benchmark_test_cluster.py
@@ -14,6 +14,7 @@ from contextlib import contextmanager
 from typing import Any, Generator, Union
 
 import requests
+import semver
 from requests.auth import HTTPBasicAuth
 from retry.api import retry_call  # type: ignore
 
@@ -121,6 +122,7 @@ class BenchmarkTestCluster:
 
     def setup_cdk_params(self, config: dict) -> dict:
         suffix = ''
+        need_strong_password = False
         if self.args.stack_suffix and self.manifest:
             suffix = self.args.stack_suffix + '-' + self.manifest.build.id + '-' + self.manifest.build.architecture
         elif self.manifest:
@@ -132,8 +134,12 @@ class BenchmarkTestCluster:
             artifact_url = self.manifest.build.location if isinstance(self.manifest, BundleManifest) else \
                 f"https://artifacts.opensearch.org/snapshots/core/opensearch/{self.manifest.build.version}/opensearch-min-" \
                 f"{self.manifest.build.version}-linux-{self.manifest.build.architecture}-latest.tar.gz"
+            if not self.args.insecure and semver.compare(self.manifest.build.version, '2.12.0') != -1:
+                need_strong_password = True
         else:
             artifact_url = self.args.distribution_url.strip()
+            if not self.args.insecure and semver.compare(self.args.distribution_version, '2.12.0') != -1:
+                need_strong_password = True
 
         return {
             "distributionUrl": artifact_url,
@@ -142,6 +148,7 @@ class BenchmarkTestCluster:
             "region": config["Constants"]["Region"],
             "suffix": suffix,
             "securityDisabled": str(self.args.insecure).lower(),
+            "adminPassword": 'myStrongPassword123!' if need_strong_password else None,
             "cpuArch": self.manifest.build.architecture if self.manifest else 'x64',
             "singleNodeCluster": str(self.args.single_node).lower(),
             "distVersion": self.manifest.build.version if self.manifest else self.args.distribution_version,
--- a/src/test_workflow/benchmark_test/benchmark_test_suite.py
+++ b/src/test_workflow/benchmark_test/benchmark_test_suite.py
@@ -10,6 +10,8 @@ import os
 import subprocess
 from typing import Any
 
+import semver
+
 from test_workflow.benchmark_test.benchmark_args import BenchmarkArgs
 
 
@@ -19,6 +21,7 @@ class BenchmarkTestSuite:
     current_workspace: str
     args: BenchmarkArgs
     command: str
+    distribution_version: str
 
     """
     Represents a performance test suite. This class runs rally test on the deployed cluster with the provided IP.
@@ -28,11 +31,13 @@ class BenchmarkTestSuite:
             self,
             endpoint: Any,
             security: bool,
+            distribution_version: str,
             args: BenchmarkArgs,
     ) -> None:
         self.endpoint = endpoint
         self.security = security
         self.args = args
+        self.distribution_version = distribution_version
         # Pass the cluster endpoints with -t for multi-cluster use cases(e.g. cross-cluster-replication)
         self.command = 'docker run --rm'
         if self.args.benchmark_config:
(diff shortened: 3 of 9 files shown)


Pipfile
```diff
@@ -39,6 +39,7 @@ zipp = "~=3.8.1"
 importlib-metadata = "~=4.12.0"
 ruamel-yaml = "~=0.17.21"
 mistune = "~=3.0.1"
+semver = ">=3,<4"
```
gaiksaya, line 42: Should we fixate this to one particular version rather than between 3 and 4?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -132,8 +134,12 @@ def setup_cdk_params(self, config: dict) -> dict:
             artifact_url = self.manifest.build.location if isinstance(self.manifest, BundleManifest) else \
                 f"https://artifacts.opensearch.org/snapshots/core/opensearch/{self.manifest.build.version}/opensearch-min-" \
                 f"{self.manifest.build.version}-linux-{self.manifest.build.architecture}-latest.tar.gz"
+            if not self.args.insecure and semver.compare(self.manifest.build.version, '2.12.0') != -1:
+                need_strong_password = True
         else:
             artifact_url = self.args.distribution_url.strip()
+            if not self.args.insecure and semver.compare(self.args.distribution_version, '2.12.0') != -1:
+                need_strong_password = True
```
gaiksaya, line 142: Since both if and else conditions needs these lines of code maybe move it to line 136?

src/test_workflow/benchmark_test/benchmark_test_suite.py
```diff
@@ -65,7 +70,9 @@ def __init__(
                 self.command += f" --telemetry-params '{self.args.telemetry_params}'"
 
     def execute(self) -> None:
-        if self.security:
+        if self.security and semver.compare(self.distribution_version, '2.12.0') != -1:
+            self.command += ' --client-options="timeout:300,use_ssl:true,verify_certs:false,basic_auth_user:\'admin\',basic_auth_password:\'myStrongPassword123!\'"'
+        elif self.security:
```
gaiksaya, line 75: Why not pass the password as per version instead of repeating the command?
```suggestion
        if self.security:
           password = admin
            if semver.compare(self.distribution_version, '2.12.0') != -1:
               password = myStrongPassword123
           self.command += ' --client-options="timeout:300,use_ssl:true,verify_certs:false,basic_auth_user:\'admin\',basic_auth_password:\'${password}!\'"'
```

gaiksaya (commented):
Can you please make sure piplock is not updating all other versions as well? And only the necessary package is installed?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -132,8 +134,12 @@ def setup_cdk_params(self, config: dict) -> dict:
             artifact_url = self.manifest.build.location if isinstance(self.manifest, BundleManifest) else \
                 f"https://artifacts.opensearch.org/snapshots/core/opensearch/{self.manifest.build.version}/opensearch-min-" \
                 f"{self.manifest.build.version}-linux-{self.manifest.build.architecture}-latest.tar.gz"
+            if not self.args.insecure and semver.compare(self.manifest.build.version, '2.12.0') != -1:
+                need_strong_password = True
```
prudhvigodithi, line 138: @rishabh6788 looks to me like this only checks for 2.12.0 which is good, but anything above 2.12.0 need this `need_strong_password` right ?

Pipfile
rishabh6788: This is to make sure version 4 of semver is not installed. Copied from their website. See https://python-semver.readthedocs.io/en/latest/install.html

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: Did not get this. This condition is evaluated based on whether manifest has been provided or distribution url + version is provided.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: this checks if the passed version is equal to or grater then 2.12.0, if it is then strong password will be used. If not then this is ignored. Anything after 2.12.0 would anyway be using strong password.

src/test_workflow/benchmark_test/benchmark_test_suite.py
rishabh6788: makes sense. Updated.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: Optimization here wouldn't really yield better results to I kept it simple.

rishabh6788:
> Can you please make sure piplock is not updating all other versions as well? And only the necessary package is installed?

I only added the new version, I am not sure how I can stop pipenv to not update other dependencies. I believe we have appropriate version restrictions on each requirements and pipenv honors that while updating. Should be safe.

gaiksaya:
> > Can you please make sure piplock is not updating all other versions as well? And only the necessary package is installed?
> 
> I only added the new version, I am not sure how I can stop pipenv to not update other dependencies. I believe we have appropriate version restrictions on each requirements and pipenv honors that while updating. Should be safe.

Sometimes that break too. Mostly because of the underlying docker images during runs being different depending on the platform. We can merge this as if and monitor for any failures.

tests/tests_test_workflow/test_benchmark_workflow/benchmark_test/test_benchmark_test_suite.py
```diff
@@ -35,8 +35,19 @@ def test_execute_default(self) -> None:
...
+            self.assertEqual(mock_check_call.call_count, 1)
+            self.assertEqual(benchmark_test_suite.command,
+                             'docker run --rm opensearchproject/opensearch-benchmark:latest execute-test '
+                             '--workload=nyc_taxis --pipeline=benchmark-only '
+                             '--target-hosts=abc.com --client-options="timeout:300,use_ssl:true,'
+                             'verify_certs:false,basic_auth_user:\'admin\',basic_auth_password:\'myStrongPassword123!\'"')
+
     def test_execute_security_enabled(self) -> None:
-        benchmark_test_suite = BenchmarkTestSuite(endpoint=self.endpoint, security=True, args=self.args)
+        benchmark_test_suite = BenchmarkTestSuite(endpoint=self.endpoint, security=True, distribution_version='2.10.0', args=self.args)
```
gaiksaya, line 50: Can we try with `2.3.0`? 😅

gaiksaya (approved):
Looks like code cov is complaining about `need_strong_password`. Maybe just add a small assert for that? Looks good overall. Need to get similar change in for cluster cdk as well.

tests/tests_test_workflow/test_benchmark_workflow/benchmark_test/test_benchmark_test_suite.py
rishabh6788: :) Done!

rishabh6788:
> Looks like code cov is complaining about `need_strong_password`. Maybe just add a small assert for that? Looks good overall. Need to get similar change in for cluster cdk as well.

Yeah, it can nit-picky. The condition is actually tested in [tests/tests_test_workflow/test_benchmark_workflow/benchmark_test/test_benchmark_test_suite.py](https://github.com/opensearch-project/opensearch-build/pull/4304/files/5e1e40ee93bf0c21abc15d72be2d318c3cc74689#diff-96808c8cb670a5753525ec9f36da237ba3f8b06b1e395a62f1115b6153cd9775) where I am asserting a string password is present in the docker command.

PR #4327, opened 2024-01-05
Make OPENSEARCH_DASHBOARDS_HOME args works correctly in dashboards dockerfile

donggyu04:
### Description

1. Fix wrong comment.
2. Make `OPENSEARCH_DASHBOARDS_HOME` argument works correctly.


docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint-1.x.sh
```diff
@@ -176,7 +173,7 @@ function setupSecurityDashboardsPlugin {
 function runOpensearchDashboards {
     longopts=()
     for opensearch_dashboards_var in ${opensearch_dashboards_vars[*]}; do
-        # 'opensearch.hosts' -> 'OPENSEARCH_URL'
+        # 'opensearch.hosts' -> 'OPENSEARCH_HOSTS'
```
donggyu04, line 176: `opensearch.hosts` is converted to `OPENSEARCH_HOSTS`. `OPENSEARCH_URL` is not correct.

docker/release/config/opensearch-dashboards/opensearch-dashboards-docker-entrypoint-1.x.sh
```diff
@@ -15,9 +15,6 @@
 #
 #       --opensearch.startupTimeout=60
 
-# Setup Home Directory
-export OPENSEARCH_DASHBOARDS_HOME=/usr/share/opensearch-dashboards
```
donggyu04, line 19: I think this depends on `ARG OPENSEARCH_DASHBOARDS_HOME` in Dockerfile. so I added this to ENV at Dockerfile

PR #4342, opened 2024-01-12
Refactor utils into a common folder

derek-ho:
### Description
Converts version check of 2.12 and later to give myStrongPassword123! instead of admin to a common util function and converts all semver checks to use this util function.

### Issues Resolved


derek-ho:
@gaiksaya @prudhvigodithi is this what you folks had in mind? Please let me know and if not I can change this PR accordingly.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -143,12 +141,12 @@ def setup_cdk_params(self, config: dict) -> dict:
             artifact_url = self.manifest.build.location if isinstance(self.manifest, BundleManifest) else \
                 f"https://artifacts.opensearch.org/snapshots/core/opensearch/{self.manifest.build.version}/opensearch-min-" \
                 f"{self.manifest.build.version}-linux-{self.manifest.build.architecture}-latest.tar.gz"
-            if not self.args.insecure and semver.compare(self.manifest.build.version, '2.12.0') != -1:
-                need_strong_password = True
+            if not self.args.insecure:
+                password = 'myStrongPassword123!' if get_password(self.manifest.build.version) == 'myStrongPassword123!' else None
         else:
```
prudhvigodithi, line 146: `password = get_password(self.manifest.build.version)` having this should solve the purpose right @derek-ho ?
Adding @peterzhuamazon

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -143,12 +141,12 @@ def setup_cdk_params(self, config: dict) -> dict:
...
             artifact_url = self.args.distribution_url.strip()
-            if not self.args.insecure and semver.compare(self.args.distribution_version, '2.12.0') != -1:
-                need_strong_password = True
+            if not self.args.insecure:
+                password = 'myStrongPassword123!' if get_password(self.args.distribution_version) == 'myStrongPassword123!' else None
```
prudhvigodithi, line 150: Same as abobe.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
peterzhuamazon: Agreed with @prudhvigodithi, we can just assign the value of get_password to password.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -157,7 +155,7 @@ def setup_cdk_params(self, config: dict) -> dict:
             "region": config["Constants"]["Region"],
             "suffix": suffix,
             "securityDisabled": str(self.args.insecure).lower(),
-            "adminPassword": 'myStrongPassword123!' if need_strong_password else None,
+            "adminPassword": 'myStrongPassword123!' if password else None,
```
peterzhuamazon, line 158: Use get_password to get the password here instead of hardcoding?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -157,7 +155,7 @@ def setup_cdk_params(self, config: dict) -> dict:
...
             "cpuArch": self.manifest.build.architecture if self.manifest else 'x64',
```
prudhvigodithi, line 159: We should use the util `get_password` here, @peterzhuamazon WDYT?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
derek-ho: I was also confused about this, but it seems like this specific test is trying to set "myStrongPassword123!" or None depending on the version, instead of "myStrongPassword123!" or "admin". Thus, I had to make the usage of the util function in this way.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
derek-ho: I was also confused about this, but it seems like this specific test is trying to set "myStrongPassword123!" or None depending on the version, instead of "myStrongPassword123!" or "admin". Thus, I had to make the usage of the util function in this way.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
derek-ho: I was also confused about this, but it seems like this specific test is trying to set "myStrongPassword123!" or None depending on the version, instead of "myStrongPassword123!" or "admin". Thus, I had to make the usage of the util function in this way.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
derek-ho: I was also confused about this, but it seems like this specific test is trying to set "myStrongPassword123!" or None depending on the version, instead of "myStrongPassword123!" or "admin". Thus, I had to make the usage of the util function in this way.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
gaiksaya: Hey! so the clusters are deployed with and without security for performance runs. Thats why password is not always required. If required, pass the right one else `None`

src/test_workflow/benchmark_test/benchmark_test_cluster.py
dblock: The comparison with an actual password is strange. Why is this hard-coded and not configuration (which can be anything, including nothing)?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
dblock: This should look like `"adminPassword": admin_password`.

gaiksaya:
@derek-ho Do you want to move forward with this PR? Looks like it could be useful.

derek-ho:
> @derek-ho Do you want to move forward with this PR? Looks like it could be useful.

Just got back from vacation. I will hopefully get this PR fixed up by end of day.

Divyaasm:
Hi @derek-ho, can I know the status of the PR?

derek-ho:
@Divyaasm I think it is ready for review - I got kind of confused with the direction the build team wants me to go with the logic in the `benchmark_test_cluster.py` file, but everything else should be fine and I cleaned up that file to what makes sense to me. Can you leave some more feedback or merge if it looks good?

derek-ho:
@Divyaasm seems like you merged in something similar recently. Should I close this out?

Divyaasm:
Hi @derek-ho, The changes are to make src/test_workflow/integ_test/utils.py more flexible. You can still refactor the code in benchmark_test workflow to re-use the `get_password` method from utils.py

src/test_workflow/benchmark_test/benchmark_test_cluster.py
Divyaasm: @derek-ho you can simply omit this [block](https://github.com/opensearch-project/opensearch-build/blob/6cb96d5a3c28d07117952b66ded7f2c318fb342f/src/test_workflow/benchmark_test/benchmark_test_cluster.py#L115C9-L122C50) and just add a new line
password = get_password(self.args.distribution_version). This will return the desired value for given version and you can eliminate using semver module in the files. Add similar approach [here](https://github.com/opensearch-project/opensearch-build/blob/6cb96d5a3c28d07117952b66ded7f2c318fb342f/src/test_workflow/benchmark_test/benchmark_test_cluster.py#L146C1-L151C44)
https://github.com/opensearch-project/opensearch-build/blob/6cb96d5a3c28d07117952b66ded7f2c318fb342f/src/test_workflow/benchmark_test/benchmark_test_cluster.py#L160

Divyaasm: Feel free to move the location of utils.py as it is used even outside test_workflow

src/test_workflow/benchmark_test/benchmark_test_cluster.py
rishabh6788: Agree, just move all the processing to util and return final password.

peterzhuamazon:
Hi @derek-ho seems like this PR is outdated now.
Closing now and please feel free to re-open.

Thanks.

PR #4343, opened 2024-01-12
Ensuring cleanup with OSD plugins when failed during continue-on-error

peterzhuamazon:
### Description
Ensuring cleanup with OSD plugins when failed during continue-on-error

This should be a better solution than adding `--single-version loose` parameter.

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5675

### DIFF
--- a/scripts/default/opensearch-dashboards/build.sh
+++ b/scripts/default/opensearch-dashboards/build.sh
@@ -91,6 +91,7 @@ else
 fi
 
 mkdir -p $OUTPUT/plugins
+PLUGIN_NAME=$(basename "$PWD")
 # TODO: [CLEANUP] Needed OpenSearch Dashboards git repo to build the required modules for plugins
 # This makes it so there is a dependency on having Dashboards pulled already.
 cp -r ../$PLUGIN_NAME/ ../OpenSearch-Dashboards/plugins
--- a/scripts/default/opensearch-dashboards/build.sh
+++ b/scripts/default/opensearch-dashboards/build.sh
@@ -9,6 +9,10 @@
 
 set -ex
 
+# vars / libs
+. ../../../lib/shell/file_management.sh
+PLUGIN_PATH=$PWD
+
 function usage() {
     echo "Usage: $0 [args]"
     echo ""
@@ -21,6 +25,13 @@ function usage() {
     echo -e "-o OUTPUT\t[Optional] Output path, default is 'artifacts'."
     echo -e "-h help"
 }
+
+
+function cleanup_all() {
+    File_Delete $PLUGIN_PATH
+}
+
+trap cleanup_all TERM INT EXIT
 
 while getopts ":h:v:q:s:o:p:a:" arg; do
     case $arg in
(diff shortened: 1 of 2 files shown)


scripts/default/opensearch-dashboards/build.sh
```diff
@@ -22,6 +26,13 @@ function usage() {
     echo -e "-h help"
 }
 
+
+function cleanup_all() {
+    File_Delete $PLUGIN_PATH
+}
```
prudhvigodithi, line 32: Thanks @peterzhuamazon where is this `File_Delete` coming from?

peterzhuamazon: See the above imported lib of file_management.

prudhvigodithi: Got it, its part of https://github.com/opensearch-project/opensearch-build/blob/main/lib/shell/file_management.sh.

prudhvigodithi:
This would be definitely a better solution for build system rather than adding `--single-version loose parameter.`, but for a user who is not using the build system and building the OSD with plugins its easy to use `--single-version loose parameter.` with `yarn` cli (including in GH runners),  I think having both should be good. 
Adding @AMoo-Miki @bbarani @gaiksaya

peterzhuamazon:
> This would be definitely a better solution for build system rather than adding `--single-version loose parameter.`, but for a user who is not using the build system and building the OSD with plugins its easy to use `--single-version loose parameter.` with `yarn` cli (including in GH runners), I think having both should be good. Adding @AMoo-Miki @bbarani @gaiksaya

So @prudhvigodithi you think we can add both `--single-version loose` and this new cleanup section together?

Thanks.

prudhvigodithi:
> So @prudhvigodithi you think we can add both `--single-version loose` and this new cleanup section together?
> 
> Thanks.

If we have this PR merged then in the build code or in component build scripts we dont have to add `--single-version loose` and will continue to work as the clean up is happening now, but for a person who is building OSD or its plugins outside of this build code still can leverage `--single-version loose`.

gaiksaya:
> > So @prudhvigodithi you think we can add both `--single-version loose` and this new cleanup section together?
> > Thanks.
> 
> If we have this PR merged then in the build code or in component build scripts we dont have to add `--single-version loose` and will continue to work as the clean up is happening now, but for a person who is building OSD or its plugins outside of this build code still can leverage `--single-version loose`.

I don't think we need to take that into consideration right? We can add `--single-version loose` for opensearch-build. For any external build, they need to fix their CI/plugin code to accommodate this.

peterzhuamazon:
> > > So @prudhvigodithi you think we can add both `--single-version loose` and this new cleanup section together?
> > > Thanks.
> > 
> > 
> > If we have this PR merged then in the build code or in component build scripts we dont have to add `--single-version loose` and will continue to work as the clean up is happening now, but for a person who is building OSD or its plugins outside of this build code still can leverage `--single-version loose`.
> 
> I don't think we need to take that into consideration right? We can add `--single-version loose` for opensearch-build. For any external build, they need to fix their CI/plugin code to accommodate this.

That is my thought as well. On our side, if we cleanup properly even with `strict` mode it will still goes through. If other users want to use the param on their side, that is on them. We can, of course, support this addition as part of the build workflow tho, but not necessary.

I will open an issue to make that as an improvement.

prudhvigodithi (approved):

peterzhuamazon:
After talk with @prudhvigodithi , we will add `--single-version loose` to other workflows such as github actions in issue #4345 afterwards.

PR #4352, opened 2024-01-18
Add opensearch cluster stdout and stderr logs to test results manifest

### LINKED ISSUE #4335: [Bug]: Logs Not available in test-report
### Describe the bug

Coming from @peternied , logs from failed autocuts do not contain actual links or logs, instead showing as "Not Available"

### To reproduce

Download test report from autocut, example: https://ci.opensearch.org/ci/dbc/integ-test/2.12.0/9167/linux/x64/tar/test-results/6950/integ-test/test-report.yml](https://ci.opensearch.org/ci/dbc/integ-test/2.12.0/9167/linux/x64/tar/test-results/6950/integ-test/test-report.yml), you will see "Not available", instead of a link to the logs or failures

### Expected behavior

_No response_

### Host / Environment

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

rishabh6788:
### Description
This PR adds opensearch cluster stdout and stderr logs to test results manifest. Now the user will be able to see opensearch cluster logs for their component's integ test run. This will be helpful when the test results show as `N/A` and the users wants to know what went wrong. 

### Issues Resolved
#4335 

### DIFF
--- a/src/report_workflow/test_report_runner.py
+++ b/src/report_workflow/test_report_runner.py
@@ -7,6 +7,7 @@
 
 import logging
 import os
+import typing
 import urllib.request
 from typing import Any
 from urllib.error import HTTPError
@@ -100,6 +101,8 @@ class TestReportRunner:
                 component_yml_ref = "URL not available"
             config_dict["yml"] = component_yml_ref
             config_dict["status"] = test_result
+            config_dict["cluster_stdout"] = get_os_cluster_logs(self.base_path, str(self.test_run_id), self.test_type, component_name, config, self.name)[0]
+            config_dict["cluster_stderr"] = get_os_cluster_logs(self.base_path, str(self.test_run_id), self.test_type, component_name, config, self.name)[1]
             component["configs"].append(config_dict)
         return component
 
@@ -120,11 +123,14 @@ class TestReportRunner:
         return templates[template_type]
 
 
-def generate_component_yml_ref(base_path: str, test_number: str, test_type: str, component_name: str, config: str) -> str:
+def generate_component_yml_ref(base_path: str, test_number: str, test_type: str, component_name: str,
+                               config: str) -> str:
     if base_path.startswith("https://"):
-        return "/".join([base_path.strip("/"), "test-results", test_number, test_type, component_name, config, f"{component_name}.yml"])
+        return "/".join([base_path.strip("/"), "test-results", test_number, test_type, component_name, config,
+                         f"{component_name}.yml"])
     else:
-        return os.path.join(base_path, "test-results", test_number, test_type, component_name, config, f"{component_name}.yml")
+        return os.path.join(base_path, "test-results", test_number, test_type, component_name, config,
+                            f"{component_name}.yml")
 
 
 def generate_test_command(test_type: str, test_manifest_path: str, artifacts_path: str, component: str = "") -> str:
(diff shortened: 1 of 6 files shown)


src/report_workflow/test_report_runner.py
```diff
@@ -135,4 +137,17 @@ def generate_test_command(test_type: str, test_manifest_path: str, artifacts_pat
     return command
 
 
+def get_os_cluster_logs(base_path: str, test_number: str, test_type: str, component_name: str, config: str) -> List[str]:
+    os_stdout = ''
+    os_stderr = ''
+    if base_path.startswith("https://"):
+        os_stdout = "/".join([base_path.strip("/"), "test-results", test_number, test_type, component_name, config, "local-cluster-logs/id-0/stdout.txt"])
```
zelinh, line 144: I think this might not be `id-0` all the time. CCR team added this `id` for the cases that there are multiple clusters so we may need to iterate through all of them. https://github.com/opensearch-project/opensearch-build/pull/4097
This would also applied to some dashboards integ tests I believe.

zelinh (commented):
I'm thinking if the component yml file actually contains the local cluster logs if it's generated. (e.g. https://ci.opensearch.org/ci/dbc/integ-test/2.12.0/9207/linux/arm64/tar/test-results/7043/integ-test/alerting/without-security/alerting.yml)
Can we generate the component yml file in this case?

src/report_workflow/test_report_runner.py
rishabh6788: Just verified with past runs and here is the structure, for all the plugins except CCR the cluster id for `with-security` runs is `id-0` and `without-security` is `id-1`. I have updated the code for the same. 
For CCR, 4 node cluster is spun up, `with-security` has `id-0, id-1, id-2, id-3` and without has 4 - 7. I think it is fine for now wrt CCR as the logs would definitely come handy, even though it comes from one configuration run, in cases where OS process fails to come up.

zelinh: But this structure will also be used for dashboards plugins. From what I'm seeing, `with-security` will have `id-0` and `id-1`; `without-security` will have `id-2` & `id-3`. We might still want to have some approaches to accommodate for dashboards test.

src/report_workflow/test_report_runner.py
```diff
@@ -135,4 +137,19 @@ def generate_test_command(test_type: str, test_manifest_path: str, artifacts_pat
     return command
 
 
+def get_os_cluster_logs(base_path: str, test_number: str, test_type: str, component_name: str, config: str) -> List[str]:
+    os_stdout: str = ''
+    os_stderr: str = ''
+    cluster_id: str = 'id-0' if config == 'with-security' else 'id-1'
```
zelinh, line 143: This hard code id will not work for the dashboards cases since dashboards will include two local cluster logs; it uses 0&1 for with-security and 2&3 for without-security.

rishabh6788: Fixed this. The stdout and stderr fields in the test report yaml will now be a list showing appropriate cluster-ids for OS and OSD.

src/report_workflow/test_report_runner.py
rishabh6788: Fixed it. @zelinh

zelinh (approved):
LGTM so far though I think we should have a more comprehensive approach for multi-cluster test components instead of hardcoding cluster id. This fix should be helpful enough at this time and I'm approving to unblock. Thanks!

PR #4377, opened 2024-01-27
Add option to validate with/without security  and new password for latest versions

Divyaasm:
### Description
Changes to existing validation workflow
1. Checks for the existence of security plugin in OS artifacts and send https/http requests as required.
2. Use latest password for validating all the version >=2.12.0
 
### Issues Resolved

### DIFF
--- a/src/test_workflow/integ_test/utils.py
+++ b/src/test_workflow/integ_test/utils.py
@@ -5,12 +5,16 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+import base64
+
 import semver
 
 
-def get_password(version: str) -> str:
+def str_to_base64(value: str) -> str:
+    return base64.b64encode(value.encode("utf-8")).decode("utf-8")
+
+
+def get_password(version: str, convert_to_base64: bool = False) -> str:
     # Starting in 2.12.0, demo configuration setup script requires a strong password
-    if semver.compare(version, '2.12.0') != -1:
-        return "myStrongPassword123!"
-    else:
-        return "admin"
+    password = "myStrongPassword123!" if semver.compare(version, '2.12.0') != -1 else "admin"
+    return str_to_base64(password) if convert_to_base64 else password
--- a/src/validation_workflow/api_request.py
+++ b/src/validation_workflow/api_request.py
@@ -6,12 +6,11 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
-import base64
 from typing import Any
 
 import requests
 
-from test_workflow.integ_test.utils import get_password
+from test_workflow.integ_test.utils import get_password, str_to_base64
 
 """
 This class is to run API test againt on local OpenSearch API URL with default port 9200.
@@ -23,8 +22,7 @@ class ApiTest:
 
     def __init__(self, request_url: str, version: str) -> None:
         self.request_url = request_url
-        self.password = base64.b64encode(f"admin:{get_password(version)}".encode("utf-8")).decode("utf-8")
-        self.apiHeaders_auth = {"Authorization": f'Basic {self.password}'}  # user/pass "admin/pass" in Base64 format fetched from get_password() method
+        self.apiHeaders_auth = {"Authorization": f'Basic {str_to_base64("admin:" + get_password(version))}'}  # user/pass "admin/pass" in Base64 format fetched from get_password() method
         self.apiHeaders_accept = {"Accept": "*/*"}
         self.apiHeaders_content_type = {"Content-Type": "application/json"}
         self.apiHeaders = {}
--- a/src/validation_workflow/api_request.py
+++ b/src/validation_workflow/api_request.py
@@ -6,9 +6,12 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+import base64
 from typing import Any
 
 import requests
+
+from test_workflow.integ_test.utils import get_password
 
 """
 This class is to run API test againt on local OpenSearch API URL with default port 9200.
@@ -18,9 +21,10 @@ It returns response status code and the response content.
 
 class ApiTest:
 
-    def __init__(self, request_url: str) -> None:
+    def __init__(self, request_url: str, version: str) -> None:
         self.request_url = request_url
-        self.apiHeaders_auth = {"Authorization": "Basic YWRtaW46YWRtaW4="}  # default user/pass "admin/admin" in Base64 format
+        self.password = base64.b64encode(f"admin:{get_password(version)}".encode("utf-8")).decode("utf-8")
+        self.apiHeaders_auth = {"Authorization": f'Basic {self.password}'}  # user/pass "admin/pass" in Base64 format fetched from get_password() method
         self.apiHeaders_accept = {"Accept": "*/*"}
         self.apiHeaders_content_type = {"Content-Type": "application/json"}
         self.apiHeaders = {}
--- a/src/validation_workflow/api_test_cases.py
+++ b/src/validation_workflow/api_test_cases.py
@@ -21,7 +21,7 @@ class ApiTestCases:
         pass
 
     @staticmethod
-    def test_apis(version: str, projects: list, allow_with_security: bool = True) -> Any:
+    def test_apis(version: str, projects: list, security_plugin_exists: bool = True) -> Any:
         pass_counter, fail_counter = 0, 0
 
         # the test case parameters are formated as ['<request_url>',<success_status_code>,'<validate_string(optional)>']
@@ -30,7 +30,8 @@ class ApiTestCases:
             ['https://localhost:9200/_cat/plugins?v', 200, ''],
             ['https://localhost:9200/_cat/health?v', 200, 'green'],
         ]
-        if not allow_with_security:
+        # Use http in the request url if the security plugin is absent
+        if not security_plugin_exists:
             for api in test_apis:
                 api[0] = "http" + api[0][5:]  # type: ignore
 
--- a/src/validation_workflow/api_test_cases.py
+++ b/src/validation_workflow/api_test_cases.py
@@ -21,7 +21,7 @@ class ApiTestCases:
         pass
 
     @staticmethod
-    def test_apis(projects: list) -> Any:
+    def test_apis(version: str, projects: list, allow_with_security: bool = True) -> Any:
         pass_counter, fail_counter = 0, 0
 
         # the test case parameters are formated as ['<request_url>',<success_status_code>,'<validate_string(optional)>']
@@ -30,6 +30,10 @@ class ApiTestCases:
             ['https://localhost:9200/_cat/plugins?v', 200, ''],
             ['https://localhost:9200/_cat/health?v', 200, 'green'],
         ]
+        if not allow_with_security:
+            for api in test_apis:
+                api[0] = "http" + api[0][5:]  # type: ignore
+
         if ("opensearch-dashboards" in projects):
             test_apis.append(['http://localhost:5601/api/status', 200, ''])
 
@@ -38,7 +42,7 @@ class ApiTestCases:
             success_status_code = test_api.__getitem__(1)
             validate_string = test_api.__getitem__(2)
 
-            status_code, response_text = ApiTest(str(request_url)).api_get()
+            status_code, response_text = ApiTest(str(request_url), version).api_get()
             logging.info(f"\nRequest_url ->{str(request_url)} \n")
             logging.info(f"\nStatus_code ->{status_code} \nresponse_text ->{response_text}")
 
--- a/src/validation_workflow/api_test_cases.py
+++ b/src/validation_workflow/api_test_cases.py
@@ -23,17 +23,14 @@ class ApiTestCases:
     @staticmethod
     def test_apis(version: str, projects: list, security_plugin_exists: bool = True) -> Any:
         pass_counter, fail_counter = 0, 0
+        protocol_prefix = "https" if security_plugin_exists else "http"
 
         # the test case parameters are formated as ['<request_url>',<success_status_code>,'<validate_string(optional)>']
         test_apis = [
-            ['https://localhost:9200/', 200, ''],
-            ['https://localhost:9200/_cat/plugins?v', 200, ''],
-            ['https://localhost:9200/_cat/health?v', 200, 'green'],
+            [f'{protocol_prefix}://localhost:9200/', 200, ''],
+            [f'{protocol_prefix}://localhost:9200/_cat/plugins?v', 200, ''],
+            [f'{protocol_prefix}://localhost:9200/_cat/health?v', 200, 'green'],
         ]
-        # Use http in the request url if the security plugin is absent
-        if not security_plugin_exists:
-            for api in test_apis:
-                api[0] = "http" + api[0][5:]  # type: ignore
 
         if ("opensearch-dashboards" in projects):
             test_apis.append(['http://localhost:5601/api/status', 200, ''])
--- a/src/validation_workflow/rpm/validation_rpm.py
+++ b/src/validation_workflow/rpm/validation_rpm.py
@@ -70,7 +70,7 @@ class ValidateRpm(Validation, DownloadUtils):
         return True
 
     def validation(self) -> bool:
-        test_result, counter = ApiTestCases().test_apis(self.args.version, self.args.projects, self.check_for_security_plugin(os.path.join(os.sep, "usr", "share", "opensearch"), "rpm") if not self.args.force_https_check else True)  # noqa: E501
+        test_result, counter = ApiTestCases().test_apis(self.args.version, self.args.projects, self.check_for_security_plugin(os.path.join(os.sep, "usr", "share", "opensearch")) if not self.args.force_https else True)  # noqa: E501
         if (test_result):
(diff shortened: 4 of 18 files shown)


src/validation_workflow/rpm/validation_rpm.py
```diff
@@ -47,7 +48,10 @@ def installation(self) -> bool:
             for project in self.args.projects:
                 self.filename = os.path.basename(self.args.file_path.get(project))
                 execute(f'sudo yum remove {project} -y', ".")
-                execute(f'sudo rpm -ivh {os.path.join(self.tmp_dir.path, self.filename)}', str(self.tmp_dir.path), True, False)
+                execute(f'sudo env OPENSEARCH_INITIAL_ADMIN_PASSWORD={get_password(str(self.args.version))} rpm -ivh {os.path.join(self.tmp_dir.path, self.filename)}', str(self.tmp_dir.path), True, False)  # noqa: 501
+
+            if self.args.allow_without_security:
+                self.args.allow_without_security = self.test_security_plugin("/usr/")
```
peterzhuamazon, line 54: Why is it different from tar?

src/validation_workflow/validation.py
```diff
@@ -38,6 +39,14 @@ def copy_artifact(self, filepath: str, tempdir_path: str) -> bool:
...
             raise Exception("Provided path for local artifacts does not exist")
 
+    def test_security_plugin(self, work_dir: str) -> bool:
+        (_, stdout_1, _) = execute(f'find {work_dir} -type f -iname \'opensearch-plugin\'', ".", True, False)
+        if (stdout_1):
+            (_, stdout_2, _) = execute("./opensearch-plugin list", stdout_1.replace("opensearch-plugin", "").rstrip("\n"), True, False)
+            return "opensearch-security" in stdout_2
+        else:
+            raise Exception("Couldn't fetch the path to plugin folder")
+
```
peterzhuamazon, line 49: Can we use better env var names, what is stdout_1 stdout_2?

src/validation_workflow/validation_args.py
```diff
@@ -109,7 +109,14 @@ def __init__(self) -> None:
...
             default="production",
-            dest="artifact_type",
+            dest="artifact_type"
+        )
+        parser.add_argument(
+            "-s",
+            "--allow-without-security",
+            action="store_true",
+            default=False,
+            help="Allow Validation without security"
```
peterzhuamazon, line 119: This is not a good design here.
-s should be parameter for things like SECURITY_ENABLED (true w/ sec, false w/o sec, default true), instead of `--allow-without-security`, which is very weird and easily get confused.

src/validation_workflow/yum/validation_yum.py
```diff
@@ -53,7 +54,9 @@ def installation(self) -> bool:
                 logging.info('Removed previous versions of Opensearch')
                 urllink = f"{self.args.file_path.get(project)} -o /etc/yum.repos.d/{os.path.basename(self.args.file_path.get(project))}"
                 execute(f'sudo curl -SL {urllink}', ".")
-                execute(f"sudo yum install '{project}-{self.args.version}' -y", ".")
+                execute(f"sudo env OPENSEARCH_INITIAL_ADMIN_PASSWORD={get_password(str(self.args.version))} yum install '{project}-{self.args.version}' -y", ".")
+            if self.args.allow_without_security:
+                self.args.allow_without_security = self.test_security_plugin("/usr/")
```
peterzhuamazon, line 59: If this is keep repeating can we move it to parent class?

peterzhuamazon (changes requested):

src/validation_workflow/api_request.py
```diff
@@ -18,9 +21,10 @@
 
 class ApiTest:
 
-    def __init__(self, request_url: str) -> None:
+    def __init__(self, request_url: str, version: str) -> None:
         self.request_url = request_url
-        self.apiHeaders_auth = {"Authorization": "Basic YWRtaW46YWRtaW4="}  # default user/pass "admin/admin" in Base64 format
+        self.password = base64.b64encode(f"admin:{get_password(version)}".encode("utf-8")).decode("utf-8")
+        self.apiHeaders_auth = {"Authorization": f'Basic {self.password}'}  # user/pass "admin/pass" in Base64 format fetched from get_password() method
```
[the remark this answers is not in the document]
Divyaasm: Used this logic without altering https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/utils.py

src/validation_workflow/api_test_cases.py
```diff
@@ -30,6 +30,10 @@ def test_apis(projects: list) -> Any:
             ['https://localhost:9200/_cat/plugins?v', 200, ''],
             ['https://localhost:9200/_cat/health?v', 200, 'green'],
         ]
+        if not allow_with_security:
+            for api in test_apis:
+                api[0] = "http" + api[0][5:]  # type: ignore
```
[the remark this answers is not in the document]
Divyaasm: Request `http` if security plugin is absent else do `https` request when security plugin is present. I'll add a comment.

src/validation_workflow/rpm/validation_rpm.py
```diff
@@ -68,8 +72,8 @@ def start_cluster(self) -> bool:
         return True
 
     def validation(self) -> bool:
-        test_result, counter = ApiTestCases().test_apis(self.args.projects)
-        if (test_result):
+        test_result, counter = ApiTestCases().test_apis(self.args.version, self.args.projects, self.args.allow_without_security)
+        if(test_result):
```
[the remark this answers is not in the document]
Divyaasm: Fixed

src/validation_workflow/rpm/validation_rpm.py
Divyaasm: I tarball, the opensearch-plugin is present in extracted path which is inside tmp folder. For rpm and yum it's inside /usr/share/opensearch/...

src/validation_workflow/validation.py
Divyaasm: Just went with stdout default stdout, will change the names. Thanks

src/validation_workflow/validation_args.py
Divyaasm: Used -s  as it is related to security, when the above argument is provided by user - we allow validation of artifacts without security plugin, by checking the presence of security plugin in the installed plugins.

src/validation_workflow/yum/validation_yum.py
Divyaasm: We might have to pass the working directory from child class only if the allow_without_security is passed by the user.

src/validation_workflow/api_request.py
```diff
@@ -6,10 +6,13 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+import base64
```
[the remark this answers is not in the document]
Divyaasm: Yeah required to convert the password.I don't think think so

src/validation_workflow/validation.py
```diff
@@ -38,6 +41,17 @@ def copy_artifact(self, filepath: str, tempdir_path: str) -> bool:
         else:
             raise Exception("Provided path for local artifacts does not exist")
 
+    def test_security_plugin(self, work_dir: str) -> bool:
+        (_, path, _) = execute(f'find {work_dir} -type f -iname \'opensearch-plugin\'', ".", True, False)
+        if (path):
+            (_, list_plugins, _) = execute("./opensearch-plugin list", path.replace("opensearch-plugin", "").rstrip("\n"), True, False)
```
gaiksaya, line 47: this should be `./bin.opensearch-plugin list` right?

Divyaasm: Yeah Sayali, but `find` retreives the filepath till `../bin/opensearch-plugin` , Just removed the `opensearch-plugin`, so now it should be in bin folder. The command executes perfectely.

src/validation_workflow/api_request.py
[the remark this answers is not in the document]
peterzhuamazon: I am not saying you need to alter, I am saying you can add a second version in return of base64 version. Instead of writing a base64 conversion within validation workflow.

src/validation_workflow/validation.py
peterzhuamazon: Can you explain what you are trying to do here?
Trying to find opensearch-plugin location? Or trying to find the plugin folder, because your line 50 `            raise Exception("Couldn't fetch the path to plugin folder")` indication you are looking for the plugin folder.

src/validation_workflow/validation_args.py
peterzhuamazon: I mean to say change `--allow-without-security` to something like `--enable-security` or `--security-enabled` instead of the other way around. True should be security enabled and false should be security disabled, instead of true meaning without security.

src/validation_workflow/yum/validation_yum.py
peterzhuamazon: The does not stop you from moving it to parent class?

src/validation_workflow/validation_args.py
peterzhuamazon: After discussing with @Divyaasm on this topic, it is clear now that what she wants is a parameter such as `--force-https-check`.

If this option is false it will check for security plugin, if present use https, else use http.
If this option is true it will always use https regardless if security plugin present.

It is an option for early release builds that security plugin is not available.

Suggesting changing the name from `--allow-without-security` to `--force-https-check` and from `-s` to `-f`.

Thanks.

src/validation_workflow/yum/validation_yum.py
peterzhuamazon: After talking to @Divyaasm on this.
I think she should not re-use `self.args.allow_without_security` and assign a different bool value.
Suggest just using a different var to save the output of `self.test_security_plugin("/usr/")`.

src/validation_workflow/validation.py
peterzhuamazon: After talking to @Divyaasm it is clear that she wants to find opensearch-plugin utilility then use it to find security plugin.

The `find` method is not efficient and is not the correct way to get the location of `opensearch-plugin`, especially that windows is using a different variation of `find` that does not run with gnu syntax.

Suggesting following https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/distribution_rpm.py or similar to implement abstract method to find the correct opensearch-plugin path, or even security plugin folder directly in child class.

Thanks.

peterzhuamazon: Checking $OPENSEARCH_HOME/plugins/opensearch-security might be an easier way so you do not need to execute any command.

src/validation_workflow/api_test_cases.py
```diff
@@ -30,6 +30,11 @@ def test_apis(projects: list) -> Any:
             ['https://localhost:9200/_cat/plugins?v', 200, ''],
             ['https://localhost:9200/_cat/health?v', 200, 'green'],
         ]
+        # Use http in the request url if the security plugin is absent
+        if not security_plugin_exists:
+            for api in test_apis:
+                api[0] = "http" + api[0][5:]  # type: ignore
```
peterzhuamazon, line 36: I understand what this is now, as @Divyaasm is trying to replace the first 5 char of `https` with `http`.

In that case, it is better to do a string concatenation such as:

```
protocol_prefix = "https" if security_plugin_exisits else "http"
test_apis = protocol_prefix + body

```

src/validation_workflow/api_request.py
```diff
@@ -18,9 +20,9 @@
 
 class ApiTest:
 
-    def __init__(self, request_url: str) -> None:
+    def __init__(self, request_url: str, version: str) -> None:
         self.request_url = request_url
-        self.apiHeaders_auth = {"Authorization": "Basic YWRtaW46YWRtaW4="}  # default user/pass "admin/admin" in Base64 format
+        self.apiHeaders_auth = {"Authorization": f'Basic {str_to_base64("admin:" + get_password(version))}'}  # user/pass "admin/pass" in Base64 format fetched from get_password() method
```
peterzhuamazon, line 25: 👍

src/test_workflow/integ_test/utils.py
```diff
@@ -5,12 +5,16 @@
...
+
+
+def get_password(version: str, convert_to_base64: bool = False) -> str:
     # Starting in 2.12.0, demo configuration setup script requires a strong password
-    if semver.compare(version, '2.12.0') != -1:
-        return "myStrongPassword123!"
-    else:
-        return "admin"
+    password = "myStrongPassword123!" if semver.compare(version, '2.12.0') != -1 else "admin"
+    return str_to_base64(password) if convert_to_base64 else password
```
peterzhuamazon, line 20: 👍

src/validation_workflow/api_test_cases.py
```diff
@@ -21,15 +21,17 @@ def __init__(self) -> None:
...
+        protocol_prefix = "https" if security_plugin_exists else "http"
 
         # the test case parameters are formated as ['<request_url>',<success_status_code>,'<validate_string(optional)>']
         test_apis = [
-            ['https://localhost:9200/', 200, ''],
-            ['https://localhost:9200/_cat/plugins?v', 200, ''],
-            ['https://localhost:9200/_cat/health?v', 200, 'green'],
+            [f'{protocol_prefix}://localhost:9200/', 200, ''],
+            [f'{protocol_prefix}://localhost:9200/_cat/plugins?v', 200, ''],
+            [f'{protocol_prefix}://localhost:9200/_cat/health?v', 200, 'green'],
```
peterzhuamazon, line 32: 👍

src/validation_workflow/validation.py
```diff
@@ -38,6 +41,14 @@ def copy_artifact(self, filepath: str, tempdir_path: str) -> bool:
         else:
             raise Exception("Provided path for local artifacts does not exist")
 
+    def check_for_security_plugin(self, work_dir: str, distribution: str) -> bool:
+        list_cmd = "dir" if distribution == "zip" else "ls"
+        (_, plugins_list, _) = execute(f"{list_cmd} plugins", work_dir, True, False)
+        return "opensearch-security" in plugins_list
```
peterzhuamazon, line 47: Can we just directly check if dir `$OPENSEARCH_HOME/plugins/opensearch-security` exist such as
```
path = '$OPENSEARCH_HOME/plugins/opensearch-security'
 
isExist = os.path.exists(path)
```

This is much easier than running ls/dir and have to specify different OS.

src/validation_workflow/validation.py
```diff
@@ -38,6 +41,14 @@ def copy_artifact(self, filepath: str, tempdir_path: str) -> bool:
...
+
+    def get_version(self, project: str) -> str:
+        return re.search(r'(\d+\.\d+\.\d+)', os.path.basename(project)).group(1)
```
peterzhuamazon, line 50: What's this for?

src/validation_workflow/validation_args.py
```diff
@@ -109,7 +109,14 @@ def __init__(self) -> None:
...
             default="production",
-            dest="artifact_type",
+            dest="artifact_type"
+        )
+        parser.add_argument(
+            "-f",
+            "--force-https-check",
+            action="store_true",
+            default=True,
+            help="If False then check for the existence of security plugin and use http/https as required, else force to check to https"
```
peterzhuamazon, line 119: 👍 
But the help msg can be fine-tuned a bit such as:
```
If False, use http/https to connect based on the existence of security plugin, else always use https, default to True
```

peterzhuamazon (commented):
A few more comments and we all good @Divyaasm .

Thanks

src/validation_workflow/validation.py
Divyaasm: Hey @peterzhuamazon. This will cover the case when only url's are provided as input, in all other cases version is provide as an argument. It will fetch the version from the url which will used in get_password(version) method.

peterzhuamazon: Is Regex here better than just do a split and check the fields?

src/validation_workflow/validation_args.py
peterzhuamazon: And now I think the long param name can be further reduce to `--force-https`.

src/validation_workflow/validation.py
```diff
@@ -38,6 +40,13 @@ def copy_artifact(self, filepath: str, tempdir_path: str) -> bool:
         else:
             raise Exception("Provided path for local artifacts does not exist")
 
+    def check_for_security_plugin(self, work_dir: str) -> bool:
+        path = os.path.exists(os.path.join(work_dir, "plugins", "opensearch-security"))
```
peterzhuamazon, line 44: 👍

src/validation_workflow/validation.py
peterzhuamazon: After discussion, @Divyaasm confirms we can keep this implementation for now.

src/validation_workflow/validation_args.py
peterzhuamazon: Thanks for changing.

peterzhuamazon (approved):

peterzhuamazon:
Thanks for following the best practice and improve code quality!
@Divyaasm

PR #4402, opened 2024-02-05
feat: remove useless config

SuZhou-Joe:
### Description
- Chatbot and Agent framework changed the implementation of how to retrieve the root agent, so the rootAgentName config is useless, remove that from config and template.
- The root agent is config in a ml system index, and it prevents FTRepo to set agent id into the index when running Cypress test(system index config can only be changed in opensearch.yml file), so move the security test out of the test manifest.

The dashboards-assistant plugin is an experimental plugin and we will address the security test in 2.13.

### Issues Resolved

### DIFF
--- a/manifests/2.12.0/opensearch-dashboards-2.12.0-test.yml
+++ b/manifests/2.12.0/opensearch-dashboards-2.12.0-test.yml
@@ -87,4 +87,3 @@ components:
         - without-security
       additional-cluster-configs:
         assistant.chat.enabled: true
-        assistant.chat.rootAgentName: 'Cypress test agent'
--- a/manifests/2.12.0/opensearch-dashboards-2.12.0-test.yml
+++ b/manifests/2.12.0/opensearch-dashboards-2.12.0-test.yml
@@ -83,7 +83,6 @@ components:
   - name: assistantDashboards
     integ-test:
       test-configs:
-        - with-security
         - without-security
       additional-cluster-configs:
         assistant.chat.enabled: true
--- a/manifests/2.12.0/opensearch-dashboards-2.12.0-test.yml
+++ b/manifests/2.12.0/opensearch-dashboards-2.12.0-test.yml
@@ -83,6 +83,7 @@ components:
   - name: assistantDashboards
     integ-test:
       test-configs:
+        - with-security
         - without-security
       additional-cluster-configs:
         assistant.chat.enabled: true
--- a/config/opensearch_dashboards-2.x.yml
+++ b/config/opensearch_dashboards-2.x.yml
@@ -199,8 +199,6 @@
 # 2.12 New experimental Assistant Dashboards Feature
 # Set the value of this setting to true to enable the assistant dashboards
 # assistant.chat.enabled: false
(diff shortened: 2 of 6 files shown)


gaobinlong (commented):
LGTM.

manifests/2.12.0/opensearch-dashboards-2.12.0-test.yml
```diff
@@ -83,8 +83,6 @@ components:
   - name: assistantDashboards
     integ-test:
       test-configs:
-        - with-security
```
peterzhuamazon, line 86: Even after reading the description I dont get why are we removing the security test. Could you elaborate?

Thanks.

peterzhuamazon (changes requested):

manifests/2.12.0/opensearch-dashboards-2.12.0-test.yml
SuZhou-Joe: assistantDashboards requires a config in system index, but we can not `put` doc into that index in our cypress test workflow as it requires a security .pem file to do so.

SuZhou-Joe: If we want to do test against a Chatbot root agent, we have to set the config inside the system index. However when security enabled, we can not do that inside Cypress test as system_index setting is a static setting that can only be modified in the opensearch.yml file.

SuZhou-Joe:
@bbarani @prudhvigodithi Could you please take a look here?

manifests/2.12.0/opensearch-dashboards-2.12.0-test.yml
gaiksaya: So you mean in order to support tests with security you need additional config to be added to opensearch.yml?

SuZhou-Joe: Yes, that should work too.

gaiksaya: Looks like a feature request. AFAIK `additional-cluster-configs` only supports the product under test (OpenSearch or Dashboards. Not both) 
https://github.com/search?q=repo%3Aopensearch-project%2Fopensearch-build+additional-cluster-configs&type=code&p=1

@peterzhuamazon Looks like time to open this issue https://github.com/opensearch-project/opensearch-build/issues/3589 :)

peterzhuamazon: Have sync up with @SuZhou-Joe and it seems we either implement #3589, which we dont have much time for 2.12, or remove the feature after @SuZhou-Joe discuss with @ylwu-amzn.

There is another option which is to comment out the `- with-security` and list it a `TODO` to be fixed in the next release.

peterzhuamazon (approved):

peterzhuamazon:
Restore security test due to this:
* https://github.com/opensearch-project/opensearch-dashboards-functional-test/commit/2f5d2f0c9e66575d793ffb45bfa9f9fbcdbb9577

Issue #4386, opened 2024-01-30
Change the Artifact Build Process for k-NN Plugin


naveentatikonda:
### Problem Statement
For the 2.12 release, in k-NN plugin a new feature [Faiss SQFP16](https://github.com/opensearch-project/k-NN/issues/1138) is being supported which also includes the [SIMD](https://en.wikipedia.org/wiki/Single_instruction,_multiple_data) optimization (Single Instruction Multiple Data) through AVX2 on x86 architecture and using NEON on ARM architecture. This SIMD optimization helps to boost the overall performance a lot. But, this might be a breaking change for the users who are upgrading from version 2.11 to 2.12 if their system or processor doesn’t support these optimizations which might result in a crash.

### Proposed Solution
To solve this issue, we will be building two different versions of Faiss library(with same name) in k-NN plugin, one with SIMD optimization enabled and the other one without SIMD optimization and store them in different sub directories (as they both have same name). By default, during runtime we will load the lib without SIMD optimization. 

For those users who know that their system supports these latest optimizations and want to try them, they can enable it by using an optional experimental feature flag something like `-Dopensearch.kn.experimental.feature.faiss.simd.enabled=true`. For this, they need to stop the existing cluster and need to pass the above optional flag during bootup which will load the other Faiss library with SIMD optimization(AVX2 or NEON depending on architecture).

gaiksaya:
[Triage] Removing the untriage label here since @peterzhuamazon @prudhvigodithi @bbarani are aware of the issue and have discussed the possible next steps. Would be helpful if someone can post them here.
Thanks!

peterzhuamazon:
Work with Naveen to help him disable avx instruction set on an existing EC2 instance:

* Open /etc/default/grub, update `GRUB_CMDLINE_LINUX_DEFAULT` line with `noxsave`.
* On Ubuntu, run `sudo update-grub`; on RHEL, run `sudo grub2-mkconfig -o /boot/grub2/grub.cfg`.
* Reboot

Before disable AVX:
```
    Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq s
[...]

PR #4407, opened 2024-02-05
Add SIMD support for k-NN in all distributions startup scripts

peterzhuamazon:
### Description
Add SIMD support for k-NN in all distributions startup scripts

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4386


peterzhuamazon:
Progress:
* https://github.com/opensearch-project/opensearch-build/issues/4386#issuecomment-1928110303

docker/release/config/opensearch/opensearch-docker-entrypoint-2.x.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash
```
dblock, line 1: These version scripts are very similar. Can one .sh call/include another? Would be nice if it wasn't a big copy paste with some small changes.

peterzhuamazon:
Close this as the method approach has been changed.
@naveentatikonda confirms that most of the changes and loading can happen through java code after the cluster is started.

Thanks.

Issue #4370, opened 2024-01-24
[Bug]: All integration tests for RPM, Debian and Windows are failing


gaiksaya:
### Describe the bug

_Note: This is an observation for OpenSearch only. OSD might be a different story_
Recently so many changes went in related to security plugin as well as in general. We enabled integration testing to be triggered via distribution pipeline for all distributions. (Earlier it was just running for tarball)

### To reproduce

All windows failed: https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test/detail/integ-test/7136/pipeline
All debian failed: https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test/detail/integ-test/7135/pipeline
All RPM failed: https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test/detail/integ-test/7134/pipeline

### Expected behavior

All tests should run. The infrastructure and test set up should be fixed.

peterzhuamazon:
Will check soon, most likely the deb/rpm is due to sudo permission not added to the ci-runner images.

peterzhuamazon:
k-NN: Test with @junqiu-lei on tar for concurrent search, resolved https://github.com/opensearch-project/k-NN/pull/1419.

peterzhuamazon:
Resolve rpm:
```
2024-01-26 23:15:19 INFO     rpm installation requires sudo, script will exit if current user does not have sudo access
LANG=C.UTF-8
HOSTNAME=1c85fa7fa882
TERM=xterm
MAIL=/var/mail/root
PATH=/sbin:/bin:/usr/sbin:/usr/bin
LOGNAME=root
USER=root
HOME=/root
SHELL=/bin/bash
SUDO_COMMAND=/bin/env OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123!
SUDO_USER=ci-runner
SUDO_UID=1000
SUDO_GID=1000
XDG_SESSION_ID=c1
XDG_RUNTIME_DIR=/run/user/0
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/0/bus
OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123!
No match for argument: opensearch
No packages marked for removal.
Dependencies resolved.
Nothing to do.
Complete!
Rocky Linux 8 - AppStream                                                                                                                                                                                                                                                                   7.1 MB/s |  10 MB     00:01
[...]

PR #4419, opened 2024-02-06
Switch ftrepo checkout based on ref instead of commit_id

peterzhuamazon:
### Description
Switch ftrepo checkout based on ref instead of commit_id

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4370

### DIFF
--- a/src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
+++ b/src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
@@ -60,7 +60,7 @@ class IntegTestSuiteOpenSearchDashboards(IntegTestSuite):
 
         self.repo = GitRepository(
             build_manifest_opensearch_dashboards.components['functionalTestDashboards'].repository,
-            build_manifest_opensearch_dashboards.components['functionalTestDashboards'].commit_id,
+            build_manifest_opensearch_dashboards.components['functionalTestDashboards'].ref,
             os.path.join(self.work_dir, self.component.name),
             test_config.working_directory
         )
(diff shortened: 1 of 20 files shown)


src/test_workflow/integ_test/integ_test_suite_opensearch_dashboards.py
```diff
@@ -60,7 +60,7 @@ def __init__(
 
         self.repo = GitRepository(
             build_manifest_opensearch_dashboards.components['functionalTestDashboards'].repository,
-            build_manifest_opensearch_dashboards.components['functionalTestDashboards'].commit_id,
+            build_manifest_opensearch_dashboards.components['functionalTestDashboards'].ref,
```
gaiksaya, line 63: Want to add a simple test verifying that ref is checked out and not commit_id?

peterzhuamazon: Updated, thanks.

tests/tests_test_workflow/test_integ_workflow/integ_test/data/build_manifest_osd.yml
```diff
@@ -0,0 +1,135 @@
+---
```
gaiksaya, line 1: Better to add tests data here: https://github.com/opensearch-project/opensearch-build/tree/main/tests/data ?

gaiksaya (approved):
Small nit with moving the test data to root folder instead of integ_test one.

Thanks!

prudhvigodithi (approved):

tests/tests_test_workflow/test_integ_workflow/integ_test/data/build_manifest_osd.yml
peterzhuamazon: This is in sync with the opensearch side of the data location.
It might be a lot more changes if we want to move.
But those test data are integTest specific so I havent touch them yet. Can move later. Thanks.

Issue #4441, opened 2024-02-11
[Bug]: Validation run with file-path silently skips tests for incorrect product name


gaiksaya:
### Describe the bug

The validation for OpenSearch and OpenSearch-Dashboards does not parse the product names and silent skips the validation if the product does not match.

### To reproduce

Run the below command:

```
./validation.sh --file-path opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.12.0/latest/linux/x64/tar/dist/opensearch/opensearch-2.12.0-linux-x64.tar.gz opensearch-dashboard=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch-dashboards/2.12.0/latest/linux/x64/tar/dist/opensearch-dashboards/opensearch-dashboards-2.12.0-linux-x64.tar.gz
```
The product name is supposed to be`opensearch-dashboards` instead of `opensearch-dashboard`

Instead of throwing the error, it passes for opensearch and just skips for OSD.

<details><summary>Details</summary>
<p>

```
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 16/16 — 00:00:04
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
Running ./src/run_validation.py --file-path opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.12.0/latest/linux/x64/tar/dist/opensearch/opensearch-2.12.0-linux-x64.tar.gz opensearch-dashboard=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch-dashboards/2.12.0/latest/linux/x64/tar/dist/opensearch-dashboards/opensearch-dashboards-2.12.0-linux-x64.tar.gz ...
Warning: Your Pipfile requires python_version 3.9, but you are using 3.7.16 (/home/g/.local/share/v/o/bin/python).
  $ pipenv --rm and rebuilding the virtual environment may resolve the issue.
  $ pipenv check will surely fail.
2024-02-11 14:38:48 INFO     Valid URL - https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.12.0/latest/linux/x64/tar/dist/opensearch/opensearch-2.12.0-linux-x64.tar.gz and Download Successful !
[...]

Issue #4424, opened 2024-02-07
Integrate validation workflow with Integtest workflow


Divyaasm:
### Is your feature request related to a problem? Please describe

Currently the integtests are triggered for OS and OSD irrespective of whether the artifacts are valid or not. While this is leading to unnecessary failures and resulting in creation of additional github issues. 

### Describe the solution you'd like

Call the validation workflow for the respective artifact and trigger integtests only if the validation is sucessful, else fail the workflow.

Divyaasm:
Closing as the Validation workflow is now integrated with Integ-test workflow.

Issue #4423, opened 2024-02-07
Add zip and debian distributions to validation workflow


Divyaasm:
### Is your feature request related to a problem? Please describe

The current workflow is lacking validation for zip and debain distributions, which needs too be separately validated during releases or when required.

### Describe the solution you'd like

Add code to support zip and debian distributions to use the validation workflow with all the supported distributions till date.

### Additional context

#4424

Divyaasm:
Closed via #4447

PR #4447, opened 2024-02-13
Support new distribution-validations and add few enhancements to validation workflow

Divyaasm:
### Description
1. Support Zip and Deb validations.
2. Move the download_artifacts method from child classes(except docker) to parent class - validation.py.
3. Changes to validation_args.py to fix the bug https://github.com/opensearch-project/opensearch-build/issues/4441
4. Cluster readiness is checked by requesting the api's instead of adding sleep().
5. Changes to docker-validation:
    1. Add security strong password changes before running the container.
    2. Bring up the services opensearchnode1 and opensearchnode2 only when OS is validated, bringup all the services in 
        compose file when both OS and OSD are provided.
    3. Pull and validate the latest staging docker image in the given version if build number is not provided.
    
Addresses #4424 and #4423

### DIFF
--- a/src/validation_workflow/deb/validation_deb.py
+++ b/src/validation_workflow/deb/validation_deb.py
@@ -16,32 +16,26 @@ from validation_workflow.validation import Validation
 from validation_workflow.validation_args import ValidationArgs
 
 
-class ValidateRpm(Validation, DownloadUtils):
-
+class ValidateDeb(Validation, DownloadUtils):
     def __init__(self, args: ValidationArgs) -> None:
         super().__init__(args)
 
     def installation(self) -> bool:
         try:
-            execute('sudo rpm --import https://artifacts.opensearch.org/publickeys/opensearch.pgp', str(self.tmp_dir.path), True, False)
             for project in self.args.projects:
-                self.filename = os.path.basename(self.args.file_path.get(project))
-                execute(f'sudo yum remove {project} -y', ".")
-                execute(f'sudo env OPENSEARCH_INITIAL_ADMIN_PASSWORD={get_password(str(self.args.version))} rpm -ivh {os.path.join(self.tmp_dir.path, self.filename)}', str(self.tmp_dir.path), True, False)  # noqa: 501
+                set_password = f' env OPENSEARCH_INITIAL_ADMIN_PASSWORD={get_password(str(self.args.version))}' if project == "opensearch" else ""
+                execute(f'sudo dpkg --purge {project}', ".")
+                execute(f'sudo {set_password} dpkg -i {os.path.basename(self.args.file_path.get(project))}', str(self.tmp_dir.path))
         except:
-            raise Exception('Failed to install Opensearch')
+            raise Exception("Failed to install OpenSearch/OpenSearch-Dashboards")
         return True
 
     def start_cluster(self) -> bool:
         try:
             for project in self.args.projects:
+                execute(f'sudo systemctl enable {project}', ".")
                 execute(f'sudo systemctl start {project}', ".")
-                (stdout, stderr, status) = execute(f'sudo systemctl status {project}', ".")
-                if(status == 0):
-                    logging.info(stdout)
-                else:
-                    logging.info(stderr)
-
+                execute(f'sudo systemctl status {project}', ".")
         except:
             raise Exception('Failed to Start Cluster')
         return True
@@ -60,8 +54,7 @@ class ValidateRpm(Validation, DownloadUtils):
     def cleanup(self) -> bool:
         try:
             for project in self.args.projects:
-                execute(f'sudo systemctl stop {project}', ".")
-                execute(f'sudo yum remove {project} -y', ".")
+                execute(f'sudo dpkg --purge {project}', ".")
         except Exception as e:
             raise Exception(f'Exception occurred either while attempting to stop cluster or removing OpenSearch/OpenSearch-Dashboards. {str(e)}')
         return True
--- a/src/validation_workflow/deb/validation_deb.py
+++ b/src/validation_workflow/deb/validation_deb.py
@@ -42,7 +42,8 @@ class ValidateDeb(Validation, DownloadUtils):
 
     def validation(self) -> bool:
         if self.check_cluster_readiness():
-            test_result, counter = ApiTestCases().test_apis(self.args.version, self.args.projects, self.check_for_security_plugin(os.path.join(os.sep, "usr", "share", "opensearch")) if not self.args.force_https else True)  # noqa: E501
+            test_result, counter = ApiTestCases().test_apis(self.args.version, self.args.projects,
+                                                            self.check_for_security_plugin(os.path.join(os.sep, "usr", "share", "opensearch")) if self.args.allow_http else True)
             if (test_result):
                 logging.info(f'All tests Pass : {counter}')
                 return True
--- a/src/validation_workflow/docker/validation_docker.py
+++ b/src/validation_workflow/docker/validation_docker.py
@@ -9,14 +9,10 @@ import logging
 import os
 import shutil
 import subprocess
-import time
 from subprocess import PIPE
 from typing import Any
 
-import requests
-
-from system.temporary_directory import TemporaryDirectory
-from validation_workflow.api_request import ApiTest
+from test_workflow.integ_test.utils import get_password
 from validation_workflow.api_test_cases import ApiTestCases
 from validation_workflow.docker.inspect_docker_image import InspectDockerImage
 from validation_workflow.validation import Validation
@@ -61,7 +57,7 @@ class ValidateDocker(Validation):
     def validation(self) -> bool:
         # STEP 2 . inspect image digest between opensearchproject(downloaded/local) and opensearchstaging(dockerHub)
         if not self.args.using_staging_artifact_only:
-            self.image_names_list = [self.args.OS_image, self.args.OSD_image]
+            self.image_names_list = ['opensearchproject/' + project for project in self.args.projects]
             self.image_names_list = [x for x in self.image_names_list if (os.path.basename(x) in self.args.projects)]
             self.image_digests = list(map(lambda x: self.inspect_docker_image(x[0], x[1]), zip(self.image_ids.values(), self.image_names_list)))  # type: ignore
             if all(self.image_digests):
@@ -79,7 +75,7 @@ class ValidateDocker(Validation):
                 self.args.version
             )
             if return_code:
-                logging.info('Checking if cluster is ready for API test in every 10 seconds\n\n')
+                logging.info('Checking if cluster is ready for API test in every 5 seconds\n\n')
 
                 if self.check_cluster_readiness():
                     # STEP 4 . OS, OSD API validation
@@ -94,8 +90,8 @@ class ValidateDocker(Validation):
                         raise Exception(f'Not all tests Pass : {_counter}')
                 else:
                     raise Exception("Cluster is not ready for API test.")
-        else:
-            raise Exception('The container failed to start. Exiting the validation.')
+            else:
+                raise Exception('The container failed to start. Exiting the validation.')
 
         return True
 
@@ -128,36 +124,6 @@ class ValidateDocker(Validation):
             logging.error("Error: %s - %s." % (e.filename, e.strerror))
 
         return('returncode=0' in (str(result)))
-
-    def check_http_request(self) -> bool:
-        self.test_readiness_urls = {
-            'https://localhost:9200/': 'opensearch cluster API',
-            'http://localhost:5601/api/status': 'opensearch-dashboards API',
-        }
-
-        for url, name in self.test_readiness_urls.items():
-            try:
-                status_code, response_text = ApiTest(url, self.args.version).api_get()
-                if status_code != 200:
-                    logging.error(f'Error connecting to {name} ({url}): status code {status_code}')
-                    return False
-            except (requests.exceptions.ConnectionError, requests.exceptions.ConnectTimeout) as e:
-                logging.error(f'Error connecting to {name} ({url}): {e}')
-                return False
-        return True
-
-    def check_cluster_readiness(self) -> bool:
-        max_retry = 20
-        retry_count = 0
-        while retry_count < max_retry:
-            logging.info(f'sleeping 10sec for retry {retry_count + 1}/{max_retry}')
-            time.sleep(10)
-            if self.check_http_request():
-                logging.info('\n\ncluster is now ready for API test\n\n')
-                return True
-            retry_count += 1
-        logging.error(f"Maximum number of retries ({max_retry}) reached. Cluster is not ready for API test.")
-        return False
 
     def get_artifact_image_name(self, artifact: str, using_staging_artifact_only: str) -> Any:
         self.image_names = {
@@ -241,7 +207,6 @@ class ValidateDocker(Validation):
             '2': 'docker-compose-2.x.yml'
         }
 
-        self.tmp_dir = TemporaryDirectory()
         self.target_yml_file = os.path.join(self.tmp_dir.name, 'docker-compose.yml')
 
         self.major_version_number = version[0]
@@ -252,8 +217,9 @@ class ValidateDocker(Validation):
         self.replacements = [(f'opensearchproject/{key}:{self.major_version_number}', value) for key, value in image_ids.items()]
 
         list(map(lambda r: self.inplace_change(self.target_yml_file, r[0], r[1]), self.replacements))
-
+        os.environ["OPENSEARCH_INITIAL_ADMIN_PASSWORD"] = get_password(str(version))
         # spin up containers
-        self.docker_compose_up = f'docker-compose -f {self.target_yml_file} up -d'
+        services = "opensearch-node1 opensearch-node2" if "opensearch-dashboards" not in self.args.projects else ""
+        self.docker_compose_up = f'docker-compose -f {self.target_yml_file} up -d {services}'
         result = subprocess.run(self.docker_compose_up, shell=True, stdout=PIPE, stderr=PIPE, universal_newlines=True)
         return ('returncode=0' in (str(result)), self.target_yml_file)
--- a/src/validation_workflow/docker/validation_docker.py
+++ b/src/validation_workflow/docker/validation_docker.py
@@ -58,7 +58,6 @@ class ValidateDocker(Validation):
         # STEP 2 . inspect image digest between opensearchproject(downloaded/local) and opensearchstaging(dockerHub)
         if not self.args.using_staging_artifact_only:
             self.image_names_list = ['opensearchproject/' + project for project in self.args.projects]
-            self.image_names_list = [x for x in self.image_names_list if (os.path.basename(x) in self.args.projects)]
             self.image_digests = list(map(lambda x: self.inspect_docker_image(x[0], x[1]), zip(self.image_ids.values(), self.image_names_list)))  # type: ignore
             if all(self.image_digests):
                 logging.info('Image digest is validated.\n\n')
--- a/src/validation_workflow/rpm/validation_rpm.py
+++ b/src/validation_workflow/rpm/validation_rpm.py
@@ -48,7 +48,8 @@ class ValidateRpm(Validation, DownloadUtils):
 
     def validation(self) -> bool:
         if self.check_cluster_readiness():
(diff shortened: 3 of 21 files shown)


src/validation_workflow/docker/validation_docker.py
```diff
@@ -61,7 +61,7 @@ def start_cluster(self) -> bool:
     def validation(self) -> bool:
         # STEP 2 . inspect image digest between opensearchproject(downloaded/local) and opensearchstaging(dockerHub)
         if not self.args.using_staging_artifact_only:
-            self.image_names_list = [self.args.OS_image, self.args.OSD_image]
+            self.image_names_list = ['opensearchproject/' + project for project in self.args.projects]
             self.image_names_list = [x for x in self.image_names_list if (os.path.basename(x) in self.args.projects)]
```
jordarlu, line 65: what does these two lines of `self.image_name_list` do?

src/validation_workflow/deb/validation_deb.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def installation(self) -> bool:
+        try:
+            for project in self.args.projects:
+                set_password = f' env OPENSEARCH_INITIAL_ADMIN_PASSWORD={get_password(str(self.args.version))}' if project == "opensearch" else ""
+                execute(f'sudo dpkg --purge {project}', ".")
+                execute(f'sudo {set_password} dpkg -i {os.path.basename(self.args.file_path.get(project))}', str(self.tmp_dir.path))
+                time.sleep(80)
+        except:
+            raise Exception("Failed to install Opensearch")
```
gaiksaya, line 32: I believe the product might be anything? OpenSearch or Dashboards? Recommending to make the error message generic.

src/validation_workflow/rpm/validation_rpm.py
```diff
@@ -22,26 +21,6 @@ class ValidateRpm(Validation, DownloadUtils):
...
-                else:
-                    self.args.version = self.get_version(self.args.file_path.get(project))
-                    self.check_url(self.args.file_path.get(project))
-            else:
-                if (self.args.artifact_type == "staging"):
-                    self.args.file_path[project] = f"{self.base_url_staging}{project}/{self.args.version}/{self.args.build_number[project]}/linux/{self.args.arch}/{self.args.distribution}/dist/{project}/{project}-{self.args.version}-linux-{self.args.arch}.rpm"  # noqa: E501
-                else:
-                    self.args.file_path[project] = f"{self.base_url_production}{project}/{self.args.version}/{project}-{self.args.version}-linux-{self.args.arch}.rpm"
-                self.check_url(self.args.file_path.get(project))
-        return True
```
gaiksaya, line 44: Why are we removing the entire RPM method?

src/validation_workflow/tar/validation_tar.py
```diff
@@ -23,29 +22,9 @@ class ValidateTar(Validation, DownloadUtils):
...
-                    self.args.version = self.get_version(self.args.file_path.get(project))
-                    self.check_url(self.args.file_path.get(project))
-            else:
-                if (self.args.artifact_type == "staging"):
-                    self.args.file_path[project] = f"{self.base_url_staging}{project}/{self.args.version}/{self.args.build_number[project]}/linux/{self.args.arch}/{self.args.distribution}/dist/{project}/{project}-{self.args.version}-linux-{self.args.arch}.tar.gz"  # noqa: E501
-                else:
-                    self.args.file_path[project] = f"{self.base_url_production}{project}/{self.args.version}/{project}-{self.args.version}-linux-{self.args.arch}.tar.gz"
-                self.check_url(self.args.file_path.get(project))
-        return True
-
```
gaiksaya, line 48: Same question. Why are removing the entire code block?

src/validation_workflow/validation_args.py
```diff
@@ -123,22 +123,30 @@ def __init__(self) -> None:
...
+            if 'opensearch' not in args.file_path.keys() or not set(args.file_path.keys()) <= {'opensearch', 'opensearch-dashboards'}:
+                raise Exception("Missing OpenSearch artifact details! Please provide the valid product names among opensearch and opensearch-dashboards")
             args.distribution = self.get_distribution_type(args.file_path)
             args.projects = args.file_path.keys()
-        if (('opensearch' not in args.projects)):
-            raise Exception("Missing OpenSearch OpenSearch artifact details! Please provide the same along with OpenSearch-Dashboards to validate")
+
+        if args.distribution == "docker":
+            if args.osd_build_number != "latest" and "opensearch-dashboards" not in args.projects:
+                raise Exception("Provide opensearch-dashboards in projects argument to validate OSD")
```
gaiksaya, line 147: nit:
```suggestion
                raise Exception("Provide opensearch-dashboards in projects argument to validate OpenSearch-Dashboards")
```

gaiksaya:
Hi Divya, can you please add more details in the PR description explaining what changes are being made. Looks like it is doing a lot more than debian and zip validation.
Also please attach the issue number related to this PR change.

Thanks!

src/validation_workflow/deb/validation_deb.py
```diff
@@ -0,0 +1,60 @@
...
+from test_workflow.integ_test.utils import get_password
+from validation_workflow.api_test_cases import ApiTestCases
+from validation_workflow.download_utils import DownloadUtils
+from validation_workflow.validation import Validation
+from validation_workflow.validation_args import ValidationArgs
+
+
+class ValidateDeb(Validation, DownloadUtils):
+    def __init__(self, args: ValidationArgs) -> None:
+        super().__init__(args)
```
peterzhuamazon, line 22: Missing tests.

src/validation_workflow/deb/validation_deb.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def installation(self) -> bool:
+        try:
+            for project in self.args.projects:
+                set_password = f' env OPENSEARCH_INITIAL_ADMIN_PASSWORD={get_password(str(self.args.version))}' if project == "opensearch" else ""
+                execute(f'sudo dpkg --purge {project}', ".")
+                execute(f'sudo {set_password} dpkg -i {os.path.basename(self.args.file_path.get(project))}', str(self.tmp_dir.path))
```
peterzhuamazon, line 29: Why using `project` instead of `product`?

src/validation_workflow/deb/validation_deb.py
peterzhuamazon: ^^

src/validation_workflow/deb/validation_deb.py
```diff
@@ -0,0 +1,60 @@
...
+        except:
+            raise Exception("Failed to install Opensearch")
+        return True
+
+    def start_cluster(self) -> bool:
+        try:
+            for project in self.args.projects:
+                execute(f'sudo systemctl enable {project}', ".")
+                execute(f'sudo systemctl start {project}', ".")
+                time.sleep(20)
```
peterzhuamazon, line 40: Why is there a need to sleep 20 since opensearch systemd is using notify type and will wait until success.
Even if you have to wait why not just do a check of API calls every 5sec instead of blindly waiting for 20sec?
Different server spec can change the delay of this startup.

src/validation_workflow/docker/validation_docker.py
peterzhuamazon: Second this.

src/validation_workflow/docker/validation_docker.py
```diff
@@ -252,8 +250,9 @@ def run_container(self, image_ids: dict, version: str) -> Any:
         self.replacements = [(f'opensearchproject/{key}:{self.major_version_number}', value) for key, value in image_ids.items()]
 
         list(map(lambda r: self.inplace_change(self.target_yml_file, r[0], r[1]), self.replacements))
-
+        os.environ["OPENSEARCH_INITIAL_ADMIN_PASSWORD"] = get_password(str(version))
         # spin up containers
-        self.docker_compose_up = f'docker-compose -f {self.target_yml_file} up -d'
+        services = "opensearch-node1 opensearch-node2" if "opensearch-dashboards" not in self.args.projects else ""
```
peterzhuamazon, line 255: Why are we separating these instead of just spin up the entire docker compose?

src/validation_workflow/rpm/validation_rpm.py
peterzhuamazon: ^^

src/validation_workflow/tar/validation_tar.py
peterzhuamazon: ^^

peterzhuamazon (changes requested):
Hi @Divyaasm ,

As @gaiksaya also pointed out here, the PR is huge with a lot of information but little comments or descriptions on what is happening.

I would suggest you separate this PR into two separate ones, lets review and get deb merge and we understand your approach before going back on win zip, or you can also avoid duplicate changes.

You are also missing tests on the PR. With a big PR like this I am hesitate to just merge through without any tests.

Please take a look at the comments and let us know. 😄 

Thanks.

src/validation_workflow/docker/validation_docker.py
Divyaasm: I have unassigned the args self.args.OS_image, self.args.OSD_image in args.py  and declared a list which stores the images names based projects=["opensearch", "opensearchdashboards"]

src/validation_workflow/tar/validation_tar.py
Divyaasm: I moved the entire download_artifacts method from child classes(tar, rpm, yum, deb, zip) to parent class - validation.py

Divyaasm:
> Hi Divya, can you please add more details in the PR description explaining what changes are being made. Looks like it is doing a lot more than debian and zip validation. Also please attach the issue number related to this PR change.
> 
> Thanks!

Sure Sayali

src/validation_workflow/docker/validation_docker.py
Divyaasm: Added changes to bring up the services opensearchnode1 and opensearchnode2 only when OS is validated, bringup all the services in compose file when both OS and OSD are provided.

src/validation_workflow/deb/validation_deb.py
Divyaasm: I'm working on the test cases. Will update it in next commit. Thanks

src/validation_workflow/deb/validation_deb.py
Divyaasm: The argument is `project` from the beginning

src/validation_workflow/deb/validation_deb.py
Divyaasm: Implemented this for all the distributions

src/validation_workflow/validation.py
```diff
@@ -53,9 +60,59 @@ def run(self) -> Any:
...
+                self.args.file_path[project] = self.get_filepath(project)
+                self.check_url(self.args.file_path.get(project))
+        return True
+
+    def get_filepath(self, project: str) -> str:
+        file_name_suffix = "tar.gz" if self.args.distribution == "tar" else self.args.distribution
+        if self.args.artifact_type == "staging":
+            if self.args.distribution == "yum":
+                return f"{self.base_url_staging}{project}/{self.args.version}/{self.args.build_number[project]}/{self.args.platform}/{self.args.arch}/rpm/dist/{project}/{project}-{self.args.version}.staging.repo"  # noqa: E501
+            return f"{self.base_url_staging}{project}/{self.args.version}/{self.args.build_number[project]}/{self.args.platform}/{self.args.arch}/{self.args.distribution}/dist/{project}/{project}-{self.args.version}-{self.args.platform}-{self.args.arch}.{file_name_suffix}"  # noqa: E501
```
peterzhuamazon, line 82: I think you can still make this multi-lines right?

src/validation_workflow/validation_args.py
```diff
@@ -114,7 +114,7 @@ def __init__(self) -> None:
         parser.add_argument(
             "-f",
             "--force-https",
-            action="store_true",
+            action="store_false",
```
peterzhuamazon, line 117: Why this change?

src/validation_workflow/validation_args.py
```diff
@@ -123,22 +123,30 @@ def __init__(self) -> None:
...
-            help="(optional) Validate digest only; will not run docker to test API")
+            help="(optional) Validate digest only; will not run docker to test API",
+            dest="validate_digest_only")
         group.add_argument(
             "--using-staging-artifact-only",
             action="store_true",
             default=False,
-            help="(optional) Use only staging artifact to run docker and API test, will not validate digest")
+            help="(optional) Use only staging artifact to run docker and API test, will not validate digest",
+            dest="using_staging_artifact_only")
```
peterzhuamazon, line 133: Why?
```
dest="using_staging_artifact_only")
```

peterzhuamazon (commented):
More questions.

src/validation_workflow/validation.py
```diff
@@ -53,9 +60,59 @@ def run(self) -> Any:
...
+        retry_count = 0
+        while retry_count < max_retry:
+            logging.info(f'sleeping 5sec for retry {retry_count + 1}/{max_retry}')
+            time.sleep(5)
+            if self.check_http_request():
+                logging.info('\n\ncluster is now ready for API test\n\n')
+                return True
+            retry_count += 1
+        logging.error(f"Maximum number of retries ({max_retry}) reached. Cluster is not ready for API test.")
+        return False
```
peterzhuamazon, line 98: Cool 👍

src/validation_workflow/validation_args.py
Divyaasm: I have just added it for easy readability, as the value using-staging-artifact-only is using '-'.

src/validation_workflow/docker/validation_docker.py
jordarlu: thanks, Divya, I just feel a bit redundant here becasue you constructed the image_names_list from args.projects on the first line of code, and then you do a filtering from the same args.projects on the next line of code. Could you give me an exmaple of expected output from the 2nd line of code assuming the image_names_list has ["opensearchproject/opensearch", "opensearchproject/opensearchdashboards"]?

src/validation_workflow/validation_args.py
Divyaasm: Bit confused on this parameter actually. This is always true with current implementation. Now , It is true by default, when -----force-https is passed it will store false.

To remove the confusion we can change the name to --do-not force-https or something similar and assign a value False by default, if provided we check for security plugin and run http/https as required.

peterzhuamazon: Why do we keep doing negative params? I still dont understand your point here. If --force-https, then it will always check https, if not / default it will check https or http based on security plugin presence, isnt it?

src/validation_workflow/validation_args.py
peterzhuamazon: I mean to say why it ended up with `)`? For both dest you added there?

Divyaasm: help="(optional) Use only staging artifact to run docker and API test, will not validate digest")  `)` comes from here 
            help="(optional) Use only staging artifact to run docker and API test, will not validate digest",
            dest="using_staging_artifact_only")

src/validation_workflow/validation_args.py
Divyaasm: I think we want to make the validation workflow confined to `https` all the time. Unless an external argument  --force-https is provided we do not want to validate without security plugin(i.e http)
Adding @gaiksaya

gaiksaya: Right! Security is always assumed to be enabled by default especially for validation.

src/validation_workflow/docker/validation_docker.py
Divyaasm: I get it now. Sorry missed this line while editing. It's of no use. Will remove it Thanks for pointing.

src/validation_workflow/validation_args.py
peterzhuamazon: Then correct the location of the `)`?

src/validation_workflow/validation_args.py
Divyaasm: Hi @peterzhuamazon, addressed all the changes except this. Can I know your thoughts?

peterzhuamazon: After talking to @Divyaasm as there are some confusion.

The param was meant to `disable` https-only check, aka allow http if security plugin not present.
See this flow:

validation force-https drawio

Therefore, the initial naming of `--force-https` is not applicable anymore.

Suggest using `--allow-http` so that, if such parameter not present it will always force to check https no matter what; else, if the parameter present it will allow for `http` check if security plugin is not present.

Thanks.

src/validation_workflow/validation_args.py
```diff
@@ -123,22 +123,32 @@ def __init__(self) -> None:
...
+            if 'opensearch' not in args.file_path.keys() or not set(args.file_path.keys()) <= {'opensearch', 'opensearch-dashboards'}:
+                raise Exception("Missing OpenSearch artifact details! Please provide the valid product names among opensearch and opensearch-dashboards")
             args.distribution = self.get_distribution_type(args.file_path)
             args.projects = args.file_path.keys()
-        if (('opensearch' not in args.projects)):
-            raise Exception("Missing OpenSearch OpenSearch artifact details! Please provide the same along with OpenSearch-Dashboards to validate")
+
+        if args.distribution == "docker":
+            if args.osd_build_number != "latest" and "opensearch-dashboards" not in args.projects:
+                raise Exception("Provide opensearch-dashboards in projects argument to validate OpenSearch-Dashboards")
```
peterzhuamazon, line 149: Probably change this msg to be much more clear such as:
```
You cannot provide a osd_build_number without specifying opensearch-dashboards in --projectsProbably change this msg to be much more clear such as:
```
You cannot provide a osd_build_number without specifying opensearch-dashboards in --projects
```

Or later we can just skip osd test if user provided osd build number without specifying projects as osd.
```

Or later we can just skip osd test if user provided osd build number without specifying projects as osd.

src/validation_workflow/validation_args.py
```diff
@@ -123,22 +123,32 @@ def __init__(self) -> None:
...
+            if not(args.validate_digest_only or args.using_staging_artifact_only):
```
peterzhuamazon, line 150: Nit: `if not(A) and not(B)` or a better condition for understanding.
We will raise exception if both of these params are empty.

src/validation_workflow/validation_args.py
```diff
@@ -146,14 +146,14 @@ def __init__(self) -> None:
 
         if args.distribution == "docker":
             if args.osd_build_number != "latest" and "opensearch-dashboards" not in args.projects:
-                raise Exception("Provide opensearch-dashboards in projects argument to validate OpenSearch-Dashboards")
-            if not(args.validate_digest_only or args.using_staging_artifact_only):
+                raise Exception("osd_build_number argument cannot be provided without specifying opensearch-dashboards in --projects")
```
peterzhuamazon, line 149: Probably should use `--osd-build-number`

src/validation_workflow/validation_args.py
peterzhuamazon: Resolved now.

src/validation_workflow/validation.py
```diff
@@ -53,9 +60,66 @@ def run(self) -> Any:
...
+        return f"{self.base_url_production}{project}/{self.args.version}/{project}-{self.args.version}-{self.args.platform}-{self.args.arch}.{file_name_suffix}"
+
+    def check_cluster_readiness(self) -> bool:
+        max_retry = 20
+        retry_count = 0
+        while retry_count < max_retry:
+            logging.info(f'sleeping 5sec for retry {retry_count + 1}/{max_retry}')
+            time.sleep(5)
+            if self.check_http_request():
+                logging.info('\n\ncluster is now ready for API test\n\n')
```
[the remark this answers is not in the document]
peterzhuamazon: Resolved now.

src/validation_workflow/validation_args.py
peterzhuamazon: Resolved now.

peterzhuamazon (approved):

peterzhuamazon:
Thanks @Divyaasm for addressing these comments.
Approved now.

Thanks.

src/validation_workflow/validation_args.py
peterzhuamazon: Thanks.

jordarlu (approved):

peterzhuamazon:
Hi @Divyaasm you need to address the python test failure.

Issue #4445, opened 2024-02-13
[Bug]: [Docker] Security plugin demo config install script executes even if security plugin is disabled


DarshitChanpura:
### Describe the bug

The code in `opensearch-docker-entrypoint.sh` currently installs demo config first if enabled and then check whether security plugin is disabled. See [here](https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/config/opensearch/opensearch-docker-entrypoint.sh#L30C1-L51C2)

### To reproduce

Run:
```
docker pull opensearchstaging/opensearch:2.12.0.9399 && docker run -it -p 9200:9200 -e "DISABLE_SECURITY_PLUGIN=true" -e "discovery.type=single-node" opensearchstaging/opensearch:2.12.0.9399
2.12.0.9399: Pulling from opensearchstaging/opensearch
```

### Expected behavior

Cluster should start without any errors related to security-plugin.

### Relevant log output

```shell
➜  ~ docker pull opensearchstaging/opensearch:2.12.0.9399 && docker run -it -p 9200:9200 -e "DISABLE_SECURITY_PLUGIN=true" -e "discovery.type=single-node" opensearchstaging/opensearch:2.12.0.9399
2.12.0.9399: Pulling from opensearchstaging/opensearch
Digest: sha256:87b2c2bed78237714aea05572b71757470b9b0622867615dde14c1ac3e2eeb27
Status: Image is up to date for opensearchstaging/opensearch:2.12.0.9399
docker.io/opensearchstaging/opensearch:2.12.0.9399

What's Next?
  1. Sign in to your Docker account → docker login
  2. View a summary of image vulnerabilities and recommendations → docker scout quickview opensearchstaging/opensearch:2.12.0.9399
Enabling execution of install_demo_configuration.sh for OpenSearch Security Plugin
OpenSearch 2.12.0 onwards, the OpenSearch Security Plugin a change that requires an initial password for 'admin' user.
Please define an environment variable 'OPENSEARCH_INITIAL_ADMIN_PASSWORD' with a strong password string.
If a password is not provided, the setup will quit.
 For more details, please visit: https://opensearch.org/docs/latest/install-and-configure/install-opensearch/docker/
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
OpenSearch install type: rpm/deb on Linux 6.6.12-linuxkit aarch64
OpenSearch config dir: /usr/share/opensearch/config/
[...]

PR #4446, opened 2024-02-13
[Docker] Fixes security enabled logic to execute demo config only if security plugin is enabled

DarshitChanpura:
### Description
The current docker entrypoint executes the demo security config setup script regardless of whether security plugin is enabled. Then it checks whether security plugin is enabled. I'm not sure why this design was implemented but I feel like this is wrong. This PR restructures the logic to install demo config only if security plugin is enabled.

### Issues Resolved
- Resolves https://github.com/opensearch-project/opensearch-build/issues/4445

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4452, opened 2024-02-14
Properly separate 1.x/2.x/default opensearch docker entrypoint like in opensearch-dashboards

peterzhuamazon:
### Description
Properly separate 1.x/2.x/default opensearch docker entrypoint like in opensearch-dashboards

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4445
https://github.com/opensearch-project/opensearch-build/issues/4115
A follow up to #4446 and #4450.


docker/release/README.md
```diff
@@ -103,11 +103,11 @@ Here are three example scenarios of using above variables:
 #### Scenario 2: No demo certs/configs + disable security on both OpenSearch and OpenSearch-Dashboards:
   * OpenSearch:
      ```
-     docker run -it -p 9200:9200 -p 9600:9600 -e "discovery.type=single-node" -e "DISABLE_INSTALL_DEMO_CONFIG=true" -e "DISABLE_SECURITY_PLUGIN=true" opensearchproject/opensearch:1.1.0
+     $ docker run -it -p 9200:9200 -p 9600:9600 -e "discovery.type=single-node" -e "DISABLE_INSTALL_DEMO_CONFIG=true" -e "DISABLE_SECURITY_PLUGIN=true" opensearchproject/opensearch:1.1.0
      ```
-    Note: For OpenSearch > 2.11.1 and > 1.3.14, `DISABLE_SECURITY_PLUGIN` when set to true will automatically disable the security demo configuration setup and will no longer require `DISABLE_INSTALL_DEMO_CONFIG` to be explicitly set.
+    Note: For OpenSearch 2.12 and later, `DISABLE_SECURITY_PLUGIN` when set to true will automatically disable the security demo configuration setup and will no longer require `DISABLE_INSTALL_DEMO_CONFIG` to be explicitly set.
```
DarshitChanpura, line 108: qq, what happens for release 1.3.15 and 2.11.2?

docker/release/config/opensearch/opensearch-docker-entrypoint-1.x.sh
```diff
@@ -0,0 +1,117 @@
...
+# that cgroup statistics are available for the container this process
+# will run in.
+export OPENSEARCH_JAVA_OPTS="-Dopensearch.cgroups.hierarchy.override=/ $OPENSEARCH_JAVA_OPTS"
+
+# Security Plugin
+function setupSecurityPlugin {
+    SECURITY_PLUGIN="opensearch-security"
+
+    if [ -d "$OPENSEARCH_HOME/plugins/$SECURITY_PLUGIN" ]; then
+        if [ "$DISABLE_INSTALL_DEMO_CONFIG" = "true" ]; then
```
DarshitChanpura, line 34: why not keep it same as 2.x? If I understand correctly this is the only difference i see between this file and one for 2.x

DarshitChanpura (commented):
this seems like a big change for docker, not sure how it could be tested. Maybe with an RC?

docker/release/README.md
peterzhuamazon: 1. There is no 2.11.2 if 2.12 is out.
2. 1.3.15 use the same entrypoint as 1.3.14 since 1.x/2.x/default.x are separate now.
3. 1.x entrypoint exactly the same as 1.3.14 while 2.x and default are the same following the latest one.

docker/release/config/opensearch/opensearch-docker-entrypoint-1.x.sh
peterzhuamazon: The point to do the separation is to keep 1.x behavior while only change 2.x since admin:admin only touch 2.x. I am not saying the original implementation is the best, but I try to keep 1.x as it is, not getting affected.

DarshitChanpura:
the changes look good to me based on the responses from the comments. One final question I have is will this be test-able only after RC generation?

peterzhuamazon:
> the changes look good to me based on the responses from the comments. One final question I have is will this be test-able only after RC generation?

I test the build on local for both 1 and 2 and they both goes up.

peterzhuamazon:
Adding @bbarani @prudhvigodithi for review here.

prudhvigodithi (approved):

PR #4481, opened 2024-02-27
Add a validation stage before triggering the Integration tests for OS and OSD

Divyaasm:
### Description
Currently the integration tests are trigerred for the given component list irrespective of whether the built distributions are valid or not.
Add a stage to Validate artifacts before triggering the integ tests. 

### Issues Resolved
[_List any issues this PR will resolve, e.g. Closes [...]._ 
](https://github.com/opensearch-project/opensearch-build/issues/4424)


peterzhuamazon (commented):
Is it possible to add a param so that if people want to skip validation before integ-test? Since this would likely take some time to complete.

Divyaasm:
> Is it possible to add a param so that if people want to skip validation before integ-test? Since this would likely take some time to complete.

Yes we can do that, but validation workflow takes upto 5 mins extra time which is very minimal. Can I make the validation step optional or keep it as a mandatory stage?
 @bbarani thoughts?

peterzhuamazon:
Adding @rishabh6788 @prudhvigodithi into the above discussion as well ^^

Thanks.

Issue #4506, opened 2024-03-05
[BUG] Windows bundle manifest is using \ not / in url location


peterzhuamazon:
[BUG] Windows bundle manifest is using \ not / in url location

See [manifest (1).txt](https://github.com/opensearch-project/opensearch-build/files/14502231/manifest.1.txt).

```
---
schema-version: '1.1'
build:
  name: OpenSearch Dashboards
  version: 2.12.0
  platform: windows
  architecture: x64
  distribution: zip
  location: https://ci.opensearch.org/ci/dbc/distribution-build-opensearch-dashboards/2.12.0/7326/windows/x64/zip/dist/opensearch-dashboards/opensearch-dashboards-2.12.0-windows-x64.zip
  id: '7326'
components:
  - name: OpenSearch-Dashboards
    repository: https://github.com/opensearch-project/OpenSearch-Dashboards.git
    ref: '2.12'
    commit_id: 9ec9a677af5f28e5450926ce07e9d6c3273717a7
    location: https://ci.opensearch.org/ci/dbc/distribution-build-opensearch-dashboards/2.12.0/7326/windows/x64/zip/builds/opensearch-dashboards/dist\opensearch-dashboards-min-2.12.0-windows-x64.zip
......
```

It is due to build manifest saving relative path as `dist\<>.zip` or `plugins\<>.zip` for windows. We cannot fix this on root as the manifest can be either local path based, or url based for official release through Jenkins.

Therefore, we need to fix the bundle_url_location.py directly.

Thanks.

PR #4508, opened 2024-03-05
Ensure that artifact url is valid for Windows dist manifest

peterzhuamazon:
### Description
Ensure that artifact url is valid for Windows dist manifest

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-build/issues/4506

### DIFF
--- a/src/assemble_workflow/bundle_url_location.py
+++ b/src/assemble_workflow/bundle_url_location.py
@@ -16,4 +16,8 @@ class BundleUrlLocation(BundleLocation):
 
     def join(self, *args: str) -> str:
         sub_path = "/".join(args)
-        return urljoin(self.path + "/", sub_path)
+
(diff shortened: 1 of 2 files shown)


prudhvigodithi (approved):

dblock:
This works around the problem, but where does it come from in the first place? Somewhere you get a path separator instead of a `/` that's platform specific, but should actually be a `/`.

peterzhuamazon:
> This works around the problem, but where does it come from in the first place? Somewhere you get a path separator instead of a `/` that's platform specific, but should actually be a `/`.

It is coming from the build manifest of windows, where everything is using `\` due to local path on windows is using `\` instead of `/`.

dblock:
@peterzhuamazon I understand, however the replacement should happen at the time when the value is passed into anything that takes a relative **url**, which is likely in a different/higher level place.

Issue #4478, opened 2024-02-26
[Benchmark] Add support to run benchmark against user provided cluster endpoint


rishabh6788:
### Is your feature request related to a problem? Please describe

As of now the nightly benchmark platform supports running end to end benchmark, meaning, the workflow creates the opensearch cluster as per user provided configuration, runs the benchmark using opensearch-benchmark tool and then deletes the cluster once the benchmark is finished. 

There are cases when the user wants to preserve the cluster to post-analysis after the benchmark has run but the benchmark doesn't support preserving the cluster after the benchmark. This is also not scalable and will become difficult to manage as more and more users want to preserve the cluster, and managing access will be another pain. 

### Describe the solution you'd like

The solution that we are proposing is to allow users to just provide an endpoint to their own clusters and use the benchmark platform to run the benchmark against it. 
This way the users can manage the life-cycle of their clusters and leverage benchmark platform to execute benchmarks and also be able to analyze benchmark metrics data using metrics datastore cluster.

Divyaasm:
Closing via https://github.com/opensearch-project/opensearch-build/pull/4579

PR #4541, opened 2024-03-18
Add mechanism to run benchmark tests against a given cluster endpoint

Divyaasm:
### Description
1. Runs benchmark tests against any given opensearch cluster
2. Refactor new password changes for versions >= 2.12.0
3. Use semver.Version instead of semver.compare in utils.py to avoid warnings as semver.compare is depreciated

### Issues Resolved
Adresses https://github.com/opensearch-project/opensearch-build/issues/4478

### DIFF
--- a/src/test_workflow/benchmark_test/benchmark_create_cluster.py
+++ b/src/test_workflow/benchmark_test/benchmark_create_cluster.py
@@ -1,0 +1,156 @@
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+
+import json
+import logging
+import os
+import subprocess
+from contextlib import contextmanager
+from typing import Any, Generator, Union
+
+from manifests.build_manifest import BuildManifest
+from manifests.bundle_manifest import BundleManifest
+from test_workflow.benchmark_test.benchmark_args import BenchmarkArgs
+from test_workflow.benchmark_test.benchmark_test_cluster import BenchmarkTestCluster
+from test_workflow.integ_test.utils import get_password
+
+
+class BenchmarkCreateCluster(BenchmarkTestCluster):
+    manifest: Union[BundleManifest, BuildManifest]
+    work_dir: str
+    current_workspace: str
+    output_file: str
+    params: str
+    is_endpoint_public: bool
+    cluster_endpoint: str
+
+    """
+    Represents a performance test cluster. This class deploys the opensearch bundle with CDK. Supports both single
+    and multi-node clusters
+    """
+
+    def __init__(
+            self,
+            args: BenchmarkArgs,
+            bundle_manifest: Union[BundleManifest, BuildManifest],
+            config: dict,
+            current_workspace: str
+    ) -> None:
+        super().__init__(args)
+        self.manifest = bundle_manifest
+        self.current_workspace = current_workspace
+        self.output_file = "output.json"
+        role = config["Constants"]["Role"]
+        params_dict = self.setup_cdk_params(config)
+        params_list = []
+        for key, value in params_dict.items():
+            if value:
+                '''
+                TODO: To send json input to typescript code from command line it needs to be enclosed in
+                single-quotes, this is a temp fix to achieve that since the quoted string passed from command line in
+                tesh.sh wrapper script gets un-quoted and we need to handle it here.
+                '''
+                if key == 'additionalConfig':
+                    params_list.append(f" -c {key}=\'{value}\'")
+                else:
+                    params_list.append(f" -c {key}={value}")
+        role_params = (
+            f" --require-approval=never --plugin cdk-assume-role-credential-plugin"
+            f" -c assume-role-credentials:writeIamRoleName={role} -c assume-role-credentials:readIamRoleName={role} "
+        )
+        self.params = "".join(params_list) + role_params
+        self.is_endpoint_public = False
+        self.stack_name = f"opensearch-infra-stack-{self.args.stack_suffix}"
+        if self.manifest:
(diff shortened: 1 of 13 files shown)


src/test_workflow/benchmark_test/benchmark_create_cluster.py
```diff
@@ -0,0 +1,156 @@
...
+            load_output = json.load(read_file)
+            self.create_endpoint(load_output)
+        self.wait_for_processing()
+        logging.info("Wait for processing executed")
+
+    def create_endpoint(self, cdk_output: dict) -> None:
+        load_balancer_url = cdk_output[self.stack_name].get('loadbalancerurl', None)
+        if load_balancer_url is None:
+            raise RuntimeError("Unable to fetch the cluster endpoint from cdk output")
+        self.args.cluster_endpoint = load_balancer_url
```
rishabh6788, line 87: Why are we overwriting the args.cluster_endpoint?
Need better way to handle this.

Divyaasm: Assuming cluster_endpoint is provided in self.args.cluster_endpoint, In cases where manifest and distribution _url are provided this is empty. I will remove the above line. 
`self.cluster_endpoint_with_port` should do the work for us

rishabh6788:
How about we work with existing implementation of `BenchmarkTestCluster` class and do the following:
1. in the `init` method check if cluster_endpoint arg is set, if yes, skip the entire logic and do nothing. 
2. in the start method check again for cluster_endpoint arg, if yes, skip cluster creation, just do wait_for_processing method to test if cluster is healthy. 
3. Same for terminate method. 
4. Adjust getting the version accordingly. 
@Divyaasm

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -8,94 +8,46 @@
...
-                    params_list.append(f" -c {key}=\'{value}\'")
-                else:
-                    params_list.append(f" -c {key}={value}")
-        role_params = (
-            f" --require-approval=never --plugin cdk-assume-role-credential-plugin"
-            f" -c assume-role-credentials:writeIamRoleName={role} -c assume-role-credentials:readIamRoleName={role} "
-        )
-        self.params = "".join(params_list) + role_params
-        self.is_endpoint_public = False
-        self.cluster_endpoint = None
```
rishabh6788, line 73: declare `self.cluster_endpoint = args.cluster_endpoint` and then override this class var in BenchmarkCreateCluster class?

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -8,94 +8,46 @@
...
-        if loadbalancer_url is None:
-            raise RuntimeError("Unable to fetch the cluster endpoint from cdk output")
-        self.cluster_endpoint = loadbalancer_url
-        self.cluster_endpoint_with_port = "".join([loadbalancer_url, ":", str(self.port)])
+        self.cluster_endpoint_with_port = "".join([self.args.cluster_endpoint, ":", str(self.port)])
 
     @property
     def endpoint(self) -> str:
-        return self.cluster_endpoint
+        return self.args.cluster_endpoint
```
rishabh6788, line 50: return self.cluster_endpoint as per above comment.

src/test_workflow/benchmark_test/benchmark_create_cluster.py
rishabh6788: override the self.cluster_endpoint class var here as mentioned in other comments.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -8,94 +8,46 @@
...
+            self.args.distribution_version = res_dict['version']['number']
         self.wait_for_processing()
-
-    def create_endpoint(self, cdk_output: dict) -> None:
-        loadbalancer_url = cdk_output[self.stack_name].get('loadbalancerurl', None)
-        if loadbalancer_url is None:
-            raise RuntimeError("Unable to fetch the cluster endpoint from cdk output")
-        self.cluster_endpoint = loadbalancer_url
-        self.cluster_endpoint_with_port = "".join([loadbalancer_url, ":", str(self.port)])
+        self.cluster_endpoint_with_port = "".join([self.args.cluster_endpoint, ":", str(self.port)])
```
rishabh6788, line 46: This may cause an issue. We will have to explicitly state that the endpoint should be passed without http or https. 
Verify if benchmark command accepts --target-url in the form of `https://<endpoint>:443`. If yes then ignore else we need to clarify it.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
```diff
@@ -8,94 +8,46 @@
...
-        self.cluster_endpoint = None
         self.cluster_endpoint_with_port = None
-        self.stack_name = f"opensearch-infra-stack-{self.args.stack_suffix}"
-        if self.manifest:
-            self.stack_name += f"-{self.manifest.build.id}-{self.manifest.build.architecture}"
+        self.password = None
 
     def start(self) -> None:
-        command = f"npm install && cdk deploy \"*\" {self.params} --outputs-file {self.output_file}"
+        command = f"curl -X GET http://{self.args.cluster_endpoint}" if self.args.insecure else f"curl -X GET https://{self.args.cluster_endpoint} -u 'admin:{get_password('2.12.0')}' --insecure"
```
rishabh6788, line 36: use the form `https://<endpoint> -ku user:pass`, remove `--insecure` flag.

src/test_workflow/benchmark_test/benchmark_test_cluster.py
Divyaasm: Sure Rishabh

rishabh6788 (approved):

PR #4574, opened 2024-03-26
URL encode test result files

### LINKED ISSUE #4569: URL encode links in test result files

Relatively minor issue: when Cypress tests fail and are put in the test report, the links aren't URL encoded, which means spaces in the link cause pyyaml to word wrap and messes with clicking links if trying to navigate through several failing tests.

URL encoding the file names in the [manifest output generator](https://github.com/opensearch-project/opensearch-build/blob/97770ce8a3fde527f366d91f12af2f2d432dc690/src/test_workflow/test_recorder/test_recorder.py#L75-L76) would make it easier to navigate the test manifest. We could update the linked line that creates the test paths to `return [os.path.join(base_path, relative_path, urllib.parse.quote_plus(file)) for file in files]`.

I'm happy to implement this myself, feel free to assign the issue to me.

_No response_

### Additional context

_No response_

Swiddis:
### Description
Following up on #4569: URL encode the filenames in the test manifest file to better support tools scanning these files for links.

### Issues Resolved
Closes #4569.

### DIFF
--- a/tests/tests_test_workflow/test_recorder/test_test_recorder.py
+++ b/tests/tests_test_workflow/test_recorder/test_test_recorder.py
@@ -99,6 +99,19 @@ class TestTestRecorder(unittest.TestCase):
         file_path = test_recorder._update_absolute_file_paths(["file1", "file2"], "working-directory", "sub-directory")
         self.assertEqual(file_path, [os.path.join("working-directory", "sub-directory", "file1"), os.path.join("working-directory", "sub-directory", "file2")])
 
+    @patch("test_workflow.test_recorder.test_recorder.TestResultsLogs")
+    @patch("test_workflow.test_recorder.test_recorder.RemoteClusterLogs")
+    @patch("test_workflow.test_recorder.test_recorder.LocalClusterLogs")
+    def test_update_absolute_file_paths(self, mock_local_cluster_logs: Mock, mock_remote_cluster_logs: Mock, mock_test_results_logs: Mock, *mock: Any) -> None:
+        test_recorder = TestRecorder(
+            "1234",
+            "integ-test",
+            "working-directory",
+            "https://ci.opensearch.org/ci/dbc/integ-test/"
+        )
+        file_path = test_recorder._update_absolute_file_paths(["A long test case name"], "working-directory", "sub-directory")
+        self.assertEqual(file_path, [os.path.join("working-directory", "sub-directory", "A+long+test+case+name")])
+
     @patch("os.walk")
     @patch("test_workflow.test_recorder.test_recorder.TestResultsLogs")
     @patch("test_workflow.test_recorder.test_recorder.RemoteClusterLogs")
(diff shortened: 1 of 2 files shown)


zelinh (commented):
Thanks for the PR @Swiddis.
I am aware of this issue and this might cause confusion; but I believe the link should be accessible if you include all of the URL address with spaces.

One thing I'm worried about this PR is that our implementation should also be compatible with local directory address when we pass in local `base_path` so replacing space here may not be working for local test record.

Swiddis:
Hm, good point on local test suites... I could try introducing a check for if `base_path` starts with `http(s)://` if that would suffice? Or maybe more generally `[scheme]://`.

I'm aware that the link is accessible if you copy the address with spaces -- it's mostly annoying with some text editors when copying directly from yaml, since they typically won't respect the yaml unquoted-string whitespace conventions and this ends up in needing to find and delete stray spaces.

zelinh:
> Hm, good point on local test suites... I could try introducing a check for if `base_path` starts with `http(s)://` if that would suffice? Or maybe more generally `[scheme]://`.
> 
> I'm aware that the link is accessible if you copy the address with spaces -- it's mostly annoying with some text editors when copying directly from yaml, since they typically won't respect the yaml unquoted-string whitespace conventions and this ends up in needing to find and delete stray spaces.

Yeah I think you could have another check to see if `base_path` starts with `https://` as what we did here. https://github.com/opensearch-project/opensearch-build/blob/b86cf756be5df4f32d73753bf45c12377792e234/src/test_workflow/test_recorder/test_recorder.py#L40-L44

Swiddis:
Updated the logic and associated test cases

peterzhuamazon (changes requested):
Does it make sense to change on line 40 instead?

peterzhuamazon:
> Does it make sense to change on line 40 instead?
> 
> https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/test_recorder/test_recorder.py#L40

Ignore it, I think I didnt realize the `file` value is appended later on.

peterzhuamazon (approved):

peterzhuamazon:
Thanks.

tests/tests_test_workflow/test_recorder/test_test_recorder.py
```diff
@@ -99,6 +99,19 @@ def test_update_absolute_file_paths(self, mock_local_cluster_logs: Mock, mock_re
...
+    @patch("test_workflow.test_recorder.test_recorder.LocalClusterLogs")
+    def test_update_absolute_file_paths_escaping(self, mock_local_cluster_logs: Mock, mock_remote_cluster_logs: Mock, mock_test_results_logs: Mock, *mock: Any) -> None:
+        test_recorder = TestRecorder(
+            "1234",
+            "integ-test",
+            "working-directory",
+            "https://ci.opensearch.org/ci/dbc/integ-test/"
+        )
+        file_path = test_recorder._update_absolute_file_paths(["A long test case name"], "https://working-directory", "sub-directory")
+        self.assertEqual(file_path, [os.path.join("https://working-directory", "sub-directory", "A+long+test+case+name")])
```
zelinh, line 113: For this case, I think we should just join them with `/` instead of `os.path.join`.

zelinh (approved):
LGTM! Thanks.

Issue #4504, opened 2024-03-05
[Improvement] Release Notes workflow Quality of Life Improvement and Standardization


peterzhuamazon:
Hi,

As of now, the consolidated release notes workflows has been automated using MarkdownTableWriter from pytablewriter.

There are, however, multiple aspects to be improved for the future.

1. The release notes is now in html format instead of the markdown, which add up complexity on maintenance and reduce readability without a render. Example: https://raw.githubusercontent.com/opensearch-project/opensearch-build/2.12.0/release-notes/opensearch-release-notes-2.12.0.md.
2. The workflow can compile either OpenSearch or OpenSearch-Dashboards consolidated release notes, but is not able to put them both together into one file. This add up efforts for the release manager to manually combine both into a final release notes PR.

We are looking for ways to resolve the above two issues with some quality of life improvements.

Thanks.

bbarani:
@peterzhuamazon Do you have any recommendation?

rishabh6788:
Will take a look.

rishabh6788:
Fixed as part of PR #4627

PR #4627, opened 2024-04-11
Refactor release-notes workflow

### LINKED ISSUE #4549: [Bug]: Release notes checker does not check release notes for core engines
### Describe the bug

With recent enhancement to compile consolidated release notes [PR](https://github.com/opensearch-project/opensearch-build/pull/4196), a change was introduced to skip few components [here](https://github.com/opensearch-project/opensearch-build/pull/4196/files#diff-367096ff0c801397154d58cb4c8fea65125f353b543f5bd686f55bf187499afc)

However, we do not add core engine (OpenSearch and OpenSearch-Dashboards) release notes, we do however need to check their presence in release notes tracker workflow that posts the comments on the release issue. 
Example: https://github.com/opensearch-project/opensearch-build/issues/4433#issuecomment-2004998714

### To reproduce

See https://github.com/opensearch-project/opensearch-build/issues/4433#issuecomment-2004998714

### Expected behavior

The table should also component: OpenSearch and OpenSearch-Dashboards. 

The condition is true for `action_type: compile` but the`action_type: check` should only skip `notifications-core` and not others

### Host / Environment

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

rishabh6788:
### Description
This PR:
* Fixes bug to also include core components while checking and compiling release notes. 
* Adds feature to check and compile release notes for OpenSearch + components and OpenSearch Dashboards + components in a single run. 
* Converts the html format back to markdown. 

### Issues Resolved
#4549 #4504 

### DIFF
--- a/src/release_notes_workflow/release_notes.py
+++ b/src/release_notes_workflow/release_notes.py
@@ -19,18 +19,19 @@ from system.temporary_directory import TemporaryDirectory
 
 class ReleaseNotes:
 
-    def __init__(self, manifest: InputManifest, date: str, action_type: str) -> None:
-        self.manifest = manifest
+    def __init__(self, input_manifests: List[InputManifest], date: str, action_type: str) -> None:
+        self.manifests = input_manifests  # type: ignore[assignment]
         self.date = date
         self.action_type = action_type
 
     def table(self) -> MarkdownTableWriter:
         table_result = []
-        for component in self.manifest.components.select():
-            if component.name == 'OpenSearch' or component.name == 'OpenSearch-Dashboards' or component.name == 'notifications-core':
-                continue
-            if hasattr(component, "repository"):
-                table_result.append(self.check(component))  # type: ignore[arg-type]
+        for manifest in self.manifests:
+            for component in manifest.components.select():
+                if (component.name == 'OpenSearch' or component.name == 'OpenSearch-Dashboards' or component.name == 'notifications-core') and self.action_type == 'compile':
+                    continue
+                if hasattr(component, "repository"):
+                    table_result.append(self.check(component, manifest.build.version))  # type: ignore[arg-type]
 
         # Sort table_result based on Repo column
         table_result.sort(key=lambda x: (x[0], x[1]) if len(x) > 1 else x[0])
@@ -43,13 +44,13 @@ class ReleaseNotes:
             raise ValueError("Invalid action_type. Use 'check' or 'compile'.")
 
         writer = MarkdownTableWriter(
-            table_name=f" {self.manifest.build.name} CommitID(after {self.date}) & Release Notes info",
+            table_name=f"Core Components CommitID(after {self.date}) & Release Notes info",
             headers=headers,
             value_matrix=table_result
         )
         return writer
 
-    def check(self, component: InputComponentFromSource) -> List:
+    def check(self, component: InputComponentFromSource, build_version: str) -> List:
         results = []
         with TemporaryDirectory(chdir=True) as work_dir:
             results.append(component.name)
@@ -61,7 +62,7 @@ class ReleaseNotes:
                     component.working_directory,
             ) as repo:
                 logging.debug(f"Checked out {component.name} into {repo.dir}")
-                release_notes = ReleaseNotesComponents.from_component(component, self.manifest.build.version, repo.dir)
+                release_notes = ReleaseNotesComponents.from_component(component, build_version, repo.dir)
                 commits = repo.log(self.date)
                 if len(commits) > 0:
                     last_commit = commits[-1]
--- a/src/run_releasenotes_check.py
+++ b/src/run_releasenotes_check.py
@@ -10,7 +10,9 @@ import os
 import re
 import shutil
 from collections import defaultdict
+from typing import List
 
+import markdownify
 import mistune
 import requests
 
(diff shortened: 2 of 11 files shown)


rishabh6788:
Attached sample release-notes for OS and OSD combined. 
@gaiksaya @peterzhuamazon 
[release_notes-2.13.0.md](https://github.com/opensearch-project/opensearch-build/files/14941980/release_notes-2.13.0.md)

src/release_notes_workflow/release_notes.py
```diff
@@ -19,18 +19,19 @@
...
-        self.manifest = manifest
+    def __init__(self, input_manifests: List[InputManifest], date: str, action_type: str) -> None:
+        self.manifests = input_manifests  # type: ignore[assignment]
         self.date = date
         self.action_type = action_type
 
     def table(self) -> MarkdownTableWriter:
         table_result = []
-        for component in self.manifest.components.select():
-            if component.name == 'OpenSearch' or component.name == 'OpenSearch-Dashboards' or component.name == 'notifications-core':
```
peterzhuamazon, line 30: I think we do want to ignore OS/OSD core because core has their own format and link reference isnt it?
cc: @bbarani

src/run_releasenotes_check.py
```diff
@@ -23,18 +25,33 @@
...
-
+    manifests: List[InputManifest] = []
     # storing temporary release notes for testing purposes
     BASE_FILE_PATH = "release_notes_workflow/results"
+
+    for input_manifests in args.manifest:
+        manifests.append(InputManifest.from_file(input_manifests))
+
+    #  Assuming that the OS and OSD manifests will be provided for the same release version
+    BUILD_VERSION = manifests[0].build.version
```
peterzhuamazon, line 36: Can we do a check between OS and OSD if user input both, and error out if it is not matching?

src/run_releasenotes_check.py
```diff
@@ -23,18 +25,33 @@
...
     table_filename = f"{BASE_FILE_PATH}/release_notes_table-{BUILD_VERSION}.md"
     urls_filename = f"{BASE_FILE_PATH}/release_notes_urls-{BUILD_VERSION}.txt"
 
+    def check_if_exists_then_delete(file_path: List[str]) -> None:
+        for file in file_path:
+            if os.path.exists(os.path.join(os.path.dirname(__file__), file)):
+                print(f"file {file} exists. Deleting")
+                os.remove(os.path.join(os.path.dirname(__file__), file))
+            else:
+                logging.info(f"The file {file} does not exist, creating new.")
```
peterzhuamazon, line 47: Shall we have a param to allow user to override or keep the old generated copy?

src/run_releasenotes_check.py
```diff
@@ -71,15 +88,17 @@ def create_urls_file_if_not_exists() -> None:
 
         urls_filepath = os.path.join(os.path.dirname(__file__), urls_filename)
         os.makedirs(os.path.dirname(urls_filepath), exist_ok=True)
-        with open(urls_filepath, "w") as urls_file:
+        with open(urls_filepath, "a") as urls_file:
```
peterzhuamazon, line 91: Benefit of changing from w to a?

src/run_releasenotes_check.py
```diff
@@ -188,11 +207,19 @@ def create_urls_file_if_not_exists() -> None:
             for item in temp_content:
                 outfile.write(markdown(item))
 
+    with open(RELEASE_NOTE_MD_PATH, 'r') as f:
+        html_content = f.read()
+
+    markdown_content = markdownify.markdownify(html_content, heading_style="ATX")
```
peterzhuamazon, line 213: So we basically still write as html then restore it back to md?
If so, do we think there is a case where we want user to choose if they want html or md output with a param?

Thanks.

peterzhuamazon (commented):
Hi @rishabh6788 ,

Have some comments about the PR.

Thanks.

src/release_notes_workflow/release_notes.py
rishabh6788: I can keep it for `check` and ignore at `compile`.

gaiksaya: Yeah that sounds better. We for sure need it for check but can ignore for compile.

src/run_releasenotes_check.py
rishabh6788: Seems redundant. If a user wants to generate a new one for the same version, better to overwrite the existing one, in case user wants to generate for another version, a new file will be created. There is already an `--output` option to save files elsewhere, the user can use that one to keep older files.

src/run_releasenotes_check.py
rishabh6788: Previously we were processing only one manifest, now since we process more than one and process them one by one in a loop we need to append, as `w` will overwrite the contents of the previous manifest.

src/run_releasenotes_check.py
rishabh6788: Lets follow one standard format in the code. 
If a user wants to convert, they can do it from command line tools.

rishabh6788:
> Mend show issues with this cve:
> 
> ```
> [CVE-2024-22949](https://www.mend.io/vulnerability-database/CVE-2024-22949)
> 	Medium 	5.5 	jfreechart-1.0.19.jar 		None
> ```

I think this is coming from jenkins core.

src/run_releasenotes_check.py
rishabh6788: Added checks.

src/release_notes_workflow/release_notes.py
rishabh6788: Done

src/run_releasenotes_check.py
peterzhuamazon: Got it. Make sense. Close now.

src/run_releasenotes_check.py
peterzhuamazon: Got it!

src/run_releasenotes_check.py
peterzhuamazon: Got it.

src/run_releasenotes_check.py
```diff
@@ -23,18 +25,41 @@
...
-    BUILD_VERSION = manifest_file.build.version
-
+    manifests: List[InputManifest] = []
     # storing temporary release notes for testing purposes
     BASE_FILE_PATH = "release_notes_workflow/results"
+
+    for input_manifests in args.manifest:
+        manifests.append(InputManifest.from_file(input_manifests))
+
+    if len(args.manifest) == 2:
```
peterzhuamazon, line 35: Do you have a check to ensure user input max of 2 manifests? What if someone entered more than 2?

peterzhuamazon: nvm, missed the check on more than 2.

peterzhuamazon (approved):
Thanks @rishabh6788 , approved!

Issue #4442, opened 2024-02-11
[Bug]: Validation run for OpenSearch Dashboard fails with unauthorized error


gaiksaya:
### Describe the bug

Tried running validation workflow for 2.3.0. It passed for OpenSearch but failed for Dashboards with below error:

```
2024-02-11 16:26:36 INFO
Status_code ->401
response_text ->
/home/gaiksaya/.local/share/virtualenvs/opensearch-build-Jk4AL_lu/lib/python3.7/site-packages/urllib3/connectionpool.py:1068: InsecureRequestWarning: Unverified HTTPS request is being made to host 'localhost'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  InsecureRequestWarning,
2024-02-11 16:26:37 INFO
Request_url ->https://localhost:9200/_cat/health?v

2024-02-11 16:26:37 INFO
Status_code ->401
response_text ->
2024-02-11 16:26:37 INFO
Request_url ->http://localhost:5601/api/status

2024-02-11 16:26:37 INFO
Status_code ->401
response_text ->{"statusCode":401,"error":"Unauthorized","message":"Authentication Exception"}
Traceback (most recent call last):
  File "/local/home/gaiksaya/opensearch-build/src/validation_workflow/validation.py", line 52, in run
    return self.download_artifacts() and self.installation() and self.start_cluster() and self.validation() and self.cleanup()
  File "/local/home/gaiksaya/opensearch-build/src/validation_workflow/tar/validation_tar.py", line 75, in validation
    raise Exception(f'Not all tests Pass : {counter}')
Exception: Not all tests Pass : There are 0/4 test cases Pass

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./src/run_validation.py", line 28, in <module>
    sys.exit(main())
  File "./src/run_validation.py", line 22, in main
    test_result = ValidationTestRunner.dispatch(args, args.distribution).run()
  File "/local/home/gaiksaya/opensearch-build/src/validation_workflow/validation.py", line 54, in run
    raise Exception(f'An error occurred while running the validation tests: {str(e)}')
Exception: An error occurred while running the validation tests: Not all tests Pass : There are 0/4 test cases Pass
```

### To reproduce

Run the below command:

```
[...]

PR #4664, opened 2024-04-30
Validation for OS and OSD fail with error when executed multiple times on local

Divyaasm:
### Description
Handle complete cleanup after running the workflow.

### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-build/issues/4442

### DIFF
--- a/src/run_validation.py
+++ b/src/run_validation.py
@@ -9,6 +9,7 @@ import logging
 import sys
 
 from system import console
+from system.temporary_directory import TemporaryDirectory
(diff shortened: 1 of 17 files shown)


gaiksaya (approved):
So the entire issue was because of incomplete clean up?

Divyaasm:
> So the entire issue was because of incomplete clean up?

Tar and zip handles the cluster startup differently. When an artifact is not valid we still need to kill the processes that started the cluster. But the others - rpm, deb and yum fail at the startup which doesn't affect the consecutive runs.

PR #4739, opened 2024-05-30
Remove nyc_taxis multi client/shard cases and add big5 multi client cases

jed326:
### Description
After analyzing the multi client cases for `nyc_taxis`, we determined the data from that is not very good as 4/5 operations complete too quickly to get a good picture of resource utilization vs. performance. Instead will use big5 test case as there are longer running operations in that workload. Also removed all multi shard cases.


jed326:
Tested this on staging endpoint: https://ci-staging.opensearch.org/job/benchmark-test-endpoint/50/parameters/

rishabh6788 (approved):

PR #4784, opened 2024-06-17
feat: split osd core test into ci groups

SuZhou-Joe:
### Description

OSD core contains more than 50 specs, with hundreds of tests getting ran sequentially. While many side effects are performed across test cases, It is usually the case that core dashboards test cases are failed or flaky, and hard to troubleshoot. In order to maintain a more stable test env and more reproducible CI, this PR split sanity test cases of core dashboards into CI groups.

In order to make this work, https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/1411 has to be merged in advance.

### Issues Resolved

### DIFF
--- a/manifests/2.15.0/opensearch-dashboards-2.15.0-test.yml
+++ b/manifests/2.15.0/opensearch-dashboards-2.15.0-test.yml
@@ -5,7 +5,111 @@ ci:
   image:
     name: opensearchstaging/ci-runner:ci-runner-rockylinux8-opensearch-dashboards-integtest-v4
 components:
-  - name: OpenSearch-Dashboards
+  - name: OpenSearch-Dashboards-ci-group-1
+    integ-test:
+      test-configs:
+        - with-security
+        - without-security
+      additional-cluster-configs:
+        vis_builder.enabled: true
+        data_source.enabled: true
+        savedObjects.maxImportPayloadBytes: 10485760
+        server.maxPayloadBytes: 1759977
+        logging.json: false
+        data.search.aggs.shardDelay.enabled: true
+        csp.warnLegacyBrowsers: false
+  - name: OpenSearch-Dashboards-ci-group-2
+    integ-test:
+      test-configs:
+        - with-security
+        - without-security
+      additional-cluster-configs:
+        vis_builder.enabled: true
+        data_source.enabled: true
+        savedObjects.maxImportPayloadBytes: 10485760
+        server.maxPayloadBytes: 1759977
+        logging.json: false
+        data.search.aggs.shardDelay.enabled: true
+        csp.warnLegacyBrowsers: false
+  - name: OpenSearch-Dashboards-ci-group-3
+    integ-test:
+      test-configs:
+        - with-security
+        - without-security
+      additional-cluster-configs:
+        vis_builder.enabled: true
+        data_source.enabled: true
+        savedObjects.maxImportPayloadBytes: 10485760
+        server.maxPayloadBytes: 1759977
+        logging.json: false
+        data.search.aggs.shardDelay.enabled: true
+        csp.warnLegacyBrowsers: false
+  - name: OpenSearch-Dashboards-ci-group-4
+    integ-test:
+      test-configs:
+        - with-security
+        - without-security
+      additional-cluster-configs:
+        vis_builder.enabled: true
+        data_source.enabled: true
+        savedObjects.maxImportPayloadBytes: 10485760
+        server.maxPayloadBytes: 1759977
+        logging.json: false
+        data.search.aggs.shardDelay.enabled: true
+        csp.warnLegacyBrowsers: false
+  - name: OpenSearch-Dashboards-ci-group-5
+    integ-test:
+      test-configs:
+        - with-security
+        - without-security
+      additional-cluster-configs:
+        vis_builder.enabled: true
+        data_source.enabled: true
+        savedObjects.maxImportPayloadBytes: 10485760
+        server.maxPayloadBytes: 1759977
+        logging.json: false
+        data.search.aggs.shardDelay.enabled: true
+        csp.warnLegacyBrowsers: false
+  - name: OpenSearch-Dashboards-ci-group-6
+    integ-test:
+      test-configs:
+        - with-security
+        - without-security
+      additional-cluster-configs:
+        vis_builder.enabled: true
+        data_source.enabled: true
+        savedObjects.maxImportPayloadBytes: 10485760
+        server.maxPayloadBytes: 1759977
+        logging.json: false
+        data.search.aggs.shardDelay.enabled: true
+        csp.warnLegacyBrowsers: false
+  - name: OpenSearch-Dashboards-ci-group-7
+    integ-test:
+      test-configs:
+        - with-security
+        - without-security
+      additional-cluster-configs:
+        vis_builder.enabled: true
+        data_source.enabled: true
+        savedObjects.maxImportPayloadBytes: 10485760
+        server.maxPayloadBytes: 1759977
+        logging.json: false
+        data.search.aggs.shardDelay.enabled: true
+        csp.warnLegacyBrowsers: false
+  - name: OpenSearch-Dashboards-ci-group-8
+    integ-test:
+      test-configs:
+        - with-security
+        - without-security
+      additional-cluster-configs:
+        vis_builder.enabled: true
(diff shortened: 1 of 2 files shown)


peterzhuamazon:
Convert back to draft because this would not work.
Due to our Jenkinsfile is checking if the component exist.
It is comparing the name of the component with the one in build manifest.

https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/integ-test.jenkinsfile#L188-L192

We still need more changes if we want to get this in.

Thanks.

peterzhuamazon:
Taking a look on the jenkinsfile changes.

manifests/2.15.0/opensearch-dashboards-2.15.0-test.yml
```diff
@@ -5,7 +5,111 @@ ci:
   image:
     name: opensearchstaging/ci-runner:ci-runner-rockylinux8-opensearch-dashboards-integtest-v4
 components:
-  - name: OpenSearch-Dashboards
+  - name: OpenSearch-Dashboards-ci-group-1
```
gaiksaya, line 8: I don't think this would throw error saying component not found as there is a mismatch.

SuZhou-Joe:
@peterzhuamazon https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/integ-test.jenkinsfile#L188-L192

rishabh6788:
> @peterzhuamazon https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/integ-test.jenkinsfile#L188-L192 I do not have enough context on the code in build repo but is this line necessary? What will happen if we remove these lines.

This ensures that for the component for which the integration test is being run is actually present in the distribution artifact that is being used to spin up the OSD process. In the present scenario this is helpful when a component is passed in COMPONENT_NAME field but it is not present in build manifest of distribution artifact selected for that run. 

When nothing is passed this list is same as component list in build manifest. 
In your case I believe you can add an `OR` condition to check if the component name is of `OpenSearch-Dashboards-ci-group` pattern then continue as it corresponds to OSD and OSD will always be present in a valid build manifest. 

Please correct me if I'm wrong @peterzhuamazon

manifests/2.15.0/opensearch-dashboards-2.15.0-test.yml
rishabh6788: I think we are good from the python side as it only checks test manifest and gets list of components. If we can handle it from Jenkinsfile side as mentioned in my comment below then we are good. Right?

manifests/2.15.0/opensearch-dashboards-2.15.0-test.yml
```diff
@@ -5,7 +5,111 @@ ci:
...
+        - without-security
+      additional-cluster-configs:
+        vis_builder.enabled: true
+        data_source.enabled: true
+        savedObjects.maxImportPayloadBytes: 10485760
+        server.maxPayloadBytes: 1759977
+        logging.json: false
+        data.search.aggs.shardDelay.enabled: true
+        csp.warnLegacyBrowsers: false
+  - name: OpenSearch-Dashboards-ci-group-3
```
prudhvigodithi, line 34: Hey @SuZhou-Joe may I know what tests or components part of these groups ?  example for `OpenSearch-Dashboards-ci-group-3`, coming from this PR https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/1411/files, how to know what tests or components are part of `OpenSearch-Dashboards-ci-group-3`?

peterzhuamazon:
> @peterzhuamazon Update the jenkins file to get pass the check, how do you like the change?

This will still having issues due to the lib is using `name` to find components when calling test.sh.

Please wait for a bit before we sort this out.

We are very close to the release date and lets just rush to get this change in. 

Thanks.

manifests/2.15.0/opensearch-dashboards-2.15.0-test.yml
SuZhou-Joe: The specs are written in the `package.json`. 
https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/package.json#L28-L36

prudhvigodithi: I would like to get this change documentation somewhere (may be in @peterzhuamazon should we update the https://github.com/opensearch-project/opensearch-build/wiki/Testing-the-Distribution with this change, adding few lines on how does this grouping work).
Thanks

rishabh6788:
Dived a little deeper into the code and it is not as straight forward as it looks like. Simply adding `OpenSearch-Dashboards-ci-group-*` component in the test manifest will not work. 
This is because [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/integ_test_runner.py#L42) the passed component is verified against build manifest. If the component exists in build-manifest then the flow goes through, but since `OpenSearch-Dashboards-ci-group-*` is not a valid component the code will always throw an exception [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/manifests/component_manifest.py#L67-L70). 
Even though we try to fool it by adding a condition like `invalid = [item for item in focus if item not in self and 'ci-group' not in item]` it will error out  [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/manifests/component_manifest.py#L72-L76). 

This is not a straight forward problem to solve and we may need to revisit how to handle ci-groups in test manifest, maybe add a new key to the schema. 
Thoughts @gaiksaya @peterzhuamazon 
CC: @getsaurabh02

peterzhuamazon:
> Dived a little deeper into the code and it is not as straight forward as it looks like. Simply adding `OpenSearch-Dashboards-ci-group-*` component in the test manifest will not work. This is because [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/integ_test_runner.py#L42) the passed component is verified against build manifest. If the component exists in build-manifest then the flow goes through, but since `OpenSearch-Dashboards-ci-group-*` is not a valid component the code will always throw an exception [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/manifests/component_manifest.py#L67-L70). Even though we try to fool it by adding a condition like `invalid = [item for item in focus if item not in self and 'ci-group' not in item]` it will error out [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/manifests/component_manifest.py#L72-L76).
> 
> This is not a straight forward problem to solve and we may need to revisit how to handle ci-groups in test manifest, maybe add a new key to the schema. Thoughts @gaiksaya @peterzhuamazon CC: @getsaurabh02

Agreed with @rishabh6788 , we will need to implement this properly, without rushing for a quick change here.
Let's discuss more on this topic.

Thanks.

gaiksaya:
> Dived a little deeper into the code and it is not as straight forward as it looks like. Simply adding `OpenSearch-Dashboards-ci-group-*` component in the test manifest will not work. This is because [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/integ_test/integ_test_runner.py#L42) the passed component is verified against build manifest. If the component exists in build-manifest then the flow goes through, but since `OpenSearch-Dashboards-ci-group-*` is not a valid component the code will always throw an exception [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/manifests/component_manifest.py#L67-L70). Even though we try to fool it by adding a condition like `invalid = [item for item in focus if item not in self and 'ci-group' not in item]` it will error out [here](https://github.com/opensearch-project/opensearch-build/blob/main/src/manifests/component_manifest.py#L72-L76).
> 
> This is not a straight forward problem to solve and we may need to revisit how to handle ci-groups in test manifest, maybe add a new key to the schema. Thoughts @gaiksaya @peterzhuamazon CC: @getsaurabh02

Yes that is what I meant by my comment here https://github.com/opensearch-project/opensearch-build/pull/4784#discussion_r1643539535
One approach I can think of adding `ci_group` to the test manifest schema as an optional key. In this way you can add `OpenSearch-Dashboards` multiple times with different ci_group each time or just add range. But yeah needs to be thought through and not a straight forward implementation.

ashwin-pc:
@rishabh6788 @gaiksaya what are the next steps here?

peterzhuamazon:
Adding new changes and test with:
* https://github.com/opensearch-project/opensearch-build/pull/4796

Issue #4837, opened 2024-07-10
[BENCHMARK] Add support for opensearch-benchmark's `compare` feature


rishabh6788:
### Is your feature request related to a problem? Please describe

As of today we are using [opensearch-benchmark's ](https://github.com/opensearch-project/opensearch-benchmark) `execute-test` api to run performance benchmark tests on nightly and as well as ad-hoc basis. 
We are also adding support to run performance benchmark on OpenSearch [pull requests](https://github.com/opensearch-project/opensearch-build/pull/4831). 

So, while a developer is working on their changes, apart from just getting performance benchmark numbers published on the PR, they might also want to know how their changes compare against a baseline. 

### Describe the solution you'd like

Fortunately, opensearch-benchmark's [compare](https://opensearch.org/docs/latest/benchmark/reference/commands/compare/) command solves this problem for us. It only requires `baseline` and `contender` `test-execution-ids` as input params and publishes a detailed comparison result. 

We need to add this feature into existing [benchmark workflow](https://github.com/opensearch-project/opensearch-build/blob/main/src/run_benchmark_test.py) and add a job in jenkins, similar to [benchmark-test](https://build.ci.opensearch.org/job/benchmark-test/) to execute this workflow on ad-hoc basis and as well as on pull request.

rishabh6788:
Assigning this to @OVI3D0

OVI3D0:
Commenting for assignment

PR #4864, opened 2024-07-18
Add support for opensearch-benchmark's `compare` feature

OVI3D0:
### Description
Adds support for developers to use the `compare` feature from opensearch-benchmark to compare their changes against a baseline.

### Issues Resolved
[#4837](https://github.com/opensearch-project/opensearch-build/issues/4837#issue-2401462817)


src/test_workflow/compare_benchmark/compare_args.py
```diff
@@ -0,0 +1,41 @@
...
+    id2: str
+    results_format: str
+    results_numbers_align: str
+    results_file: str
+    show_in_results: str
+
+    def __init__(self) -> None:
+        self.parser = argparse.ArgumentParser(description="Compare two IDs")
+        self.parser.add_argument("id1", type=str, help="The baseline ID to compare")
+        self.parser.add_argument("id2", type=str, help="The contender ID to compare")
```
peterzhuamazon, line 22: Can we update the README file of test_workflow to know how exactly this feature work?

src/test_workflow/compare_benchmark/compare_args.py
```diff
@@ -0,0 +1,41 @@
...
+        self.parser.add_argument("--results-format", type=str, help="Defines the output format for the results, markdown or csv (default: markdown)")
+        self.parser.add_argument("--results-numbers-align", type=str, help="Defines the format for the command line results. (Default: right)")
+        self.parser.add_argument("--results-file", type=str, help="File path to write the results file to")
+        self.parser.add_argument("--show-in-results", type=str, help="Determines whether to include the comparison in the results file")
```
peterzhuamazon, line 26: Can you add (Defaults) to all these optional parameters?
Also use Capital for all of them.

src/test_workflow/compare_benchmark/compare_args.py
```diff
@@ -0,0 +1,41 @@
...
+        self.results_numbers_align = self.args.results_numbers_align if self.args.results_numbers_align else "right"
+        self.results_file = self.args.results_file if self.args.results_file else None
+        self.show_in_results = self.args.show_in_results if self.args.show_in_results else None
+
+
+if __name__ == "__main__":
+    try:
+        compare_args = CompareArgs()
+    except argparse.ArgumentError as e:
+        print(f"Error: {e}")
```
peterzhuamazon, line 41: Using logging, instead of print here.

src/test_workflow/compare_benchmark/compare_test_runner.py
```diff
@@ -0,0 +1,62 @@
...
+from test_workflow.compare_benchmark.compare_args import CompareArgs
+
+
+class CompareTestRunner:
+    def __init__(self) -> None:
+        self.container_name = "compare-container"
+
+    def run_comparison(self, compare_args: CompareArgs) -> None:
+        """
+        Perform the comparison between two test results using the OSB tool.
```
peterzhuamazon, line 19: Use `OpenSearch-Benchmark` instead of OSB here.

src/test_workflow/compare_benchmark/compare_test_runner.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+
+        # construct the command to compare two test executions
+        command = (
+            f"docker run --name {self.container_name} "
+            "-v ~/.benchmark/benchmark.ini:/opensearch-benchmark/.benchmark/benchmark.ini "
+            f"opensearchproject/opensearch-benchmark:1.6.0 "
+            f"compare --baseline={compare_args.id1} --contender={compare_args.id2} "
+        )
```
peterzhuamazon, line 28: Just my 2cents:

We should assume user is running `opensearch-benchmark` command in an env with such command already available. We should not help user setup docker, especially if someone is using system such as windows that does not always support linux docker by default. Also, you are assuming user must have docker installed by default. And you probably need to take case of the cleanup issue if not using `--rm` in the run commands.

I would suggest just invote `opensearch-benchmark` call directly, and let user take care of their own env setup. You could, however, mention in your readme file and refer to `opensearch-benchmark` repo on how to set it up.

src/test_workflow/compare_benchmark/compare_test_runner.py
```diff
@@ -0,0 +1,62 @@
...
+            subprocess.run(copy_command, shell=True, check=True)
+
+        except subprocess.CalledProcessError as e:
+            print(f"Error executing command: {e.output}")
+
+        finally:
+            self.cleanup()
+
+    def cleanup(self) -> None:
+        subprocess.check_call(f"docker rm {self.container_name}", cwd=os.getcwd(), shell=True)
```
peterzhuamazon, line 62: Same as above.

tests/test_run_compare.py
```diff
@@ -0,0 +1,43 @@
...
+import unittest
+from typing import Any
+from unittest.mock import Mock, patch
+
+import pytest
+
+from run_compare_tests import main
+
+
+class TestRunCompareTests(unittest.TestCase):
```
peterzhuamazon, line 17: Need more test.

peterzhuamazon (changes requested):
HI @OVI3D0 ,

Please see comments on the review.

Thanks.

test.sh
```diff
@@ -23,8 +23,11 @@ case $1 in
   "benchmark-test")
   "$DIR/run.sh" "$DIR/src/run_benchmark_test.py" "${@:2}"
   ;;
+  "compare")
```
rishabh6788, line 26: nit: `benchmark-compare`?

src/test_workflow/compare_benchmark/compare_args.py
```diff
@@ -0,0 +1,41 @@
...
+    id1: str
+    id2: str
+    results_format: str
+    results_numbers_align: str
+    results_file: str
+    show_in_results: str
+
+    def __init__(self) -> None:
+        self.parser = argparse.ArgumentParser(description="Compare two IDs")
+        self.parser.add_argument("id1", type=str, help="The baseline ID to compare")
```
rishabh6788, line 21: rename id1 and id2 match OSB, `baseline` and `contender`.

src/test_workflow/compare_benchmark/compare_args.py
rishabh6788: Te defaults will be set by OSB itself when these parameters are not passed. 
i'm okay with anything here.

src/test_workflow/compare_benchmark/compare_test_runner.py
rishabh6788: nit: `opensearch-benchmark`

src/test_workflow/compare_benchmark/compare_test_runner.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if compare_args.results_format:
+            command += f"--results-format={compare_args.results_format} "
+
+        if compare_args.results_numbers_align:
+            command += f"--results-numbers-align={compare_args.results_numbers_align} "
+
+        if compare_args.results_file:
+            # create a temporary directory for the compare results file in the container
+            container_results_dir = "/tmp/results"
```
rishabh6788, line 38: The final results file is actually written inside the container which then should be copied to local once the process has completed. 
this argument should just be `final_results.md` so that the file is created inside the container at `/opensearch-benchmark/` path. 
Same is happening with nightly benchmark runs, see https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/benchmark_test/benchmark_test_suite.py#L80. 

The file is then copied to local, see https://github.com/opensearch-project/opensearch-build/blob/main/src/test_workflow/benchmark_test/benchmark_test_suite.py#L96-L101

src/test_workflow/compare_benchmark/compare_test_runner.py
```diff
@@ -0,0 +1,62 @@
...
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+import os
+import subprocess
+
+from test_workflow.compare_benchmark.compare_args import CompareArgs
+
+
+class CompareTestRunner:
```
rishabh6788, line 13: Instead of a new class let us try to leverage the existing `BenchmarkTestSuite` class. 
You can modify the existing `BenchmarkArgs` class and leverage [Subparsers](https://dev.to/taikedz/ive-parked-my-side-projects-3o62) to add different targets with each having its own set of arguments. 

you can then modify the `BenchmarkTestSuite` class and make it more modular to format and execute commands based on the sub-parser passed. 
Example:
```
def main():
    parser = argparse.ArgumentParser(description="Test runner and comparator")
    subparsers = parser.add_subparsers(dest="command", help="Available commands")

    # run-test command
    run_test_parser = subparsers.add_parser("execute-test", help="Run a test")
    run_test_parser.add_argument("-i", "--input-file", required=True, help="Input file for the test")
    run_test_parser.add_argument("-o", "--output-file", required=True, help="Output file for the test results")
    run_test_parser.add_argument("-v", "--verbose", action="store_true", help="Enable verbose mode")

    # compare command
    compare_parser = subparsers.add_parser("compare", help="Compare two files")
    compare_parser.add_argument("file1", help="First file to compare")
    compare_parser.add_argument("file2", help="Second file to compare")
    compare_parser.add_argument("-t", "--tolerance", type=float, default=0.01, help="Comparison tolerance")
```
so now you would run the script like `python ./run_benchmark_test execute-test <existing arguments>` and `python ./run_benchmark_test compare <compare arguments>`. 

Follow the existing code and make necessary changes.

rishabh6788: `BenchmarkTestSuite` can be [Abstract Base Class](https://www.geeksforgeeks.org/abstract-classes-in-python/) and execute method can be made abstract to allow each method to implement it as per their need. you can then have `BenchmarkTestSuiteExecute` and `BenchmarkTestSuiteCompare` that extend `BenchmarkTestSuite`.

src/run_benchmark_test.py
```diff
@@ -13,21 +13,31 @@
...
     """
     benchmark_args = BenchmarkArgs()
-    console.configure(level=benchmark_args.logging_level)
-    if benchmark_args.bundle_manifest:
-        manifest: Union[BundleManifest, BuildManifest] = BundleManifest.from_file(benchmark_args.bundle_manifest) if not benchmark_args.min_distribution else \
-            BuildManifest.from_file(benchmark_args.bundle_manifest)
-        BenchmarkTestRunners.from_args(benchmark_args, manifest).run()
+
+    if benchmark_args.command == "execute-test":
+        console.configure(level=benchmark_args.logging_level)
```
rishabh6788, line 28: This can stay outside of the if condition as logging is required in all the cases.

src/run_compare_tests.py
rishabh6788, line 1: Is this file required given we are using the same args class to classify `compare` and `execute-test` commands?

src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -50,117 +50,162 @@ class BenchmarkArgs:
...
+                                         ' E.g. --include-tasks="scroll"',)
+        execute_test_parser.add_argument("--capture-node-stat", dest="telemetry", action="append_const", const="node-stats",
+                                         help="Enable opensearch-benchmark to capture node stat metrics such as cpu, mem, jvm etc as well.")
+        execute_test_parser.add_argument("--capture-segment-replication-stat", dest="telemetry", action="append_const", const="segment-replication-stats",
+                                         help="Enable opensearch-benchmark to segment_replication stat metrics such as replication lag.")
+        execute_test_parser.add_argument("--telemetry-params", dest="telemetry_params",
+                                         help="Allows to set parameters for telemetry devices. Accepts json input.")
+        execute_test_parser.add_argument("-v", "--verbose", action="store_const", default=logging.INFO, const=logging.DEBUG, dest="logging_level",
+                                         help="Show more verbose output.")
+        execute_test_parser.add_argument("--ml-node-storage", dest="ml_node_storage",
```
rishabh6788, line 143: remove this, duplicate.

src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -50,117 +50,162 @@ class BenchmarkArgs:
...
+                                         help="User provided ml-node ebs block storage size defaults to 100Gb")
+
+        # command to run comparison
+        compare_parser = subparsers.add_parser("compare", help="Compare two IDs")
```
rishabh6788, line 147: add `--benchmark-config` arg for compare command as well, since we need to use the same config to fetch datastore details.

src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -50,117 +50,162 @@ class BenchmarkArgs:
...
-        if self.bundle_manifest is None and self.distribution_url is None and self.cluster_endpoint is None:
-            raise Exception('Please provide either --bundle-manifest or --distribution-url  or --cluster_endpoint to run the performance test.')
-        elif self.distribution_url and self.distribution_version is None:
-            raise Exception("--distribution-version is required parameter while using --distribution-url param.")
+            if self.bundle_manifest is None and self.distribution_url is None and self.cluster_endpoint is None:
+                raise Exception("Please provide either --bundle-manifest or --distribution-url  or --cluster_endpoint to run the performance test.")
+            elif self.distribution_url and self.distribution_version is None:
+                raise Exception("--distribution-version is required parameter while using --distribution-url param.")
+        elif args.command == "compare":
+            self.baseline = args.baseline if hasattr(args, "baseline") else None
```
rishabh6788, line 203: raise exception if either `baseline` or `contender` value is not provided as they are mandatory arguments for OSB compare command.

src/run_compare_tests.py
OVI3D0: No, it's not required. Removing

src/run_benchmark_test.py
rishabh6788: All the comments can be resolved once they have been addressed. @OVI3D0

src/test_workflow/benchmark_test/benchmark_test_base.py
rishabh6788, line 1: rename this file for `benchmark_test_suite_runners` and class `BenchmarkTestSuiteRunners` and do following. 
```
class BenchmarkTestSuiteRunners:
    SUITES = {
        "execute-test": ExecuteTestSuite,
        "compare": CompareTestSuite
    }
    
    @classmethod
    def from_args(cls, args: BenchmarkArgs, endpoint=None, security=False, password=''):
        print("I AM IN ARGS")
        print(f"stack suffix is {args.stack_suffix}")
        test_class = cls.SUITES.get(args.command)
        if test_class is None:
            raise ValueError(f"Unknown command: {args.command}")
        
        if args.command == 'execute-test':
            print("HERE!!")
            return test_class(args, endpoint, security, password)
        else:
            return test_class(args)
```

Basically use this class to return what kind of OSB command class you want to return.

src/test_workflow/benchmark_test/benchmark_test_suite.py
rishabh6788, line 1: Make this class ABC and do the following. 
```
import os
import subprocess
from abc import ABC, abstractmethod
from test_workflow.benchmark_test.benchmark_args import BenchmarkArgs


class BenchmarkTestSuite(ABC):
    
    def __init__(self, args: BenchmarkArgs) -> None:
        self.args = args
        self.command = ''

    @abstractmethod
    def execute(self) -> None:
        "Implement execute logic in respective base classes"
        pass
    
    @abstractmethod
    def form_command(self) -> str:
        pass
    
    def cleanup(self) -> None:
        subprocess.check_call(f"docker rm -f docker-container-{self.args.stack_suffix}", cwd=os.getcwd(), shell=True)
```

src/test_workflow/benchmark_test/benchmark_test_base.py
```diff
@@ -0,0 +1,45 @@
...
+    password: str
+
+    def __init__(
+            self,
+            endpoint: Any,
+            security: bool,
+            args: BenchmarkArgs,
+            password: str
+    ) -> None:
+        self.endpoint = endpoint
```
rishabh6788, line 30: only `args` can be initialized in this init method and remaining values can be initialized in respective command subclasses. You do not need to initialize any of these for `compare` command class, all the required parameters will be available from `args` in the super class.

src/test_workflow/benchmark_test/benchmark_test_base.py
```diff
@@ -0,0 +1,45 @@
...
+        self.security = security
+        self.args = args
+        self.password = password
+        self.command = None
+
+    @abstractmethod
+    def form_command(self, args: BenchmarkArgs) -> str:
```
rishabh6788, line 37: don't need args argument here.

src/test_workflow/benchmark_test/benchmark_test_suite.py
rishabh6788, line 1: This can be called `benchmark_test_suite_runners` and can be called to return respective command class. 
Something like:
```
class BenchmarkTestSuiteRunners:
    SUITES = {
        "execute-test": ExecuteTestSuite,
        "compare": CompareTestSuite
    }
    
    @classmethod
    def from_args(cls, args: BenchmarkArgs, endpoint=None, security=False, password=''):
        print("I AM IN ARGS")
        print(f"stack suffix is {args.stack_suffix}")
        test_class = cls.SUITES.get(args.command)
        if test_class is None:
            raise ValueError(f"Unknown command: {args.command}")
        
        if args.command == 'execute-test':
            print("HERE!!")
            return test_class(args, endpoint, security, password)
        else:
            return test_class(args)
```

src/test_workflow/benchmark_test/benchmark_test_suite_execute.py
```diff
@@ -0,0 +1,91 @@
...
+import subprocess
+
+import pandas as pd
+
+from system.temporary_directory import TemporaryDirectory
+from test_workflow.benchmark_test.benchmark_args import BenchmarkArgs
+from test_workflow.benchmark_test.benchmark_test_base import BenchmarkTestBase
+
+
+class ExecuteTestSuite(BenchmarkTestBase):
```
rishabh6788, line 22: Rename this to match the file name, `BenchmarkTestSuiteExecute`. 
Once you have made `BenchmarkTestSuite` class ABC class, replace it here. 
Also add `init` method to initialize required variables like `endpoint`, `password`, etc. 
E.g.
```
def __init__(self, args: BenchmarkArgs, endpoint: str, security: bool, password: str) -> None:
        super().__init__(args)
        self.endpoint = endpoint
        self.security = security
        self.password = password
```
Do the same for `compare` subclass. You may not need to set any other variable apart from doing `super().__init__(args)` as all the parameters for compare command are available from `args`.

src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -50,117 +50,172 @@ class BenchmarkArgs:
...
+                                         help="Enable opensearch-benchmark to segment_replication stat metrics such as replication lag.")
+        execute_test_parser.add_argument("--telemetry-params", dest="telemetry_params",
+                                         help="Allows to set parameters for telemetry devices. Accepts json input.")
+        execute_test_parser.add_argument("-v", "--verbose", action="store_const", default=logging.INFO, const=logging.DEBUG, dest="logging_level",
+                                         help="Show more verbose output.")
+        execute_test_parser.add_argument("--ml-node-storage", dest="ml_node_storage",
+                                         help="User provided ml-node ebs block storage size defaults to 100Gb")
+
+        # command to run comparison
+        compare_parser = subparsers.add_parser("compare", help="Compare two tests using their test execution IDs")
```
rishabh6788, line 147: `suffix` is required for compare as well. 
`compare_parser.add_argument("--suffix", dest="suffix", help="Suffix to be added to stack name for performance comparison")`.

src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -50,117 +50,172 @@ class BenchmarkArgs:
...
+        elif args.command == "compare":
+            if not args.baseline or not args.contender:
+                raise ValueError("Both 'baseline' and 'contender' arguments are required for the 'compare' command.")
+            self.baseline = args.baseline
+            self.contender = args.contender
+            self.results_format = args.results_format if hasattr(args, "results_format") else None
+            self.results_numbers_align = args.results_numbers_align if hasattr(args, "results_numbers_align") else None
+            self.results_file = args.results_file if hasattr(args, "results_file") else None
+            self.show_in_results = args.show_in_results if hasattr(args, "show_in_results") else None
+            self.stack_suffix = "comparison"
```
rishabh6788, line 216: replace this with `self.stack_suffix = args.suffix if args.suffix else None`

rishabh6788:
The proposed changes would require you to update `benchmark_test_runner_opensearch` to get the respective command class. 
`benchmark_test_suite = BenchmarkTestSuiteRunners.from_args(self.args, cluster.endpoint_with_port, self.security, cluster.fetch_password())` for `execute-test`. For compare you would just need to pass `self.args`.

src/test_workflow/benchmark_test/benchmark_test_suite_execute.py
```diff
@@ -0,0 +1,91 @@
...
+    def execute(self) -> None:
+        self.form_command(self.args)
+        log_info = f"Executing {self.command.replace(self.endpoint, len(self.endpoint) * '*').replace(self.args.username, len(self.args.username) * '*')}"
+        logging.info(log_info.replace(self.password, len(self.password) * "*") if self.password else log_info)
+        try:
+            subprocess.check_call(f"{self.command}", cwd=os.getcwd(), shell=True)
+            if self.args.cluster_endpoint:
```
rishabh6788, line 29: This is supposed to be `if self.args.cluster_endpoint or self.args.distribution_url:`. 
Can you please make sure the execute logic is not changed for existing code?

src/test_workflow/benchmark_test/benchmark_args.py
```diff
@@ -50,117 +50,173 @@ class BenchmarkArgs:
...
-                            help="Enable opensearch-benchmark to capture node stat metrics such as cpu, mem, jvm etc as well.")
-        parser.add_argument("--capture-segment-replication-stat", dest="telemetry", action="append_const",
-                            const="segment-replication-stats",
-                            help="Enable opensearch-benchmark to segment_replication stat metrics such as replication lag.")
-        parser.add_argument("--telemetry-params", dest="telemetry_params",
-                            help="Allows to set parameters for telemetry devices. Accepts json input.")
-        parser.add_argument("-v", "--verbose", help="Show more verbose output.", action="store_const", default=logging.INFO,
-                            const=logging.DEBUG, dest="logging_level")
+        parser = argparse.ArgumentParser(description="Test an OpenSearch Bundle or compare two tests")
+        subparsers = parser.add_subparsers(dest="command", required=True)
```
rishabh6788, line 63: add `help="Please provide either 'execute-test' to run benchmark or 'compare' command to compare benchmark runs."` to `command`

tests/tests_test_workflow/test_benchmark_workflow/benchmark_test/test_benchmark_test_suite.py
rishabh6788, line 1: Please add a test for compare command as well.

tests/tests_test_workflow/test_benchmark_workflow/benchmark_test/test_benchmark_test_suite_compare.py
rishabh6788, line 1:
```# Copyright OpenSearch Contributors
#
# The OpenSearch Contributors require contributions made to
# this file be licensed under the Apache-2.0 license or a
# compatible open source license.
```
add this to the new files you added

rishabh6788 (approved):

rishabh6788:
LGTM, great job on your first PR @OVI3D0!
@peterzhuamazon Let us know if you have any comments or concerns.

peterzhuamazon:
I have some offline conversation with @rishabh6788, and now understand the requirement of using docker, due to barebone OSB binary is preventing multiple processes run in parallel.

However, I would suggest these things if you must use docker this way:
1. Add a docker check before the run so it can inform user if it does not detect docker
2. Verify it is on either mac or linux, hard fail on any other related systems for now (Not hard blocker)
3. Add corresponding requirements to the readme file, such as docker, cpu/ram, system for this specific implementation since the code is not generalized to run on any system

We need to make sure users are aware of the limited condition of this compare feature. Since this is a rather big change, would appreciate if we can get these checks in the same PR or in a quick followup later.
Thanks.

src/run_benchmark_test.py
```diff
@@ -13,21 +14,44 @@
...
+
+def check_docker() -> None:
+    if shutil.which('docker') is None:
+        raise Exception("Docker is not installed.")
+
+    # Check if Docker daemon is running
+    try:
+        subprocess.run(["docker", "info"], check=True, capture_output=True, text=True)
+    except subprocess.CalledProcessError:
+        raise Exception("Docker is installed but not running")
```
peterzhuamazon, line 28: 👍

peterzhuamazon (approved):

peterzhuamazon:
Thanks @OVI3D0 / @rishabh6788 👍

PR #4887, opened 2024-07-24
Add 1.1 version of test report yml with bundle and rc information

peterzhuamazon:
### Description
Add 1.1 version of test report yml with bundle and rc information
Adding version, platform, architecture, distribution, build id, rc number information.

### Issues Resolved
https://github.com/opensearch-project/opensearch-metrics/issues/51

### DIFF
--- a/src/report_workflow/report_args.py
+++ b/src/report_workflow/report_args.py
@@ -19,6 +19,8 @@ class ReportArgs:
     test_manifest_path: str
     artifact_paths: dict
     test_type: str
+    schema_version: str
+    rc_number: int
     logging_level: int
 
     def __init__(self) -> None:
@@ -33,6 +35,8 @@ class ReportArgs:
         parser.add_argument("--output-path", type=str, help="Specify the path location for the test-report manifest.")
         parser.add_argument("--test-run-id", type=int, help="The unique execution id for the test")
         parser.add_argument("--component", type=str, dest="components", nargs='*', help="Test a specific component or components instead of the entire distribution.")
+        parser.add_argument("--schema-version", type=str, default="1.1", help="The schema version for the test-report manifest")
+        parser.add_argument("--rc-number", type=int, default=-1, help="The release candidate (rc) number of the artifacts, -1 means not a release candidate")
         parser.add_argument(
             "-v", "--verbose", help="Show more verbose output.", action="store_const", default=logging.INFO, const=logging.DEBUG, dest="logging_level"
         )
@@ -45,4 +49,6 @@ class ReportArgs:
         self.base_path = args.base_path
         self.test_type = args.test_type
         self.components = args.components
+        self.schema_version = args.schema_version
+        self.rc_number = args.rc_number
         self.output_path = args.output_path
--- a/src/report_workflow/report_args.py
+++ b/src/report_workflow/report_args.py
@@ -20,7 +20,7 @@ class ReportArgs:
     artifact_paths: dict
     test_type: str
     schema_version: str
-    rc_number: int
+    release_candidate: str
     logging_level: int
 
     def __init__(self) -> None:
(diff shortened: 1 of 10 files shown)


src/report_workflow/test_report_runner.py
```diff
@@ -48,6 +52,14 @@ def __init__(self, args: ReportArgs, test_manifest: TestManifest) -> None:
 
     def update_data(self) -> dict:
         self.test_run_data["name"] = self.product_name
+        if float(self.schema_version) > 1.0:
```
zelinh, line 55: Any situation that we still want to run on 1.0 schema version? 
or could we have a better mechanism to work between previous schema_version vs current? This seems to be hardcoded. I feel like having a boolean flag might be slightly better but still wonder do we really need this.

src/report_workflow/report_args.py
```diff
@@ -33,6 +35,8 @@ def __init__(self) -> None:
         parser.add_argument("--output-path", type=str, help="Specify the path location for the test-report manifest.")
         parser.add_argument("--test-run-id", type=int, help="The unique execution id for the test")
         parser.add_argument("--component", type=str, dest="components", nargs='*', help="Test a specific component or components instead of the entire distribution.")
+        parser.add_argument("--schema-version", type=str, default="1.1", help="The schema version for the test-report manifest.")
+        parser.add_argument("--release-candidate", type=str, default="0", help="The release candidate (rc) information of the artifacts, added in schema-version='1.1'.")
```
zelinh, line 39: Why do we need this? I assume this is the RC number which should be same as the ID we fetched from the bundle manifest right?

src/report_workflow/test_report_runner.py
```diff
@@ -48,6 +52,14 @@ def __init__(self, args: ReportArgs, test_manifest: TestManifest) -> None:
...
+            self.bundle_manifest = BundleManifest.from_urlpath(self.dist_manifest)
+            self.test_run_data["version"] = self.bundle_manifest.build.version
+            self.test_run_data["platform"] = self.bundle_manifest.build.platform
+            self.test_run_data["architecture"] = self.bundle_manifest.build.architecture
+            self.test_run_data["distribution"] = self.bundle_manifest.build.distribution
+            self.test_run_data["id"] = self.bundle_manifest.build.id
+            self.test_run_data["rc"] = self.release_candidate
```
zelinh, line 62: Same question here. The RC seems to be the same as `id` right? I think we don't need the argument for RC.

src/report_workflow/report_args.py
peterzhuamazon: RC number is not build id they are different.
On Jenkins distribution build release manager manually enter the RC number of the release candidate.

src/report_workflow/test_report_runner.py
peterzhuamazon: It is for the purpose of BC.
Note for all of the manifest right now we are following this way.
I do think there should be better way to handle the schema_version > 1.0, but it is not easy at the moment. Happy for any advice right here.

src/report_workflow/test_report_runner.py
peterzhuamazon: Same as above.

src/report_workflow/test_report_runner.py
zelinh: I think the most case we worry about BC because we are importing the manifest from `.yml` file so we have a different manifest class as `test_report_manifest_1_0.py`. But in this report workflow, we are generating the yml file from data map. Still think it's not necessary to introduce another argument in this case. We could for example make `--release-candidate` an optional argument; if it's provided, we are generating `1.1` version report manifest; else keep creating old version manifest if we would still need that (which I think we probably won't). WDYT?

src/report_workflow/report_args.py
zelinh: Oh I see that was recently introduced into the build job. https://github.com/opensearch-project/opensearch-build/pull/4876
Thanks.

prudhvigodithi (approved):

prudhvigodithi:
Thanks for this change @peterzhuamazon, Once @zelinh is ok with this PR we can go ahead and merge it.
Thanks

src/report_workflow/test_report_runner.py
peterzhuamazon: I think we can still keep the schema-version param as we should not define the version based on new entries, but the other way around tho.

By default the schema version will always point to the latest version, which is 1.1 in this case, unless someone want to fall back to 1.0, which they can with the parameter.

peterzhuamazon: Have offline discussion and will remove schema-version params, will keep 1.1 as the only default.

Thanks.

zelinh (approved):
LGTM. Thanks @peterzhuamazon

PR #4905, opened 2024-08-02
Skipping integTest for rpm and deb distribution for CCR

nandnkum:
Skipping integTest for cross-cluster-replication component when running on deb and rpm distribution.

### DIFF
--- a/jenkins/opensearch/integ-test.jenkinsfile
+++ b/jenkins/opensearch/integ-test.jenkinsfile
@@ -172,6 +172,11 @@ pipeline {
                             echo "Skipping tests for ${component_check} as is not present in the provided build manifest."
                             componentList -= component_check
                         }
+                    // Skipping integTest for cross-cluster-replication component when running on deb and rpm distribution
+                        if ((distribution.equals('rpm') || distribution.equals('deb')) && component_check == 'cross-cluster-replication') {
+                            echo "Skipping integTest for ${distribution} distribution for cross-cluster-replication"
+                            componentList -= component_check
+                        }
                     }
                     echo "Testing components: ${componentList}"
                     currentBuild.description = "$TEST_MANIFEST, $version, $architecture, $platform, $buildId, $distribution, $componentList"
--- a/jenkins/opensearch/integ-test.jenkinsfile
+++ b/jenkins/opensearch/integ-test.jenkinsfile
@@ -172,8 +172,11 @@ pipeline {
                             echo "Skipping tests for ${component_check} as is not present in the provided build manifest."
                             componentList -= component_check
                         }
-                    // Skipping integTest for cross-cluster-replication component when running on deb and rpm distribution
-                        if ((distribution.equals('rpm') || distribution.equals('deb')) && component_check == 'cross-cluster-replication') {
+                        // Due to inability to install multiple versions of deb/rpm packages on the same EC2 host.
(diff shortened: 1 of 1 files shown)


jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -172,6 +172,11 @@ pipeline {
                             echo "Skipping tests for ${component_check} as is not present in the provided build manifest."
                             componentList -= component_check
                         }
+                    // Skipping integTest for cross-cluster-replication component when running on deb and rpm distribution
```
peterzhuamazon, line 175: Spacing.

peterzhuamazon: Please also add `TODO` to this comment and link the corresponding issue.
Eventually we would need to fix this as this change is temporary.

Thanks.

jenkins/opensearch/integ-test.jenkinsfile
```diff
@@ -172,6 +172,11 @@ pipeline {
...
+                        if ((distribution.equals('rpm') || distribution.equals('deb')) && component_check == 'cross-cluster-replication') {
```
peterzhuamazon, line 176: Can we keep consistent on using `equals` vs `==`.

peterzhuamazon (changes requested):
Hi @nandnkum ,

Add some comments.

Once you resolve those I will check if we need test cases update. Thanks.

peterzhuamazon (approved):
Thanks @nandnkum .

cc: @prudhvigodithi whether we need to merge this next week after 2.16.0 is out. (convert back to draft just in case)

Or we can test earlier on this. Thanks.

gaiksaya:
Please wait until 2.16.0 is out. If the pipeline breaks due to syntax, release process might be stuck.
Thanks!

Also @nandnkum Is there no plan to support CCR on these distributions long term?

nisgoel-amazon:
>Also @nandnkum Is there no plan to support CCR on these distributions in the long term?

@gaiksaya for now we don't have any plan to fix these tests on deb/rpm. we need support from release infra team to help over here to create multiple cluster on same ec2.

gaiksaya:
> > Also @nandnkum Is there no plan to support CCR on these distributions in the long term?
> 
> @gaiksaya for now we don't have any plan to fix these tests on deb/rpm. we need support from release infra team to help over here to create multiple cluster on same ec2.

I see! Trying to think of a better way to skip for specific use-cases (like this one) instead of adding conditional statements. @peterzhuamazon @prudhvigodithi Do you think we can add something in test-manifest to handle this?
We can create an issue and discuss there too.

peterzhuamazon:
> > > Also @nandnkum Is there no plan to support CCR on these distributions in the long term?
> > 
> > 
> > @gaiksaya for now we don't have any plan to fix these tests on deb/rpm. we need support from release infra team to help over here to create multiple cluster on same ec2.
> 
> I see! Trying to think of a better way to skip for specific use-cases (like this one) instead of adding conditional statements. @peterzhuamazon @prudhvigodithi Do you think we can add something in test-manifest to handle this? We can create an issue and discuss there too.

We can, but that would take significantly time.
For now I am ok to add this condition as we already added a TODO section to fix later.
We can always add improvements later. Thanks.

nisgoel-amazon (approved):

peterzhuamazon:
Able to skip, tho not perfect, we should look into directly making the call in python code later.
https://build.ci.opensearch.org/job/integ-test/8532/console

Thanks.

PR #4927, opened 2024-08-07
Fix for only adding modified files in release-notes-check compile script

bshien:
### Description
Previously the release-notes-check compile script was only adding modified file, but it was causing fails because the file would be added. This change will fix this.

### Issues Resolved


prudhvigodithi (approved):

peterzhuamazon:
Not fully getting what is the actual problem.
Could you please give more details?

Thanks.

gaiksaya (approved):

bshien:
> Not fully getting what is the actual problem. Could you please give more details?
> 
> Thanks.

The gradle workflow release-notes-check is failing with the compile parameter because the file that it creates is untracked, when the existing code only adds tracked, modified files. This change fixes that by adding all unstaged changes.

Issue #4943, opened 2024-08-13
[Bug]: Test Report logger broke for OpenSearch Dashboards.


prudhvigodithi:
### Describe the bug

Coming from https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/1411 and https://github.com/opensearch-project/opensearch-build/pull/4796 after adding the ci-groups changes for OSD tests, the test report logger for OSD tests shows as `not available` this is because of the component name is no longer as `OpenSearch-Dashboards`. 

Example the `OpenSearch-Dashboards-ci-group-6` the log is now generated with the URL link https://ci.opensearch.org/ci/dbc/integ-test-opensearch-dashboards/2.16.0/7858/linux/arm64/tar/test-results/6200/integ-test/OpenSearch-Dashboards-ci-group-6/with-security/OpenSearch-Dashboards-ci-group-6.yml  and not as https://ci.opensearch.org/ci/dbc/integ-test-opensearch-dashboards/2.16.0/7858/linux/arm64/tar/test-results/6200/integ-test/OpenSearch-Dashboards/with-security/OpenSearch-Dashboards.yml.

[...]

PR #4960, opened 2024-08-21
Fix the report workflow to work with ci-group

zelinh:
### Description
Fix the report workflow to work with ci-group

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4943

### DIFF
--- a/src/manifests/test_manifest.py
+++ b/src/manifests/test_manifest.py
@@ -31,6 +31,7 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
                 - with-less-security
               additional-cluster-configs:
                 - key : value
+              ci-group: 6
             bwc-test:
               test-configs:
                 - with-security
--- a/tests/tests_report_workflow/test_test_report_runner_dashboards.py
+++ b/tests/tests_report_workflow/test_test_report_runner_dashboards.py
@@ -59,6 +59,21 @@ class TestTestReportRunnerDashboards(unittest.TestCase):
             test_run_runner.generate_report(test_run_runner_data, path.name)
             self.assertTrue(os.path.isfile(output_path))
 
+    @patch("report_workflow.report_args.ReportArgs")
+    def test_ci_group(self, report_args_mock: MagicMock) -> None:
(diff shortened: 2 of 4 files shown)


src/manifests/test_manifest.py
```diff
@@ -31,6 +31,7 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
                 - with-less-security
               additional-cluster-configs:
                 - key : value
+              ci-group: 6
```
peterzhuamazon, line 34: Why hardcode here?

peterzhuamazon (changes requested):

tests/tests_report_workflow/test_test_report_runner_dashboards.py
```diff
@@ -51,14 +51,29 @@ def test_generate_file(self, report_args_mock: MagicMock) -> None:
...
+        report_args_mock.test_run_id = 123
+        report_args_mock.base_path = self.DATA_DIR
+        report_args_mock.test_type = "integ-test"
+        report_args_mock.release_candidate = "100"
+
+        test_report_runner = TestReportRunner(report_args_mock, self.TEST_MANIFEST_OPENSEARCH_DASHBOARDS)
+        test_report_runner_data = test_report_runner.update_data()
+
+        self.assertEqual(len(test_report_runner_data["components"]), 9)
+        self.assertEqual(test_report_runner_data["components"][0]["name"], "OpenSearch-Dashboards-ci-group-1")
```
gaiksaya, line 75: Can you also add assertions for all ci-groups? Until 9?

src/manifests/test_manifest.py
gaiksaya: Looks like an example of the schema

zelinh: This is the schema for the test manifest. All these are commented out. Basically showing the structure and format for the test manifest.

tests/tests_report_workflow/test_test_report_runner_dashboards.py
zelinh: Already assert the total number of components in the report manifest though. I would add assertions for all of them.

peterzhuamazon (approved):
Thanks.

PR #4985, opened 2024-08-30
Update the clients release docker image to AlmaLinux 8

Xtansia:
### Description
Update the clients release docker image to AlmaLinux 8.

Allows upgrading to .NET SDK 8.0. [opensearch-net](https://github.com/opensearch-project/opensearch-net) now uses .NET SDK 8.0 to build, but current image only has 6.0, leading to a [release failure](https://build.ci.opensearch.org/job/opensearch-net-release/11/console). CentOS 7 is not supported by Microsoft anymore and so .NET 8.0 is unavailable without upgrading.


peterzhuamazon:
Taking a look.

gaiksaya:
How about keep both files under current? As other clients are using the current ones. You can override the default one by new image using https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/standardReleasePipelineWithGenericTrigger.groovy#L15

peterzhuamazon:
> How about keep both files under current? As other clients are using the current ones. You can override the default one by new image using https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/standardReleasePipelineWithGenericTrigger.groovy#L15

I am ok to actually start migrating to alma8 for other clients now.
As CentOS7 is already deprecated per our announcement.
Also it is not easy to maintain it overtime due to not having updated pkgs everywhere.

The CentOS7 image is on dockerhub already so we can move the dockerfile to legacy to prevent new builds and can start the migrations as needed.
WDYT @gaiksaya ?

Thanks.

docker/ci/dockerfiles/current/release.almalinux8.clients.x64.arm64.dockerfile
```diff
@@ -0,0 +1,141 @@
...
+
+# Tools setup
+COPY --chown=0:0 config/yq-setup.sh config/gh-setup.sh /tmp/
+RUN dnf install -y go && /tmp/yq-setup.sh && /tmp/gh-setup.sh
+
+# Install JDK
+RUN curl -SL https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.15%2B10/OpenJDK11U-jdk_x64_linux_hotspot_11.0.15_10.tar.gz -o /opt/jdk11.tar.gz && \
+    mkdir -p /opt/java/openjdk-11 && \
+    tar -xzf /opt/jdk11.tar.gz --strip-components 1 -C /opt/java/openjdk-11/ && \
+    rm /opt/jdk11.tar.gz
```
peterzhuamazon, line 47: Let's remove this setup and use the config/jdk-setup.sh like any others.
Since this only has x64 arch not gonna support arm64 arch.

peterzhuamazon: Or you can have a branch to install corresponding jdk version based on arch.

docker/ci/dockerfiles/current/release.almalinux8.clients.x64.arm64.dockerfile
```diff
@@ -0,0 +1,141 @@
...
+ENV RUBY_HOME=/usr/local/rvm/rubies/ruby-2.6.0/bin
+ENV RVM_HOME=/usr/local/rvm/bin
+ENV GEM_HOME=$CONTAINER_USER_HOME/.gem
+ENV GEM_PATH=$GEM_HOME
+ENV PATH=$RUBY_HOME:$RVM_HOME:$PATH
+
+# Installing Python binary
+RUN update-alternatives --set python /usr/bin/python3.9 && \
+    update-alternatives --set python3 /usr/bin/python3.9 && \
+    pip3 install pip==23.1.2 && pip3 install twine==4.0.2 cmake==3.24.1.1
```
peterzhuamazon, line 97: cmake should be 3.26.4 as of now.

peterzhuamazon, line 97: Seems like pipenv, awscli are missing in this file.

docker/ci/dockerfiles/current/release.almalinux8.clients.x64.arm64.dockerfile
```diff
@@ -0,0 +1,141 @@
...
+CMD ["/bin/bash", "-l"]
+
+# Install ruby / rpm / fpm related dependencies
+RUN . /etc/profile.d/rvm.sh && \
+    rvm install 2.6.0 && \
+    rvm install 3.1.2 && \
+    rvm install jruby-9.3.0.0 && \
+    rvm --default use 2.6.0 && \
+    dnf install -y rpm-build rpm-sign createrepo pinentry && \
+    dnf clean all
```
peterzhuamazon, line 86: This section is moved and now under User0 not User1000 anymore.

peterzhuamazon (changes requested):
Hi @Xtansia ,

The file has been altered a bit compares to the original dockerfile.

Could you please take a look on it? Thanks.

docker/ci/dockerfiles/current/release.almalinux8.clients.x64.arm64.dockerfile
Xtansia: Sure changed to jdk-setup

docker/ci/dockerfiles/current/release.almalinux8.clients.x64.arm64.dockerfile
Xtansia: Oops not sure how I missed those, added in

Xtansia: Have updated it

docker/ci/dockerfiles/current/release.almalinux8.clients.x64.arm64.dockerfile
Xtansia: Was referring to setup in one of the other almalinux8 images, have switched it back to single user install

peterzhuamazon:
Thanks @Xtansia !

peterzhuamazon:
Will build image.

peterzhuamazon:
https://build.ci.opensearch.org/job/docker-build/5786/consolea

peterzhuamazon:
https://build.ci.opensearch.org/job/docker-copy/1441/console

peterzhuamazon:
Image online.

PR #5043, opened 2024-09-19
Add new stage in gradle-check to check and abort stale runs

### LINKED ISSUE #5008: Optimize gradle-check CI Pipeline to Handle Rapid PR Updates
### Description:

We're experiencing inefficiencies in our gradle-check CI pipeline due to multiple workflow triggers on rapid PR updates. This results in redundant GitHub Actions runs and Jenkins job executions that cause unneccessary resource consumption and queue exhaustion.

**Current Behavior:**
1. Each new commit to a PR triggers a GitHub Actions workflow.
2. The workflow submits a Jenkins job via generic webhook trigger.
3. GitHub Actions waits for Jenkins job completion (SUCCESS/FAILURE).
4. Multiple commits in quick succession lead to parallel, redundant workflow/job executions, thus, resulting in queue exhaustion as there can be only 30 runs in parallel.
5. Multiple force-push commits in quick succession also result in git ref failures in gradle-check while checking out the code.

**Problem:**
- Resource wastage due to multiple concurrent runs.
- Only the latest run's results are relevant.
- Potential for confusing or conflicting CI feedback.

**Desired Behavior:**
- Only the most recent commit should trigger a full gradle-check CI pipeline run.
- Earlier runs,  should be cancelled/aborted.

### Proposed Solutions:

1. **GitHub Actions:** Implement workflow concurrency to cancel outdated runs at individual PR level. Github provides a mechanism to control [concurrency](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency) at each workflow level. For e.g. in our case if a gradle-check is triggered due commit push and then a few minutes later a new commit is force-pushed on the same PR, there will be a new instance of gradle-check workflow. 
At this point we are no longer interested in the outcome of previous instance of the running workflow. Github provides out-of-the-box solution to cancel all the previous running instances the same workflow. 
In the example below, the workflow has a concurrency check, where for a particular PR if there is already a github actions workflow running, and new one is submitted, it will auto-cancel the former one. 
```
concurrency:
  # For a given workflow, if we push to the same PR, cancel all previous builds on that PR.
  group: "${{ github.workflow }}-${{ github.event.pull_request.number}}"
  cancel-in-progress: true
```

2. **Jenkins:** Modify pipeline to abort previous builds of the same job based on PR description. Even if we auto-cancel the github actions workflow when a new one is submitted for the same PR, it will not cancel/abort the already submitted jenkins job. To handle this we need to have custom solution apart from above to clean up stale/outdated gradle-check runs. 
I am proposing an extra stage in the `gradle-check` jenkinsfile which will run through the builds for the job, check if it is for the same PR, may be based on build description, and abort all the previously running builds. 

```
pipeline {
    agent any
    
    stages {
        stage('Abort previous builds') {
            steps {
                script {
                    def jobName = env.JOB_NAME
                    def buildNumber = env.BUILD_NUMBER.toInteger()
                    def currentJob = Jenkins.instance.getItemByFullName(jobName)
                    
                    for (build in currentJob.builds) {
                        if (!build.isBuilding()) { continue; }
                        if (buildNumber.equals(build.getNumber().toInteger())) { continue; }
                        
                        build.doStop()
                        echo "Aborted build #${build.getNumber()}"
                    }
                }
            }
        }
        
        stage('Your actual job steps') {
            steps {
                echo 'Your job steps go here'
                // Add your actual job steps
            }
        }
    }
}
```

This can be a generic groovy script which can take job name as an argument to be easily extended to other jobs if needed. A common script to handle many other operational tasks or generic checks which could be useful to multiple jobs. 

### Action Items:

- [ ] Research GitHub Actions concurrency controls.
- [ ] Investigate Jenkins job management via API or pipeline scripts.
- [ ] Prototype solution for cancelling outdated runs.
- [ ] Update documentation for new CI/CD behavior.
- [ ] Plan testing strategy for the optimized pipeline.

### Additional Considerations:

- Impact on PR checks and status reporting.
- Performance overhead of iterating over several builds of the job and aborting them. 
- Potential race conditions in cancellation logic.

We welcome input from the team on these proposed solutions and any additional ideas to streamline our CI/CD process.

rishabh6788:
### Description
This PR adds a new stage in gradle-check to check and abort previous in-progress runs for the same pull-request. 

### Issues Resolved
#5008 

### DIFF
--- a/jenkins/gradle/gradle-check.jenkinsfile
+++ b/jenkins/gradle/gradle-check.jenkinsfile
@@ -7,7 +7,7 @@
  * compatible open source license.
  */
 
-lib = library(identifier: 'jenkins@6.6.0', retriever: modernSCM([
+lib = library(identifier: 'jenkins@6.9.1', retriever: modernSCM([
     $class: 'GitSCMSource',
     remote: 'https://github.com/opensearch-project/opensearch-build-libraries.git',
 ]))
@@ -79,6 +79,13 @@ pipeline {
         BUILD_CAUSE = currentBuild.getBuildCauses()
     }
     stages {
+        stage('Check and abort stale runs') {
+            steps {
+                script {
+                    abortStaleJenkinsJobs(jobName: 'gradle-check', lookupTime: 3)
+                }
+            }
+        }
         stage('Run Gradle Check') {
             steps {
                 script {


rishabh6788:
@peterzhuamazon Please provide your feedback and review.

jenkins/gradle/gradle-check.jenkinsfile
```diff
@@ -79,6 +79,13 @@ pipeline {
         BUILD_CAUSE = currentBuild.getBuildCauses()
     }
     stages {
+        stage('Check and abort stale runs') {
+            steps {
+                script {
+                    abortStaleJenkinsJobs(jobName: 'gradle-check', lookupTime: 3)
```
gaiksaya, line 85: Just thinking out loud, do you think there will be an edge case scenario that the PR author/user wants to continue running gradle check for previous commit?
I don't think it would matter anyway because the GHA check would consider the most recent run right?

rishabh6788: I don't see any reason to keep an old job running if you have pushed a new commit. Only the result on the new run matters for approval.

gaiksaya (approved):
We should definitely move this script to core repo for more control by the core maintainers over gradle check. 😅

peterzhuamazon (approved):
I agree that if a PR author push a new commit, they would most likely want to see the result of the new commit, not old commit, thus we can start with this change, and monitor the outcome of it.

Thanks @rishabh6788

PR #5111, opened 2024-10-14
Record OpenSearch-Dashboards failed test cases in test-report.yml

peterzhuamazon:
### Description
Record OpenSearch-Dashboards failed test cases in test-report.yml

### Issues Resolved
https://github.com/opensearch-project/opensearch-metrics/issues/68


src/report_workflow/test_report_runner.py
```diff
@@ -207,48 +215,63 @@ def get_os_cluster_logs(base_path: str, test_number: str, test_type: str, compon
     return [os_stdout, os_stderr]
 
 
-def get_failed_tests(base_path: str, test_number: str, test_type: str, component_name: str, config: str,
-                     product_name: str) -> typing.List[list]:
-    failed_test: list = []
-    result_path: str = ''
-    result_content: str = ''
+def get_failed_tests(product_name: str, test_result: str, test_result_files: list) -> typing.List[list]:
```
prudhvigodithi, line 218: This method defaults to `.xml` and if OpenSearch will use the `html.parser` right ?

peterzhuamazon:
> Hey @peterzhuamazon thanks, do we need all the `.xml` files to be part of the this change or can we just test on one `.xml`?

We need all of the xml, due to in cypress each plugin would run multiple test js file, and each xml corresponding to one js.

src/report_workflow/test_report_runner.py
peterzhuamazon: See code:
```
        if self.name == "opensearch":
            soup = BeautifulSoup(result_content, "html.parser")
        else:
            soup = BeautifulSoup(result_content, "xml")

```

prudhvigodithi (approved):

prudhvigodithi:
Just a thought, I think for testing peter we can remove the all the `xml` files just keep one and update here https://github.com/opensearch-project/opensearch-build/pull/5111/files#diff-a44612e37c0352c6226189afb70d84c146f4dfe5982cfc6caf445e462a3664a3, then the tests should pass as well right?

peterzhuamazon:
> Just a thought, I think for testing peter we can remove the all the `xml` files just keep one and update here https://github.com/opensearch-project/opensearch-build/pull/5111/files#diff-a44612e37c0352c6226189afb70d84c146f4dfe5982cfc6caf445e462a3664a3, then the tests should pass as well right?

We need to test the for loop with multiple xmls instead of only limit to one.

Issue #5126, opened 2024-10-22
Add new smoke test section into test manifest schema


zelinh:
Subtask from https://github.com/opensearch-project/opensearch-build/issues/4551

A new `smoke-test` section should be added to the test manifest to specify which component were onboarded for smoke tests. 

It could be looked like this
```
components:
  - name: opensearch
    smoke-test:
      test-spec: opensearch.yml
  - name: alerting
    integ-test:
      ...
    bwc-test:
      ...
    smoke-test:
      test-spec: alerting.yml
  - ...
 ```

In the above example, `opensearch & alerting` are onboarded with smoke test. In the `test-spec` section, it would point to a component spec yaml file which specifies the API request path for the smoke tests to run on.

zelinh:
The `smoke_test` is added into our test manifest. Close this issue for now.

PR #5127, opened 2024-10-22
Add smoke tests to the test manifest schema

zelinh:
### Description
Add smoke tests to the test manifest schema

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5126

### DIFF
--- a/src/manifests/test_manifest.py
+++ b/src/manifests/test_manifest.py
@@ -37,7 +37,7 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
                 - with-security
                 - without-security
             smoke-test:
-              test-spec: spec.yaml
+              test-spec: spec.yml
     """
 
     SCHEMA = {
--- a/src/manifests/test_manifest.py
+++ b/src/manifests/test_manifest.py
@@ -36,6 +36,9 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
               test-configs:
                 - with-security
                 - without-security
+            smoke-test:
+              test-spec:
+                - spec.yaml
     """
 
     SCHEMA = {
@@ -86,6 +89,12 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
                         "schema": {
                             "build-dependencies": {"type": "list"},
                             "test-configs": {"type": "list", "allowed": ["with-security", "without-security", "with-less-security"]},
+                        },
+                    },
+                    "smoke-test": {
+                        "type": "dict",
+                        "schema": {
+                            "test-spec": {"type": "list"},
                         },
                     },
                 },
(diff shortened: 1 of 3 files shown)


src/manifests/test_manifest.py
```diff
@@ -36,6 +36,9 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
               test-configs:
                 - with-security
                 - without-security
+            smoke-test:
+              test-spec:
```
gaiksaya, line 40: Do we want to keep this as `test-configs`? Similar to integ-test and bwc test?

zelinh: I'm thinking it's not technically config? Keep it as `test-configs` may cause some confusion to users since we don't modify any of the cluster configuration.

tests/tests_manifests/data/opensearch-2.18.0-test.yml
```diff
@@ -0,0 +1,152 @@
...
+name: OpenSearch
+ci:
+  image:
+    name: opensearchstaging/ci-runner:ci-runner-al2-opensearch-build-v1
+    args: -e JAVA_HOME=/opt/java/openjdk-21
+components:
+  - name: opensearch
+    smoke-test:
+      test-spec:
+        - opensearch.yml
```
peterzhuamazon, line 12: Where is this file located here?

zelinh: I'm thinking to have this file stored in a new spec folder inside the smoke tests workflow directory. Open about any other suggestions.

src/manifests/test_manifest.py
```diff
@@ -88,6 +91,12 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
                             "test-configs": {"type": "list", "allowed": ["with-security", "without-security", "with-less-security"]},
                         },
                     },
+                    "smoke-test": {
+                        "type": "dict",
+                        "schema": {
+                            "test-spec": {"type": "list"},
```
peterzhuamazon, line 97: Do we need to use list here? More than 1 spec file for each test?

gaiksaya (approved):
We can start with one spec per component and then iterate if we need more. Would recommend to add multiple specs under same file rather than having multiple spec files.

zelinh:
> We can start with one spec per component and then iterate if we need more. Would recommend to add multiple specs under same file rather than having multiple spec files for one component.

Sure let me update it to a single string of spec for now. We can enhance later if we really need more. Thanks.

src/manifests/test_manifest.py
```diff
@@ -37,8 +37,7 @@ class TestManifest(ComponentManifest['TestManifest', 'TestComponents']):
                 - with-security
                 - without-security
             smoke-test:
-              test-spec:
-                - spec.yaml
+              test-spec: spec.yaml
```
peterzhuamazon, line 40: Can we keep it consistent here?
`.yaml` or `.yml`?

peterzhuamazon (approved):

src/manifests/test_manifest.py
zelinh: Sure updated to `.yml`.

zelinh:
@gaiksaya @peterzhuamazon Any final thoughts or concerns before we merge this PR?

peterzhuamazon:
> @gaiksaya @peterzhuamazon Any final thoughts or concerns before we merge this PR?

LGTM

PR #5175, opened 2024-11-08
Add release note entry for new phone number analyzers

andrross:
This is a noteworthy new feature that was missed in the release notes but is worthy of a mention, in my opinion.

### DIFF
--- a/release-notes/opensearch-release-notes-2.18.0.md
+++ b/release-notes/opensearch-release-notes-2.18.0.md
@@ -14,6 +14,7 @@ OpenSearch 2.18 includes a number of updates that help you build machine learnin
 * Grouping top N queries by similarity now supports field names and data types, allowing you to group queries at a finer level of detail so that you can better identify and analyze resource-intensive query patterns across query types.
 * Workspaces provide a new multi-tenant environment to increase team collaboration. Granular access controls let you manage or remove collaborators and control their access at the workspace level with role-based permissions.
 * A redesigned homepage and navigation structure provide a centralized toolkit for accessing and creating Workspaces, with a more cleanly organized navigation bar that adapts to the selected workspace and supports tailored workflows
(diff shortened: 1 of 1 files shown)


release-notes/opensearch-release-notes-2.18.0.md
```diff
@@ -14,6 +14,7 @@ OpenSearch 2.18 includes a number of updates that help you build machine learnin
 * Grouping top N queries by similarity now supports field names and data types, allowing you to group queries at a finer level of detail so that you can better identify and analyze resource-intensive query patterns across query types.
 * Workspaces provide a new multi-tenant environment to increase team collaboration. Granular access controls let you manage or remove collaborators and control their access at the workspace level with role-based permissions.
 * A redesigned homepage and navigation structure provide a centralized toolkit for accessing and creating Workspaces, with a more cleanly organized navigation bar that adapts to the selected workspace and supports tailored workflows
+* Two new analyzers, `phone` and `phone-search`, have been added to simplify the complex task of parsing global phone numbers.
```
rursprung, line 17: it might make sense to explicitly call out the fact that these are part of an optional plugin?
```suggestion
* Two new analyzers, `phone` and `phone-search`, have been added to simplify the complex task of parsing global phone numbers. This requires installing the optional `analysis-phonenumber` plugin.
```

also, would it make sense to [link to the documentation](https://opensearch.org/docs/latest/analyzers/supported-analyzers/phone-analyzers/)? (i see that other entries are also not doing that, but i think this would be helpful in general so that people interested in a topic can find more information on it faster?)

andrross: Hmm, I was under the impression all the plugins were installed by default in the distribution, but it appears not to be the case for the analysis plugins. I'll make the update.

prudhvigodithi: Yes that is correct, in the release bundle plugins part of the manifest are installed by default, the native plugins (inside the core repo `plugins/` folder) has to be installed explicitly as required.

gaiksaya (approved):
@andrross Are we good to merge this?

PR #5192, opened 2024-11-20
Refactor distribution build job

### LINKED ISSUE #5173: Build distribution docker images immediately after tarballs

Check the most recent run for 3.0.0 here: https://build.ci.opensearch.org/blue/organizations/jenkins/distribution-build-opensearch/detail/distribution-build-opensearch/10494/pipeline

The docker images are build at the end of the pipeline when all the dist/arch/platform builds are complete. Often times, we have seen windows failing mainly due to shell glitch or other operating system issues. Failure in any step before docker won't trigger a docker build as it is sequential at this point.

In order to build the docker images for OpenSearch and OpenSearch-Dashboards, we only need tarball linux distribution for x64 and arm64 to be ready. Hence the docker stage can be moved from sequential run at the end of all builds/assemble to parallel stage once tarball x64 and arm64 builds are complete.

Keep what we have. It is not a blocker but definitely lacks the consistent builds due to other distribution failures.

### Additional context

_No response_

rishabh6788:
### Description
This PR moves the `docker build` stage into parallel processing stages. 
The stage will wait on x64 and arm64 tar builds to complete, and once the artifact urls for both are available it will process the docker build job. 

### Issues Resolved
#5173 

### DIFF
--- a/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
+++ b/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
@@ -872,44 +872,55 @@ pipeline {
                         }
                     }
                 }
-            }
-        }
-        stage('docker build') {
-            when {
-                beforeAgent true
-                allOf {
-                    expression {
-                        params.BUILD_DOCKER != 'do_not_build_docker'
+                stage('docker build') {
+                    when {
+                        beforeAgent true
+                        allOf {
+                            expression {
+                                params.BUILD_DOCKER != 'do_not_build_docker'
+                            }
+                            expression {
+                                params.BUILD_PLATFORM.contains('linux')
+                            }
+                            expression {
+                                params.BUILD_DISTRIBUTION.contains('tar')
+                            }
+                        }
                     }
-                    expression {
-                        params.BUILD_PLATFORM.contains('linux')
+                    options {
+                        timeout(time: 90, unit: 'MINUTES')
                     }
-                    expression {
-                        params.BUILD_DISTRIBUTION.contains('tar')
+                    agent {
+                        docker {
+                            label AGENT_LINUX_X64
+                            image dockerAgent.image
+                            args dockerAgent.args
+                            registryUrl 'https://public.ecr.aws/'
+                            alwaysPull true
+                        }
                     }
-                }
-            }
-            agent {
-                docker {
-                    label AGENT_LINUX_X64
-                    image dockerAgent.image
-                    args dockerAgent.args
-                    registryUrl 'https://public.ecr.aws/'
-                    alwaysPull true
-                }
-            }
-            steps {
-                script {
-                    echo "env.ARTIFACT_URL_LINUX_X64_TAR: ${env.ARTIFACT_URL_LINUX_X64_TAR}"
-                    echo "env.ARTIFACT_URL_LINUX_ARM64_TAR: ${env.ARTIFACT_URL_LINUX_ARM64_TAR}"
+                    steps {
+                        script {
+                            while(true) {
+                                if (env.ARTIFACT_URL_LINUX_X64_TAR != null && env.ARTIFACT_URL_LINUX_ARM64_TAR != null) {
+                                    echo "env.ARTIFACT_URL_LINUX_X64_TAR: ${env.ARTIFACT_URL_LINUX_X64_TAR}"
+                                    echo "env.ARTIFACT_URL_LINUX_ARM64_TAR: ${env.ARTIFACT_URL_LINUX_ARM64_TAR}"
 
-                    buildDockerImage(
-                        inputManifest: "manifests/${INPUT_MANIFEST}",
-                        buildNumber: "${BUILD_NUMBER}",
-                        buildOption: "${BUILD_DOCKER}",
-                        artifactUrlX64: env.ARTIFACT_URL_LINUX_X64_TAR,
-                        artifactUrlArm64: env.ARTIFACT_URL_LINUX_ARM64_TAR
-                    )
+                                    buildDockerImage(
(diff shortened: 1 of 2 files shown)


jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -872,44 +872,55 @@ pipeline {
...
+                                        inputManifest: "manifests/${INPUT_MANIFEST}",
+                                        buildNumber: "${BUILD_NUMBER}",
+                                        buildOption: "${BUILD_DOCKER}",
+                                        artifactUrlX64: env.ARTIFACT_URL_LINUX_X64_TAR,
+                                        artifactUrlArm64: env.ARTIFACT_URL_LINUX_ARM64_TAR
+                                    )
+                                    break
+                                } else {
+                                    echo "Waiting for x64 and arm64 tar builds to complete, sleeping 120 seconds..."
+                                    sleep(time: 120, unit: 'SECONDS')
```
gaiksaya, line 919: I had tried something like this https://github.com/gaiksaya/opensearch-build/blob/6d4745692422ba7c7c89a55db110a6789ab97e9c/jenkins/opensearch/test-dist.jenkinsfile#L65-L70

Can you check if this is a better option than that? It will automatically start the build as soon those variables are set. If not the timeout will take care of timing out.

rishabh6788: Both are the same, in this we don't need to set additional env variables as we can leverage the ones already available.

gaiksaya: Wont that start the build as soon as the variables are set instead of waiting for next 120 seconds? Also wondering, if both or one of the stages failed, we should skip building docker altogether. However,in this case, it will wait for 90 minutes before timing out. Looks like we cannot check stage status (only job status). Can we add some check for that scenario?

rishabh6788: Once the variables are set that means the artifacts are available in S3 and can be used immediately, I don't see an issue here. Also as per logic it will only trigger the docker build once both, x64 and arm64 artifacts are available else timeout eventually. 
Stage status check is not trivial and there is no straight-forward way to do the same, will need some hack.

rishabh6788: We shouldn't be building docker just for x64 or arm64, either build for both else skip.

gaiksaya: Right! My point being, the average build time for tarball x64 and arm64 is 20-30min. So if they fail, we would have to wait more 60min (since docker timeout is 90min) for docker stage to timeout eventually.

rishabh6788: Sure, may be you can PR it since you have the changes done in your local. I will cancel or you can edit this as well. 
I am okay with both.

rishabh6788: On a second look, `env.TAR_X64_SUCCESS = 'true'` would only be set if the groovy lib that executes the actual workflow returns successfully. In case of failure it will exit out of the job without setting this variable. 
So it is same as what I'm doing, in case if success/failure both the solutions work same. 
Please correct me if I'm wrong.

peterzhuamazon (commented):
So the idea is to have everything in one parallel stage, then keep waiting until both link available?

rishabh6788:
> So the idea is to have everything in one parallel stage, then keep waiting until both link available?

Yes!

rishabh6788:
If no more concerns can we approve and merge this? @gaiksaya @peterzhuamazon

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
gaiksaya: How about setting `env.TAR_X64_SUCCESS = 'false'` in post failure stage? Or any value you see fit?

```
post {
  success {
      script {
          env.TAR_ARM64_SUCCESS = 'succeeded'
      }
  }
  failure {
    script {
      env.TAR_ARM64_SUCCESS = 'failed'
    }
  }
}
```

Check for both values in the docker build succeeded and failed and act accordingly. The main motive of this change is to lower the build time and immediately build docker when tarball is ready. Looks like in case of failure we would be waiting 90min for the job to complete.

gaiksaya: Synced up offline, looks like this would be too much of hack. We can see how the change in this PR goes and iterate on it later. Thanks!

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -872,44 +872,55 @@ pipeline {
...
-                    alwaysPull true
-                }
-            }
-            steps {
-                script {
-                    echo "env.ARTIFACT_URL_LINUX_X64_TAR: ${env.ARTIFACT_URL_LINUX_X64_TAR}"
-                    echo "env.ARTIFACT_URL_LINUX_ARM64_TAR: ${env.ARTIFACT_URL_LINUX_ARM64_TAR}"
+                    steps {
+                        script {
+                            while(true) {
```
gaiksaya, line 904: Would recommend to use inbuilt feature of [waitUntil](https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#waituntil-wait-for-condition). We can take up it up as an enhancement if you like. Would avoid us adding code for checking and waiting for 120seconds

gaiksaya (approved):

Issue #5224, opened 2025-01-06
Integrate the smoke test workflow with CI/CD pipeline


zelinh:
Subtask of https://github.com/opensearch-project/opensearch-build/issues/5223

The new Jenkins pipeline would be created to integrate smoke tests with existing build workflow for automation to catch major 
issue in earlier stage.

This new CI integration will include a new groovy library script, a new jenkinsfile and required test cases. 

Task:
- [x] Create the groovy library in opensearch-build-library repo to run the script for smoke tests.
- [x] Introduce a new Jenkins job `Smoke Test` with a distinct jenkinsfile.
- [x] Enable auto integration with the release candidates.

zelinh:
Groovy library for smoke test script is added here. https://github.com/opensearch-project/opensearch-build-libraries/pull/569

zelinh:
The PR is merged. We now have the new smoke test job on Jenkins. https://build.ci.opensearch.org/job/smoke-test/

Issue #4233, opened 2023-11-22
[Enhancement] CI docker images for build and test should be maintained per version


gaiksaya:
### Describe the bug

The CI images used to build and test OpenSearch and OpenSearch-Dashboards are all over the place. 

Example:
1. Build-images: https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L24-L26
2. Test-images: https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/integ-test.jenkinsfile#L16-L21

These images are susceptible to changes depending on the the version and new changes coming in. Even though these docker images are source coded and versioned, there is no way to tell today which of those versions were used in a specific release unless someone goes and check the tag of that specific release.

### To reproduce

Just look at the configurations in various files. 

### Expected behavior

These images needs to be a part of the manifest https://github.com/opensearch-project/opensearch-build/blob/main/manifests/2.11.1/opensearch-2.11.1-test.yml#L6-L7 or maintained separately per version. 

The manifest schema can be modified for each distribution, something like:
```yml
ci:
  image:
    linux:
      tar:
        name: opensearchstaging/ci-runner:ci-runner-centos7-opensearch-build-v3
        args: -e JAVA_HOME=/opt/java/openjdk-17
      rpm:
        name: opensearchstaging/ci-runner:ci-runner-rockylinux8-opensearch-build-v4
        args: -e JAVA_HOME=/usr/bin/java/openjdk-17
      deb:
        name: opensearchstaging/ci-runner:ci-runner-ubuntu2004-opensearch-build-v2
        args: -e JAVA_HOME=/opt/java/openjdk-17
    windows:
      zip: 
        name: <name_nere>
        args: <args_here>
      msi: 
        name: <name_nere>
        args: <args_here>    
    macos:
      pkg:
        name: <name_nere>
        args: <args_here>   

```

Another solution is to maintain another `yml` file with all things related to CI images and args and pass that in current manifest as:

```yml
---
schema-version: '1.0'
build:
  name: OpenSearch
  version: 2.11.1
ci:
  file: ci_images.yml
```

### Acceptance Criteria

- Use the CI images for build/test from the manifest instead of hardcoding it in the jenkinsfiles. With current setup the same jenkinsfiles are used for building multiple release versions.
- Having all the CI Images as part of the manifest.

gaiksaya:
[...]

PR #5225, opened 2025-01-06
Add a new Jenkins job for the smoke test

zelinh:
### Description
Add a new jenkinsfile for the smoke test Jenkins job intended for CI with distribution build job.

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/5224


jenkins/opensearch/smoke-test.jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+                    def buildManifestObj = lib.jenkins.BuildManifest.new(readYaml(file: BUILD_MANIFEST))
+                    def testManifestObj = lib.jenkins.TestManifest.new(readYaml(file: "manifests/${TEST_MANIFEST}"))
+                    currentBuild.description = "$TEST_MANIFEST, $version, $architecture, $platform, $buildId, $distribution"
+
+                    // Stash the current working directory files, aka opensearch-build repo
+                    // Unstash later in each triggered stage to run integTest
+                    stash includes: '**', name: "smoketest-opensearch-$BUILD_NUMBER"
+
+                    timeout(time: 1, unit: 'HOURS') {
+                        node(AGENT_LABEL) {
```
peterzhuamazon, line 113: Is this similar to integTest where each plugin will have its own container?

jenkins/opensearch/smoke-test.jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+                        url: BUILD_MANIFEST_URL,
+                        path: BUILD_MANIFEST
+                    )
+
+                    def buildManifestObj = lib.jenkins.BuildManifest.new(readYaml(file: BUILD_MANIFEST))
+                    def testManifestObj = lib.jenkins.TestManifest.new(readYaml(file: "manifests/${TEST_MANIFEST}"))
+                    currentBuild.description = "$TEST_MANIFEST, $version, $architecture, $platform, $buildId, $distribution"
+
+                    // Stash the current working directory files, aka opensearch-build repo
+                    // Unstash later in each triggered stage to run integTest
```
peterzhuamazon, line 109: `integTest`

jenkins/opensearch/smoke-test.jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+                    stash includes: '**', name: "smoketest-opensearch-$BUILD_NUMBER"
```
peterzhuamazon, line 110: Do you just need the build repo? No artifacts?

Then calling scm would be easier than stash/unstash?

jenkins/opensearch/smoke-test.jenkinsfile
zelinh: Not exactly. All testing plugins would be tested with only one deployed container.

jenkins/opensearch/smoke-test.jenkinsfile
zelinh: We will need the build manifests though. It was downloaded through `downloadBuildManifest`.

jenkins/opensearch/smoke-test.jenkinsfile
zelinh: Updated. Thanks.

jenkins/opensearch/smoke-test.jenkinsfile
```diff
@@ -0,0 +1,154 @@
...
+    $class: 'GitSCMSource',
+    remote: 'https://github.com/opensearch-project/opensearch-build-libraries.git',
+]))
+
+def docker_images = [
+    'tar': 'opensearchstaging/ci-runner:ci-runner-al2-opensearch-build-v1',
+    'rpm': 'opensearchstaging/ci-runner:ci-runner-almalinux8-systemd-base-integtest-v1',
+    'deb': 'opensearchstaging/ci-runner:ci-runner-ubuntu2004-systemd-base-integtest-v3',
+    'zip': 'opensearchstaging/ci-runner:ci-runner-windows2019-opensearch-build-v1',
+]
```
gaiksaya, line 20: Looks like we need to add these docker images sooner as workflows are increasing. https://github.com/opensearch-project/opensearch-build/issues/4233

jenkins/opensearch/smoke-test.jenkinsfile
gaiksaya: Agree with @peterzhuamazon. Instead of stashing the entire repo, better stash just the build-manifest then and checkout scm.

jenkins/opensearch/smoke-test.jenkinsfile
gaiksaya: You mean all plugins would run sequentially one after the other?

zelinh: So far yes. All tests should be lightweight.

jenkins/opensearch/smoke-test.jenkinsfile
zelinh: Sure let me take a look here.

zelinh: Updated here with `checkout scm` so we don't need stash anymore. Thanks.

jenkins/opensearch/smoke-test.jenkinsfile
```diff
@@ -0,0 +1,153 @@
...
+        stage('smoke-test') {
+            // Need to run this directly on agent node here in order to trigger stages with docker container and avoid docker within docker situation
+            // Can only be run in runner that is at least 50GB per container
+            agent { label AGENT_LABEL }
+            steps {
+                script {
+                    currentBuild.description = "$TEST_MANIFEST, $version, $architecture, $platform, $buildId, $distribution"
+
+                    timeout(time: 1, unit: 'HOURS') {
+                        node(AGENT_LABEL) {
```
peterzhuamazon, line 102: Just general question, why do we use `node` here if you only call the executor once, maybe just use the normal docker block? A bit confused.

jenkins/opensearch/smoke-test.jenkinsfile
```diff
@@ -0,0 +1,153 @@
...
+                            docker.withRegistry('https://public.ecr.aws/') {
+                                docker.image(docker_images["$distribution"]).inside(docker_args["$distribution"]) {
+                                    try {
+                                        stage("Smoke_tests") {
+                                            checkout scm
+                                            sleep 20
```
peterzhuamazon, line 108: Why do we need sleep here?

jenkins/opensearch/smoke-test.jenkinsfile
zelinh: Good point. I was thinking maybe in the future we can enable separate images for multiple executions. For now, I'm ok we remove it. I will update it in the latest commit.

jenkins/opensearch/smoke-test.jenkinsfile
zelinh: I'm thinking it may need time for the job to checkout source code

jenkins/opensearch/smoke-test.jenkinsfile
```diff
@@ -0,0 +1,151 @@
...
+            // Need to run this directly on agent node here in order to trigger stages with docker container and avoid docker within docker situation
+            // Can only be run in runner that is at least 50GB per container
+            agent { label AGENT_LABEL }
+            steps {
+                script {
+                    currentBuild.description = "$TEST_MANIFEST, $version, $architecture, $platform, $buildId, $distribution"
+
+                    timeout(time: 1, unit: 'HOURS') {
+                        docker.withRegistry('https://public.ecr.aws/') {
+                            docker.image(docker_images["$distribution"]).inside(docker_args["$distribution"]) {
```
peterzhuamazon, line 103: Maybe we should just switch to the pipeline standard:
https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L157C21-L164C22

Since you just need one runner and you just need 2+ minutes to run all the plugins.

Also suggest to set the timeout at the job level and set it to 30min instead of 1 hour. Thanks.

zelinh: Updated in the latest commit. Thanks!

peterzhuamazon (approved):

PR #5233, opened 2025-01-13
Query insights dashboard 2.19

derek-ho:
### Description
Adds query insights dashboards plugin to 2.19 release, adds the backend and frontend plugin to 3.0.0 manifest. 

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5234, opened 2025-01-13
Allow all manifest PRs to run manifest-check workflow

gaiksaya:
### Description
This PR removes the conditional check to see if the label is present in the given PR. Adding a label was manual overhead and is often missed. Also if the event is PR synchronize `github.event.label.name` condition is never met. Example: #5233 

### Issues Resolved

### DIFF
--- a/.github/workflows/manifests.yml
+++ b/.github/workflows/manifests.yml
@@ -3,7 +3,7 @@ name: manifests-ci-check
 
 on:
   pull_request:
-    types: [labeled, synchronize]
+    types: [opened, synchronize]
     paths:
       - 'manifests/**/*.yml'
       - '!manifests/templates/**/'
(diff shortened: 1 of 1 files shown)


.github/workflows/manifests.yml
```diff
@@ -3,15 +3,15 @@ name: manifests-ci-check
 
 on:
   pull_request:
-    types: [labeled, synchronize]
+    types: [opened, synchronize]
     paths:
       - 'manifests/**/*.yml'
       - '!manifests/templates/**/'
       - 'legacy-manifests/**/*.yml'
```
zelinh, line 10: Maybe we can remove this line? Since it's in legacy-manifest folder, we shouldn't modify any of the existing manifests. We will be only move manifests to the legacy folder right?

gaiksaya: Right but incase we move files from manifest to legacy, the checks will also test for validations. Example, all tags exists and manifests are valid.

zelinh (approved):

Issue #5223, opened 2025-01-06
[Meta] Enable smoke tests for OpenSearch distribution + plugins


zelinh:
We proposed smoke test workflow provides a foundational framework to run quick checks on the OpenSearch bundle. These tests run separately from the more complex integration tests but follow a similar deployment process.

Introduce a smoke test framework to the CI/CD process. The workflow will pass parameters like the distribution path, architecture, and components.
![st2](https://github.com/user-attachments/assets/86f4ae21-6275-42bd-9e0c-792331ab512a)

### Tasks:
- https://github.com/opensearch-project/opensearch-build/issues/5126
A new "smoke-test" section will be added to the manifest file to specify which components are onboarded for smoke testing.

- https://github.com/opensearch-project/opensearch-build/issues/5164
Onboard component test spec YAML and spec framework. Each plugin will have a YAML file defining the smoke tests and associated API requests based on the OpenSearch API specification. The test specs will account for version differences in APIs.
The new introduced smoke test runner automates OpenSearch cluster deployment, loads test manifests, sends API requests, and validates responses against the OpenSearch API specification with third-party [OpenAPI-core](https://openapi-core.readthedocs.io/en/stable/index.html) python client. It ensures all key components function immediately after a build.

- Integrate the smoke test workflow with CI/CD pipeline
The new Jenkins pipeline would be created to integrate smoke tests with existing build workflow for automation to catch major issue in earlier stage.

zelinh:
We had an analysis on [Consideration for Onboarding Smoke Tests to OpenSearch Dashboards](https://github.com/opensearch-project/opensearch-build/issues/5268).

zelinh:
To onboard more OpenSearch plugins, we created a sub issue to keep track of the status. 
[Onboard Core and plugins for the Smoke test](https://github.com/opensearch-project/opensearch-build/issues/5247)

prudhvigodithi:
@zelinh can we close this issue?

PR #5244, opened 2025-01-16
Update README for smoke tests with usages and examples

zelinh:
### Description
Update README for smoke tests with usages and examples

### Issues Resolved
Part of #5223 


src/test_workflow/README.md
```diff
@@ -290,6 +315,14 @@ After the performance test completes, it will report back the test results as we
 
 The development is tracked by meta issue [#126](https://github.com/opensearch-project/opensearch-build/issues/126)
 
+#### smokeTest job
+
+It is a Jenkins job designed to run smoke tests on bundle artifacts. It retrieves the bundle artifacts using the provided manifest URL parameter and spins up a cluster with all plugins installed. The cluster is deployed using the default configuration without any customizations. Smoke tests for all components are executed sequentially, and the results are returned at the end.
```
gaiksaya, line 320:
```suggestion
It is a Jenkins job designed to run smoke tests on distribution artifacts. It retrieves the distribution artifacts using the provided manifest URL parameter and spins up a cluster with all plugins installed. The cluster is deployed using the default configuration without any customization. Smoke tests for all components are executed sequentially, and the results are returned at the end.  
```

gaiksaya: Is the cluster cleaned up after every component?

zelinh: Updated based on suggestions. Thanks.


> Is the cluster cleaned up after every component?

No. We will only clean up after all components test completed.

gaiksaya: That might cause conflicts right?

zelinh: We have test requirements that if it causes conflicts, we shouldn't add it into the framework.

gaiksaya: I dont think that is feasible in the long term. Adding @prudhvigodithi @rishabh6788 @peterzhuamazon to get some more inputs.

zelinh: Our current design involves spinning up a single cluster and accommodating non-conflicting smoke test cases for multiple plugins. Perhaps we could create a separate issue to discuss this further?

gaiksaya (approved):

src/test_workflow/README.md
zelinh: Cut an enhancement issue here. We may revisit it to analyze its benefits further.
https://github.com/opensearch-project/opensearch-build/issues/5274

PR #5251, opened 2025-01-23
Adding 3.0.0-alpha1 in manifests and Jenkins runs

peterzhuamazon:
### Description
Adding 3.0.0-alpha1 in manifests and Jenkins runs

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3747


manifests/3.0.0-alpha1/opensearch-3.0.0-alpha1.yml
```diff
@@ -3,6 +3,7 @@ schema-version: '1.1'
 build:
   name: OpenSearch
   version: 3.0.0
+  qualifier: alpha1
```
gaiksaya, line 6: Should it be rc1? Looking at past releases https://opensearch.org/releases.html#release-history.
cc: @reta @dblock

peterzhuamazon: Hi @gaiksaya ,

No, we have proposed to have alpha1 and beta1 just like previously in 1.x/2.x.

The differences is, 1.x and 2.x have alpha1/beta1/rc1 before we come up with the RC concept for every release.

Therefore to avoid confusion we will have alpha1/beta1/GA(no qualifier) for 3.x.

https://github.com/opensearch-project/.github/issues/252

Thanks.

gaiksaya: Is there a possibility of `alpha2`? If not why not just move ahead with alpha/beta/GA?

peterzhuamazon: Because in core there is a rule in code, that qualifier has to be `char+number`, therefore alpha will not even allow core to be compiled and error our.

peterzhuamazon: This has already been communicated with Andrew and Saurabh.

peterzhuamazon:
I will fix the test soon.

peterzhuamazon:
Instead of fixing the test I will restore the 3.0.0 folder we are fine to add a 3.0.0-alpha1 instead of renaming.

jenkins/check-for-build.jenkinsfile
```diff
@@ -23,12 +23,11 @@ pipeline {
...
-            H 1 * * * %INPUT_MANIFEST=2.18.1/opensearch-2.18.1.yml;TARGET_JOB_NAME=distribution-build-opensearch;BUILD_PLATFORM=linux windows;BUILD_DISTRIBUTION=tar rpm deb zip;TEST_MANIFEST=2.18.1/opensearch-2.18.1-test.yml;TEST_PLATFORM=linux;TEST_DISTRIBUTION=tar
             H 1 * * * %INPUT_MANIFEST=2.19.0/opensearch-2.19.0.yml;TARGET_JOB_NAME=distribution-build-opensearch;BUILD_PLATFORM=linux windows;BUILD_DISTRIBUTION=tar rpm deb zip;TEST_MANIFEST=2.19.0/opensearch-2.19.0-test.yml;TEST_PLATFORM=linux;TEST_DISTRIBUTION=tar
-            H 4 * * * %INPUT_MANIFEST=3.0.0/opensearch-3.0.0.yml;TARGET_JOB_NAME=distribution-build-opensearch;BUILD_PLATFORM=linux windows;BUILD_DISTRIBUTION=tar rpm deb zip
-            H 4 * * * %INPUT_MANIFEST=3.0.0/opensearch-dashboards-3.0.0.yml;TARGET_JOB_NAME=distribution-build-opensearch-dashboards;BUILD_PLATFORM=linux windows;BUILD_DISTRIBUTION=tar rpm deb zip
+            H 1 * * * %INPUT_MANIFEST=2.18.1/opensearch-2.18.1.yml;TARGET_JOB_NAME=distribution-build-opensearch;BUILD_PLATFORM=linux windows;BUILD_DISTRIBUTION=tar rpm deb zip
```
zelinh, line 28: Can we remove this cron for 2.18.1? I don't think we have plan for this patch version right?

peterzhuamazon: Until 2.19.0 is out we will keep it.

zelinh:
Could you look into the CI failure? Seems like manifest ci check is failing with qualifier.

peterzhuamazon:
> Could you look into the CI failure? Seems like manifest ci check is failing with qualifier.

···

2025-01-23 00:57:45 INFO     Executing "./gradlew publishToMavenLocal -Dopensearch.version=3.0.0-alpha1-SNAPSHOT -Dbuild.snapshot=true -Dbuild.version_qualifier=alpha1" in /tmp/tmptiar_71o/common-utils
Downloading https://services.gradle.org/distributions/gradle-8.10.2-bin.zip
.............10%.............20%.............30%.............40%.............50%.............60%.............70%.............80%.............90%.............100%

Welcome to Gradle 8.10.2!

Here are the highlights of this release:
 - Support for Java 23
 - Faster configuration cache
 - Better configuration cache reports

For more details see https://docs.gradle.org/8.10.2/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)

FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* Where:
Build file '/tmp/tmptiar_71o/common-utils/build.gradle' line: 156

* What went wrong:
A problem occurred evaluating root project 'common-utils'.
> Could not find method shadowJar() for arguments [build_auab109pypb5ye49zkgd5auvx$_run_closure11@69772b32] on root project 'common-utils' of type org.gradle.api.Project.
···

The issue is not on core as core just merged the lucene 10 changes in https://github.com/opensearch-project/OpenSearch/pull/16366, which passed mavenlocal. The issue is on any plugin following as they havent start support the breaking changes in core yet on main.

This means we would need to merge this PR and ignore the manifest check failures on 3.0.0-alpha1, so that we can start building min and bundle for plugin teams to pass their checks.

Since we have continue-on-error already I am ok for us to merge this without removing the plugins. Thanks.

zelinh (approved):
Manifest CI fails because of the recent breaking change. I don't think this should be a blocker for OS core. 
Start to build on min should be a good step so far. I'm ok with this PR. Thanks.

peterzhuamazon:
Will merge tomorrow and build 3.0.0-alpha1. 

Thanks.

peterzhuamazon:
Re-generate 3.0.0-alpha1 based on 2.19.0 to have all the exact same components and settings.

Add a warning file in 3.0.0 manifest folder as we should not touch it until GA.

Thanks.

Issue #5171, opened 2024-11-07
Make the OpenSearch Release Process Accessible and Executable by External Contributors


gaiksaya:
## Overview

As of 2.18.0 release of OpenSearch and OpenSearch-Dashboards, we have most processes in the release automated individually. This includes almost all major release steps such as building, assembling, testing, signing and promoting artifacts. We also have one step release for publishing the artifacts to all the platforms with one click. 
However, what we currently lack is the ability to link these automations end-to-end, streamline communication between stakeholders throughout the process and the ability of an external non-amazonian community member to be a release manager. 

This issue discusses the current process, the gaps as well as an approach to make the release process more efficient and hands-free. It also argues how 1-click release process is not a feasible option for OpenSearch distribution releases.

## What is a 1-click release

1-click release terminology came from a universal release process that was introduced for standalone components in the OpenSearch-project. Check the [Github issue](https://github.com/opensearch-project/opensearch-build/issues/1234) and the [onboarding guide](https://github.com/opensearch-project/opensearch-build/blob/main/ONBOARDING.md#onboarding-to-universal--1-click-release-process) for more details.
TL;DR: When a component is ready to be released, the maintainer of the component repository initiates a release by pushing a tag to the repository. This triggers a 2 person review of the release using a GitHub issue. Once approved, a draft release is created on the GitHub with the release artifacts attached to it. The draft release triggers the component jenkins workflow that helps to sign and publish the release to the right platform.

## Analysis

OpenSearch and OpenSearch-Dashboards consist of multiple components which includes core + plugins. As of 2.18.0, we have 24 backend plugins and 15 front-end plugins that are bundled together to form various distributions. 
Comparing to the existing 1-click release process where the process only involves publishing to the right platform after the product has been tested and validated, the release process for OpenSearch is fairly complicated. 
The overall release process involves a series of steps, including building, assembling, thorough testing, and meeting various criteria across multiple components. A fully automated, one-click release process may be challenging to achieve, but a more realistic and valuable goal would be to create a streamlined release process that any external community member can manage with minimal effort. This approach ensures accessibility and efficiency while still maintaining control and quality.

## What we have

Below is the list of all the automation we have w.r.t release:

[...]

Issue #5022, opened 2024-09-13
Automate manual steps while creating release candidate


gaiksaya:
### Is your feature request related to a problem? Please describe

We have automated most of the steps required to generate a release candidate. However, below steps are still manual:

1. Click docker build option to check `build-with-tag` in distribution build jobs on jenkins.
2. Post comment on release issue that new RC is ready for build. Per RC all that changes are the build numbers. 
3. Number of the RC in the distribution build job
4. RC build does not validate the artifact at the end of the job. It should validate all artifacts it generated and only then declare it as a release candidate.
5. Should also post the docker scan details in the GH comment about RC

P1 items:
1. Should have the code coverage data for current release compared to last release (Not RC, last release) It is one of the entry criteria today but we do not have a mechanism to track it.

### Acceptance criteria:
Automate the following manual actions

1. Automatically select `build-with-tag` for docker in distribution build jobs on jenkins.
2. Post comment on release issue that new RC is ready for build 
3. Number of the RC in the distribution build job
4. Validate RC
5. Post the docker scan details in the GH comment about RC

### Describe the solution you'd like

See above

gaiksaya:
@opensearch-project/engineering-effectiveness Please feel free to add any more ad-hoc steps that needs to be automated while creating RC.
Thanks!

prudhvigodithi:
One more additional point.
To consider it a valid RC the build and validation should pass, later the parameter RC_NUMBER should be updated based on the results of the build and validation.

PR #5269, opened 2025-01-27
Validate distribution while building RC

gaiksaya:
### Description
When the given build is a release candidate build, it should be validated before declaring the workflow as successful. This chnage adds the step to do so. 

### Issues Resolved
part of https://github.com/opensearch-project/opensearch-build/issues/5022 https://github.com/opensearch-project/opensearch-build/issues/5171

### DIFF
--- a/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
+++ b/jenkins/opensearch-dashboards/distribution-build.jenkinsfile
@@ -148,6 +148,9 @@ pipeline {
                     paramType.each { key, value ->
                         verifyParameterPlatformDistribution(key, value)
                     }
+
+                    def inputManifestObj = lib.jenkins.InputManifest.new(readYaml(file: "manifests/${INPUT_MANIFEST}"))
+                    env.version = inputManifestObj.build.version
                 }
             }
         }
@@ -953,6 +956,9 @@ pipeline {
         success {
             node(AGENT_LINUX_X64) {
                 script {
+                    if (params.RC_NUMBER.toInteger() > 0) {
+                        triggerDistributionValidationWorkflow(env.version)
+                    }
                     postCleanup()
                 }
             }
@@ -1006,6 +1012,22 @@ def triggerBWCTests(String buildManifestUrl, String platform, String distributio
                     string(name: 'AGENT_LABEL', value: AGENT_LINUX_X64)
                 ]
     }
+}
+
+def triggerDistributionValidationWorkflow(String version) {
+    def triggerValidationWorkflow =
+            build job: 'distribution-validation',
+            propagate: true,
+            wait: true,
+            parameters: [
+                string(name: 'VERSION', value: version),
+                string(name: 'OS_BUILD_NUMBER', value: 'latest'),
+                string(name: 'OSD_BUILD_NUMBER', value: BUILD_NUMBER),
+                string(name: 'OPTIONAL_ARGS', value: 'using-staging-artifact-only'),
+                string(name: 'DOCKER_SOURCE', value: 'dockerhub'),
+                string(name: 'PROJECTS', value: 'Both'),
+                string(name: 'ARTIFACT_TYPE', value: 'staging')
+            ]
 }
 
 def verifyParameterPlatformDistribution(String paramName, String allowedValue) {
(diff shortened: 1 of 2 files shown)


jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -1008,6 +1014,22 @@ def triggerBWCTests(String buildManifestUrl, String platform, String distributio
...
+    def triggerValidationWorkflow =
+            build job: 'distribution-validation',
+            propagate: true,
+            wait: true,
+            parameters: [
+                string(name: 'VERSION', value: version),
+                string(name: 'OS_BUILD_NUMBER', value: 'latest'),
+                string(name: 'OSD_BUILD_NUMBER', value: BUILD_NUMBER),
+                string(name: 'OPTIONAL_ARGS', value: 'using-staging-artifact-only'),
+                string(name: 'DOCKER_SOURCE', value: 'dockerhub'),
```
Divyaasm, line 1027: We can ignore this parameter Sayali, DOCKER_SOURCE defaults to dockerhub https://github.com/opensearch-project/opensearch-build/blob/c33d51b133c909927d62d9e0cebdba320b19fd71/src/validation_workflow/validation_args.py#L80

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -1008,6 +1014,22 @@ def triggerBWCTests(String buildManifestUrl, String platform, String distributio
     }
 }
 
+def triggerDistributionValidationWorkflow(String version) {
```
peterzhuamazon, line 1017: 1. Can we add a param for people to disable this behavior.
2. if OSD built before OS and failed validation due to OS being old version or not exist, how do we avoid such issue?

zelinh: +1. We should have a mechanism to disable the validation.
Besides that, I'm thinking do we skip this validation if the workflow itself fails?

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
```diff
@@ -1008,6 +1014,22 @@ def triggerBWCTests(String buildManifestUrl, String platform, String distributio
...
+    def triggerValidationWorkflow =
+            build job: 'distribution-validation',
+            propagate: true,
+            wait: true,
+            parameters: [
+                string(name: 'VERSION', value: version),
+                string(name: 'OS_BUILD_NUMBER', value: 'latest'),
```
zelinh, line 1024: Do we need to consider when users don't select the current build as the `latest`? Can we set this build number to the current job number?

Divyaasm: But this is for OSD build right! current job number will be OSD_BUILD_NUMBER
I think we have to consider the case where we first trigger OS build, eventually wait for a while and build OSD . This way it will pick up the latest OS build while generating RC.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
gaiksaya: 1. We want the RC build to fail if validation fails for any reason. We added too many params at this point and RM usually forgets what to check mark and what not to. Example: docker build-with-tag. (Forgot it many times)
2. I discussed with @rishabh6788 in order to not wait for entire OpenSearch build to complete, I passed the `latest` to validation which should still be okay. Else we would have to wait for backend to complete building.

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
gaiksaya: Its a choice in jenkins workflow. And defaults whatever is first choice. In case we change the order in future, it would default to whatever is first. Hence added this

jenkins/opensearch-dashboards/distribution-build.jenkinsfile
gaiksaya: Thats what we are doing right now. RC is not the first time we build a artifact for the given version. For validating OSD we just need some backend even if plugins are missing. Hence added `latest`. Also have not seen a scenario yet where we do not hit do not update latest

Divyaasm (approved):

PR #5263, opened 2025-01-25
Resolve plus sign causing access denied to node20 tarball

peterzhuamazon:
### Description
Resolve plus sign causing access denied to node20 tarball
Need to use `%2b` if we want to use literal `+` in encoding.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5248

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5280, opened 2025-01-30
Update codecov to v5 with new token

peterzhuamazon:
### Description
Update codecov to v5 with new token
Also remove jacoco codecov upload due to jacoco already removed here: https://github.com/opensearch-project/opensearch-build/pull/3298

### Issues Resolved


.github/workflows/groovy-tests.yml
```diff
@@ -15,7 +15,3 @@ jobs:
       - name: Run Tests
         run: |
           ./gradlew test --info
-      - name: Upload Coverage Report
-        uses: codecov/codecov-action@v2
-        with:
-          files: ./build/reports/jacoco/test/jacocoTestReport.xml
```
gaiksaya, line 21: We need codecov for this as well right?

peterzhuamazon: No because in https://github.com/opensearch-project/opensearch-build/pull/3298 you removed jacoco.

.github/workflows/python-tests.yml
```diff
@@ -76,7 +76,26 @@ jobs:
...
+  python-tests-codecov:
+    needs: python-tests-linux
+    strategy:
+      matrix:
+        os:
+          - ubuntu-24.04-arm  # arm64-preview
+          - ubuntu-24.04  # x64
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/download-artifact@v4
```
gaiksaya, line 93: Should be up `upload-artifact`?

peterzhuamazon: No, because we are downloading the coverage.xml that we uploaded from the actual pytest run job section.

gaiksaya: Also we dont need to upload twice. It will just overwrite the previous results right?

gaiksaya: Sorry not getting why we are downloading the coverage?

peterzhuamazon: Because AL2 does not support codecov actions over v3.
And only v5 have fixed this issue.

So we need to run the test on AL2, upload the coverage to github artifact, download it in a normal ubuntu server, then send to codecov using v5.

gaiksaya: Ah okay! In that case, why run the matrix if we are just uploading the already generated report?

peterzhuamazon: Because arm64 and x64 have two coverage yml, not one.
So I just for the sake of it upload twice, through matrix, instead of manually write two similar steps.

gaiksaya: I dont think it makes sense to upload the coverage twice. The tests are same and when you upload the coverage to the codecov it is for per branch. So the results will be overwritten whoever wins the race. Please correct if I am wrong.

peterzhuamazon: It is just to avoid hardcoding, upload twice is very common even in the original setup.
It wont cause any issue and codecov can easily get around it.

gaiksaya: Better to save resources for actual tests than using for minor uploads? Does not give an value IMO.

peterzhuamazon: It is possible at times where one test on x64 success while arm64 fail, so I dont think this is necessary.

gaiksaya: In that case code coverage will not be uploaded as the step abruptly exits? Only on successful runs, code coverage gets uploaded to the codecov. [Example](https://github.com/opensearch-project/opensearch-build/pull/5263#issuecomment-2613640365)

peterzhuamazon: Updated to use only one upload.

.github/workflows/python-tests.yml
```diff
@@ -76,7 +76,26 @@ jobs:
...
+          name: coverage-report-${{ matrix.os }}
+          path: ./coverage.xml
+
+  python-tests-codecov:
+    needs: python-tests-linux
+    strategy:
+      matrix:
+        os:
+          - ubuntu-24.04  # x64
+    runs-on: ${{ matrix.os }}
```
gaiksaya, line 91:
```suggestion
  python-tests-codecov:
    needs: python-tests-linux
    runs-on: ubuntu-24.04
```

peterzhuamazon: Resolved. thanks.

peterzhuamazon:
Hi @gaiksaya ready to merge please review.

Thanks.

gaiksaya (approved):

Issue #5320, opened 2025-02-13
[Enhancement] Enable test record and results report for smoke test framework


zelinh:
## Overview
With the Smoke Test Framework ([issue #5223](https://github.com/opensearch-project/opensearch-build/issues/5223)) now established, we have successfully onboarded multiple OpenSearch plugins—including Index Management, Security, SQL, and K-NN—to ensure reliable pre-release validation.

To further improve integration and visibility, we aim to enhance the smoke tests with a reporting and notification mechanism. Ideally, after each run, the framework should generate a structured test report similar to the [integration test report](https://build.ci.opensearch.org/job/integ-test/lastSuccessfulBuild/artifact/test-report.yml/*view*/), capturing test results and associated logs.

The generated report will be:
- Improving visibility into test results.
- Consumed by metric dashboards for more insights.
- Used for automated failure notifications, enabling faster issue resolution.

### Tasks
- [ ] Enhance the smoke test workflow to record test results and generate a structured component YAML file.
- [ ] Integrate with the reporting workflow to produce an indexable test-report.yml containing test outcomes.
- [ ] Work on indexing the new test report into existing metric dashboards for monitoring and analysis.

This improvement will provide greater transparency, automation, and efficiency in OpenSearch distribution testing.

zelinh:
I'm planning to record the component smoke test results using the following hierarchy for `test_results`:
```
-- test_results
  -- smoke_tests
       -- local_cluster_logs
            -- stdout.txt
            -- stderr.txt
       -- security 
            -- GET_plugins_security_api_account
                -- stdout.txt
                -- stderr.txt
                -- security.yml
            -- PUT_plugins_security_api_actiongroups_my-action-group
                -- stdout.txt
                -- stderr.txt
                -- security.yml
            -- POST_...
       -- ...
       -- ...
```
This structure ensures that logs and test results are well-organized by component and API, making it easier to track failures and debug issues.

zelinh:
We have completed the test report generation implementation for smoke tests in recent commits. I will make sure add the report generation during the Jenkins test and attach in each Jenkins job.

PR #5333, opened 2025-02-20
Recording components logs and generate component yaml for smoke tests framework

zelinh:
### Description
Recording components logs and generate component yaml for smoke tests framework

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/5320

### DIFF
--- a/src/test_workflow/smoke_test/smoke_test_runner_opensearch.py
+++ b/src/test_workflow/smoke_test/smoke_test_runner_opensearch.py
@@ -5,10 +5,12 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+import io
 import logging
 import os
+from logging import Handler
 from pathlib import Path
-from typing import Any
+from typing import Any, Tuple
 
 import requests
 from openapi_core import Spec, validate_request, validate_response
@@ -17,6 +19,7 @@ from openapi_core.contrib.requests import RequestsOpenAPIRequest, RequestsOpenAP
 from manifests.test_manifest import TestManifest
 from test_workflow.smoke_test.smoke_test_runner import SmokeTestRunner
 from test_workflow.test_args import TestArgs
+from test_workflow.test_recorder.test_result_data import TestResultData
 from test_workflow.test_result.test_component_results import TestComponentResults
 from test_workflow.test_result.test_result import TestResult
 from test_workflow.test_result.test_suite_results import TestSuiteResults
@@ -64,6 +67,31 @@ class SmokeTestRunnerOpenSearch(SmokeTestRunner):
         validate_response(response=response, spec=self.spec_, request=request)
         logging.info("Response is validated.")
 
+    def record_test_result(self, component: str, test_api: str, api_action: str, stdout: str, stderr: str) -> None:
+        test_config = f"{api_action}_{test_api.replace('/', '_')}"
+        logging.info(f"Recording test result for {component} component, config {test_config}")
+        file_path = self.test_recorder._create_base_folder_structure(component, test_config)
+        self.test_recorder._generate_std_files(stdout, stderr, file_path)
+
+    def setup_logging_buffers(self) -> Tuple[io.StringIO, io.StringIO, logging.StreamHandler, logging.StreamHandler]:
+        info_buffer = io.StringIO()
+        error_buffer = io.StringIO()
+
+        info_handler = logging.StreamHandler(info_buffer)
+        error_handler = logging.StreamHandler(error_buffer)
+
+        info_handler.setLevel(logging.INFO)
+        error_handler.setLevel(logging.ERROR)
+
+        logging.getLogger().addHandler(info_handler)
+        logging.getLogger().addHandler(error_handler)
+
+        return info_buffer, error_buffer, info_handler, error_handler
+
+    def cleanup_logging_handlers(self, info_handler: Handler, error_handler: Handler) -> None:
+        logging.getLogger().removeHandler(info_handler)
+        logging.getLogger().removeHandler(error_handler)
+
     def start_test(self, work_dir: Path) -> TestSuiteResults:
         url = "https://localhost:9200"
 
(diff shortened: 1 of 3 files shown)


src/test_workflow/smoke_test/smoke_test_runner_opensearch.py
```diff
@@ -64,6 +67,31 @@ def validate_response_swagger(self, response: Any) -> None:
         validate_response(response=response, spec=self.spec_, request=request)
         logging.info("Response is validated.")
 
+    def record_test_result(self, component: str, test_api: str, api_action: str, stdout: str, stderr: str) -> None:
+        test_config = f"{api_action}_{test_api.replace('/', '_')}"
```
peterzhuamazon, line 71: Can you explain what the actual line would be?

zelinh: In this case, test results folder will be looked like this:
```
-- test_results
  -- smoke_tests
       -- local_cluster_logs
            -- stdout.txt
            -- stderr.txt
       -- security 
            -- GET___plugins__security_api_account
                -- stdout.txt
                -- stderr.txt
                -- security.yml
            -- PUT___plugins__security_api_actiongroups_my-action-group
                -- stdout.txt
                -- stderr.txt
                -- security.yml
            -- POST_...
       -- ...
       -- ...
```

src/test_workflow/smoke_test/smoke_test_runner_opensearch.py
```diff
@@ -64,6 +67,31 @@ def validate_response_swagger(self, response: Any) -> None:
...
+        info_handler = logging.StreamHandler(info_buffer)
+        error_handler = logging.StreamHandler(error_buffer)
+
+        info_handler.setLevel(logging.INFO)
+        error_handler.setLevel(logging.ERROR)
+
+        logging.getLogger().addHandler(info_handler)
+        logging.getLogger().addHandler(error_handler)
+
+        return info_buffer, error_buffer, info_handler, error_handler
```
gaiksaya, line 89: What values are re getting from this?

src/test_workflow/smoke_test/smoke_test_runner_opensearch.py
gaiksaya: Do we really need stdout and stderr for each API? Feels like too many files for just a single API response.

zelinh: If the API test passes, the stderr will be empty and stdout will display all logging for this specific API path. There will be just two std files and a component yml file generated for each API.

zelinh: Each of the component yaml file(e.g. security.yml) will contain specific test results so I believe it's better to have them separately. WDYT @rishabh6788

zelinh: Another reason we have component yml file is report workflow loads each of the test results from it. If we want to create a new consolidated component yml file including all test results for one plugin component, we have to refractor the report workflow logic as well.

zelinh: Had an offline discussion. This PR is to generate component yaml files (e.g. security.yml) which will be consumed by report workflow to generate consolidated test reports to include all test results for all components. 
Another PR to update the report workflow is pending.

gaiksaya: Thinking from component's team point of view, I still feel having component.yml for each API is too much. componet.yml as the name suggests should be per component. An aggregated result of all the APIs run. Also I believe we would have with and without security in the future as well?
Just an example of what I am asking:
```yaml
cat security.ml
- GET
  - api_name: account
     result: PASS
     logs:
  - api_name: actiongroups_my-action-group
    result: fail
    logs:
- PUT
- DELET
```
If you think this is too complicated we can move with what we have and iterate on it but just want to make sure its a two way door. 
@rishabh6788 @peterzhuamazon  What do you think?

zelinh: This proposal is a valuable enhancement. However, it requires careful consideration of multiple fields since it modifies the structure of the component YAML file.

Here are a few areas to take into account:
1. We will need to update the test recorder to generate this new YAML format based on the current implementation: https://github.com/opensearch-project/opensearch-build/blob/a7b2f8de96df09e2fa5430c64d03444b20a71e9d/src/test_workflow/test_recorder/test_recorder.py#L58-L74.
2. Currently, test results are recorded immediately after each test completes. We may need to adjust the workflow logic to record results after an entire component has completed, though this seems manageable.
3. The reporting workflow currently reads from the test manifest and loads each component’s YAML file based on its relative path. If we modify the test result structure, significant updates will be required in the reporting workflow to ensure accurate report generation for smoke tests.

I believe the consolidated results for each component should still be reflected in test-report.yml, while individual component YAMLs can serve as references.

I'm happy to proceed with these changes if we determine that this approach is essential. Let me know your thoughts!

gaiksaya: I understand the additional refactoring for smoke testing, however, thinking from user's perspective (who often avoid clicking multiple links in the past), it would be good to have a consolidated view of results especially for 1 single component and one config. We would also be adding with and without security in future right? So the number of files for each API would be doubling. 

@rishabh6788 @peterzhuamazon Let me know what you think.

zelinh: > I understand the additional refactoring for smoke testing, however, thinking from user's perspective (who often avoid clicking multiple links in the past), it would be good to have a consolidated view of results especially for 1 single component and one config. We would also be adding with and without security in future right? So the number of files for each API would be doubling.
> 
> @rishabh6788 @peterzhuamazon Let me know what you think.

Yeah, the thing is most user will firstly look into the later generated `test-report.yml` will include all consolidated test results. They can simply find any failing tests from there and click into the exact failing test component yml without need to check others. Making a consolidated component yaml and consolidated test report seems like duplicate to me in addition to require multiple refactoring on our different workflows. I think we can think about a better way once we are onto adding with/without security in the future. WDYT. @gaiksaya Thanks!

zelinh: @gaiksaya Do you have any other concerns on this PR? Are we OK to move forward with this?

gaiksaya: Not just metrics but even the test-report would be too much information. Anyway we can proceed with this and see how it goes. We can accordingly adjust to new reviews.

gaiksaya (approved):

peterzhuamazon (approved):

PR #5353, opened 2025-02-27
Update version increment workflow to remove main/3.x updates

peterzhuamazon:
### Description
Update version increment workflow to remove main/3.x updates

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3747
As of now 3.x version increment are crafted specifically to support qualifiers with a lot of tweaks from plugin side.
Disable this for now until after GA so we do not receive PRs.

### DIFF
--- a/.github/workflows/os-increment-plugin-versions.yml
+++ b/.github/workflows/os-increment-plugin-versions.yml
@@ -53,9 +53,6 @@ jobs:
           - {repo: opensearch-remote-metadata-sdk}
           - {repo: opensearch-learning-to-rank-base}
         branch:
-          - 2.x
-          - main
-          - '2.18'
           - '2.19'
           - '2.20'
         exclude:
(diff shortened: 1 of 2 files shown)


.github/workflows/os-increment-plugin-versions.yml
```diff
@@ -53,11 +53,8 @@ jobs:
           - {repo: opensearch-remote-metadata-sdk}
           - {repo: opensearch-learning-to-rank-base}
         branch:
-          - 2.x
-          - main
```
prudhvigodithi, line 57: We cans till keep `main` right Peter as it will handle future version increments?

peterzhuamazon: I will add main once 3.0.0 is out.

prudhvigodithi (approved):

PR #5407, opened 2025-03-18
Adding User Behavior Insights API tests

jzonthemtn:
### Description
Adds User Behavior Insights (UBI) smoke tests per https://github.com/opensearch-project/user-behavior-insights/issues/77.

### Issues Resolved
Part of https://github.com/opensearch-project/user-behavior-insights/issues/77 but does not close it until [opensearch-api-specification#844](https://github.com/opensearch-project/opensearch-api-specification/issues/844) in opensearch-api-specification is also merged.

### DIFF
--- a/src/test_workflow/smoke_test/smoke_tests_spec/2.x/ubi.yml
+++ b/src/test_workflow/smoke_test/smoke_tests_spec/2.x/ubi.yml
@@ -1,0 +1,9 @@
+---
+info:
+  title: OpenSearch UBI plugin smoke tests
+  version: default
+name: ubi
+paths:
+  /_plugins/ubi/initialize:
+    POST:
+      parameters: []


src/test_workflow/smoke_test/smoke_tests_spec/2.x/ubi.yml
zelinh, line 1: Hey @jzonthemtn, will UBI plugin be part of OpenSearch 2.x distribution bundle? I only see it in 3.0.0 manifest. https://github.com/opensearch-project/opensearch-build/blob/75babb59ac2e64de356a40730adc8caa046c3df1/manifests/3.0.0/opensearch-3.0.0.yml#L253C11-L253C33

jzonthemtn: No, thanks for catching that. I have moved the `ubi.yaml` file to be under the `default` directory. I assume that's the directory for 3.0.

epugh:
Hi all, I think the UBI plugin is much better positioned to participate in release tests.  I think this PR can now be merged?  Since some other plugins like search-relevance have a optional dependency on ubi, it would be good to make sure this gets integrated.  We can expand the smoke tests from here if needed.

gaiksaya:
Looks like we still have UBI pending to be on-boarded to api-spec https://github.com/opensearch-project/opensearch-api-specification/pull/845
We do not have failure notifications for smoke testing in place yet but tests would fail until the PR gets in.

PR #5423, opened 2025-03-24
Add credentials to retrieve jenkins data

gaiksaya:
### Description
Add credentials to retrieve jenkins data using curl GET requests as we plan to switch read-only access to authenticated users.

### Issues Resolved

### DIFF
--- a/scripts/benchmark/benchmark-pull-request.sh
+++ b/scripts/benchmark/benchmark-pull-request.sh
@@ -17,7 +17,7 @@ TRIGGER_TOKEN=""
 GITHUB_USER=""
 GITHUB_TOKEN=""
 
-while getopts "u:t:" opt; do
+while getopts "u:t:p:" opt; do
   case $opt in
     t)
       TRIGGER_TOKEN="$OPTARG"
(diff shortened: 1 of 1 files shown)


peterzhuamazon:
Hi @gaiksaya assuming we also need to update gradle check?
As that one requires jenkins api as well.

Thanks.

gaiksaya:
> Hi @gaiksaya assuming we also need to update gradle check? As that one requires jenkins api as well.
> 
> Thanks.

Yes workflow has low blast radius in case this does not work as expected. We will exercise this first and then move to gradle check. We will switch from anonymous to auth only after all credentials are in place.

scripts/benchmark/benchmark-pull-request.sh
```diff
@@ -11,11 +11,51 @@
...
 
-
+# Default values
 JENKINS_URL="https://build.ci.opensearch.org"
-TRIGGER_TOKEN=$1
+TRIGGER_TOKEN=""
+GITHUB_USER=""
+GITHUB_TOKEN=""
+
+while getopts "u:t:p:" opt; do
```
rishabh6788, line 20: would this affect the way script is executed? Do we need to pass some flags as well now?

gaiksaya: Yes just seems cleaner instead of positional args. Will be making change in core repo before merging this.
It will look like this
```bash
./benchmark-pull-request.sh -t genericToken -u user -p githubToken

rishabh6788 (approved):

PR #5442, opened 2025-04-14
Add support for java agent in deb/rpm packages in 3.x

peterzhuamazon:
### Description
Add support for java agent in deb/rpm packages in 3.x

### Issues Resolved
Closes https://github.com/opensearch-project/OpenSearch/issues/17914


zelinh (commented):
Looks good to me. Wonder we don't change anything to build templates from previous 1.x to 2.x right?

peterzhuamazon:
> Looks good to me. Wonder we don't change anything to build templates from previous 1.x to 2.x right?

No that is why we rename legacy to 1.x, and make a copy of current as 2.x, so that 3.x and above will be using current.

zelinh (approved):
Thanks. @peterzhuamazon 
We only made changes to 3.x, so we need to separate it from 2.x. For 2.x, we generate build templates by duplicating the ones from 1.x.

peterzhuamazon:
> Thanks. @peterzhuamazon We only made changes to 3.x, so we need to separate it from 2.x. For 2.x, we generate build templates by duplicating the ones from 1.x.

Correction: duplicate `current` to get 2.x instead of `legacy` which is reserved only for 1.x.

Thanks.

PR #5525, opened 2025-05-19
Fix BuildManifestBuilder.append_artifact to safely initialize artifact lists

aravindkesavan:
### Type of change
- Bug Fix / Improvement

### Description
This change fixes a bug in the BuildManifestBuilder.append_artifact method where the artifact list was not safely initialized. Previously, the code retrieved the list using artifacts.get(type, []) but did not always assign the empty list back to the dictionary, which could lead to cases where the list was not properly stored or updated.

The fix ensures that if the artifact type key does not exist or is not a list, an empty list is explicitly assigned before appending the new artifact path. 

### Issues Resolved
No specific issue number was referenced, but this fix improves the stability of the build manifest creation logic.

Note: Fixes initialization of artifact lists in BuildManifestBuilder.append_artifact. Ensures the artifact type key always holds a list before appending, preventing potential errors during build manifest creation.


peterzhuamazon:
Hi @aravindkesavan ,

Thanks for PR.
Do you have an example of the failures before this fix?
Trying to understand about it because I have not see issues related to this before. Thanks.

PR #5541, opened 2025-06-01
Add --dry-run option to AssembleArgs to support non-destructive runs

aravindkesavan:
### Description

This PR adds a `--dry-run` option to the assemble tool to support non-destructive runs, allowing users to simulate the assembly process without downloading or extracting artifacts. Additionally, it introduces the capability to generate a manifest file dynamically via the `--generate-manifest` option.

With these enhancements, users can:

- Dynamically generate a manifest YAML file with customizable parameters such as version, platform, architecture, distribution, and component.
- Perform dry runs using an existing or dynamically generated manifest to validate the assemble logic safely without modifying disk or network state.

### Generate a Manifest File Dynamically

Run the following command to create a manifest file named `dry-run-manifest.yml` with your desired parameters (all optional):

**Generate a new manifest file for version 2.14.0:**

```bash
python src/run_assemble.py --generate-manifest output-manifest.yml --version 2.14.0 --platform linux --arch x64 --dist tar --component opensearch
```

Using the below command to run the assemble tool in dry-run mode with the generated manifest:

```bash
python3 src/run_assemble.py dry-run-manifest.yml --dry-run
```

Other enhancements include new command-line arguments to specify version, platform, architecture, distribution type, and component name for the manifest generation feature.

### Issues Resolved
- Adds safer testing and debugging capabilities for the assemble workflow.
- Provides dynamic manifest generation to simplify build and release workflows.
- Improves CLI usability and feature set.

### DIFF
--- a/src/assemble_workflow/README.md
+++ b/src/assemble_workflow/README.md
@@ -28,6 +28,53 @@ The following options are available in `assemble.sh`.
 | --keep             | Do not delete the temporary working directory on both success or error. |
 | -v, --verbose      | Show more verbose output.                                               |
 
+
+### run_assemble.py Options
+
+You can also run the assembling logic directly using Python:
+
+```bash
+python src/run_assemble.py <manifest-file> [options]
+```
+
+If you're generating a manifest dynamically, you can omit the input file and use `--generate-manifest`.
+
+#### CLI Flags
+
+| Option                  | Description                                                                 |
+|--------------------------|-----------------------------------------------------------------------------|
+| `--generate-manifest`   | Generate a manifest YAML file dynamically (default filename: `manifest.yml`). |
+| `--dry-run`             | Simulate the assemble step without downloading or extracting artifacts.      |
+| `--version`             | OpenSearch version (e.g., `2.14.0`). Used with `--generate-manifest`.        |
+| `--platform`            | Target platform (`linux`, `windows`, `macos`).                              |
+| `--arch`                | Target architecture (`x64`, `arm64`).                                       |
+| `--dist`                | Distribution type (`tar`, `zip`, `rpm`).                                    |
+| `--component`           | Component name (defaults to `opensearch`).                                  |
+| `-b`, `--base-url`      | Base URL to download artifacts from.                                        |
+| `--keep`                | Do not delete the working temporary directory.                              |
+| `-v`, `--verbose`       | Enable verbose logging.                                                     |
+
+#### Examples
+
+**Assemble using an existing manifest:**
+
+```bash
+python src/run_assemble.py builds/opensearch/manifest.yml --verbose
+```
+
+**Dry run (simulate without downloading or extracting):**
+
+```bash
+python src/run_assemble.py builds/opensearch/manifest.yml --dry-run
+```
+
+**Generate a new manifest file for version 2.14.0:**
+
+```bash
+python src/run_assemble.py --generate-manifest output-manifest.yml --version 2.14.0 --platform linux --arch x64 --dist tar --component opensearch
+```
+
+
 ### Custom Install Scripts
 
 You can perform additional plugin install steps by adding an `install.sh` script. By default the tool will look for a script in [scripts/bundle-build/components](../../scripts/bundle-build/components), then default to a noop version implemented in [scripts/default/install.sh](../../scripts/default/install.sh).
--- a/src/run_assemble.py
+++ b/src/run_assemble.py
@@ -9,6 +9,8 @@
 import logging
 import os
 import sys
+import platform
+import yaml
 
 from assemble_workflow.assemble_args import AssembleArgs
 from assemble_workflow.bundle_locations import BundleLocations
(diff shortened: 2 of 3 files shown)


aravindkesavan:
@gaiksaya Could you please review this feature and provide your opinion on the next GO

gaiksaya:
> @gaiksaya Could you please review this feature and provide your opinion on the next GO

Will do in next few days. Just returned from my vacation today so catching up.

src/run_assemble.py
```diff
@@ -18,12 +20,55 @@
...
 from system import console
 
+def generate_manifest_file(args: AssembleArgs):
+    component = args.component or "opensearch"
+    platform_name = args.platform or platform.system().lower()
+    arch = args.arch or platform.machine().lower()
+    version = args.version or "0.0.1"
+    dist = args.dist or "tar"
+
+    url = f"https://artifacts.opensearch.org/releases/bundle/{component}/{version}/{component}-{version}-{platform_name}-{arch}.{dist}"
```
peterzhuamazon, line 30: We should not hard code this to prod url as if you see other similar features, we only use prod url substitutions in Jenkins lib. As opensearch-build was meant to run in any env and expected the files to be linked to local file paths.

src/assemble_workflow/README.md
```diff
@@ -28,6 +28,53 @@ The following options are available in `assemble.sh`.
...
 | -v, --verbose      | Show more verbose output.                                               |
 
+
+### run_assemble.py Options
+
+You can also run the assembling logic directly using Python:
+
+```bash
+python src/run_assemble.py <manifest-file> [options]
+```
```
peterzhuamazon, line 38: I do not think specifying this as we use the shell script as entrypoint with everything linked to pipenv.
This would add up confusions and remove the abstraction layer.
Users may not installed all the libs beforehand.

src/assemble_workflow/README.md
```diff
@@ -28,6 +28,53 @@ The following options are available in `assemble.sh`.
...
+| `--generate-manifest`   | Generate a manifest YAML file dynamically (default filename: `manifest.yml`). |
+| `--dry-run`             | Simulate the assemble step without downloading or extracting artifacts.      |
+| `--version`             | OpenSearch version (e.g., `2.14.0`). Used with `--generate-manifest`.        |
+| `--platform`            | Target platform (`linux`, `windows`, `macos`).                              |
+| `--arch`                | Target architecture (`x64`, `arm64`).                                       |
+| `--dist`                | Distribution type (`tar`, `zip`, `rpm`).                                    |
+| `--component`           | Component name (defaults to `opensearch`).                                  |
+| `-b`, `--base-url`      | Base URL to download artifacts from.                                        |
+| `--keep`                | Do not delete the working temporary directory.                              |
+| `-v`, `--verbose`       | Enable verbose logging.                                                     |
```
peterzhuamazon, line 55: These could be add in the similar chart above when discussing `assemble.sh`.

src/assemble_workflow/README.md
```diff
@@ -28,6 +28,53 @@ The following options are available in `assemble.sh`.
...
+
+```bash
+python src/run_assemble.py builds/opensearch/manifest.yml --dry-run
+```
+
+**Generate a new manifest file for version 2.14.0:**
+
+```bash
+python src/run_assemble.py --generate-manifest output-manifest.yml --version 2.14.0 --platform linux --arch x64 --dist tar --component opensearch
+```
```
peterzhuamazon, line 75: Same as above, can be added as examples for assemble.sh.

src/run_assemble.py
```diff
@@ -18,12 +20,55 @@
...
     args = AssembleArgs()
 
+    if args.generate_manifest:
+        generate_manifest_file(args)
+        return 0
+
     console.configure(level=args.logging_level)
 
+    if args.dry_run:
+        print("Dry-run enabled. Skipping actual download or extraction.")
```
peterzhuamazon, line 64: We should use logging instead of print.

src/run_assemble.py
```diff
@@ -18,12 +20,55 @@
...
+        if args.manifest:
+            print(f"Manifest file: {args.manifest.name}")
+        else:
+            print("No manifest file provided.")
+        return 0
```
peterzhuamazon, line 69: Did not get, what is the usage of --dry-run here?
I can understand generate_manifest without actual assembling, but for dry_run itself it just print three times?

src/run_assemble.py
```diff
@@ -18,12 +20,55 @@
...
+            "distribution": dist
+        },
+        "components": [
+            {
+                "name": component,
+                "version": version,
+                "url": url
+            }
+        ]
+    }
```
peterzhuamazon, line 47: Are we not reusing the recorder logic for assemble_workflow?

src/run_assemble.py
peterzhuamazon: Similar for all the prints here.

peterzhuamazon (changes requested):
Thanks for contribution.
Tho I have a few doubt about this PR.

Thanks.

Issue #5542, opened 2025-06-02
[On-boarding]: Search Relevance


heemin32:
### What is the name of your plugin?

Search Relevance

### What is the link to your GitHub repo?

https://github.com/opensearch-project/search-relevance

### Targeted release version

3.1

### Does the plugin require documentation?

Yes. Documentation is not ready yet.

### What type of the plugin are we on-boarding?

OpenSearch

### Did you read the on-boarding document?

Yes

### Does this plugin has all the necessary automated tests?

Yes. BWC test will be added later.

### Have you completed the required reviews including Security reviews, UX reviews?

Security review is not completed. UX review is not required as this is a backend component.

### Have you on-boarded automated security scanning for the GitHub repo associated with this plugin?

Yes

rishabh6788:
Closing this.

PR #5543, opened 2025-06-02
Add search relevance plugin to the distribution

heemin32:
### Description
Add search relevance plugin to the distribution

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5542

### DIFF
--- a/manifests/3.1.0/opensearch-3.1.0.yml
+++ b/manifests/3.1.0/opensearch-3.1.0.yml
@@ -224,3 +224,9 @@ components:
     platforms:
       - linux
       - windows
+  - name: search-relevance
+    repository: https://github.com/opensearch-project/search-relevance.git
+    ref: main
+    platforms:
+      - linux
+      - windows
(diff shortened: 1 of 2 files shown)


manifests/3.1.0/opensearch-3.1.0.yml
```diff
@@ -224,3 +224,9 @@ components:
     platforms:
       - linux
       - windows
+  - name: search-relevance
```
zelinh, line 227: Does this plugin have any dependency components? If so, we should list them out in this manifest.
e.g. https://github.com/opensearch-project/opensearch-build/blob/2565848ec951b413e821b688cc402c1bf6d4960b/manifests/3.1.0/opensearch-3.1.0.yml#L78-L86

heemin32: There is no hard dependency on other plugins. https://github.com/opensearch-project/search-relevance/blob/main/build.gradle

zelinh (approved):

PR #5554, opened 2025-06-04
deb - Listen to DISABLE_INSTALL_DEMO_CONFIG

indykoning:
### Description
This PR allows users installing the package on debian to request no demo config. Taken from the docker install

### Issues Resolved
- https://github.com/opensearch-project/security/issues/4199

### DIFF
--- a/scripts/pkg/build_templates/1.x/opensearch/deb/debian/postinst
+++ b/scripts/pkg/build_templates/1.x/opensearch/deb/debian/postinst
@@ -21,7 +21,9 @@ pid_dir=/var/run/opensearch
 
 # Apply Security Settings
 if [ -d ${product_dir}/plugins/opensearch-security ]; then
-    bash ${product_dir}/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s > ${log_dir}/install_demo_configuration.log 2>&1
+    if [ ! "$DISABLE_INSTALL_DEMO_CONFIG" = "true" ]; then
+        bash ${product_dir}/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s > ${log_dir}/install_demo_configuration.log 2>&1
+    fi
 fi
 
 # Apply PerformanceAnalyzer Settings
--- a/scripts/pkg/build_templates/2.x/opensearch/deb/debian/postinst
+++ b/scripts/pkg/build_templates/2.x/opensearch/deb/debian/postinst
@@ -21,7 +21,9 @@ pid_dir=/var/run/opensearch
 
 # Apply Security Settings
 if [ -d ${product_dir}/plugins/opensearch-security ]; then
-    bash ${product_dir}/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s > ${log_dir}/install_demo_configuration.log 2>&1 || (echo "ERROR: Something went wrong during demo configuration installation. Please see the logs in ${log_dir}/install_demo_configuration.log" && exit 1)
+    if [ ! "$DISABLE_INSTALL_DEMO_CONFIG" = "true" ]; then
+        bash ${product_dir}/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s > ${log_dir}/install_demo_configuration.log 2>&1 || (echo "ERROR: Something went wrong during demo configuration installation. Please see the logs in ${log_dir}/install_demo_configuration.log" && exit 1)
+    fi
 fi
 # Apply PerformanceAnalyzer Settings
 chmod a+rw /tmp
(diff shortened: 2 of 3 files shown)


peterzhuamazon:
Adding @DarshitChanpura @derek-ho I remember we discussed this before but at the time response it we need to have it installed by default.

Thanks.

inflatador:
Hello, this is Brian with the Wikimedia Foundation. I was wondering if anyone had a chance to look at this yet? Our organization requires us to build our own Docker images using Debian packages. 

Since the package requires a password, we're [currently hard-coding it in our Docker build files](https://gitlab.wikimedia.org/repos/data-engineering/opensearch/-/merge_requests/9/diffs#1f86a055f040df8f5ad29740de495d716a1d75d2_12_17). 

While this is safe enough (the datadir never gets copied to the image), it looks bad enough that I've had to explain myself a few times. If we could get this merged, I think it would simplify a lot of similar workflows (such as installing via config management).

krisfreedain:
@peterzhuamazon @gaiksaya @rishabh6788 @zelinh @prudhvigodithi @Divyaasm @tianleh -- can you look at @inflatador 's input? thanks

peterzhuamazon:
As previously stated, the request to mandate demo installation with specific password was requested by the security team back in 2.12.0.
This change required the inputs from @DarshitChanpura @derek-ho who was working on the password requirement back then.

Thanks.

bugfood:
Additional bugs about this which I have found are:
https://github.com/opensearch-project/security/issues/4199 (already mentioned above)
https://github.com/opensearch-project/security/issues/4344
https://github.com/opensearch-project/security/issues/4965

My position on this is that either the demo configuration should not be created by default, or it should default to a random password (generated on demand).

For example, the Debian package for bacula generates random passwords:
https://salsa.debian.org/bacula-team/bacula/-/blob/master/debian/additions/common-functions.dpkg#L6-L46

Either way, package installation should succeed without requiring the user to specify _any_ environment variables.

Imagine if using an environment variable this way were standard--installing packages would be utter chaos. For example:
```
sudo env MARIADB_PASSWD=baz apt install mariadb-server
sudo env BACULA_DIRPASSWD=foo BACULA_DIRMPASSWD=bar ... apt install bacula-director
sudo env POSTGRES_PASSWD=qux apt install postgres
```

Admins and configuration management systems would need to know and support all that, and installation would still be doomed if another package pulled in any such package as a dependency.

The opensearch situation is worse on Debian than it is on RHEL (at least as I tested on AlmaLinux 8).
* Debian considers post-install failure fatal, so the package ends up with a `half-configured` status, and the failure results the package manager returning a non-zero exit status (failure).
* RHEL reports post-install failures as a warning; the post-install script exits early, leaving later steps unfinished, but the package manager returns a zero exit status (ok).

At least with RHEL, for better or for worse, tolerating the failure lets configuration management (or an admin) proceed onward and properly configure opensearch.

zaeemarshad:
Chiming in with my own 2 cents here - we use Puppet to setup Debian based Opensearch clusters. The hardcoded demo configuration means that the puppet manifest is littered with execs like `remove_demo_users`, `remove_demo_configs` etc. The original proposal is a decent one which will allow the users to control how their packages should be installed.

inflatador:
Thank you for the link, @zaeemarshad ! I'm about to start writing Puppet code for [a 1.x->2.x update](https://phabricator.wikimedia.org/T421757) and that module looks very useful.

scripts/pkg/build_templates/1.x/opensearch/deb/debian/postinst
```diff
@@ -21,7 +21,9 @@ pid_dir=/var/run/opensearch
 
 # Apply Security Settings
 if [ -d ${product_dir}/plugins/opensearch-security ]; then
-    bash ${product_dir}/plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s > ${log_dir}/install_demo_configuration.log 2>&1
+    if [ ! "$DISABLE_INSTALL_DEMO_CONFIG" = "true" ]; then
```
DarshitChanpura, line 24: Since the default password requirement applies only to demo config, and it is currently enabled on the install path, the correct path forward would be to have it disabled by default and have users configure it manually.

What do you think about flipping this logic to `ENABLE_INSTALL_DEMO_CONFIG` which only installs demo config if set to true? Since, production clusters do not use demo setup, this will prevent most admins from having to specify a flag to disable install.

inflatador:
Since this is time-sensitive and we haven't heard back from the original submitter in awhile, I've opened https://github.com/opensearch-project/opensearch-build/pull/6124 , incorporating feedback from @DarshitChanpura 's review above. Apologies for the confusion!

scripts/pkg/build_templates/1.x/opensearch/deb/debian/postinst
peterzhuamazon: This is not needed anymore since 1.x is already deprecated and we will not release newer version of 1.x.

peterzhuamazon:
Realized there was another similar one in : https://github.com/opensearch-project/opensearch-build/pull/6124/changes.

peterzhuamazon:
Superseded by https://github.com/opensearch-project/opensearch-build/pull/6131.

Thanks.

PR #5610, opened 2025-06-25
Add native plugin validation for staged artifacts

Divyaasm:
### Description
Add native plugin validation for staged artifacts of distribution type tarball, rpm, debian, zip. And enables native plugin validation by default for all the production artifacts.

### Issues Resolved
https://build.ci.opensearch.org/blue/organizations/jenkins/distribution-validation/detail/distribution-validation/2808/pipeline


src/validation_workflow/validation.py
```diff
@@ -54,9 +54,16 @@ def check_for_security_plugin(self, work_dir: str) -> bool:
...
-        for native_plugin in self.native_plugins_list:
-            execute('.' + os.sep + f'opensearch-plugin install --batch {native_plugin}', os.path.join(path, "bin"))
+        native_plugins_list = self.get_native_plugin_list(path, installed_plugins_list)
+        if self.args.artifact_type == "staging":
+            self.download_and_install_native_plugin_zip(native_plugins_list, os.path.join(path, "bin"))
+        else:
+            try:
+                for native_plugin in native_plugins_list:
+                    execute('.' + os.sep + f'opensearch-plugin install --batch {native_plugin}', os.path.join(path, "bin"))
+
```
peterzhuamazon: Feels like you should just include this in download_and_install_native_plugin_zip func.
The func should internally check if it is staging vs production via a parameter.

src/validation_workflow/validation_args.py
```diff
@@ -117,13 +117,6 @@ def __init__(self) -> None:
             default=False,
             help="If True, use http/https to connect based on the existence of security plugin, else always use https, default to False"
         )
-        parser.add_argument(
-            "-np",
-            "--validate-native-plugin",
-            action="store_true",
-            default=False,
-            help="Set the option to True if native plugins for OS need to be installed before starting the cluster, defaults to False"
-        )
```
peterzhuamazon, line 126: So validate native plugin is enabled by default now?

Divyaasm: Yes for all except yum and docker in staging

src/validation_workflow/validation.py
Divyaasm: Then instead of moving just this block I  can combine download_and_install_native_plugin_zip and install_native_plugin into one. Let me know.

peterzhuamazon (approved):

Issue #5614, opened 2025-06-30
Generate release notes using LLMs for OS and OSD components.


rishabh6788:
This proposal outlines the development of an automated release notes generation system for OpenSearch plugins that leverages GitHub API integration and AI-powered categorization to streamline the release process.  

### Problem Statement

The current release notes workflow creates an initial draft of consolidated release notes based on separate component release notes. However, several issues have been observed that require manual fixes:

* Inconsistencies in grammar and spacing
* Errors such as missing words, incorrect URLs, wrong formats, and miscategorized content
* Lack of release details sections for hardcoded content (e.g., OS/OSD core release notes, welcome text)
* Need for further tuning of grammar and semantics before creating a pull request
* Inconsistent formatting across components
* Inconsistent URL hyperlink formatting
* Manual review process via pull requests

### Objectives

* Fetch and process commit messages and associated pull request metadata
* Generate appropriate prompt for LLM for process the commit messages and generate release notes following guidelines mentioned in https://github.com/opensearch-project/opensearch-plugins/blob/main/RELEASE_NOTES.md. 
* Create release notes pull request against the component repo. 
* Standardize formatting across all component release notes
* Normalize URL hyperlink formatting
* Maintain semantic accuracy while improving readability
* Minimize manual review overhead

## Proposed Solution

An automated system consisting of two main components:

### 1. Commit Data Extractor 
- Fetches commits from GitHub repositories within specified date ranges
- Extracts associated pull request information and labels
- Outputs structured JSON data with commit messages, PR subjects, and labels

### 2. AI-Powered Release Notes Generator
- Uses AWS Bedrock Claude Sonnet 3.5 v2 to intelligently categorize commits
- Applies OpenSearch release notes formatting conventions
- Handles edge cases with an "Unknown" category for manual review

### Data Flow
1. **Input**: Component name/manifest file, date range, GitHub token
2. **Processing**: Fetch commits, extract PR metadata, categorize using AI
3. **Output**: Formatted markdown release notes following OpenSearch conventions

#### Sample Prompt
```
I need you to generate OpenSearch plugin release notes from commit data. Please follow the OpenSearch release notes format exactly.

**Plugin Information:**
- Plugin Name: {plugin_name}
- Version: {version}
- Repository URL: {repository_url}

**Commit Data:**
{commits_text}

**Instructions:**

1. **Label-based Categorization Logic:**
   - First, check if any labels match these categories (case-insensitive, partial matches allowed):
[...]

PR #5621, opened 2025-07-03
Add AI-powered release notes generator

RileyJergerAmazon:
### Description
- Add simple_changelog_processor.py to src/release_notes_workflow/
- Implements OpenSearch release notes generation using AI analysis
- Follows existing codebase patterns with proper imports and structure
- Uses TemporaryDirectory and GitRepository for safe operations
- Integrates with AWS Bedrock Claude for intelligent categorization
- Supports both API-based and git-based repository processing
- Includes comprehensive error handling and retry logic

### Issues Resolved
Closes Generate release notes using LLMs for OS and OSD components. ([#5614](https://github.com/opensearch-project/opensearch-build/issues/5614)).


rishabh6788:
Thank you for your contribution, a few high level comments before I go any further:
* Can you make it more modular? Try to follow https://github.com/opensearch-project/opensearch-build/blob/main/src/run_releasenotes_check.py structure. 
* instead of `max_repos` use `--components` arg. 
* Move the prompt to different constants file or class. 
* Break large functions into smaller targeted functions. 
* Better error handling
* Remove hard coded values. 
@RileyJergerAmazon @gaiksaya

gaiksaya:
Agree with @rishabh6788 
Instead of a bulky single script, lets try to make it [object oriented](https://en.wikipedia.org/wiki/Object-oriented_programming). Its easier to maintain, test (need to add it) and also extend in future. 
Would also recommend to use logging instead of `print` statements. 
Also I think we can take input manifest as the input and iterate over all the repos. This can be done at CI level where everything can run in parallel or here at the workflow level.

src/release_notes_workflow/simple_changelog_processor.py
```diff
@@ -0,0 +1,692 @@
...
+        contentType='application/json'
+    )
+    
+    result = json.loads(response['body'].read())
+    return result['content'][0]['text']
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Simple OpenSearch changelog processor")
+    parser.add_argument("--token", help="GitHub token")
```
gaiksaya: Search for GITHUB_TOKEN in the env variables rather than providing it as an input. Also it should not be needed for reading files. Checkout [release_notes](https://github.com/opensearch-project/opensearch-build/tree/7a62a342418a59b2da47da6d1909177001bf7502/src/release_notes_workflow) on existing approach.

rishabh6788:
As discussed, this is how your code should look like after restructuring. 

* Starts with `run_generate_change_log.py` which will have the `main()` method, something like:
```
 def main():
      """Main workflow orchestration following established patterns."""
      # Parse arguments
      args = ReleaseNotesArgs()

      # Configure logging
      console.configure(level=logging.INFO if args.verbose else logging.WARNING)

      # Load manifest
      manifest = InputManifest.from_file(args.manifest)

     # Start processing in a temp directory context that gets cleaned up after execution
      with TemporaryDirectory(keep=args.keep, chdir=True) as work_dir:

          # Inside the temp dir context loop over all components and filter if component args are provided
          for component in manifest.components.select(focus=components, platform=target.platform):
            logging.info(f"Building {component.name}")
            # Make the decision here whether you want to go CHANGELOG route or commit route
            See https://github.com/opensearch-project/opensearch-build/blob/main/src/run_build.py#L87C9-L103
     
...
...
...
```
* You can have a github client class to handle all git related calls, like:
```
class GitHubClient:
      """GitHub API client with rate limiting and error handling."""

      def __init__(self, token: str):
          self.token = token
          self.headers = {"Authorization": f"token {token}"}
          self.session = requests.Session()
          self.session.headers.update(self.headers)

      def get_changelog(self, repo_name: str) -> Optional[str]:
          # process change log code

      def get_commits_since_date(self, repo_name: str, since_date: str) -> List[Dict]:
         # pass

      def _make_request(self, url: str, params: dict = None) -> Optional[requests.Response]:
          pass
```

* Data handling and processing can have changelog processing method and commit processing method, like:
```
class DataProcessor():

      def process_changelog(self, repo_name: str, data: str) -> Dict[str, Any]:
          """Process changelog data."""
          pass

      def extract_commits(self, repo_name: str, baseline_date: str) -> List[Dict]:
          """Extract commits since baseline."""
          pass
```

* The main execution can happen inside `AIReleaseNotesGenerator` class:
```
class AIReleaseNotesGenerator():
      """AI-powered release notes generator using AWS Bedrock."""

      def __init__(<insert params>):
          self.bedrock_client = self._create_bedrock_client()
          self.github_client = GitHubClient(target.github_token)

      def generate_release_notes(self, repositories: List[str]) -> Dict[str, Any]:
          """Generate AI-powered release notes."""

      def process_repository(self, repo_name: str) -> Dict[str, Any]:
          """Process individual repository and make decision on changelog vs commit"""
          
```

@gaiksaya Feel free add your thoughts and comments.

gaiksaya:
We already have a git client here: https://github.com/opensearch-project/opensearch-build/tree/main/src/git for git related activities.
Check if we can add to it or extend it rather than reinventing for another use case. 
Also for releaseNotesGenerator mayeb add an option of AI to existing workflow https://github.com/opensearch-project/opensearch-build/blob/main/src/release_notes_workflow/release_notes.py (main class + args, etc)

rishabh6788:
> We already have a git client here: https://github.com/opensearch-project/opensearch-build/tree/main/src/git for git related activities. Check if we can add to it or extend it rather than reinventing for another use case. Also for releaseNotesGenerator mayeb add an option of AI to existing workflow https://github.com/opensearch-project/opensearch-build/blob/main/src/release_notes_workflow/release_notes.py (main class + args, etc)

I'm good with reusing the release notes class but not included towards git because the current implementation works in the context of a checked out git repo, not using github apis, we need metadata, such as associated pull request numbers, for a git commit which afaik `git log` will not give, nor there is any git command since pull request is github concept and not git.

src/run_releasenotes_check.py
```diff
@@ -103,8 +105,156 @@ def create_urls_file_if_not_exists(manifest_files: List[InputManifest]) -> None:
         check_if_exists_then_delete([table_filename, urls_filename])
         create_urls_file_if_not_exists(manifests)
         return 0
+    
+    elif args.action == "generate":
+        # Generate AI-powered release notes
+        check_if_exists_then_delete([table_filename, urls_filename])
+        
+        def get_baseline_date_from_github_tag(current_version: str) -> str:
```
rishabh6788: This whole logic is not required. Given we have a process to generate a releases after tag cut, just call `GET /repos/{owner}/{repo}/releases/latest` and get the `published_at` date and assume that is the baseline date for now.

src/run_releasenotes_check.py
```diff
@@ -103,8 +105,156 @@ def create_urls_file_if_not_exists(manifest_files: List[InputManifest]) -> None:
...
+                raise ValueError(f"Could not find tag date for version {baseline_version}. Please use --date to specify the start date.")
+                
+            except ValueError:
+                # Re-raise ValueError as-is
+                raise
+            except Exception as e:
+                raise ValueError(f"Failed to determine baseline date: {e}. Please use --date to specify the start date.")
+        
+        # Get baseline date - use provided date or get from GitHub tag
+        current_version = manifests[0].build.version
```
rishabh6788: nit: `current_release_version`

src/release_notes_workflow/release_notes_check_args.py
```diff
@@ -40,11 +41,50 @@ def __init__(self) -> None:
...
+        parser.add_argument(
+            "--baseline-date",
+            help="Baseline date for commit analysis (format: YYYY-MM-DD)"
+        )
+        parser.add_argument(
+            "--test-mode",
+            action="store_true",
+            help="Test mode: only read from GitHub, no write operations (branches, commits, PRs)"
+        )
+        parser.add_argument(
```
rishabh6788, line 46: This should be like below:
```
parser.add_argument(
            "-c",
            "--component",
            dest="components",
            nargs='*',
            type=str,
            help="Process one or more components."
        )
```

src/run_releasenotes_check.py
```diff
@@ -103,8 +105,156 @@ def create_urls_file_if_not_exists(manifest_files: List[InputManifest]) -> None:
...
+        
+        logging.info(f"Generating AI release notes for version {current_version} since {baseline_date}")
+        
+        # Create ReleaseNotes instance with baseline date
+        release_notes = ReleaseNotes(manifests, baseline_date, args.action)
+        
+        # Generate AI release notes for each component
+        for i, manifest in enumerate(manifests):
+            manifest_path = args.manifest[i].name if i < len(args.manifest) else None
+            for component in manifest.components.select():
```
rishabh6788: This should be  `for component in manifest.components.select(focus=components, platform='linux'):` where.

rishabh6788: components is `args.components`.

src/run_releasenotes_check.py
```diff
@@ -103,8 +105,156 @@ def create_urls_file_if_not_exists(manifest_files: List[InputManifest]) -> None:
...
+                if hasattr(component, "repository"):
+                    # Filter by component if specified
+                    if args.component and args.component.lower() not in component.name.lower():
```
rishabh6788: This can deleted after fixing the component parsing logic.

src/git/github_api_extension.py
```diff
@@ -0,0 +1,148 @@
...
+        """Get CHANGELOG.md content from GitHub API."""
+        if not self.github_token:
+            logging.warning("No GitHub token provided, cannot fetch changelog via API")
+            return None
+            
+        changelog_url = f"https://api.github.com/repos/opensearch-project/{self.repo_name}/contents/CHANGELOG.md"
+        
+        response = self._make_github_request(changelog_url)
+        if response and response.status_code == 200:
+            import base64
```
gaiksaya: nit: keep all the imports at the top

src/git/github_api_extension.py
```diff
@@ -0,0 +1,148 @@
...
+        if github_token:
+            self.session.headers.update(self.headers)
+        
+        # Extract repo name from URL
+        self.repo_name = url.rstrip('/').split('/')[-1].replace('.git', '')
+    
+    def get_changelog_from_github(self) -> Optional[str]:
+        """Get CHANGELOG.md content from GitHub API."""
+        if not self.github_token:
+            logging.warning("No GitHub token provided, cannot fetch changelog via API")
```
gaiksaya: Would the warning logger be displayed during normal run mode or would it be only for debug mode?

src/git/github_api_extension.py
```diff
@@ -0,0 +1,148 @@
...
+                
+            # Enrich commits with PR information
+            for commit in commits:
+                commit['pull_request'] = self._get_pull_request_for_commit(commit['sha'])
+                
+            all_commits.extend(commits)
+            page += 1
+            
+            # Rate limiting
+            time.sleep(0.1)
```
gaiksaya: Would this be enough?

src/git/github_api_extension.py
```diff
@@ -0,0 +1,148 @@
...
+            
+        commits_url = f"https://api.github.com/repos/opensearch-project/{self.repo_name}/commits"
+        
+        params = {
+            "since": since_date,
+            "per_page": 100
+        }
+        
+        if until_date:
+            params["until"] = until_date
```
gaiksaya: Is this required? until_date by default should be `now`

src/release_notes_workflow/ai_release_notes_generator.py
```diff
@@ -0,0 +1,499 @@
...
+            self.baseline_date = "2024-01-01"  # Default baseline
+        self.baseline_tag = "custom"
+    
+    def _create_bedrock_client(self):
+        """Create AWS Bedrock client."""
+        try:
+            return boto3.client('bedrock-runtime', region_name='us-east-1')
+        except Exception as e:
+            logging.warning(f"Failed to create Bedrock client: {e}")
+            return None
```
gaiksaya: We want to hard fail if client creation does not go through. Maybe change it to error instead of returning `None`?

src/release_notes_workflow/ai_release_notes_generator.py
```diff
@@ -0,0 +1,499 @@
...
+        # Calculate previous version (3.0.0 for 3.1.0)
+        version_parts = self.version.split('.')
+        if len(version_parts) >= 2:
+            major, minor = int(version_parts[0]), int(version_parts[1])
+            if minor > 0:
+                baseline_version = f"{major}.{minor-1}.0"
+            else:
+                baseline_version = f"{major-1}.0.0" if major > 0 else "2.0.0"
+        else:
+            baseline_version = "3.0.0"
```
gaiksaya: How did we decide on 3.0.0 as baseline? Also have we considered patch version?

src/release_notes_workflow/ai_release_notes_generator.py
```diff
@@ -0,0 +1,499 @@
...
+        pr_match = re.search(r'\(#(\d+)\)', commit_message)
+        if not pr_match:
+            return []
+        
+        pr_number = pr_match.group(1)
+        
+        # Scrape PR labels from GitHub page
+        try:
+            import requests
+            from bs4 import BeautifulSoup
```
gaiksaya: nit: imports at the top

src/release_notes_workflow/ai_release_notes_generator.py
```diff
@@ -0,0 +1,499 @@
...
+            repo.execute(f"git checkout -b {branch_name}")
+            
+            # Add and commit
+            repo.execute("git add release-notes/")
+            repo.execute(f'git commit -m "Add {self.version} release notes for {component_name}"')
+            
+            logging.info(f"Committed release notes on branch {branch_name}")
+            
+        except Exception as e:
+            logging.warning(f"Failed to commit release notes: {e}")
```
gaiksaya: @rishabh6788 Since we already have the token for retrieval wondering do we want to generate PR from here or let CI take care of it?

src/release_notes_workflow/release_notes_check_args.py
```diff
@@ -40,11 +41,50 @@ def __init__(self) -> None:
...
+            "--ai-model",
+            default="claude-3-7-sonnet",
+            help="AI model to use for analysis (default: claude-3-7-sonnet)"
+        )
+        parser.add_argument(
+            "--baseline-date",
+            help="Baseline date for commit analysis (format: YYYY-MM-DD)"
+        )
```
gaiksaya: The `--date` on line 35 can be the baseline. Check out the description. Should be same Right?

gaiksaya (commented):
Yet to do go through the code in depth but here are some overview pointers.

src/release_notes_workflow/ai_release_notes_generator.py
rishabh6788: Just realized that it is easier in Ci because everything runs in the context of build repo as final release notes PR is raised in the build repo, however, here we will be creating PRs for each component repo, which is getting checked out in the context of GitRepository python class and gets deleted as soon as processing is over, it is much easier to add pull request logic in python code than jenkins.

rishabh6788:
As discussed, please move the LLM related code in `llms` directory and move all the prompts to a constant `prompts.py` file in the llms repo. 
Also, simplify the generate method in ReleaseNotes class as per below:

```
def generate(self, component: InputComponentFromSource, build_version: str, build_qualifier: str):
        with TemporaryDirectory(chdir=True) as work_dir:
            with GitRepository(
                    component.repository,
                    component.ref,
                    os.path.join(work_dir.name, component.name),
                    component.working_directory
            ) as repo:
                
                release_notes = ReleaseNotesComponents.from_component(component, build_version, build_qualifier, repo.dir)

                if os.path.isfile(os.path.join(repo.dir, 'CHANGELOG.md')):
                    # Do all the Changelog related LLM stuff here, add file in repos release-notes folder
                else:
                    # get the git commit messages and fo the LLM stuff here, add file in repo release-notes folder
               create_pr()
```

src/git/git_repository.py
```diff
@@ -23,12 +23,13 @@ class GitRepository:
...
 
-    def __init__(self, url: str, ref: str, directory: str = None, working_subdirectory: str = None) -> None:
+    def __init__(self, url: str, ref: str, directory: str = None, working_subdirectory: str = None, fetch_depth: int = 1) -> None:
         self.url = url
         self.ref = ref
+        self.fetch_depth = fetch_depth
         if directory is None:
             self.temp_dir = TemporaryDirectory()
-            self.dir = os.path.realpath(self.temp_dir.name)
+            self.dir = os.path.abspath(self.temp_dir.name)
```
peterzhuamazon: Reason for the change right here?

src/git/github_api_extension.py
```diff
@@ -0,0 +1,148 @@
...
+        self.session = requests.Session()
+        if github_token:
+            self.session.headers.update(self.headers)
+        
+        # Extract repo name from URL
+        self.repo_name = url.rstrip('/').split('/')[-1].replace('.git', '')
+    
+    def get_changelog_from_github(self) -> Optional[str]:
+        """Get CHANGELOG.md content from GitHub API."""
+        if not self.github_token:
```
peterzhuamazon: What is the alternative path if github token not found and you cant use github api to retrieve? Also cant we simply use http to get the raw content of changelog.md? Why api?

src/git/github_api_extension.py
peterzhuamazon: I really dont like using sleep here to process.
But if we really need this, probably sleep 1 instead of 0.1.

peterzhuamazon (commented):
Hi @RileyJergerAmazon @rishabh6788 , This is a huge PR so can we please get some test cases?

Also would appreciate if the README is updated in both root and subfolder of workflows.

Thanks!

src/release_notes_workflow/release_notes.py
```diff
@@ -82,3 +88,56 @@ def check(self, component: InputComponentFromSource, build_version: str, build_q
...
+        """Generate AI-powered release notes for a component."""
+        with TemporaryDirectory(chdir=True) as work_dir:
+            # Initialize processor with initial date
+            processor = Processor(build_version, self.date)
+            
+            # First, get the date of the last tag in the opensearch-build repository
+            baseline_date = self.date
+            try:
+                # Check out the opensearch-build repository to get the last tag date
+                with GitRepository(
```
rishabh6788: we don't need this, like I mentioned, just call the releases api and it should give you latest release details without checking out the repo.

src/release_notes_workflow/release_notes.py
```diff
@@ -82,3 +88,56 @@ def check(self, component: InputComponentFromSource, build_version: str, build_q
                 else:
                     results.append(None)
         return results
+
+    def generate(self, component: InputComponentFromSource, build_version: str, build_qualifier: str, manifest_path: str = None) -> None:
+        """Generate AI-powered release notes for a component."""
+        with TemporaryDirectory(chdir=True) as work_dir:
+            # Initialize processor with initial date
+            processor = Processor(build_version, self.date)
```
rishabh6788: Don't need to re-initialize for each component, please move this in the `__init__` method

src/release_notes_workflow/release_notes.py
```diff
@@ -82,3 +88,56 @@ def check(self, component: InputComponentFromSource, build_version: str, build_q
...
+                    github_token=None,
+                    version=build_version,
+                    baseline_date=baseline_date,
+                    test_mode=self.test_mode
+                )
+
+                release_notes = ReleaseNotesComponents.from_component(component, build_version, build_qualifier, repo.dir)
+                
+                # Try to fetch CHANGELOG.md directly from GitHub
+                content = processor.fetch_changelog_from_github(component)
```
rishabh6788: This can be further simplified to below: 
`changelog_exist = os.path.isfile(os.path.join(repo.dir, 'CHANGELOG.md'))`.

process.py
```diff
@@ -0,0 +1,363 @@
+#!/usr/bin/env python
```
rishabh6788: As discussed, remove this file from here and move it to git dir and rename it to GithubCommitsProcessor.

src/llms/ai_release_notes_generator.py
```diff
@@ -0,0 +1,186 @@
...
+    def _create_bedrock_client(self):
+        """Create AWS Bedrock client."""
+        try:
+            return boto3.client('bedrock-runtime', region_name='us-east-1')
+        except Exception as e:
+            logging.error(f"Failed to create Bedrock client: {e}")
+            logging.warning("Continuing without Bedrock client - will use mock responses for testing")
+            return None
+    
+    def process(self, content: str, component_name: str, manifest_path: str = None) -> Dict[str, Any]:
```
rishabh6788: as discussed, for now just take in the prompt and return response. Remove all other logic from this file

src/release_notes_workflow/release_notes.py
```diff
@@ -82,3 +89,46 @@ def check(self, component: InputComponentFromSource, build_version: str, build_q
...
+                changelog_path = os.path.join(repo.dir, 'CHANGELOG.md')
+                if changelog_exist:
+                    with open(changelog_path, 'r') as f:
+                        content = f.read()
+                    #processed_data = processor.process(content, component.name)
+                    prompt = AI_RELEASE_NOTES_PROMPT_CHANGELOG.format(
+                        repo_name=component.name,
+                        version=build_version,
+                        repository_url=component.repository
+                    )
```
rishabh6788, line 123: refactor code here to call the llm and get the response back.

src/git/git_commit_processor.py
```diff
@@ -0,0 +1,255 @@
...
+        self.component = component
+ 
+    def _make_request(self, url: str, params: Dict = None) -> Optional[Dict]:
+        """Make a GET request to GitHub API"""
+        try:
+            response = requests.get(url, headers=self.headers, params=params)
+            response.raise_for_status()
+            return response.json()
+        except requests.exceptions.RequestException as e:
+            print(f"Error making request to {url}: {e}")
```
gaiksaya: Please use a logger instead of print statement. 
Thanks!

rishabh6788 (approved):

PR #5778, opened 2025-10-14
Support module split and generate individual coverage reports in gradle check task

Divyaasm:
### Description
1. Support server module and other modules split from the gradle check tasks
2. Archive all the coverage reports generated during the run.
3. Gradle script changes to support the feature

### Issues Resolved
Addresses build side changes for the meta issue https://github.com/opensearch-project/OpenSearch/issues/19378

### DIFF
--- a/jenkins/gradle/gradle-check.jenkinsfile
+++ b/jenkins/gradle/gradle-check.jenkinsfile
@@ -158,7 +158,7 @@ pipeline {
                 }
                 always {
                     sh '''
-                        mkdir -p codeCoverage
+                        rm -rf codeCoverage && mkdir -p codeCoverage
 
                         for file in $(find search/server/build/reports/jacoco/ -name '*.xml'); do
                             cp -v "$file" codeCoverage/
--- a/jenkins/gradle/gradle-check.jenkinsfile
+++ b/jenkins/gradle/gradle-check.jenkinsfile
@@ -143,7 +143,8 @@ pipeline {
                             runGradleCheck(
                                 gitRepoUrl: "${GIT_REPO_URL}",
                                 gitReference: "${GIT_REFERENCE}",
-                                bwcCheckoutAlign: "${bwc_checkout_align}"
+                                bwcCheckoutAlign: "${bwc_checkout_align}",
+                                scope: "all"
                             )
                         }
 
--- a/jenkins/gradle/gradle-check.jenkinsfile
+++ b/jenkins/gradle/gradle-check.jenkinsfile
@@ -158,7 +158,7 @@ pipeline {
                 }
                 always {
                     sh '''
-                        rm -rf codeCoverage && mkdir -p codeCoverage/
+                        rm -rf codeCoverage && mkdir -p codeCoverage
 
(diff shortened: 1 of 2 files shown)


jenkins/gradle/gradle-check.jenkinsfile
```diff
@@ -154,9 +156,15 @@ pipeline {
...
+                    sh '''
+                    mkdir codeCoverage
+
+                    for file in $(find search/server/build/reports/jacoco/ -name '*.xml'); do
+                        cp -v "$file" codeCoverage/
+                    done
+                    '''
                     junit allowEmptyResults: true, testResults: '**/build/test-results/**/*.xml'
-                    archiveArtifacts artifacts: 'codeCoverage.xml', onlyIfSuccessful: true
+                    archiveArtifacts artifacts: 'codeCoverage/**', onlyIfSuccessful: true
```
gaiksaya, line 168: Is the value of `onlyIfSuccessful` right? Seems confusing when it is placed under `always` block but only archiveArtifacts if successful.

Divyaasm: Yeah Sayali. It's valid, only archives if the build is successful. The files aren't even created if the build fails.

gaiksaya: Isn't that what @bowenlan-amzn and we were discussing. That failure also needs to generate codecoverage? https://github.com/opensearch-project/opensearch-build-libraries/pull/744

Divyaasm: That is a different problem Sayali. Will need to expliticitly generate reports from the .exec files using jacococli or a gradle command in failed block . 
This PR handles the successful condition and generates all the required reports independently to generate coverage data for unit and integration tests.

gaiksaya: In that case should this entire block be placed under `success` block of jenkins instead of `always`?

Divyaasm: We would anyway create the reports with same names. So once I add the failure block I will remove onlyIfSuccessful condition. Archiving will be applicable to every build status in that case.

gaiksaya: Sure sounds good. I believe we are yet to add `module_name` arg in build-lib side?

Divyaasm: Yes, just a small change will send the PR.

jenkins/gradle/gradle-check.jenkinsfile
```diff
@@ -154,9 +157,15 @@ pipeline {
                     archiveArtifacts artifacts: '**/build/heapdump/*.hprof', allowEmptyArchive: true
                 }
                 always {
-                    sh ("cp -v `find search/build/reports/jacoco/ -name '*.xml' | head -n 1` codeCoverage.xml || echo")
+                    sh '''
+                        mkdir codeCoverage
```
peterzhuamazon:
```suggestion
                        mkdir -p codeCoverage
```

Issue #5815, opened 2025-10-29
[Bug]: Issues with release notes generation workflow


gaiksaya:
### Describe the bug

- [x] Base branch for the pull requests needs to be what is provided in the ref
- [ ] DCO failures
- [x] Skip plugin version increment PRs
- [ ] Remove changelog after creating PR

### To reproduce

See https://build.ci.opensearch.org/view/Release/job/release-notes-generate/48/console

### Relevant log output

```shell

```

PR #5822, opened 2025-10-30
Format LLM prompt to ignore version increments

gaiksaya:
### Description
Adding version increment PRs in the release notes do not provide any value. This change asks the LLM to specifically ignore the PR with automated title.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5815

### DIFF
--- a/src/llms/prompts.py
+++ b/src/llms/prompts.py
@@ -75,8 +75,10 @@ AI_RELEASE_NOTES_PROMPT_COMMIT = """I need you to generate OpenSearch component 
     methods non-static"
    - Do not lose any commit information, even if it doesn't match any category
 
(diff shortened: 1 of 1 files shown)


src/llms/prompts.py
```diff
@@ -76,6 +76,7 @@
    - Do not lose any commit information, even if it doesn't match any category
 
 3. **Entry Format:**
+   - Ignore the pull request that has title as "[AUTO] Increment version to".
```
rishabh6788: Add a filter messages section and then put this, will be easy to maintain all future filtering requirements.

gaiksaya: Makes sense. Made that change

Issue #5832, opened 2025-11-03
[Bug]: k-NN repo name is incorrect in auto-generated consolidated release notes


gaiksaya:
### Describe the bug

See https://github.com/opensearch-project/opensearch-build/pull/5829/files#diff-66dcf52c1265e58accadd2e4dcc9078276c082ca499397fd5915418e9009d8e2R26

The consolidated release notes were auto-generated by https://build.ci.opensearch.org/job/release-notes-tracker/373/console

### To reproduce

Run `./release_notes.sh compile manifests/2.19.4/opensearch-2.19.4.yml manifests/2.19.4/opensearch-dashboards-2.19.4.yml --output release-notes/opensearch-release-notes-2.19.4.md` and se that k-NN is written as k NN

### Expected behavior

k-NN should be correctly written as k-NN

### Additional context

AI is not used to compile consolidated release notes.

### Relevant log output

```shell

```

PR #5833, opened 2025-11-03
Handle k-NN naming and fix OpenSearch naming

gaiksaya:
### Description
Handle k-NN naming and fix OpenSearch naming

### Issues Resolved
closes https://github.com/opensearch-project/opensearch-build/issues/5832

### DIFF
--- a/src/run_releasenotes_check.py
+++ b/src/run_releasenotes_check.py
@@ -78,7 +78,7 @@ def main() -> int:
 
         if component_name == "opensearch-sql":
             component_name = "SQL"
-        elif component_name == "k-NN" or component_name.lower() == "k-nn":
+        elif component_name.lower() == "k-nn":
             return "OpenSearch k-NN"
 
         formatted_name = " ".join(word.capitalize() for word in re.split(r"[-.]", component_name))


src/run_releasenotes_check.py
```diff
@@ -73,8 +75,12 @@ def format_component_name_from_url(url: str) -> str:
         if end_index == -1:
             raise ValueError("'.release-notes' not found after 'release-notes/'")
         component_name = url[start_index + len("release-notes/"): end_index]
+
         if component_name == "opensearch-sql":
             component_name = "SQL"
+        elif component_name == "k-NN" or component_name.lower() == "k-nn":
```
peterzhuamazon: Why do we need another component_name == when we already doing component_name.lower()?

peterzhuamazon (approved):

src/run_releasenotes_check.py
gaiksaya: Just to handle cases where component name might have different capitalization. Just kept the lower one as any combination will be converted to lower cases.

rishabh6788 (approved):

PR #5903, opened 2025-12-26
Add `calcite_enabled` parameter to control histogram queries in big5 PPL benchmarks

aalva500-prog:
### Purpose

This PR is to enable baseline performance data for OpenSearch 3.2.0 with Calcite disabled by adding the `calcite_enabled` parameter to control histogram query execution in PPL test procedures.

### Problem

We need baseline performance data for version 3.2.0 with Calcite disabled to compare against Calcite-enabled performance. However, the big5 PPL histogram queries (date_histogram, range-auto-date-histo, etc.) cause failures when Calcite is disabled, preventing the baseline job from completing successfully.

### Solution

Add `"calcite_enabled"` parameter to WORKLOAD_PARAMS:
- Set to `true` for Calcite-enabled jobs → histogram queries are included
- Set to `false` for Calcite-disabled baseline → histogram queries are skipped

**Note:** This PR works in conjunction with opensearch-benchmark-workloads PR: [opensearch-benchmark-workloads#732](https://github.com/opensearch-project/opensearch-benchmark-workloads/pull/732), which implements the workload-side logic to conditionally run histogram queries based on the `calcite_enabled` parameter.

### Changes

1. **Added `"calcite_enabled":true`** to all existing Calcite-enabled PPL jobs (versions 3.2.0, 3.3.2, 3.4.0, 3.5.0)

2. **Added new Calcite-disabled baseline job** for version 3.2.0:
   ```
   WORKLOAD_PARAMS={...,"calcite_enabled":false}
   ADDITIONAL_CONFIG=plugins.calcite.enabled:false
   ```

3. **Added documentation** explaining when to include `calcite_enabled`:true for future Calcite PPL jobs

## Summary by CodeRabbit

* **New Features**
  * Added a toggle to enable/disable Calcite behavior in PPL benchmark runs to include or skip histogram queries.
* **Chores**
  * Expanded benchmark job coverage for versions 3.2.0, 3.3.2, 3.4.0, 3.5.0 across x64 and arm64.
  * Added Calcite-disabled baseline jobs to support A/B performance comparisons.
* **Documentation**
  * Included guidance for configuring Calcite-enabled benchmark runs.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


tests/jenkins/jenkinsjob-regression-files/opensearch/benchmark-test-distribution-url.jenkinsfile.txt
```diff
@@ -95,10 +95,19 @@
...
+        # ===== PPL Calcite-Enabled Benchmark Jobs =====
+        # IMPORTANT: When adding new OpenSearch versions to Calcite PPL jobs below, you MUST include "calcite_enabled":true in WORKLOAD_PARAMS
+        # This parameter controls whether histogram operations (date_histogram, range-auto-date-histo, etc.) are included in the PPL test procedure
+        # Both settings are required for proper Calcite testing:
+        #   - ADDITIONAL_CONFIG=plugins.calcite.enabled:true (enables Calcite in OpenSearch cluster)
+        #   - "calcite_enabled":true in WORKLOAD_PARAMS (enables histogram queries in workload)
```
dai-chen, line 103: As I understand, this new param must be the same value as `plugins.calcite.enabled` in `ADDITIONAL_CONFIG`. Do we have access to the latter in workload file in https://github.com/opensearch-project/opensearch-benchmark-workloads/pull/732? If possible, this PR is not needed?

dai-chen: Alternatively, is more than one workload file allowed? If so, we can just separate histogram test workload out without this param?

OVI3D0: @dai-chen yes, we can always separate the histogram enabled test into a separate test procedure. Workloads are evaluated before runtime so I don't think there's any practical way to access the additional config from the workload file.

aalva500-prog: Hi @dai-che, @OVI3D0, thank you for your inputs. In that case, I will close this PR for now. I will test the approach of adding a different workload or test procedure to run histogram queries only when Calcite is enabled.

PR #5920, opened 2026-01-16
Fix native plugin validation faliure from 3.5.0 version

Divyaasm:
### Description
Checks for identity-shiro plugin only for versions less than 3.5 which recent [changes](https://github.com/opensearch-project/OpenSearch/pull/20305) in core to remove if from plugin directory

### Issues Resolved
Resolves nighlty build failure https://github.com/opensearch-project/opensearch-devops/actions/runs/21051266651/job/60537883007

### DIFF
--- a/src/validation_workflow/validation.py
+++ b/src/validation_workflow/validation.py
@@ -91,8 +91,7 @@ class Validation(ABC):
             plugin_list.remove("examples")
             plugin_list.remove("build.gradle")
             if semver.compare(self.args.version, "3.5.0") < 0:
-                # Since the security plugin is enabled in the artifacts and identity-shiro is also an identity plugin, we cannot have both the plugins installed together.
-                plugin_list.remove("identity-shiro")
+                plugin_list.remove("identity-shiro")  # Since the security plugin is enabled in the artifacts and identity-shiro is also an identity plugin, we cannot have both the plugins installed together. # noqa: E501
             return plugin_list
         except Exception:
             logging.exception(


peterzhuamazon:
Hi @Divyaasm ,

I am not getting this, are we hardcoding the name of the plugins?
A better way would be get the list from the commit id in the manifest.yml in the bundle. Thanks.

Divyaasm:
> Hi @Divyaasm ,
> 
> I am not getting this, are we hardcoding the name of the plugins? A better way would be get the list from the commit id in the manifest.yml in the bundle. Thanks.

Yes, We are getting plugin names from manifest commit id. From that list, we ignore only examples folder and identity-shiro(can't be installed along with security plugin) which were hard coded in for all the versions.

gaiksaya:
> Hi @Divyaasm ,
> 
> I am not getting this, are we hardcoding the name of the plugins? A better way would be get the list from the commit id in the manifest.yml in the bundle. Thanks.

I agree. Repos are dynamic source. If we parse the manifest and the get from there, it would handle the per version plugin list efficiently as well.

src/validation_workflow/validation.py
```diff
@@ -78,19 +79,25 @@ def install_native_plugin(self, path: str, installed_plugins_list: list) -> None
...
+            commit_id = bundle_manifest.components["OpenSearch"].commit_id
+            plugin_url = f"https://api.github.com/repos/opensearch-project/OpenSearch/contents/plugins?ref={commit_id}"
+            api_response = requests.get(plugin_url)
+            if api_response.status_code != 200:
+                raise Exception("Github Api returned error code while retrieving the list of native plugins")
             response = api_response.json()
             plugin_list = [i["name"] for i in response if i["name"] not in installed_plugins_list]
             plugin_list.remove("examples")
             plugin_list.remove("build.gradle")
-            plugin_list.remove("identity-shiro")  # Since the security plugin is enabled in the artifacts and identity-shiro is also an identity plugin, we cannot have both the plugins installed together. # noqa: E501
```
gaiksaya, line 90: If we were already removing this plugin, what changed?

tests/tests_validation_workflow/test_validation.py
```diff
@@ -207,6 +212,7 @@ def test_get_native_plugin_list(self, mock_temporary_directory: Mock, mock_valid
         mock_response.status_code = 200
         mock_get.return_value = mock_response
         mock_temporary_directory.return_value.path = "/tmp/trytytyuit/"
+        mock_validation_args.return_value.version = "3.0.0"
```
gaiksaya: Can you please also add a test for lower version?  Example, 2.19.4?

src/validation_workflow/validation.py
Divyaasm: this commit removed it from main branch. So upcoming versions don't have it any more. So added condition

src/validation_workflow/validation.py
```diff
@@ -16,6 +16,7 @@
 from typing import Any
 
 import requests
+from packaging.version import Version
```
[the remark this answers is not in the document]
gaiksaya: Use semver instead. https://github.com/opensearch-project/opensearch-build/blob/25740e9f9e1da93d8c600eff8dd20b613ebb94f0/src/test_workflow/perf_test/perf_test_cluster.py#L114

src/validation_workflow/validation.py
```diff
@@ -78,19 +79,25 @@ def install_native_plugin(self, path: str, installed_plugins_list: list) -> None
...
+            api_response = requests.get(plugin_url)
+            if api_response.status_code != 200:
+                raise Exception("Github Api returned error code while retrieving the list of native plugins")
             response = api_response.json()
             plugin_list = [i["name"] for i in response if i["name"] not in installed_plugins_list]
             plugin_list.remove("examples")
             plugin_list.remove("build.gradle")
-            plugin_list.remove("identity-shiro")  # Since the security plugin is enabled in the artifacts and identity-shiro is also an identity plugin, we cannot have both the plugins installed together. # noqa: E501
+            if semver.compare(self.args.version, "3.5.0") < 0:
+                plugin_list.remove("identity-shiro")  # Since the security plugin is enabled in the artifacts and identity-shiro is also an identity plugin, we cannot have both the plugins installed together. # noqa: E501
```
gaiksaya, line 94:
```suggestion
# Since the security plugin is enabled in the artifacts and identity-shiro is also an identity plugin, we cannot have both the plugins installed together.
                plugin_list.remove("identity-shiro")
```

src/validation_workflow/validation.py
```diff
@@ -78,19 +79,25 @@ def install_native_plugin(self, path: str, installed_plugins_list: list) -> None
...
             return plugin_list
-        else:
-            raise Exception("Github Api returned error code while retrieving the list of native plugins")
+        except Exception:
+            logging.exception(
+                "Couldn't fetch native plugin list from Opensearch repository"
```
gaiksaya: Since we already have an exception for GitHub API above, this captures a more generic issue that can be caused by anything.
```suggestion
                "Error while validating native plugin list."
```

src/validation_workflow/validation.py
```diff
@@ -78,19 +79,25 @@ def install_native_plugin(self, path: str, installed_plugins_list: list) -> None
...
+            plugin_url = f"https://api.github.com/repos/opensearch-project/OpenSearch/contents/plugins?ref={commit_id}"
+            api_response = requests.get(plugin_url)
+            if api_response.status_code != 200:
+                raise Exception("Github Api returned error code while retrieving the list of native plugins")
             response = api_response.json()
             plugin_list = [i["name"] for i in response if i["name"] not in installed_plugins_list]
             plugin_list.remove("examples")
             plugin_list.remove("build.gradle")
-            plugin_list.remove("identity-shiro")  # Since the security plugin is enabled in the artifacts and identity-shiro is also an identity plugin, we cannot have both the plugins installed together. # noqa: E501
+            if semver.compare(self.args.version, "3.5.0") < 0:
```
rishabh6788, line 93: are we not recording native plugins list in any manifest?

gaiksaya: We are recording in build manifest but I believe dynamically "referring to git" approach was chosen so that we do due diligence in case some plugin even fails to build. For other plugins there are part of input manifest but that is not the case for native plugins.

Issue #5912, opened 2026-01-09
[Proposal] Use Agentic AI to help with the backlog of flaky tests in the core repo


cwperks:
### Summary

The idea for this issue was initially discussed here: https://github.com/opensearch-project/OpenSearch/issues/18108#issuecomment-3724122586

Flaky tests continue to be a major blocker for improving automation and developer velocity in the OpenSearch project. They prevent us from safely enabling workflows such as automated version bumps, dependency upgrades, and more aggressive CI automation.

With the flaky test dashboard (and auto-filed GH issues), we have a way to identify which tests are flaky — but we still rely on manual triage and debugging, which does not scale.

This issue proposes introducing an **agentic workflow** (e.g., using AI agents capable of executing tests) to automatically investigate flaky tests, summarize findings, and in some cases even propose fixes via PRs.

### Motivation

Modern coding agents (e.g., ChatGPT Codex-style agents) can:

* Run tests repeatedly with different seeds and environments
* Detect patterns across failures
* Differentiate between:

  1. True test flakiness
  2. Resource-constraint failures
  3. Failures tied to specific random seeds

Instead of only reporting that a test is flaky, we could automatically attempt to **diagnose and resolve** the root cause.

### Proposed Workflow

When the flaky test dashboard opens a GitHub issue, the issue body should include a standardized **agent instruction prompt**, enabling automation to pick up and process the issue.

#### Suggested Agent Tasks

1. **Reproduce the failure**

   * Run the test with multiple iterations (e.g., N=10, N=100)
   * Capture failing seeds and logs

2. **Classify the failure**

   * Random seed–specific
   * Intermittent timing / race condition
   * Resource or environment sensitivity

3. **Post analysis to the issue**

   * Summary of findings
   * Whether the issue is reproducible and under what conditions

4. **Attempt remediation**

   * Modify test or code if root cause is identifiable

5. **Open a PR if successful**

   * Validate that flakiness is reduced or eliminated
   * Reference the flaky test issue

To avoid noise, agents should **only comment if reproduction succeeds or actionable conclusions are found**.

### Benefits

* Reduces manual debugging workload
* Accelerates stabilization of CI
* Provides consistent, structured diagnostics instead of ad-hoc comments

### Open Questions

* Where should agents run? (self-hosted CI vs managed infra)
* What resource limits should be applied?
* Which test suites should be eligible initially?
* Security and PR approval workflow

### Next Steps

* Define a minimal pilot for a subset of flaky tests
[...]

PR #5958, opened 2026-02-02
Add code diff analyzer github action to be re-used by others

peterzhuamazon:
### Description
Add code diff analyzer github action to be re-used by others
Example: https://github.com/peterzhuamazon/OpenSearch/pull/2#issuecomment-3844667108
Pre-requisite to https://github.com/opensearch-project/OpenSearch/pull/20504

* Note: this is early stage and only review for malicious code. We can extend it to also review on code enhancement just like other AI review tools, through bedrock.

### Issues Resolved

### DIFF
--- a/.github/workflows/code-diff-analyzer.yml
+++ b/.github/workflows/code-diff-analyzer.yml
@@ -14,7 +14,19 @@ on:
         required: false
         type: string
         default: 'skip-diff-analyzer'
-      update_pr_comment_with_analyzer_report:  # Only take if event is pull_request_target
+      # all = every mode
+      # report = analyze diff to find malicious code
+      # TODO: review = analyze diff to suggest code improvement
+      diff_analyzer_mode:
+        required: false
+        type: string
+        default: 'report'
+      update_pr_comment_with_analyzer_report:  # Only update if event is pull_request_target
+        required: false
+        type: boolean
+        default: false
+      # TODO: Implement review
+      update_pr_conversation_with_analyzer_review:  # Only update if event is pull_request_target
         required: false
         type: boolean
         default: false


.github/workflows/code-diff-analyzer.yml
```diff
@@ -0,0 +1,317 @@
...
+        required: true
+    inputs:
+      skip_validation_on_push:
+        required: false
+        type: boolean
+        default: true
+      skip_validation_label_name:
+        required: false
+        type: string
+        default: 'skip-diff-validation'
```
jainankitk: I am wondering if `skip-diff-analyzer` is better name, given the workflow is called `code-diff-analyzer`? `skip-code-diff-analyzer` is probably bit long

peterzhuamazon: There are still a lot of inconsistent namings across the files as it was initially only setup as a validator, will update.

.github/workflows/code-diff-analyzer.yml
```diff
@@ -0,0 +1,317 @@
...
+              echo "Diff validation skipped due to inputs.skip_validation_on_push is set to 'true'"
+              echo "diff_validation=9" >> $GITHUB_ENV
+            fi
+          elif [ "${{ github.event_name }}" = "pull_request_target" ]; then
+            base_sha=${{ github.event.pull_request.base.sha }}
+            head_sha=${{ github.event.pull_request.head.sha }}
+            PR_LABELS=$(curl -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+              -H "Accept: application/vnd.github.v3+json" \
+              "https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.number }}/labels" \
+              | jq -r '.[].name')
```
cwperks, line 90: Are these not already provided as part of the github event?

.github/workflows/code-diff-analyzer.yml
```diff
@@ -0,0 +1,317 @@
...
+            echo "wrong github event: ${{ github.event_name }}, must be 'push' or 'pull_request_target'"
+            echo "diff_validation=5" >> $GITHUB_ENV
+            exit 1
+          fi
+
+          echo "BASE_SHA=$base_sha" >> $GITHUB_ENV
+          echo "HEAD_SHA=$head_sha" >> $GITHUB_ENV
+
+      - name: Get diff details
+        if: ${{ env.diff_validation != '5' && env.diff_validation != '9' }}
```
cwperks: What is the significance of 5 and 9? I can see we set them above in higher steps. wdyt about making these human readable rather than numbers?

cwperks (approved):
Thank you for this PR @peterzhuamazon! Approving as the changes look good to me, just had a couple of questions.

I think this will be a novel way to improve the experience for first-time contributors. Most of our repos are configured not to allow CI to run automatically for first-time contributors until a maintainer has reviewed the code and manually allows the CI to be run.

While that is a reasonable setting (IMO), I have seen varying degrees of responsiveness across the repos and in many instances first-time contributors have to wait a long time prior to receiving feedback from CI checks because it takes a long time for a maintainer to approve the checks to run.

cwperks:
This also seems like its in the direction of elevating Claude to maintainer status. i.e. this is currently a decision that a maintainer opts to take (run CI checks for first time contributors) and we are now delegating to Claude. Obviously, Claude is still not accountable as a full maintainer but this moves in that direction and at the rate AI is progressing it could get there quickly.

.github/workflows/code-diff-analyzer.yml
peterzhuamazon: The reason I use numbers is because 0 is pass, everything above 0 (1/2/3/4/5) is error, 9 is skipped.
This is modeling the os runlevel so I think it is fine for the time being.
As long as the input var name is unchanged, internal logic can be update later if needed.

peterzhuamazon:
Yes @cwperks, just to add a few things here:
1. I will update the var names to be more aligned with the purpose of this action, which is also suggested by @jainankitk 
2. We are not specifically limited to claude as this is just a first draft. It is just easier to use claude to connect to bedrock at the moment. In my recent search online seems like some other models perform better in `one-shot` task than claude, whereas claude relies a bit more on context history
3. This action is flexible to be added into any existing repo and actions if needed. Of course currently it only support `pull_request_target` if it is a PR since we really need the secret to assume roles to use bedrock. Let's test this through on OS core and improve it overtime.

Thanks.

.github/workflows/code-diff-analyzer.yml
peterzhuamazon: Using event is not reliable. For example, we dont want to use `labeled` because it will trigger gradle check every time when a label being added. If we dont, then `${{ github.event }}` never update and get cached until the next new commit being pushed, which means re-run the same workflow will not actually update the label list.

We want the flow to be:
1. Action failed
2. Maintainer review code
3. Maintainer think it is valid to skip, add label
4. Maintainer re-run failed action
5. Action detect new label being added
6. Action skip validation, straight to gradle check

Thanks.

peterzhuamazon:
@coderabbitai review

Issue #5979, opened 2026-02-10
[3.6.0] Build min and default distribution with -Pcrypto.standard=FIPS-140-3 for 3.6.0 release


cwperks:
Opening this issue after discussion on https://github.com/opensearch-project/technical-steering/issues/77. 

### Summary

For the **OpenSearch 3.6.0 release**, we should build both the **default distribution** and the **minimal distribution** with:

```
-Pcrypto.standard=FIPS-140-3
```

as part of the official release artifacts.

This issue is to track the work required to make this change, validate the resulting distributions, and resolve any build or runtime issues that arise from producing FIPS-140-3–configured artifacts by default for 3.6.0.

### Background

There has been ongoing discussion in the TSC about OpenSearch’s FIPS posture and the long-term direction of FIPS support in the default distribution:
[https://github.com/opensearch-project/technical-steering/issues/77](https://github.com/opensearch-project/technical-steering/issues/77)

Today, OpenSearch distributions are typically built in a non-FIPS configuration by default, with FIPS-enabled artifacts produced only when explicitly building with `-Pcrypto.standard=FIPS-140-3` (e.g., `./gradlew localDistro -Pcrypto.standard=FIPS-140-3`).

For 3.6.0, we propose that the **release artifacts themselves** (both default and minimal) be built with `-Pcrypto.standard=FIPS-140-3`.

### Describe the solution you'd like

Always build with `-Pcrypto.standard=FIPS-140-3`

### Describe alternatives you've considered

Status quo

peterzhuamazon:
Hi @cwperks, regarding all OS plugins. Do they also need to update their build params?

Thanks.

cwperks:
> Hi [@cwperks](https://github.com/cwperks), regarding all OS plugins. Do they also need to update their build params?
> 
> Thanks.

Some plugins will need to become aware of the build param, yes. 

One of the key differences with this build param is that when `-Pcrypto.standard=FIPS-140-3` is provided then core will include `bcFips` dependencies under the `lib/` folder. If the build param is absent then they are not there.

So for plugins that rely on these same dependencies:

1. If `-Pcrypto.standard=FIPS-140-3` is _present_ then treat bcFips as a **provided** dependency (i.e. treat as _compileOnly_ because its available at runtime)
2. If `-Pcrypto.standard=FIPS-140-3` is _absent_ then bcFips deps are **not provided**. (i.e. treat as _implementation_ because the plugin will need to include these in its assembly)

[...]

PR #5995, opened 2026-02-18
Toggle -Pcrypto.standard=FIPS-140-3 by default on all build scripts (Build)

peterzhuamazon:
### Description
Toggle -Pcrypto.standard=FIPS-140-3 by default on all build scripts (Build)

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5979

### DIFF
--- a/scripts/components/OpenSearch-DataFusion/build.sh
+++ b/scripts/components/OpenSearch-DataFusion/build.sh
@@ -75,11 +75,11 @@ mkdir -p $OUTPUT/maven/org/opensearch
 
 # Build project and publish to maven local.
 echo "Building and publishing OpenSearch project to Maven Local"
-./gradlew publishToMavenLocal -Dbuild.snapshot=$SNAPSHOT -Dbuild.version_qualifier=$QUALIFIER -PrustRelease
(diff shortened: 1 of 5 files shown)


gaiksaya:
@peterzhuamazon Same scripts will be used for 2.19.x patch releases as well. And I am assuming this only applies to >=3.6 version.

scripts/gradle/request.json
gaiksaya, line 1: Looks like an empty file was added?

peterzhuamazon:
Hi @gaiksaya 

1. This is ok as that param will not take any effect in 2.x as the code is not commited in 2.x so it is just a no-op. Confirmed with @cwperks.
2. Removed empty file.

Thanks!

gaiksaya:
>
>>    1. This is ok as that param will not take any effect in 2.x as the code is not commited in 2.x so it is just a no-op. Confirmed with @cwperks.

Since it is just a gradle property should not cause any issue. Please keep watch in case anyone backports to 2.19 by mistake.

peterzhuamazon:
> > > 1. This is ok as that param will not take any effect in 2.x as the code is not commited in 2.x so it is just a no-op. Confirmed with @cwperks.
> 
> Since it is just a gradle property should not cause any issue. Please keep watch in case anyone backports to 2.19 by mistake.

It should be fine as the core code was merged back in 3.1.

Issue #5928, opened 2026-01-20
[Bug]: Incremental build does not remove old artifacts if the core is rebuilding


gaiksaya:
### Describe the bug

We had an edge case where a native plugin was removed from core https://github.com/opensearch-project/OpenSearch/pull/20305
Since incremental builds were activated, it downloads the previous builds and then builds on top of it depending on the changes in the commit

The current logic is if the core(s) commit has changed, everything gets rebuilt. In that case, it should be a clean build. Which is not the case today. 

### To reproduce

See build https://build.ci.opensearch.org/job/distribution-build-opensearch/11626/ which was incremental.

### Expected behavior

Any component which is rebuilding should have a clean build. 
Depending on overriding artifact is hazardous. 

### Relevant log output

```shell

```

moaead:
I've opened a PR to address this issue: https://github.com/opensearch-project/opensearch-build/pull/5929

**Changes made:**
- Added `is_full_rebuild` flag to `BuildIncremental` class to track when core is rebuilding
- Modified `run_build.py` to use a clean build manifest (pass `None` to `BuildRecorder`) when core is rebuilding, instead of reusing the previous build manifest

**How to verify:**
1. Trigger an incremental build where core (OpenSearch) has new changes
2. Verify that the build output does not contain artifacts from plugins that were removed from the manifest
3. Confirm the log shows "Full rebuild triggered, using clean build manifest."

Happy to iterate based on feedback.

krisfreedain:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

ThyTran1402:
Hi @gaiksaya 
I tried to work on this issue. I was able to find the root causes of the issue are:

+ output directory not cleaned on full rebuild: when all components rebuild (rebuild_plugins returns all components because core changed), the outputdir at {distribution}/builds/{filename}/ is never cleaned. Stale files from the previous build persist   
alongside the new ones.
+ removed components from old manifest persist in new build manifest: if a component was removed from the input manifest but exists in the previous build manifest, it is never overwritten, so it persists in the new build manifest with stale artifact paths. The assemble step would then try to install it.
+ crash when previous manifest doesn't exist

[...]

PR #6003, opened 2026-02-22
fixed incremental build logic

ThyTran1402:
### Description
- When incremental builds are active and a native plugin is removed from core, it causes 3 bugs:
Output_dir not cleaned on full rebuild  when all components rebuild because core changed, stale files from the previous build persist in the output directory alongside new artifacts.       
Removed components persist in new manifest: If a component is removed from the input manifest but exists in the previous build manifest, it survives into the new build manifest and the assemble step tries to install it.    
Crash when no previous manifest: run_build.py logs a warning about a missing manifest file, then unconditionally calls BuildManifest.from_path() on the missing file, crashing with FileNotFoundError.                      

- Fixed:
Main orchestrator, where all three fixes land                       
Fix removed components persist in new manifest by valid_component_names filter   
Reference only, rebuild_plugins() already returns all components when core changes
Update + add integration tests   
Add unit test for stale-component filtering 

### Issues Resolved
- #5928 

### DIFF
--- a/src/run_build.py
+++ b/src/run_build.py
@@ -57,6 +57,7 @@ def main() -> int:
             core in list_of_updated_plugins for core in ("OpenSearch", "OpenSearch-Dashboards")
         )
 
+        build_manifest = None
         if is_full_rebuild:
             logging.info("Full rebuild triggered. Cleaning output directory and skipping previous build manifest.")
             if os.path.exists(output_dir):


ThyTran1402:
Hi @peterzhuamazon 

Can you help to review it please?

Thank you!

src/run_build.py
```diff
@@ -80,7 +90,11 @@ def main() -> int:
             architecture=args.architecture or manifest.build.architecture,
         )
 
-        build_recorder = BuildRecorder(target, build_manifest) if args.incremental else BuildRecorder(target)
+        if args.incremental and build_manifest:
+            valid_names = {c.name for c in manifest.components.select()}
```
gaiksaya, line 95: Can you please confirm what valid_names mean here? Is it all components?

gaiksaya (commented):
Hi @ThyTran1402 

Thank you for the contribution. Can you please check the code review suggestions here: https://github.com/opensearch-project/opensearch-build/pull/6003#issuecomment-3941395830

Also can you please fix the tests so that expected behavior is validated?

Thanks!

src/run_build.py
ThyTran1402: Yes. So valid_names here is all components in the current build manifest like {"OpenSearch", "common-utils", ...}, and it used to filter out any components that were in the previous build but are no longer valid.

ThyTran1402:
> Hi @ThyTran1402
> 
> Thank you for the contribution. Can you please check the code review suggestions here: [#6003 (comment)](https://github.com/opensearch-project/opensearch-build/pull/6003#issuecomment-3941395830)
> 
> Also can you please fix the tests so that expected behavior is validated?
> 
> Thanks!

I just initialized build_manifest to None. Can you review it again please?

Thank you!

PR #6010, opened 2026-02-25
Temporarily set file size limit before moving from -p to --file as parameter allows only 100kb inputs

peterzhuamazon:
### Description
Temporarily set file size limit before moving from -p to --file as parameter allows only 100kb inputs

### Issues Resolved

### DIFF
--- a/.github/workflows/code-diff-analyzer.yml
+++ b/.github/workflows/code-diff-analyzer.yml
@@ -49,7 +49,7 @@ jobs:
     env:
       AWS_REGION: 'us-east-1'
       CLAUDE_CODE_USE_BEDROCK: '1'
-      CLAUDE_CODE_MAX_INPUT_TOKENS: '150000'
+      CLAUDE_CODE_MAX_INPUT_TOKENS: '100000'
       CLAUDE_CODE_MAX_OUTPUT_TOKENS: '4096'
       MAX_THINKING_TOKENS: '1024'
       DIFF_CONTENT_PATH: 'git_diff_content.txt'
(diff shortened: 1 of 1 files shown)


peterzhuamazon:
> ## PR Code Suggestions ✨
> 
> Explore these optional code suggestions:
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue 	
> Comparing bytes to token limit
> 
> 	High

This is expected at the moment as a temp measure.
Since parameters accept max 100KB of chars, before I migrate from -p to --file I need to set this way to fail PRs early.

Thanks.

.github/workflows/code-diff-analyzer.yml
```diff
@@ -107,12 +107,11 @@ jobs:
...
-          DIFF_TOKEN_EST=$((DIFF_SIZE / 3))
-          echo "Check diff size of '$DIFF_SIZE' with token number '$DIFF_TOKEN_EST' over max token limit '$CLAUDE_CODE_MAX_INPUT_TOKENS'"
-          if [ "$DIFF_TOKEN_EST" -gt "$CLAUDE_CODE_MAX_INPUT_TOKENS" ]; then
-            echo "Diff too large, requires skip by maintainers after manual review"
+          echo "Check diff size of '$DIFF_SIZE' max token limit '$CLAUDE_CODE_MAX_INPUT_TOKENS'"
+          if [ "$DIFF_SIZE" -ge "$CLAUDE_CODE_MAX_INPUT_TOKENS" ]; then
+            echo "Diff too large, reduce your PR size or requires skip by maintainers after manual review"
             echo "diff_analyzer=5" >> $GITHUB_ENV
-            echo "diff_report='Diff too large, requires skip by maintainers after manual review'" >> $GITHUB_ENV
+            echo "diff_report='Diff too large, reduce your PR size or requires skip by maintainers after manual review'" >> $GITHUB_ENV
```
gaiksaya, line 114: Anyway manual review would be needed right?
```suggestion
            echo "diff_report='Diff too large, reduce your PR size to utilize this feature'" >> $GITHUB_ENV
```

.github/workflows/code-diff-analyzer.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     env:
       AWS_REGION: 'us-east-1'
       CLAUDE_CODE_USE_BEDROCK: '1'
-      CLAUDE_CODE_MAX_INPUT_TOKENS: '150000'
+      CLAUDE_CODE_MAX_INPUT_TOKENS: '100000'
```
gaiksaya, line 52: Are we reducing the tokens here?

.github/workflows/code-diff-analyzer.yml
peterzhuamazon: It needed anyway, but in order to get out of the limit, maintainer need to manually skip the workflow which is why I add that line to be specific.

.github/workflows/code-diff-analyzer.yml
peterzhuamazon: Yes that is expected because the current -p param only support up to that.
We would need to migrate to --file in order to send larger files.

PR #6033, opened 2026-03-12
Retry Jenkins trigger on failure in gradle-check

### LINKED ISSUE #5890: [Bug]: Gradle check script does not fail fast on 403
### Describe the bug

When running the [gradle check workflow in the OpenSearch repo](https://github.com/opensearch-project/opensearch-build/blob/f607b5ff4034e17fbf3614c6fd2aa98dbf6b520a/scripts/gradle/gradle-check.sh), sometimes Jenkins seems to respond with a 403. I don't know the root cause of this, but it appears to be transient and a retry will usually work. However, the gradle check workflow will not fail gracefully and instead proceed to poll for a couple hours before timing out, despite the fact that the script can't actually poll the status of a Jenkins job. See [this script output](https://github.com/opensearch-project/OpenSearch/actions/runs/20380368754/job/58569115936) as an example:

```
Trigger Jenkins workflows
{"pr_from_sha": "ad5c02da1a3df638e78dc058c1f9b6197434fde6", "pr_from_clone_url": "https://github.com/andrross/OpenSearch.git", "pr_to_clone_url": "https://github.com/opensearch-project/OpenSearch.git", "pr_title": "fix allocationconstraintstests", "pr_number": "20292", "post_merge_action": "false", "pr_owner": "andrross"}
<html>
 <head><title>403 Forbidden</title></head>
 <body>
 <center><h1>403 Forbidden</h1></center>
 </body>
 </html>
jq: parse error: Invalid numeric literal at line 1, column 7
QUEUE_URL
wait for jenkins to start workflow
Check if queue exist in Jenkins after triggering
Use queue information to find build number in Jenkins if available
jq: parse error: Invalid numeric literal at line 1, column 7
WORKFLOW_URL
Job not started yet. Waiting for 60 seconds before next attempt.
time passed: 60
Use queue information to find build number in Jenkins if available
jq: parse error: Invalid numeric literal at line 1, column 7
WORKFLOW_URL
Job not started yet. Waiting for 60 seconds before next attempt.
time passed: 120
Use queue information to find build number in Jenkins if available
jq: parse error: Invalid numeric literal at line 1, column 7
WORKFLOW_URL
Job not started yet. Waiting for 60 seconds before next attempt.
time passed: 180
```

### To reproduce

I don't know of a reliable reproduction, but I have seen it multiple times.

### Expected behavior

If the following line

```
QUEUE_URL=$(echo $JENKINS_REQ | jq --raw-output '.jobs."gradle-check".url')
```

is unable to get a valid URL, it should retry or at minimum fail the script immediately if it can't determine a valid URL to poll.

andrross:
When Jenkins responds with a transient 403, the script fails to parse the response with jq but proceeds to poll with an invalid URL for hours before timing out. Add retry logic (3 attempts) for the trigger request and fail fast if a valid queue URL cannot be obtained.

Also fix the guard condition to be "non-empty AND not 'null'" so that it only enters the polling loop if we have a valid URL.

### Issues Resolved
Resolves #5890

### DIFF
--- a/scripts/gradle/gradle-check.sh
+++ b/scripts/gradle/gradle-check.sh
@@ -92,55 +92,74 @@ perform_curl_and_process_with_jq() {
 }
 
 echo "Trigger Jenkins workflows"
-JENKINS_REQ=`curl -s -XPOST \
-     -H "Authorization: Bearer $TRIGGER_TOKEN" \
-     -H "Content-Type: application/json" \
-     "$JENKINS_URL/generic-webhook-trigger/invoke" \
-     --data "$(echo $PAYLOAD_JSON)"`
-
 echo $PAYLOAD_JSON
-echo $JENKINS_REQ
 
-QUEUE_URL=$(echo $JENKINS_REQ | jq --raw-output '.jobs."gradle-check".url')
-echo QUEUE_URL $QUEUE_URL
+TRIGGER_RETRIES=3
+for i in $(seq 1 $TRIGGER_RETRIES); do
+    RESPONSE=$(curl -s -w "\n%{http_code}" -XPOST \
+         -H "Authorization: Bearer $TRIGGER_TOKEN" \
+         -H "Content-Type: application/json" \
+         "$JENKINS_URL/generic-webhook-trigger/invoke" \
+         --data "$(echo $PAYLOAD_JSON)")
+    HTTP_CODE=$(echo "$RESPONSE" | tail -1)
+    JENKINS_REQ=$(echo "$RESPONSE" | sed '$d')
+    echo "$JENKINS_REQ"
+
+    if [ "$HTTP_CODE" -ge 200 ] && [ "$HTTP_CODE" -lt 400 ]; then
+        break
+    fi
+
+    echo "Trigger attempt $i/$TRIGGER_RETRIES returned HTTP $HTTP_CODE. Retrying in 30 seconds..."
+    if [ "$i" -eq "$TRIGGER_RETRIES" ]; then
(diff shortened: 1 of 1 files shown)


andrross:
Full disclosure: kiro mostly wrote this code because writing complex logic in bash makes my head hurt.

andrross:
@gaiksaya Could you take a look here when you get a few minutes? Note it helps to hide whitespace on the diff because there's a big block where only the indent changed from removing an if statement.

scripts/gradle/gradle-check.sh
```diff
@@ -92,55 +92,74 @@ perform_curl_and_process_with_jq() {
...
+    RESPONSE=$(curl -s -w "\n%{http_code}" -XPOST \
+         -H "Authorization: Bearer $TRIGGER_TOKEN" \
+         -H "Content-Type: application/json" \
+         "$JENKINS_URL/generic-webhook-trigger/invoke" \
+         --data "$(echo $PAYLOAD_JSON)")
+    HTTP_CODE=$(echo "$RESPONSE" | tail -1)
+    JENKINS_REQ=$(echo "$RESPONSE" | sed '$d')
+    echo "$JENKINS_REQ"
+
+    if [ "$HTTP_CODE" -ge 200 ] && [ "$HTTP_CODE" -lt 300 ]; then
```
peterzhuamazon: I wonder if we should make it `-lt 400` since 301 is also legit.

cwperks:
@peterzhuamazon Can we get this PR and the one in the associated opensearch-build-libraries merged? It would be really nice to start using this to break the ./gradlew check monolith of the core and get average time down from 80 min to much lower.

PR #6043, opened 2026-03-23
Add -e (--exclude-tasks) param to gradle-check.sh

cwperks:
### Description

This PR is part of a series of 3:

- https://github.com/opensearch-project/OpenSearch/pull/20946
- https://github.com/opensearch-project/opensearch-build/pull/6043
- https://github.com/opensearch-project/opensearch-build-libraries/pull/814

This PR adds a new param `-e` to gradle-check.sh so that OpenSearch core's invocation can pass a list of tasks to exclude: https://github.com/opensearch-project/OpenSearch/blob/main/.github/workflows/gradle-check.yml#L131

### Issues Resolved

Related to: https://github.com/opensearch-project/OpenSearch/issues/19378

### DIFF
--- a/jenkins/gradle/gradle-check.jenkinsfile
+++ b/jenkins/gradle/gradle-check.jenkinsfile
@@ -70,7 +70,8 @@ pipeline {
                 [key: 'pr_title', value: '$.pr_title'],
                 [key: 'pr_number', value: '$.pr_number'],
                 [key: 'post_merge_action', value: '$.post_merge_action'],
-                [key: 'pr_owner', value: '$.pr_owner']
+                [key: 'pr_owner', value: '$.pr_owner'],
+                [key: 'exclude_tasks', value: '$.exclude_tasks', defaultValue: '']
             ],
             tokenCredentialId: 'jenkins-gradle-check-generic-webhook-token',
             causeString: 'Triggered by PR on OpenSearch core repository',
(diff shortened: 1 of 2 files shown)


peterzhuamazon:
Dont think I have any issue with this, but need @gaiksaya to co-approve as she was recently working on this and more familiar.

Thanks.

andrross:
@cwperks As opposed to passing excludes to gradle, I think it would be better to let OpenSearch itself split `check` into multiple targets, which would then be run via separate Jenkins jobs. Hypothetically, that could be something like:

- `./gradlew check-1` (all `:qa` tasks)
- `./gradlew check-2` (all test tasks in `:plugins:*`)
- `./gradlew check-3` (everything not included in checks 1 and 2)

The OpenSearch repo would then be free to moving things between these various tasks as appropriate. This seems more straightforward to me than using exclusions. It also keeps everything defined in the gradle configuration without having certain exclusions in the GitHub action definition.

cwperks:
@andrross I don't disagree, but does that preclude this change and corresponding change in opensearch-build-libraries from being added?

Might be nice to have in the future for flexibility.

andrross:
> but does that preclude this change and corresponding change in opensearch-build-libraries from being added?

@cwperks Definitely not, it's just a question of whether we want to add the complexity here.

andrross:
@cwperks There's the version to parameterize the command: #6074

cwperks:
Closing in favor of https://github.com/opensearch-project/opensearch-build/pull/6074

cwperks:
> @cwperks There's the version to parameterize the command: #6074

Yea this is definitely the direction we should go in. Thank you for raising that @andrross !

PR #6055, opened 2026-03-27
Improve AI release notes generation for OpenSearch

andrross:
- Fetch PR body from GitHub API and include it (truncated to 2000 chars) in the commit data sent to the LLM for better categorization and filtering decisions.

- Add OpenSearch-specific prompt that filters non-user-facing changes (test additions, CI action bumps, test fixture deps, internal refactoring, release machinery) before categorization. Removes Infrastructure/Documentation/Refactoring categories that contradicted the exclusion policy. Only Breaking Changes, Features, Enhancements, Bug Fixes, and Maintenance categories are used.

- Add borderline calls section: the LLM appends an HTML comment documenting debatable inclusion/exclusion/categorization decisions. This is extracted from the LLM output and written to a separate -borderline.md file.

- Update Jenkinsfile to include borderline calls in the PR description when creating or updating release notes PRs. Also fix the existing PR path to update the PR body via gh pr edit.

### DIFF
--- a/src/llms/prompts.py
+++ b/src/llms/prompts.py
@@ -55,7 +55,7 @@ AI_RELEASE_NOTES_PROMPT_COMMIT = """I need you to generate OpenSearch component 
      * "refactor" or "refactoring" → Refactoring
 
 2. **Fallback Message Analysis:**
-   - If no labels match, analyze the Message content and PullRequestSubject to determine the appropriate category following below guidelines:
+   - If no labels match, analyze the Message content, PullRequestSubject, and PR Description to determine the appropriate category following below guidelines:
     * Features: A change that introduce a net new unit of functionality of a software system that satisfies a requirement,
     represents a design decision, and provides a potential configuration option. As for improvement on existing features,
     use the Enhancement category. As for fixes on existing features, use the Bug Fixes category.
@@ -79,7 +79,9 @@ AI_RELEASE_NOTES_PROMPT_COMMIT = """I need you to generate OpenSearch component 
    - Do not add the pull request that has title starting with "[AUTO] Increment version to".
 
 4. **Entry Format:**
-   - Use PullRequestSubject as the main content for each entry
+   - Use the PullRequestSubject and PR Description as input, but rewrite each entry as a concise, clear one-line
+     summary. The target audience is end-users, operators, and system administrators — not developers.
+     Do not simply copy the PR subject verbatim; improve clarity and consistency.
    - Extract PR number from PullRequestSubject (format: (#123))
    - Format: `* <description> ([#<number>]({repository_url}/pull/<number>))`
    - Always use asterisk (*) for bullet points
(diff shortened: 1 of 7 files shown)


src/llms/prompts.py
```diff
@@ -101,10 +103,128 @@
...
+   - Use the commit message, PR description, and labels to make this judgment.
+   - When uncertain whether a change is user-facing, **include it** — a human reviewer can remove it later.
+
+2. **Categorization — only for PRs that survive filtering:**
+   - First, check if any labels match these categories (case-insensitive, partial matches allowed):
+     * "breaking change" or "breaking" → Breaking Changes
+     * "feature" or "feat" → Features
+     * "enhancement" or "improve" → Enhancements
+     * "bug" or "fix" or "bugfix" → Bug Fixes
+     * "maintenance" or "version" or "support" → Maintenance
```
gaiksaya, line 166: The categories also include infrastructure and documentation changes as per guidelines. 
https://github.com/opensearch-project/opensearch-plugins/blob/main/RELEASE_NOTES.md#categories
I agree they do no provide any value. @peterzhuamazon maybe we also need to exclude them from consolidated ones.

src/llms/prompts.py
```diff
@@ -101,10 +103,128 @@
...
+   For each, reference the PR number and briefly explain the decision and alternatives. Example:
+   ```
+   <!-- BORDERLINE_CALLS
+   - #1234: Placed in **Enhancements** — could also be **Bug Fixes** since it changes existing behavior to fix a usability issue.
+   - #5678: Placed in **Maintenance** — borderline with **Infrastructure**, chose Maintenance because it's a dependency update.
+   -->
+   ```
+   If there are no borderline calls, omit this section entirely.
+
+Generate the release notes in proper OpenSearch format:"""
```
gaiksaya, line 124: Duplicate of line 230?
`"""` starts here but not sure where it is ending.

src/llms/prompts.py
```diff
@@ -101,10 +103,128 @@
...
+   - Sort categories in this order: Breaking Changes, Features, Enhancements, Bug Fixes, Maintenance
+   - Each entry should be a single line with proper PR link formatting
+
+5. **PR Link Format:**
+   - Extract PR number from PullRequestSubject
+   - Format as: `([#<number>]({repository_url}/pull/<number>))`
+   - Example: `([#456](https://github.com/opensearch-project/OpenSearch/pull/456))`
+
+6. **Important Notes:**
+   - If you cannot determine the appropriate category from labels OR content analysis, place the entry in "Maintenance"
```
gaiksaya: In the consolidation workflow where we combine all non-core components' release notes into one, we place the uncategorized data in `non-compliant` section? Would doing the same here avoid confusing between actual maintenance?

andrross: That's probably a good idea.

The reason I hesitated is that anytime the LLM creates an "unknown" or "non-compliant" section it means manual work will be required. Probably not the end of the world if it truly has no idea how to categorize.

src/llms/prompts.py
andrross: This is ending the prompt that started on line 34. Because of the difference with excluding commits, the prompt that starts on line 126 is only being used with the OpenSearch repo in this PR. See the`commit_prompt = AI_RELEASE_NOTES_PROMPT_COMMIT_OPENSEARCH if component.name == 'OpenSearch' else AI_RELEASE_NOTES_PROMPT_COMMIT` line in release_notes.py

src/llms/prompts.py
```diff
@@ -101,10 +103,128 @@
...
+     * Features: A net new unit of functionality that satisfies a requirement, represents a design decision,
+       and provides a potential configuration option. For improvements on existing features, use Enhancements.
+       For fixes on existing features, use Bug Fixes.
+     * Enhancements: Improves the performance, usability, or reliability of an existing feature without
+       changing its core functionality.
+     * Bug Fixes: Resolves an issue or defect in the software.
+     * Maintenance: Routine upkeep such as dependency updates that ship in the distribution.
+   - Do not use "Infrastructure", "Documentation", or "Refactoring" as categories. Changes that would
+     belong to those categories should have been excluded by the filtering step. If a PR survived filtering
+     but does not fit any of the above categories, place it in "Maintenance".
```
gaiksaya:
```suggestion
     but does not fit any of the above categories, place it in "Unknown".
```

andrross: Actually I think I should strike the last sentence. I find it confusing given the instruction at the end.

PR #6068, opened 2026-03-31
Tighten release notes prompt and clean input data

andrross:
- Strip Signed-off-by/Co-authored-by tags from commit messages
- Strip HTML comments, Check List, Related Issues sections, and DCO boilerplate from PR descriptions
- Drop PR body for dependabot PRs (upstream changelogs are noise)
- Tighten AI_RELEASE_NOTES_PROMPT_COMMIT_OPENSEARCH: remove redundant sections, add example rewrite, use template variables for version/URL
- Simplify Maintenance category to include all dependency updates

### DIFF
--- a/src/llms/prompts.py
+++ b/src/llms/prompts.py
@@ -123,107 +123,61 @@ AI_RELEASE_NOTES_PROMPT_COMMIT = """I need you to generate OpenSearch component 
 
 Generate the release notes in proper OpenSearch format:"""
 
-AI_RELEASE_NOTES_PROMPT_COMMIT_OPENSEARCH = """I need you to generate OpenSearch component release notes from commit data. Please follow the OpenSearch release notes format exactly.
+AI_RELEASE_NOTES_PROMPT_COMMIT_OPENSEARCH = """Generate release notes for {component_name} {version} from the commit data below.
 
-**Component Information:**
-- Component Name: {component_name}
-- Version: {version}
-- Repository URL: {repository_url}
+Repository: {repository_url}
 
 **Commit Data:**
 {commits_text}
 
-**Instructions:**
+**Filtering (apply first):**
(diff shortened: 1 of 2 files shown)


andrross:
@gaiksaya @rishabh6788 Kiro suggested improvements to the prompt to be more direct and remove redundancy. I figure kiro knows what the models want :) The only substantive change in the prompt is to remove exclusion rule for "Dependency bumps that only affect test fixtures or build tooling". That actually seems like a difficult judgement call and there seems to be little harm to including all dependency updates so on balance it seems simpler without it.

Stripping things like the checklist and DCO out of the input data actually cuts down on the amount of text to bedrock by half, which should help since that was just noise.

src/llms/prompts.py
```diff
@@ -123,107 +123,61 @@
...
+- Match labels first (case-insensitive, partial match):
+  "breaking"→Breaking Changes, "feature"/"feat"→Features, "enhancement"/"improve"→Enhancements,
+  "bug"/"fix"/"bugfix"→Bug Fixes, "maintenance"/"version"/"support"→Maintenance
+- If no label matches, categorize by content:
+  Features=net new functionality, Enhancements=improves existing feature,
+  Bug Fixes=fixes a defect, Maintenance=dependency updates and routine upkeep.
+- Only use these 5 categories. If none fit, use "Unknown".
+
+**Entry format:**
+`* <concise one-line summary for end-users/operators> ([#<number>]({repository_url}/pull/<number>))`
```
gaiksaya, line 153: Does it make sense to add something like `Strictly follow this format only`? Its known to follows its own mind from time to time.

andrross: We can add that if necessary. I think with this plus the examples below the formatting should be clear.

PR #6075, opened 2026-03-31
Update code diff analyzer on the prompts

peterzhuamazon:
### Description
Update code diff analyzer on the prompts

### Issues Resolved

### DIFF
--- a/.github/workflows/code-diff-analyzer.yml
+++ b/.github/workflows/code-diff-analyzer.yml
@@ -144,6 +144,17 @@ jobs:
           	
           	**PRIMARY FOCUS: Detect deliberate attempts to compromise security, not coding mistakes.**
           	
+          	**MANDATORY RULE — SUPPLY CHAIN / DEPENDENCY CHANGES:**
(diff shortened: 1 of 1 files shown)


.github/workflows/code-diff-analyzer.yml
```diff
@@ -144,6 +144,17 @@ jobs:
           	
           	**PRIMARY FOCUS: Detect deliberate attempts to compromise security, not coding mistakes.**
           	
+          	**MANDATORY RULE — SUPPLY CHAIN / DEPENDENCY CHANGES:**
+          	- Any dependency, package registry, or build plugin change MUST be flagged as **high** severity
+          	- Do NOT judge whether a dependency name looks "legitimate" — you cannot verify artifact authenticity
+          	- Namespace hijacking and typosquatting are common attack techniques
+          	- Always flag and let maintainers verify
+          	- This applies to changes including but not limited to:
+          	  - Adding, removing, or changing versions of libraries, packages, or modules
```
cwperks, line 153: If we do this, then can we also exempt dependabot and mend? We should be able to do that in their respective configurations to automatically at the skip-diff-analyzer label.

PR #6088, opened 2026-04-02
Add gradle_check_command webhook parameter

andrross:
My previous PR only added a pipeline parameter. I need a webhook parameter to be able to start this job from the OpenSearch GitHub Action.

This also adds validation to the gradle check command to reject invalid characters.


andrross:
> Directly interpolating $GRADLE_CHECK_COMMAND into a JSON string is unsafe if the
value contains double quotes, backslashes, or other special JSON characters, which
could break the JSON structure or introduce injection issues. Use a tool like jq to
safely construct the JSON payload.

I think this is a good idea but I'd rather do it as a follow up PR as it is not something new introduced here

PR #6115, opened 2026-04-10
Improve issue dedupe detect workflow reliability

qianheng-aws:
## Summary
- Change default model to `claude-sonnet-4-5`
- Move comment posting from Claude to workflow steps to avoid Claude Code shell security validation rejecting multi-line markdown (`Newline followed by # inside a quoted argument`)
- Extract duplicate list from `execution_file` output instead of temp file, removing need for `Write` permission
- Add `show_full_output` input parameter for debugging
- Add `GH_TOKEN` env to label step

## Context
When Claude posts comments directly via `gh issue comment --body`, the `\n###` pattern in markdown triggers Claude Code's shell injection protection (`Newline followed by # inside a quoted argument can hide arguments from path validation`). This caused inconsistent failures depending on issue title content. Moving comment formatting to workflow steps makes it deterministic.

## Test plan
- [ ] Trigger dedupe detection on an issue with special characters in title (e.g., backticks)
- [ ] Verify comment format is correct with proper markdown headers and bullet points
- [ ] Verify `show_full_output: 'true'` shows CC logs in workflow output

### DIFF
--- a/.github/workflows/issue-dedupe-detect.yml
+++ b/.github/workflows/issue-dedupe-detect.yml
@@ -45,8 +45,6 @@ jobs:
       - name: Run Claude Code for duplicate detection
         id: claude
         uses: anthropics/claude-code-action@2ff1acb3ee319fa302837dad6e17c2f36c0d98ea # v1
-        env:
-          DUPLICATE_GRACE_DAYS: ${{ inputs.grace_days }}
         with:
           use_bedrock: 'true'
           github_token: ${{ secrets.GITHUB_TOKEN }}
@@ -56,12 +54,13 @@ jobs:
             --allowedTools
             "Bash(gh issue view *)"
             "Bash(gh search issues *)"
+            --json-schema '{"type":"object","properties":{"duplicates":{"type":"array","items":{"type":"integer"}},"skip_reason":{"type":"string"}},"required":["duplicates"]}'
           prompt: |
             Find up to 3 likely duplicate issues for issue #${{ env.ISSUE_NUMBER }} in ${{ github.repository }}.
 
             Follow these steps precisely:
 
-            1. Use `gh issue view ${{ env.ISSUE_NUMBER }} --repo ${{ github.repository }} --comments` to read the issue and its comments. If the issue is closed, or is broad product feedback without a specific bug/feature request, or already has a duplicate detection comment (containing `<!-- duplicate-detection -->`), stop immediately and output nothing.
+            1. Use `gh issue view ${{ env.ISSUE_NUMBER }} --repo ${{ github.repository }} --comments` to read the issue and its comments. If the issue is closed, or is broad product feedback without a specific bug/feature request, or already has a duplicate detection comment (containing `<!-- duplicate-detection -->`), return an empty duplicates array with skip_reason explaining why.
 
             2. Summarize the issue's core problem in 2-3 sentences. Identify the key terms, error messages, and affected components.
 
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-dedupe-detect.yml
```diff
@@ -74,36 +73,77 @@ jobs:
...
+            - Do not flag the issue as a duplicate of itself
 
-               d. Post the comment. You MUST use this EXACT command with the EXACT format below (only replace N with the count and <dup_list> with lines like `- #123 — Title`):
-                  ```
-                  gh issue comment ${{ env.ISSUE_NUMBER }} --repo ${{ github.repository }} --body "<!-- duplicate-detection -->
+      - name: Build duplicate list
+        id: build-list
+        shell: bash
+        run: |
+          EXECUTION_FILE="${{ steps.claude.outputs.execution_file }}"
```
peterzhuamazon: Didnt get where does this file being created?
Is this a file that is provided directly by claude action?
Could you give me the reference link to their docs? Thanks.

qianheng-aws: >Is this a file that is provided directly by claude action?

Yes and I discovered it from the log of every workflows, e.g. https://github.com/opensearch-project/opensearch-build/actions/runs/24189746250/job/70603502795
```
Log saved to /home/runner/work/_temp/claude-execution-output.json
```

> Could you give me the reference link to their docs?

Find related output definition  and README about execution file:
https://github.com/anthropics/claude-code-action/blob/1c8b699d43e9bfed42b48ef15da85d89bab70960/action.yml#L154-L157

https://github.com/anthropics/claude-code-action/blob/1c8b699d43e9bfed42b48ef15da85d89bab70960/base-action/README.md?plain=1#L119

qianheng-aws: https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md#structured-outputs,

There is another output file seems to be more proper for our case. It's generated only when --json-schema arguments is provided. I'll try this way as well.

qianheng-aws:
> https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md#structured-outputs,
There is another output file seems to be more proper for our case. It's generated only when --json-schema arguments is provided. I'll try this way as well.

@peterzhuamazon  This approach works fine as well and should be more robust. 

Test workflow: https://github.com/opensearch-project/sql/actions/runs/24380112736/job/71201609512

Test issue: https://github.com/opensearch-project/sql/issues/5326

.github/workflows/issue-dedupe-detect.yml
```diff
@@ -74,36 +72,85 @@ jobs:
...
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Add duplicate label
+        if: steps.process.outputs.has_duplicates == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh label create "duplicate" --description "Issue is a duplicate of an existing issue" --color "cccccc" --repo "${{ github.repository }}" 2>/dev/null || true
+          gh issue edit "${{ env.ISSUE_NUMBER }}" --repo "${{ github.repository }}" --add-label "duplicate"
+          echo "Added duplicate label to issue #${{ env.ISSUE_NUMBER }}"
```
peterzhuamazon, line 137: This is not blocking for now but may need to update, if there is already duplicate label will this command fail?

peterzhuamazon (approved):

.github/workflows/issue-dedupe-detect.yml
qianheng-aws: No, it won't fail per my test. Adding a label is an idempotent operation. It will success with nothing change.

Issue #5810, opened 2025-10-28
[Enhancement] Supporting building specific feature branch distributions


peterzhuamazon:
[Enhancement] Supporting building specific feature branch distributions

Ex: build OpenSearch core engine with https://github.com/opensearch-project/OpenSearch/tree/feature/datafusion with a specific feature name.

Need to update both jenkins and libs to support changing version to feature name.

Need to make sure build repo has a separate feature-manifests folder.

Need to have a new jenkins workflow to support feature specific builds.

Thanks.

------

PRs:
* https://github.com/opensearch-project/opensearch-build-libraries/pull/780
* https://github.com/opensearch-project/opensearch-build/pull/5811

peterzhuamazon:
Pending docker images.

peterzhuamazon:
Issues: https://build.ci.opensearch.org/job/feature-build-opensearch/9/console

```
+ ./assemble.sh tar/builds/opensearch/manifest.yml --base-url https://ci.opensearch.org/ci/dbc/feature-build-opensearch/3.3.0/9/linux/x64
Installing dependencies in . ...
Installing dependencies from Pipfile.lock (d51f6f)...
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
Running ./src/run_assemble.py tar/builds/opensearch/manifest.yml --base-url https://ci.opensearch.org/ci/dbc/feature-build-opensearch/3.3.0/9/linux/x64 ...
2025-11-03 22:39:23 INFO     Bundling OpenSearch (x64) on linux into /var/jenkins/workspace/feature-build-opensearch/tar/dist/opensearch ...
2025-11-03 22:39:23 INFO     Copied min bundle to /tmp/tmp7bskrnzy/opensearch-min-3.3.0-linux-x64.tar.gz.
2025-11-03 22:39:23 INFO     Start creating distribution tar for OpenSearch-DataFusion.
2025-11-03 22:39:23 INFO     Extracting dist into /tmp/tmp7bskrnzy.
Traceback (most recent call last):
  File "/var/jenkins/workspace/feature-build-opensearch/./src/run_assemble.py", line 58, in <module>
    sys.exit(main())
  File "/var/jenkins/workspace/feature-build-opensearch/./src/run_assemble.py", line 42, in main
    with Bundles.create(build_manifest, artifacts_dir, bundle_recorder, args.keep) as bundle:
[...]

PR #6175, opened 2026-05-02
Sort the build recorder on plugin list so plugins can be installed by order

peterzhuamazon:
### Description
Sort the build recorder on plugin list so plugins can be installed by order

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5810

### DIFF
--- a/scripts/components/core-plugins-sandbox/build.sh
+++ b/scripts/components/core-plugins-sandbox/build.sh
@@ -72,6 +72,7 @@ pwd
 ../../gradlew assemble -Dbuild.snapshot="$SNAPSHOT" -Dbuild.version_qualifier=$QUALIFIER -Dsandbox.enabled=true -PrustRelease -Pcrypto.standard=FIPS-140-3
 
 [ -z "$OUTPUT" ] && OUTPUT=artifacts
+INSTALL_ORDER=1
 mkdir -p $OUTPUT/plugins
 for plugin in ./*; do
   PLUGIN_NAME=$(basename "$plugin")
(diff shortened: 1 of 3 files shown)


scripts/components/core-plugins-sandbox/build.sh
```diff
@@ -84,7 +85,11 @@ for plugin in ./*; do
        [ "$PLUGIN_NAME" = "dsl-query-executor" ] ||
        [ "$PLUGIN_NAME" = "parquet-data-format" ]; then
       PLUGIN_ARTIFACT_BUILD_NAME=`ls "$plugin"/build/distributions/ | grep "$PLUGIN_NAME-$VERSION.zip"`
-      cp -v "$plugin"/build/distributions/"$PLUGIN_ARTIFACT_BUILD_NAME" "${OUTPUT}"/plugins/
+      if [ "$PLUGIN_NAME" = "analytics-engine" ]; then
+        PLUGIN_NAME=$INSTALL_ORDER-$PLUGIN_NAME
+        INSTALL_ORDER=$((INSTALL_ORDER + 1))
```
expani, line 90: Given that this needs to be a DAG eventually and the order of determination will be hierarchical in nature.

Should we consider moving this to `buidler_from_source` for easy maintenance in Python compared to Shell scripts ?

expani (commented):
Thanks for helping with this @peterzhuamazon

scripts/components/core-plugins-sandbox/build.sh
peterzhuamazon: This needs proper planning before make it to python. The python code is for general builds. Previously we never require a core plugin to be installed before a third party plugin in the released bundle. This means either we still follow that behavior, or we decided to install a small set of core plugin beforehand. Until that decision is made and all the datafusion code officially made it to main, we will not onboard that logic to python code.

rishabh6788 (approved):

scripts/components/core-plugins-sandbox/build.sh
gaiksaya: The DAG is being introduced here: https://github.com/opensearch-project/opensearch-build/pull/6190. This is for component level though and builds the graph based on `depends_on`. We can see where this dependency tree needs to be (manifest or core build.gradle)

PR #6177, opened 2026-05-04
Bundle native library into sandbox plugin zips and wire library path at startup

bowenlan-amzn:
## Summary

The DataFusion sandbox plugins depend on `libopensearch_native` (built from Rust via `sandbox/libs/dataformat-native`), but the native library was never packaged into plugin distribution zips. This caused `bin/opensearch` to crash at startup:

```
java.lang.RuntimeException: Failed to load native library 'opensearch_native'
    at NativeLibraryLoader.loadLibrary(NativeLibraryLoader.java:168)
    at NativeBridge.<clinit>(NativeBridge.java:63)
    at DataFusionService.doStart(DataFusionService.java:64)
    at DataFusionPlugin.createComponents(DataFusionPlugin.java:129)
```

`./gradlew run` worked because it explicitly sets `-Dnative.lib.path=...`, but any assembled distribution (tar/deb/rpm) failed because `NativeLibraryLoader` could not find the library through any of its three search tiers (`java.library.path`, classpath resource, `native.lib.path` sysprop).

### `scripts/components/OpenSearch-DataFusion/build.sh`
- Add `sandbox/plugins/*` assemble and zip collection loop (previously only `plugins/*` was collected)
- For plugins that load the native lib at runtime (`analytics-backend-datafusion`, `parquet-data-format`), inject `libopensearch_native` into the plugin zip at `lib/` using `zip -ur` — same pattern as k-NN's `scripts/build.sh`
- Fail fast if the native library is not found after build

### `scripts/startup/tar/linux/opensearch-tar-install.sh`
- Add `DATAFUSION_LIB_DIR` for `plugins/analytics-backend-datafusion/lib`
- Refactor KNN-only native library path export into a loop over all native lib directories
- Skip directories that don't exist (graceful when plugin is not installed)
- Export `JAVA_LIBRARY_PATH` (macOS) or `LD_LIBRARY_PATH` (Linux) so `NativeLibraryLoader`'s tier-1 search finds the lib

## Companion Change (OpenSearch repo)

`sandbox/libs/dataformat-native/build.gradle` needs:
```gradle
assemble.dependsOn buildRustLibrary
```
Without this, the Rust native library is only built as a test dependency and won't exist when `build.sh` looks for it after assemble.

## How It Works (k-NN pattern)

```
gradlew assemble → produces plugin zip (JARs only, no native lib)
    ↓
build.sh → zip -ur injects lib/libopensearch_native.{so,dylib} into zip
    ↓
plugin install → extracts to plugins/<name>/lib/
    ↓
opensearch-tar-install.sh → exports plugins/<name>/lib/ to JAVA_LIBRARY_PATH/LD_LIBRARY_PATH
    ↓
NativeLibraryLoader tier-1 → scans java.library.path dirs → finds lib → success
```

## Test plan

- [ ] Verify `build.sh` produces sandbox plugin zips with `lib/libopensearch_native.*` inside
- [ ] Verify `opensearch-tar-install.sh` exports correct paths on both macOS and Linux
- [ ] Verify `bin/opensearch` starts successfully with all sandbox plugins installed
- [ ] Verify graceful behavior when datafusion plugin is not installed (lib dir doesn't exist, loop skips)

### DIFF
--- a/scripts/components/core-plugins-sandbox/build.sh
+++ b/scripts/components/core-plugins-sandbox/build.sh
@@ -71,6 +71,19 @@ pwd
 
 ../../gradlew assemble -Dbuild.snapshot="$SNAPSHOT" -Dbuild.version_qualifier=$QUALIFIER -Dsandbox.enabled=true -PrustRelease -Pcrypto.standard=FIPS-140-3
 
+# Determine native lib path and extension
+[ -z "$PLATFORM" ] && PLATFORM=$(uname -s | awk '{print tolower($0)}')
+LIB_EXT="so"
+if [[ "$PLATFORM" == "darwin" ]]; then LIB_EXT="dylib"; fi
+NATIVE_LIB="../../sandbox/libs/dataformat-native/rust/target/release/libopensearch_native.${LIB_EXT}"
+
+if [ ! -f "$NATIVE_LIB" ]; then
+    echo "Error: Native library not found at $NATIVE_LIB"
+    exit 1
+fi
+
+NATIVE_LIB_PLUGINS="analytics-backend-datafusion parquet-data-format"
+
 [ -z "$OUTPUT" ] && OUTPUT=artifacts
 INSTALL_ORDER=1
 mkdir -p $OUTPUT/plugins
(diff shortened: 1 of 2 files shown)


scripts/components/OpenSearch-DataFusion/build.sh
```diff
@@ -186,3 +186,39 @@ for plugin in plugins/*; do
   fi
 done
 
+echo "Building sandbox plugins..."
+cd sandbox/plugins
+../../gradlew assemble -Dbuild.snapshot="$SNAPSHOT" -Dbuild.version_qualifier=$QUALIFIER -Dsandbox.enabled=true -PrustRelease -Pcrypto.standard=FIPS-140-3
```
rishabh6788: Assuming assemble task now also compiles the library

bowenlan-amzn: https://github.com/opensearch-project/OpenSearch/pull/21475

peterzhuamazon: DO NOT do this there are ways to already achieve this in core-plugins-sandbox dir.
We can discuss more on how to approach but I would like to separate them for the time being because they are sandbox.

peterzhuamazon:
put to draft pending discussion with Bowen before approaching, since this is very big changes to non-mustang setups.

PR #6190, opened 2026-05-07
Add an option to build components in parallel

gaiksaya:
### Description
As of today, all components are build sequentially on jenkins for OpenSearch which takes approximately 2hours. [Ref](https://build.ci.opensearch.org/view/Build/job/distribution-build-opensearch/11832/)

The `depends_on` in the manifest ([example](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/3.6.0/opensearch-3.6.0.yml#L192-L194)) is only utilized during incremental builds and NOT during regular builds.

This PR refactors the building logic to utilize the depends_on to build a dependency graph (DAG) and run builds in parallel using multiple threads controlled by `--parallel`

```mermaid
graph TD
    OpenSearch[/"OpenSearch"/]:::core

    %% Level 1 - No dependencies
    common-utils:::level1
    opensearch-learning-to-rank-base:::level1
    opensearch-remote-metadata-sdk:::level1
    job-scheduler:::level1
    security:::level1
    custom-codecs:::level1
    performance-analyzer:::level1
    query-insights:::level1
    opensearch-system-templates:::level1
    user-behavior-insights:::level1

    %% Level 2
    k-NN:::level2
    geospatial:::level2
    cross-cluster-replication:::level2
    ml-commons:::level2
    notifications-core:::level2
    notifications:::level2
    opensearch-observability:::level2
    opensearch-reports:::level2
    asynchronous-search:::level2
    anomaly-detection:::level2
    alerting:::level2
    index-management:::level2

    %% Level 3
    neural-search:::level3
    sql:::level3
    security-analytics:::level3

    %% Level 4
    flow-framework:::level4
    skills:::level4
    search-relevance:::level4

    %% Implicit deps from core
    OpenSearch -.-> common-utils
    OpenSearch -.-> opensearch-learning-to-rank-base
    OpenSearch -.-> opensearch-remote-metadata-sdk
    OpenSearch -.-> job-scheduler
    OpenSearch -.-> security
    OpenSearch -.-> custom-codecs
    OpenSearch -.-> performance-analyzer
    OpenSearch -.-> query-insights
    OpenSearch -.-> opensearch-system-templates
    OpenSearch -.-> user-behavior-insights

    %% Explicit dependencies
    common-utils --> k-NN
    security --> k-NN
    custom-codecs --> k-NN

    job-scheduler --> geospatial

    common-utils --> cross-cluster-replication

    common-utils --> ml-commons
    job-scheduler --> ml-commons
    opensearch-remote-metadata-sdk --> ml-commons
    security --> ml-commons

    ml-commons --> neural-search
    k-NN --> neural-search

    common-utils --> notifications-core
    common-utils --> notifications
    common-utils --> opensearch-observability

    common-utils --> opensearch-reports
    job-scheduler --> opensearch-reports

    ml-commons --> sql
    geospatial --> sql
    job-scheduler --> sql

    common-utils --> asynchronous-search

    common-utils --> anomaly-detection
    job-scheduler --> anomaly-detection

    common-utils --> alerting

    common-utils --> security-analytics
    alerting --> security-analytics
    job-scheduler --> security-analytics

    common-utils --> index-management
    job-scheduler --> index-management

    common-utils --> flow-framework
    opensearch-remote-metadata-sdk --> flow-framework
    ml-commons --> flow-framework
    k-NN --> flow-framework
    neural-search --> flow-framework

    job-scheduler --> skills
    anomaly-detection --> skills
    sql --> skills
    ml-commons --> skills

    job-scheduler --> search-relevance
    neural-search --> search-relevance
    k-NN --> search-relevance
    ml-commons --> search-relevance
    user-behavior-insights --> search-relevance

    %% Styles
    classDef core fill:#FFD700,stroke:#333,font-weight:bold
    classDef level1 fill:#90EE90,stroke:#333
    classDef level2 fill:#87CEEB,stroke:#333
    classDef level3 fill:#DDA0DD,stroke:#333
    classDef level4 fill:#FFA07A,stroke:#333
```

Defaults to 4 parallel threads to start with and adding parallel is optional in the build process for now.
Below are the "build" stats for 3.6.0 builds when ran using sequential build, parallel builds with 4 threads and parallel builds with 8 threads on jenkins. 

image

### Issues Resolved

### DIFF
--- a/src/run_build.py
+++ b/src/run_build.py
@@ -9,9 +9,11 @@
 import logging
 import os
 import sys
+import threading
 import uuid
 
 from build_workflow.build_args import BuildArgs
+from build_workflow.build_graph import BuildGraph
 from build_workflow.build_incremental import BuildIncremental
 from build_workflow.build_recorder import BuildRecorder
 from build_workflow.build_target import BuildTarget
@@ -21,6 +23,8 @@ from manifests.input_manifest import InputManifest
 from paths.build_output_dir import BuildOutputDir
 from system import console
 from system.temporary_directory import TemporaryDirectory
+
+CRITICAL_COMPONENTS = ['OpenSearch', 'job-scheduler', 'common-utils', 'OpenSearch-Dashboards']
 
 
 def main() -> int:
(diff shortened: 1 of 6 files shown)


src/run_build.py
```diff
@@ -22,6 +24,8 @@
 from system import console
 from system.temporary_directory import TemporaryDirectory
 
+CRITICAL_COMPONENTS = ['OpenSearch', 'job-scheduler', 'common-utils', 'OpenSearch-Dashboards']
```
peterzhuamazon: We should add opensearch-remote-metadata-sdk as a critical component.

peterzhuamazon (commented):
Hi @gaiksaya 
Generally happy with this change but with a few comments.

1. Internet might be throttling especially for home network, downloading multiple repos while later pulling dependencies from maven/npm push pressure on both host machine and remote repo (already see multiple times maven block/npm block due to too many connections in parallel from same ip)
2. whether the host has enough space / cpu to run this. Especially it has a side effect when using with --keep that can silently increase temp folder size without knowing
3. Suggest make default threads to 2 when using --parallel, good that default is still non-parallel so this does not break behavior.

Thanks.

src/run_build.py
gaiksaya: I did not touch what we already had but looks like only 2 components reply on it so not sure if it should be critical.

gaiksaya:
>
> 1. Internet might be throttling especially for home network, downloading multiple repos while later pulling dependencies from maven/npm push pressure on both host machine and remote repo (already see multiple times maven block/npm block due to too many connections in parallel from same ip)

Most of our own dependencies are pulled from local maven. Maybe we can monitor for sometime before taking the call or using 2? Atleast for all test runs and local building I did not see an issue for now. 

> 2. whether the host has enough space / cpu to run this. Especially it has a side effect when using with --keep that can silently increase temp folder size without knowing

We are not using --keep in the CI/CD so focusing on that. It's the same temp dir regardless of seq or parallel. But I could add a warning if you like. 

> 3. Suggest make default threads to 2 when using --parallel, good that default is still non-parallel so this does not break behavior.

I would like to test with 4 and decrease/increase if required. Running 2 in parallel does not provide much of a value IMO.

src/run_build.py
andrross: Is the purpose of this list that these components must succeed to create a build even if --continue-on-error is passed in? Might be worth adding a comment, and/or renaming them to "required components"

src/run_build.py
```diff
@@ -110,5 +117,63 @@ def main() -> int:
     return 0
 
 
+def _build_parallel(manifest: InputManifest, target: BuildTarget, build_recorder: BuildRecorder, work_dir: str, args: BuildArgs, components: list) -> list:
```
andrross, line 119: My AI reviewing partner tells me that `BuildRecorder` is not thread safe but is being mutated concurrently here.

src/run_build.py
```diff
@@ -110,5 +117,63 @@ def main() -> int:
...
+        critical_components=CRITICAL_COMPONENTS,
+        continue_on_error=args.continue_on_error,
+    )
+
+    if not args.continue_on_error:
+        critical_failures = [f for f in failed if f in CRITICAL_COMPONENTS]
+        if critical_failures:
+            raise Exception(f"Critical component(s) failed: {critical_failures}")
+
+    return [f for f in failed if f not in CRITICAL_COMPONENTS]
```
andrross: What happens if --continue-on-error is specified and one the critical components fails?

src/run_build.py
gaiksaya: Yes. Sure let me rename the variable.

src/run_build.py
gaiksaya: If hard fails. `--continue-on-error` does not matter as this point. Looks like there was bug with this. Fixed in recent commit. When required component fails, exception is raised no matter what. For others, only raised when continue-on-error is false.

src/run_build.py
gaiksaya: Looks like diff analyzer has same review. Removed the lock on buildRecorder altogerther.
 `record_component()` writes to `components_hash[name]` and `record_artifact()` appends to `components_hash[name]["artifacts"]`. Basically each parallel thread only operates on its own component name as the key. There's no shared mutable state between threads since the data is naturally partitioned by component name.

andrross (approved):

peterzhuamazon (approved):

Issue #6250, opened 2026-06-11
[On-boarding]: Onboard opensearch-jvector to the distribution and release automation


reta:
### What is the name of your plugin?

opensearch-jvector

### What is the link to your GitHub repo?

https://github.com/opensearch-project/opensearch-jvector

### Targeted release version

We would like to target 3.8.0+

### Does the plugin require documentation?

Yes, the documentation is in works https://github.com/opensearch-project/opensearch-jvector/issues/519 and will be submitted shortly

### What type of the plugin are we on-boarding?

Vector Search, alternative to `k-NN`

### Did you read the on-boarding document?

Yes

### Does this plugin has all the necessary automated tests?

Yes

### Have you completed the required reviews including Security reviews, UX reviews?

No but the plugin is a fork of k-NN 

### Have you on-boarded automated security scanning for the GitHub repo associated with this plugin?

Yes

### Additional context

N/A

reta:
@gaiksaya as we discussed in https://github.com/opensearch-project/opensearch-build/pull/6247#issuecomment-4674019211, thank you!

gaiksaya:
Thanks @reta 
Notifying @opensearch-project/admin about this.

peterzhuamazon:
Hi @reta,

Just want to check on the jvector onboarding.
Previously I believe Sam was just thinking of releasing jvector as is.
Also several k-NN member was talking in the original [REPO](https://github.com/opensearch-project/.github/issues/291) issue, on there might be duplication between jvector and kNN.

If we actually get jvector onboard, it will be installed by default.
I assume it has similar / same API calls with K-NN?
Will it conflict with k-NN in anyway?
And is there any reason that we want to do this right now?

Initially I was thinking just adding jvector as the version bump workflow entry, not the entire distribution.

let me know what you think.
Thanks.

cc: @getsaurabh02 @rishabh6788

peterzhuamazon:
Also seems like @navneet1v has a meta here and talking to Sam on next steps:

* https://github.com/opensearch-project/k-NN/issues/2605

navneet1v:
Agree on what @peterzhuamazon . @reta the idea was we will first build the abstraction to ensure that multiple knn related engines can be added. Otherwise plugins apis will contradict with one another

reta:
[...]

PR #6251, opened 2026-06-11
Add opensearch-jvector to 3.9.0 distribution manifests only

reta:
### Description
Add opensearch-jvector to 3.8.0 distribution manifests only

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/6250


manifests/3.8.0/opensearch-3.8.0.yml
```diff
@@ -265,3 +265,14 @@ components:
...
+    repository: https://github.com/opensearch-project/opensearch-jvector.git
+    ref: main
+    platforms:
+      - linux
+      - windows
+    depends_on:
+      - common-utils
+      - security
+      - custom-codecs
+    optional: true
```
reta: @peterzhuamazon @gaiksaya folks, would love to run this idea over you: the concept of optional plugins (== components), that eliminates the concerns we've discussed here https://github.com/opensearch-project/opensearch-build/issues/6250#issuecomment-4733789962. 

The large picture:
 - optional plugins will be build for every distribution
 - optional plugins will be tested for each distribution
 - optional plugins **will not be bundled** with the distribution but would need manual installation

Wondering what do you think folks? Thank you.

gaiksaya: Hi @reta 

I think it would be similar to core's native plugins. However, in order to test the plugin it needs to be bundled so that integration test can run against the given cluster endpoint containing the plugin

reta: Thanks @gaiksaya 

> I think it would be similar to core's native plugins. However, in order to test the plugin it needs to be bundled so that integration test can run against the given cluster endpoint containing the plugin

Correct, this could be done independently from the distribution bundle itself (the change does not include this support yet). I was wondering what do you think about the approach in principle, if the idea is sound, I will close the testing gap, thank you

reta: @gaiksaya so I have tried the number of options to support integration testing, and all of them are not looking good (both from convoluted install / uninstall flows and weird conflicting plugin chains), so I completely removed that part (== no smoke or integration tests for opensearch-jvector). So we would only build (that would help us to spot any issues beforehand), but never alter the distribution. 

What do you think?
Thanks!

gaiksaya: Sorry but I don't see a point of building it with distribution in that case. What are we trying to achieve here? If its just version increments, it can be handled separately.

reta: > Sorry but I don't see a point of building it with distribution in that case. What are we trying to achieve here? If its just version increments, it can be handled separately.

No, not just  version increments. Primarily, to track things like:
 - Apache Lucene updates, those break plugins mostly all the time
 - Dependency conflicts (including the plugins `opensearch-jvector` depends upon), those also break plugins mostly all the time

Making it part of OpenSearch distribution builds  (even with the quirks we have now, so we don't bundle) looks logical, no? thanks!

reta:
No interests from maintainers, closing

PR #6418, opened 2026-08-05
Exclude gradle dependencies from whitesource

gaiksaya:
### Description
The Mend GitHub App is filing CVE issues for our Jenkins test harness — 122 of our 284 open issues. bcprov, jenkins-core, jackson, spring, mina, jansi all come from the com.mkobit.jenkins.pipelines Gradle plugin, which exists only so ./gradlew test can run tests/jenkins/. None of it ships — there's no publish task, and the only jar in git is the Gradle wrapper. So we turn off Gradle resolution for the App and keep Python/Pipfile scanning.

Note that `.whitesource` file can't express this. Its JSON has no scope/path keys — only severity floors and check-run settings. Dependency filtering lives exclusively in Unified Agent format, and `configMode: LOCAL` is the only switch that makes the App read it. That's why the mode flip is required, not cosmetic.

Also removes old whitesource scanner as each repo is now integrated with whitesource mend app by default.

### Issues Resolved


Pipfile
```diff
@@ -42,6 +42,7 @@ certifi = "~=2024.7.4"
 types-urllib3 = "~=1.26.25.14"
 charset-normalizer = "~=2.1.1"
 beautifulsoup4 = "~=4.12.3"
+soupsieve = "~=2.8.4"
```
peterzhuamazon, line 45: Why adding explicit soupsieve 2.8.4 here?

peterzhuamazon (approved):

Pipfile
gaiksaya: To resolve https://github.com/opensearch-project/opensearch-build/issues/6302

PR #6477, opened 2026-09-16
Emit cluster output logs for validation workflows

gaiksaya:
### Description
Emit cluster output logs for validation workflows
Move validation to a stage rather than post action

### Issues Resolved
#6426 


src/validation_workflow/validation.py
```diff
@@ -153,8 +153,25 @@ def check_cluster_readiness(self) -> bool:
                 return True
             retry_count += 1
         logging.error(f"Maximum number of retries ({max_retry}) reached. Cluster is not ready for API test.")
+        self.log_cluster_output()
         return False
 
+    def log_cluster_output(self) -> None:
```
Divyaasm, line 160: After log4j2 finishes initializing, OpenSearch writes the real failure to opensearch.log, Since we currently capture only the .bat/.sh stdout/stderr, is it worth also tailing the log file?

gaiksaya: Addressed in recent commit https://github.com/opensearch-project/opensearch-build/pull/6477/changes#diff-a318e15081b06176764de2b9bd492c632948b9398c4de3b9fc000145d2c18e89R181

Divyaasm (approved):

